PR #774, opened 2014-12-31
BF: import csv files containing quoted commas

m-macaskill:
Currently, PsychoPy cannot import conditions files which have fields which contain commas, even when those fields are surrounded by quotes, which is a common CSV convention.
To fix this, the TrialHandler now imports conditions files using the pandas csv_reader rather than numpy's recfromcsv
NB this works for Coder BUT Builder's loop dialog still fails to deal with csv files which contain quoted commas. That GUI aspect needs to be updated in a similar fashion but I'm not sure how.
i.e. at this stage, data.py can deal with these csv files, but Builder users can't select them.

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -8,7 +8,7 @@ from psychopy import logging
 from psychopy.tools.arraytools import extendArr, shuffleArray
 from psychopy.tools.fileerrortools import handleFileCollision
 import psychopy
-from pandas import DataFrame
+from pandas import DataFrame, read_csv
 import cPickle, string, sys, os, time, copy
 import numpy
 from scipy import optimize, special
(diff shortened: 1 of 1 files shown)


peircej:
I guess it's time we make pandas a firm dependency for this to be possible. So far it's been an optional extra

m-macaskill:
Ahh, I assumed it was already included, as there is an existing pandas import in data.py. But maybe that just comes from Alex's recent work to avoid the cycle of having to write/read data via disk. I think if we are pushing people to have an analysis workflow via PsychoPy, pandas seems now to be a necessary package for that.

I can't follow the link to see what the Travis build failure is about…

jeremygray:
It fails at TextExpt.test_loopBlocks, line

    dat = data.importConditions(datafileBase+".csv")

with error about bad value(s) in the conditions file

m-macaskill:
So that test generates a .csv *data* file and then imports it to check its dimensions using data.importConditions(). The change to use pandas for importing conditions chokes on this because columns in a data file have headers containing punctuation, like "blocks.thisRepN".

It seems to me that we want imprtConditions to be fussy and cause an error in situations like this, because condition file headers become variable names and hence shouldn't include spaces or punctuation: it's a nice automatic check on the user.

So it seems to me that it is the test code needs to change in this situation, because it is using an inappropriate function as a shortcut to import data. i.e. it should treat it as an actual data file rather than just using importConditions for convenience. I'm reluctant to change test code, though, as I'm not at all familiar with how the tests work (and in particular how to run and debug them).

But I guess in  TextExpt.test_loopBlocks, we could replace:

dat = data.importConditions(datafileBase+".csv")

with something like:

f = open(datafileBase+".csv", 'rU')
dat = numpy.recfromcsv(f, case_sensitive=True)
f.close()

and then the test could complete as at present, with:

 assert len(dat)==8 # because 4 'blocks' with 2 trials each (3 stims per trial)

Does that make sense? I'm flying a bit blind here, dealing with tests for the first time.

peircej:
Thanks Mike, your fix worked (when combined with another to do with the relative path which was wrong on win32 machines)

PR #775, opened 2015-01-03
extraInfo was not being exported in wideText

alexholcombe:
The names of the extraInfo columns were being printed in the header, but the content wasn't! This is trialHandler only that I fixed, experimentHandler has the same bug which I can fix later, as well as having experimentHAndler also return dataframe, like trialHandler does. 

To do the fix, I broke up the header into extraInfo part and normal part. Otherwise, the code wouldn't work if extraInfo had a field name that was the same as the trials data (seems to be nothing stopping the user from doing that)

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -1161,17 +1161,19 @@ class TrialHandler(_BaseTrialHandler):
     
         # collect parameter names related to the stimuli:
         if self.trialList[0]:
-            header = self.trialList[0].keys()
+            headerExceptExtraInfo = self.trialList[0].keys()
         else:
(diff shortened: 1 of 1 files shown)


m-macaskill:
I haven't heard any complaints about that, so asssume it isn't happening in the current release. Would be really mucking up peoples' data if it was…

alexholcombe:
You're right Mike. TrialHandler.py wasn't testing this so I added some extraInfo to it, and it works. I'll investigate more why my it didn't work for my program, and I'll also suggest this modification of TrialHandler.py.

PR #784, opened 2015-01-09
install pandas via apt-get -- must be coming from neurodebian

yarikoptic:
to see what travis says


yarikoptic:
ok -- since noone is so brave -- should I merge it myself? ;)

peircej:
It's only me and you that have permission yarik, and I was busy. Sounds good.
I think I had tried installing by apt-get on another machine and had a failure that I assumed was related to  system permissions for repositories.

PR #785, opened 2015-01-09
enable mouse.setPos under pyglet

jeremygray:
Works for me on Mac with pyglet 1.2alpha1, and for Sol on Windows. Fails on Travis, possibly due to either pyglet or to xlib. So seems safest to build in a check before trying to set it.


jeremygray:
will clean up a few things and reopen as another PR

PR #786, opened 2015-01-09
mouse.setPos() works on pyglet

jeremygray:
Used Sol's idea to set the mouse under pyglet. Worked fine on Mac with pyglet1.2alpha1, but not on travis (= linux, and maybe pyglet 1.1).

travis tweaks for less verbose installation of software


jeremygray:
going to try again on this. (the whole point was a cleaner commit history!) blah.

PR #798, opened 2015-01-21
File Saving Refactoring

hoechenberger:
The most user-visible change of this patch set is that all ``saveAs*`` methods from all classes defined in the ``data.py`` module will **preserve** all existing data by default. This means that all these methods now either append to an existing file, or cleverly handle file name collision and create **new** files with increasing index if the user does not wish to append, but a file of the desired filename already exists. This behavior can be overridden, of course. It already existed for a few select methods, but was now implemented for all classes in ``data.py``.

So to sum it up:
**It should not be possible to accidentally overwrite existing data anymore. Overwriting has to be explicitly requested, and deletion has to be performed manually.**

The default filename extensions have been changed from ``.dlm``. to ``.csv``, ``.tsv``, and ``.txt``.

Also included are some minor bug fixes/enhancements, one of them changing the behavior of ``ExperimentHandler.saveAsPickle()`` and ``ExperimentHandler.saveAsWideText()`` to a more user-friendly (i.e. more reasonable) behavior.


peircej:
thanks for the clear message! ;-)

hoechenberger:
I think this is good so far, although I'd love to increase test coverage, but maybe that's another matter altogether... 

Also I'm not sure about the format of the API documentation I've added to the filetools module (mostly following [NumPy style](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt): 

https://github.com/hoechenberger/psychopy/blob/b4ceb0d0f64a8284f511f4e4e9632d134da6cbd8/psychopy/tools/filetools.py#L64

Is this okay or should it be changed to another style?

hoechenberger:
This is freshly rebased, manually tested, passing all automatic tests, and good to merge now, I think!

peircej:
Hey Richard, thanks for your input as always. Great to have you on the team!
I'm afraid I did end up building the release yesterday and pushed it to sourceforge, so this commit will wait for the next one I'm afraid (typically there's a bug-fix release soon after a major release because there's very often *some* bug that I fail to catch at release time.

hoechenberger:
Thanks Jon! The decision in which release to include these changes is of course up to you. I'm just wondering that since the code changes the behavior of PsychoPy a bit, maybe this should be part of the next major, not bugfix, release?

peircej:
There is that issue, but the number of people affected will be very 
small (the number of people using coder to create tab-delimited files 
will be small) and they are likely to be the sort of users that can cope 
with the change (they can write a script to rename their files afterwards).

It's much easier for me to stick to a single branch than to manage a 
1.82 of bug fixes alongside a master branch of all changes. I do that 
sometimes but it's no fun ;-)

On 27/01/2015 10:19, Richard Höchenberger wrote:
>
> Thanks Jon! The decision in which release to include these changes is 
> of course up to you. I'm just wondering that since the code changes 
> the behavior of PsychoPy a bit, maybe this should be part of the next 
> major, not bugfix, release?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/psychopy/psychopy/pull/798#issuecomment-71622065>.
>

-- 
Jon Peirce
http://www.peirce.org.uk

hoechenberger:
Good points, absolutely agree!

PR #801, opened 2015-01-22
NF: PSI adaptive psychophysical method

JosephGlavan:
Modified "data.py" to include new PsiHandler class

Added "psi.py" to handle back end operations of PsiHandler

Modeled after the QuestHandler methods


JosephGlavan:
This is my first time contributing to a project using GitHub so I'm not sure what merge conflicts exist or how to resolve them.

PR #804, opened 2015-01-24
ENH: always draw ISI components in routine panel; closes #799

### LINKED ISSUE #799: ISI disappears in Builder when using variable for duration
I wanted to generate a dynamic ISI and therefore used a variable for "duration". Unfortunately, this makes the ISI disappear. Well, also for other objects the visualization of the timing disappears when you do this, but for the other objects, you at least still have the icon to access the object. For the ISI, I wasn't able to access the ISI object at all after that. The object is still there though. I can change the duration variable in my code section and it has an effect, but, for now, I am just not able to reopen the component. As a workaround, you can set the "expected duration". That helps. But it might be a good idea that at least an icon of the ISI sticks around so that one can use the Builder still.

Thanks and best,

Axel

### LINKED ISSUE #802: proposal: "duration (loops)" for sound and video files
A natural thing to want to do is play a sound or movie clip for the duration of the file. It might be worth the (small?) effort to add support for duration to be specified in terms of loops through the media file, with the default being 1.

This would allow auto-terminating a routine when a sound ends.

jeremygray:
As noticed on the user list, having a $code duration or start means that an ISI (static) component
is not drawn at all in the routine panel. This made it awkward (imposssible?) to edit or delete it.

Fix: Ensure we always draw something. Append '???' to the name to signal that its timing is unknown / nonliteral.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -1421,7 +1421,7 @@ class RoutineCanvas(wx.ScrolledWindow):
         font.SetPointSize(size)
         dc.SetFont(font)
     def drawStatic(self, dc, component, yPosTop, yPosBottom):
-        """draw a static component box"""
+        """draw a static (ISI) component box"""
         #set an id for the region of this component (so it can act as a button)
         ##see if we created this already
         id=None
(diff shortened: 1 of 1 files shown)


jeremygray:
drat, wrong close number in PR, will reopen after cleaning that up

peircej:
don't worry about #802 that's about to be closed by me anyway

PR #810, opened 2015-01-26
BF: iohub use of pyHook on Windows, window.__del__ exceptions, ..

isolver:
- pyHook lib used by iohub on Windows was causing Windows system error at end of experiment when running on 64bit python. That is now fixed.

- psychopy.visual.Window.__del__ was causing unhandled exceptions at end of experiment if win.close() was not called prior to python shutting down. This was because a lot is being done in win.close(), so when called from __del__, some of the referenced objects where GCed by python already. This was especially true when iohub was in use, as win.close() informs iohub of the win id that is being removed, which takes 250 usec or so to do.


peircej:
I'm afraid I built the 1.82.00 yesterday so this commit will have to wait for the first bug-fix release (which can be soon)

isolver:
No issue; since in practice it only happens when the application is
quitting, it is just getting rig of the python error message prints that
matters.

PR #815, opened 2015-01-28
BF: Don't overwrite extraInfo variables in TrialHandler.saveAsWideText() (#814)

### LINKED ISSUE #814: extraInfo vars not recorded using TrialHandler.saveAsWideText()
I've been having trouble getting the extra variables I pass as `TrialHandler`'s
`extraInfo` argument to actually show up in the data file when using
`saveAsWideText()`. I think I've now
tracked down the bug to the lines in `data.py -> saveAsWideText()` where the entries get created:

```python
# create a dictionary representing each trial:
# this is wide format, so we want fixed information (e.g. subject ID, date, etc) repeated every line if it exists:
if (self.extraInfo != None):
    nextEntry = self.extraInfo.copy()
else:
    nextEntry = {}

# etc.

# now collect the value from each trial of the variables named in the header:
for parameterName in header:
    # the header includes both trial and data variables, so need to check before accessing:
    if self.trialList[trialTypeIndex] and parameterName in self.trialList[trialTypeIndex]:
        nextEntry[parameterName] = self.trialList[trialTypeIndex][parameterName]
    elif parameterName in self.data:
        nextEntry[parameterName] = self.data[parameterName][trialTypeIndex][repThisType]
    else: # allow a null value if this parameter wasn't explicitly stored on this trial:
        if parameterName == "TrialNumber":
            nextEntry[parameterName] = trialCount
        else:
            nextEntry[parameterName] = ''
```

Because the variables in `extraInfo` are also in `header`, that loop through the
header's parameter names overwrites them with empty strings. Should
be pretty easy to fix, I'll try to get a pull request going now.

onesandzeroes:
Fixes #814. I think the fix for this is pretty straightforward, we just fill in the `extraInfo`
values along with the rest while looping through the header. I've changed the tests
against `None` to use `is` rather than `!=` since I was changing those sections
anyway, but if people want to suggest any other improvements I'm happy to
include them.

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -1187,7 +1187,9 @@ class TrialHandler(_BaseTrialHandler):
         header.extend(self.data.dataTypes)
         # get the extra 'wide' parameter names into the header line:
         header.insert(0,"TrialNumber")
-        if (self.extraInfo != None):
+        # this is wide format, so we want fixed information
+        # (e.g. subject ID, date, etc) repeated every line if it exists:
+        if self.extraInfo is not None:
             for key in self.extraInfo:
                 header.insert(0, key)
         df = DataFrame(columns = header)
@@ -1210,11 +1212,7 @@ class TrialHandler(_BaseTrialHandler):
                 repThisType=repsPerType[trialTypeIndex]#what repeat are we on for this trial type?
 
                 # create a dictionary representing each trial:
-                # this is wide format, so we want fixed information (e.g. subject ID, date, etc) repeated every line if it exists:
-                if (self.extraInfo != None):
-                    nextEntry = self.extraInfo.copy()
-                else:
-                    nextEntry = {}
+                nextEntry = {}
 
                 # add a trial number so the original order of the data can always be recovered if sorted during analysis:
                 trialCount += 1
@@ -1226,6 +1224,8 @@ class TrialHandler(_BaseTrialHandler):
                         nextEntry[parameterName] = self.trialList[trialTypeIndex][parameterName]
                     elif parameterName in self.data:
                         nextEntry[parameterName] = self.data[parameterName][trialTypeIndex][repThisType]
+                    elif self.extraInfo is not None and parameterName in self.extraInfo:
+                        nextEntry[parameterName] = self.extraInfo[parameterName]
                     else: # allow a null value if this parameter wasn't explicitly stored on this trial:
                         if parameterName == "TrialNumber":
                             nextEntry[parameterName] = trialCount


peircej:
Hi there @onesandzeroes 
Apologies for the delay. I thought this was working already in our main datafiles, but that was maybe just for the ExperimentHandler that it was working, not the TrialHandler.
@alexholcombe you were looking at data outputs recently for ExperimentHandler and talking about porting your fixes to TrialHandler (if I remember correctly?). Would this commit give you any probs that you know of?

cheers both

onesandzeroes:
Here's some quick code to repro if you want to confirm, I've run this against the current git master and it definitely seems like the bug is still there:

```python
import random
from psychopy import data

trials = data.TrialHandler(
    [
        {'colour': 'red', 'size': 4, 'condition': 1},
        {'colour': 'blue', 'size': 5, 'condition': 2},
        {'colour': 'green', 'size': 6, 'condition': 1}
    ],
    nReps=1,
    extraInfo={'subject_id': 101, 'day': 'Tuesday'}
)

for trial in trials:
    trials.data.add('rt', random.uniform(500, 800))

trials.saveAsWideText('trialhandler_bug_example')
```

alexholcombe:
I'm sure it worked at some time before with trialHandler, but maybe my test
wasn't comprehensive enough, I really don't remember sorry and I'm under a
grant deadline now. What onesandzeroes did looks good!
Alex

On Thu, Feb 5, 2015 at 12:50 PM, onesandzeroes <notifications@github.com>
wrote:

>  Here's some quick code to repro if you want to confirm, I've run this
> against the current git master and it definitely seems like the bug is
> still there:
>
> import randomfrom psychopy import data
>
> trials = data.TrialHandler(
>     [
>         {'colour': 'red', 'size': 4, 'condition': 1},
>         {'colour': 'blue', 'size': 5, 'condition': 2},
>         {'colour': 'green', 'size': 6, 'condition': 1}
>     ],
>     nReps=1,
>     extraInfo={'subject_id': 101, 'day': 'Tuesday'}
> )
> for trial in trials:
>     trials.data.add('rt', random.uniform(500, 800))
>
> trials.saveAsWideText('trialhandler_bug_example')
>
>  —
> Reply to this email directly or view it on GitHub
> <https://github.com/psychopy/psychopy/pull/815#issuecomment-73147220>.
>



-- 
Alex Holcombe | Associate Professor
School of Psychology | Faculty of Science
The University of Sydney
Web <http://www.psych.usyd.edu.au/staff/alexh/lab/>, Twitter
<https://twitter.com/ceptional>, Map
<http://openwetware.org/wiki/Holcombe:Location>

PR #825, opened 2015-02-02
BF: move routineTimer.reset() to end of routine; fixes #822

### LINKED ISSUE #822: Builder routine: routineTimer for nonSlip can get reset too early
This is not necessarily a bug, but it tripped me up, and I think at the very least we need to tell people more about this situation: when a nonSlip routine follows a non-nonSlip (dynamic length) routine.

In a dynamic routine, the global routineTimer gets .reset() if the while loop is not exiting:

        win.flip()
    else:  # this Routine was not non-slip safe so reset non-slip timer
        routineTimer.reset()

(Seems like that could just be done once, just after the while loop when it exits, not every frame?)

However, if you have a code component in that dynamic routine and it has code in End Routine, that code will get run after the loop has ended (= after routineTimer has been reset). the routineTimer then ticks away and time elapses. In my case this was during launchScan(), which waits for user input. If the next routine is nonSlip (in my case: fMRI trials), time is then added to the routineTimer. But any already-elapsed time (during launchScan) will count against the total, which is incorrect behavior. Fortunately I noticed by eye because it was seconds, but I would not have noticed even 200ms (e.g., to load a large file). Maybe I would have seen this in the log or data files, but it would be easily missed.

To fix, I ended up putting routineTimer.reset(); routineTimer.add(someTimeValue) in code at the start of the next (= nonSlip) routine, but I think it would be better just to do routineTimer.reset() in the code at the end of the previous (dynamic) routine.

If there's a way to handle this properly in code, I think we should do it. 

If its too complicated to try to handle programmatically, we should document and warn people about it. 

Are there any other cases where routineTimer could get off from the intended behavior?

jeremygray:
- fixes my case
- not sure about all other possible cases
- more discussion might be good before pulling this is? seems ok to me even if other work is also needed

### DIFF
--- a/psychopy/app/builder/experiment.py
+++ b/psychopy/app/builder/experiment.py
@@ -1225,7 +1225,6 @@ class Routine(list):
         #are we done yet?
         buff.writeIndentedLines('\n# check if all components have finished\n')
         buff.writeIndentedLines('if not continueRoutine:  # a component has requested a forced-end of Routine\n')
-        buff.writeIndentedLines('    routineTimer.reset()  # if we abort early the non-slip timer needs reset\n')
         buff.writeIndentedLines('    break\n')
         buff.writeIndentedLines('continueRoutine = False  # will revert to True if at least one component still running\n')
         buff.writeIndentedLines('for thisComponent in %sComponents:\n' %self.name)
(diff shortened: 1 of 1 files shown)


peircej:
sounds correct to me

PR #828, opened 2015-02-03
BF: Tobii calibration window not closing in 1.82

isolver:
For some reason, in 1.82, when iohub calls win.close() on the full
screen window created for tobii calibration graphics, the window does
not actually close.

Work around is to minimize() the full screen window before calling
close, then it closes the window fine and the experiment window is
visible as expected.


isolver:
BTW. this seems to be computer (video card?) dependant, as I do not get this issue on the Sony laptop I use for testing Tobii related stuff. However it has been reported by 2 people, so it must be a somewhat common issue.

PR #831, opened 2015-02-10
ENH: TrialHandlerExt class for non-counterbalanced designs

suddhasourav:
The TrialHandlerExt class implements functionalities to run non-counterbalanced designs. Works also for usual cases. For using the advanced features, either the excel/csv file should have a column called weight, or the trialList array should have a weight key for each of the dict members.

nRep can be used as blocks, and then the random shuffling method shuffles within blocks.

Tests will be added soon, and I will provided better functionalities for saving data (currently only the default analyses work, as I inherited from BaseTrialHandler)

Link to talk: https://groups.google.com/forum/#!topic/psychopy-dev/V7cJQ8Q3aLQ


suddhasourav:
Hi Jon! I have the tests now, and was thinking of creating another pull request after this one gets accepted - does it look good to you, or should I change something? :-)

Cheers,
Suddha

peircej:
Looking at the code I have to say I'm not totally sure still. It feels like a lot of extra code for the relatively small extra feature of having a weight parameter. Is it incompatible with the original TrialHandler so that it needs a new class?
I was imagining just a few extra lines in the _makeIndices function. But I haven't actually tried to do this, so maybe you found much more was needed?

suddhasourav:
Hi Jon!

In the beginning, I thought of implementing the changes in TrialHandler too, and that can still be done. I think the functionalities are sufficiently different as to confuse someone reading the documentation when all they need is a simple TrialHandler, and therefore made a superset of its functionalities, but that’s a decision you’re in a better position to make :-D

When adding the trial weights, I did not want to change any existing behavior. In the existing TrialHandler class, the indexing is done by [trialIndex, nRepetition], but if we have the same trialIndex several times, then it’s necessary to index it in a more sophisticated way.  Therefore the getCurrentTrialPosInDataHandler() and getNextTrialPosInDataHandler() functions.

Moreover, the data matrix has to be reshaped afterwards to make the summary of one trial type belong to one single row, so I overrode _createOutputArrayData(). Same goes for saving.

I’ve tried to stick to the philosophy of as little change to existing patterns as possible, and not breaking any code that exist already. Any suggestion is welcome of course!

Best regards,
Suddha

suddhasourav:
Hi Jon, any suggestions for me? :)

peircej:
Sorry, been busy the last few weeks. I'm going to merge this in for now since it doesn't actually affect the functioning of anything else. 

Actually I think we should revamp the whole TrialHandler concept to one that produces the next trial *on demand* rather than precomputed. That might supercede your work as well as the original TrialHandler. I think I'll email the dev list about this though.

PR #832, opened 2015-02-11
DOC: Added Preferences documentation and related links

kohlman:
Hello,

I added documentation for some options in the "General Settings" that were missing. I had a couple of questions related to that:

* sound.rst: In the first paragraph, it says: "This can be set on a per-experiment basis by importing preferences, and setting the audioLib preference to use." I couldn't find any current documentation that could be linked. It would be good to know how exactly this can be done on a per-experiment basis (and not just the general settings).
* prefs.rst: I would restructure the document to reflect the sequence of tabs in the "General Settings" dialog, but I still feel uneasy about this without an OK from the core team.
* prefs.rst: Documentation for options in most other tabs are missing, but, again, I wanted to ask for an OK before I proceed with this. The only drawback will be that I might not really know what's going on with some of the options. Should I check first on the developer's mailing list and then proceed or rather just make the changes with stubs if necessary and then improve later on?
* prefs.rst: I didn't have to say much about  __audio driver__ and __flac audio compression__. Maybe somebody can improve on that.
* prefs.rst: Often lists are used in the preferences settings. Ideally, one would specify how the list is handled in each case. For example, the parallel-port setting defines a list of options that is displayed as a drop-down menu in the corresponding component. But I guess for many other settings the first option is checked and if it fails, the next options are tried (e.g., for __audio library__ or __audio driver__). Maybe somebody can confirm this behavior.

I hope that the links work as expected. As always, I am happy to extend the documentation accordingly after some feedback.

Cheers,

Axel

### DIFF
--- a/docs/source/api/sound.rst
+++ b/docs/source/api/sound.rst
@@ -6,7 +6,7 @@
 :class:`Sound`
 -------------------
 PsychoPy currently supports a choice of two sound libraries: pyo, or pygame. Select which will be
-used via the `audioLib` preference. `sound.Sound()` will then refer to either
+used via the :ref:`audioLib<generalSettings>` preference. `sound.Sound()` will then refer to either
 `SoundPyo` or `SoundPygame`. This can be set on a per-experiment basis by importing
 preferences, and setting the audioLib preference to use.
 
--- a/docs/source/general/prefs.rst
+++ b/docs/source/general/prefs.rst
@@ -16,8 +16,22 @@ fullscr:
     Should windows be created full screen by default? Can be overridden by individual experiments.
 
 allowGUI:
-    When the window is created, should the frame of the window and the mouse pointer be visible.
-    If set to False then both will be hidden.
+	    When the window is created, should the frame of the window and the mouse pointer be visible. If set to False then both will be hidden.
+
+paths:
+    Paths for additional Python packages can be specified. See more information :ref:`here<addModules>`.
+
+audioLib:
+    As explained in the :doc:`Sound</api/sound>` documentation, currently two sound libraries are available, pygame and pyo.
+
+audioDriver:
+    Also, different audio drivers are available.
+
+audioFlac:
+    Set flac audio compression.
+
+parallelPorts:
+    This list determines the addresses available in the drop-down menu for the :doc:`/builder/components/parallelout`.
 
 .. _applicationSettings:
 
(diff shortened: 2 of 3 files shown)


jeremygray:
Hi Axel,

I agree with your idea to restructure to follow the tab. I'll take a look at the other things as I can.

Here's an outline for the sound.rst part:

For sound.rst the code part to set the pref per-experiment:

    from psychopy import prefs
    prefs.general['audioLib'] = ['pyo']
    from psychopy import sound

You must import the sound module AFTER setting the pref. To check that you are getting what you want (don't do this in your actual experiment):

   print sound.Sound

and this will be `<class 'psychopy.sound.SoundPyo'>` for pyo, or  `<class 'psychopy.sound.SoundPygame'>` for pygame.

kohlman:
Hi Jeremy,

Thanks for your comments. I will make further adjustments accordingly. I 
guess I'll better wait until my first changes are pulled and then modify 
the overall structure and include your hints. I think I'll manage to 
provide reasonable placeholders for most of the option settings. I think 
it's better than not finding the option in the documentation. With 
further hints from you and others, I can then improve on that in a 
continuous fashion.

Best,

Axel

kohlman:
Hi Jeremy,

Concerning the sound preferences, I thought the simplest solution would 
be to link to the API reference, but I realized that there is no 
documentation included for Preferences. My guess is that this is because 
one should rather operate on the class instance "prefs" than on the 
class "Preferences" itself (just referring to the class might be 
confusing for users). Nevertheless, it would be good to also have 
documentation on those parts. Is there a way to do this? Maybe include 
"Preferences", but still inform about the correct usage in a preceding 
paragraph?

Best,

Axel

GraymatterCode:
Yes, you are thinking about things clearly. The idea with setting the pref value is exactly as you say, to change the value in the class instance at run-time, so that a) psychopy.sound will be sure to use the value ('pyo' or 'pygame') and NOT default to whatever the user's pref might be, and b) it is only set for the duration of the script execution, and does not overwrite the user's prefs.

I'm less sure that the documentation itself should only steer people toward prefs (and away from preferences). Most people will not be concerned with the instance at runtime vs prefs as saved on disk distinction. So I agree with your suggestion to have documentation on Preferences, and mention setting pref values as a special usage, and when its appropriate. Currently I think selecting the lib to use for sound is the only common reason to assign values to the prefs instance.

PR #845, opened 2015-02-23
BF: fix mac laptop sound latency 0.5s for first sound; fixes #844

### LINKED ISSUE #844: sound play() lag of 0.5s on mac laptop
There can be a 0.5s lag, at least on Mac laptops, when playing a sound if no sound has been played for a minute or so. Appears to be an OS-level thing [Edit: apparently not on further testing]. Solution: play very short, very quiet sound before sound is needed, in order to "wake up" the sound card.

Its not just a laggy start, but can clip the end of the sound by 0.5s. The thread that stops pyo sounds is based on time, and starts at the right time, terminating the sound before its actually finished.

Complexities: Its easy to play a micro-sound at the start of the experiment, but hard to guarantee that this will take care of the issue (instructions screen might be up for a while, etc). So to fix the general case more is needed. Could start a thread that plays a micro sound, sleeps for 30s, and then loops until the experiment quits.

jeremygray:
- problem: see #844
- fix: play a short ~silent sound every 30s from a thread, launched by initPyo()
- limitations: pyo only (I can't easily test with pygame)

### DIFF
--- a/psychopy/sound.py
+++ b/psychopy/sound.py
@@ -481,6 +481,27 @@ class SoundPyo(_SoundBase):
         # a DataTable has no .getDur() method, so just store the duration:
         self.duration = float(len(thisArray)) / self.sampleRate
 
+class _SilentSoundLoop(threading.Thread):
+    """function to wake up the sound card and keep it awake (Mac only)
+
+    Needed on mac (laptops only?) to avoid 0.5s latency to play first sound if
+    no sound has been played for ~45sec or so.
+
+    Sound volume == 0.0, but there still might be some slight distortion
+    introduced (not guaranteed to be completely silent).
+    """
+    def __init__(self):
+        super(_SilentSoundLoop, self).__init__(None, '_SilentSoundLoop', None)
+        self.daemon = True
+        self.microsound = Sound('A', 0.001, volume=0.000)
+        self.microsound.play()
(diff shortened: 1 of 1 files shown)


peircej:
I'm not convinced this is a good idea. I'm certainly not happy with it using core.wait() since that has a period of cpu hogging towards the end to increase accuracy (time.sleep() would be more acceptable). I think the knock-on effect of this will be sloppier timing everywhere else and it means another thread that needs cleaning up (ie. more to go wrong).

peircej:
Maybe there are commands to turn this off in the apple API (like there are commands to stop the screen from sleeping)

jeremygray:
Interestingly, I don't get the latency issue when using pyaudio instead of pyo or pysoundcard for playing a sound. This suggests that my original thinking was off, its not low-level Mac OS thing (and so is unlikely to be able to be turned on or off with a system setting). But its still an issue for Mac users who want sound in PsychoPy. time.sleep() would be fine, since its Mac only and there's no need for precision in the timing (core.wait(hogCPU=0) is not needed for cross-platform compatibility). using a daemon-ized thread will mean it cleans itself up on program exit--it certainly allows immediate program shutdown without any special handholding of the thread, unlike when self.daemon = False. Will think more about this.

PR #849, opened 2015-02-26
BF: Condition keys should be in ASCII

hoechenberger:
As Jeremy [pointed out](https://github.com/psychopy/psychopy/commit/6b293561edcd778d6de925662cb3c573b9a59d74#commitcomment-9915433), the keys in experimental condition specifications should consist of ASCII characters only, because at least the Builder creates variable names from the keys, and the test might yield misleading results in the future.

### DIFF
--- a/psychopy/tests/test_data/test_ExperimentHandler.py
+++ b/psychopy/tests/test_data/test_ExperimentHandler.py
@@ -86,8 +86,8 @@ class TestExperimentHandler():
         )
 
         conds = [
-            {'id': '01', u'näme': u'umlauts-öäü'},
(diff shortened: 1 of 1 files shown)


jeremygray:
Looking over the page, the majority of the prefs would never be set manually by a user from within a .py program in order to override the default settings. Definitely all of those within Coder, Builder, App, and Key bindings. And not checkForUpdates. Those are all about the app or used by the app, not an experiment created by the app. The other prefs within General and Connections just might possibly need tweaking by a user at experiment run time, so I would leave those and remove the others to simplify.

PR #850, opened 2015-02-26
BF: ParallelPort Class gives informative warnings

kohlman:
* During import, warnings are thrown if on a Mac or Windows drivers could not be found
* Added __init__ for ParallelPort, so that informative warning is thrown during instance creation

### DIFF
--- a/psychopy/parallel/__init__.py
+++ b/psychopy/parallel/__init__.py
@@ -16,8 +16,9 @@ port.
 There is also a legacy API which consists of the routines which are directly in this
 module.  That API assumes you only ever want to use a single parallel port at once.
 """
-from psychopy import logging
+
 import sys
+import warnings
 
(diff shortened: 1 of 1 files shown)


peircej:
I'm not sure about the change to using `warnings` rather than `psychopy.logging`
The advantage about logging is that it allows users to control whether or not they get this level of information from a single point (ie. by setting the logging level).

kohlman:
Yeah, Jon. I thought about that. The problem is that the log files were not even written because the uninformative error messages came up (using proposed standard code). But thinking about it, maybe adding the ParallelPort \__init__ is sufficient because it will preclude the error from happening. Then, just logging should be alright. I'll add another commit re-inserting the logging messages.

The only drawback is that users won't be aware of the warnings if they just include the code without taking care of proper logging etc. I don't know how acceptable this is.

PR #859, opened 2015-03-03
FIX: Icon and tooltip of UnknownComponent were not correctly displayed.

hsogo:
I've been concerned that UnknownComponent is not correctly displayed.

For example, a wrong tooltip is displayed when mouse cursor is put on the UnknownComponent icon as shown in the following screen shot.
fig1

When custom components are loaded using 'Prenerence->components folders' option, icon of a custom component is displayed as icon of UnknownComponent.
fig2

At present, UnknownComponent is defined in app/builder/components/_base.py. I think it tricky to define tooltip and icon for  UnknownComponent in _base.py. It would be better to define UnknownComponent in a separate file (e.g. unknown.py).


hsogo:
Thank you. I've fixed it.

PR #865, opened 2015-03-05
Tobii eyex refactor

dgfitch:
This changeset moves the Tobii EyeX eyetracker back in with the rest of the Tobii devices, where they can live in harmony. (Sort of. The APIs are quite different, unfortunately.)

Docs are updated; to use the EyeX you just need to change the model_name to "Tobii EyeX" in the YAML. 

I have worked with Sol and also confirmed that this branch works on the Tobii T120 and the EyeX, but more tests are welcome.

[I messed up merging from upstream in my other branch so I moved the history here. Sorry it looks weird, I'm git-stupid sometimes.]


dgfitch:
I think there's an unrelated typo in gamma.py. Once that's fixed, I think this should build fine - let me know if it does not.

peircej:
Thanks Dan. I'm waiting for Sol to check this over before pulling it in.

isolver:
Tested with the T120 I have; works great now with both included demo's. 
Good to go @peircej ! 
Thanks again @dgfitch !

njvack:
:tada:

PR #870, opened 2015-03-10
BF: ratingscale could show "False" for the scale description

jeremygray:


jeremygray:
the one test failure was just one of those seemingly random quirks

PR #872, opened 2015-03-11
ENH: TA and **kwargs arguments added to SyncGenerator to allow for proto...

thompsonj:
...cols with silent gaps (e.g. auditory experiments) and to facilitate the use of the settings dictionary for other experiment-specific settings. The mock scanner noise is now only played for the length of the acquisition (TA) and will repeat every TR with silent gaps of TR-TA seconds between acquisitions.

### DIFF
--- a/psychopy/hardware/emulator.py
+++ b/psychopy/hardware/emulator.py
@@ -55,7 +55,7 @@ class ResponseEmulator(threading.Thread):
         self.stopflag = True
 
 class SyncGenerator(threading.Thread):
-    def __init__(self, TR=1.0, volumes=10, sync='5', skip=0, sound=False):
+    def __init__(self, TR=1.0, TA=1.0, volumes=10, sync='5', skip=0, sound=False, **kwargs):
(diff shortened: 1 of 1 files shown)


peircej:
Hi Jessica, welcome to the team!. 
@jeremygray do you want to take a quick look since it's your code being extended? I can't see any issues (unless users have been putting args in using order rather than using keyword) but let me know if you want it pulled.
j

jeremygray:
Welcome, Jessica! Nice to have TA for auditory gap pulse sequences. Perhaps just add a note to the changelog about the order of named arguments changing?

I agree that putting TA right after TR makes sense conceptually. Yes, it might trip someone who has been using positional args rather than kwargs, but that is maybe not so likely and hopefully easily fixed on their part. So sure, lets go for it, with a note to the changelog. 

emulators.launchScan is fine: `syncPulse = SyncGenerator(**settings)`. I suppose the ability to allow that form of usage is what motivates having kwargs as well as TA (easy / safe for dicts with lots of stuff).

PR #874, opened 2015-03-11
TESTS: travis tweaks

jeremygray:
- less logging
- better set-up (update pip, use sudo -H, pyglet 1.2.2)


peircej:
I have a slight worry that if we turn off auto-logging in too many places it won't get tested very thoroughly? But maybe we aren't there yet. e.g. could make sure that at least one of the window contexts keeps logging on?

jeremygray:
Yeah, good point. We could have an explicit test for logging to test the overall approach.

PR #878, opened 2015-03-16
BF: Actually use ‘skip’ setting in launchScan

kastman:
Previously skip was used for the sync generator keypress thread, but not when waiting to launch.

@jeremygray Would you take a look at this one when you get a chance? I've been testing it with this basic compiled builder script: https://gist.github.com/kastman/fa1adbaf95da1d65c175  Something is bonked with my current install (pyglet 1.2alpha from conda not playing nicely?) so it may not work, but I think the logic is ok.

Although you can pass in a 'skip' and the doc sounds like it should wait, in practice the waiting is ignoring skip and going on the first key received. This will count the number of received TRs to wait to go, in line with the doc's intention.

### DIFF
--- a/psychopy/hardware/emulator.py
+++ b/psychopy/hardware/emulator.py
@@ -260,7 +260,7 @@ def launchScan(win, settings, globalClock=None, simResponses=None,
     timeoutClock = core.Clock() # zeroed now
     allKeys = []
     receivedTRs = 0
-    while not settings['sync'] in allKeys and receivedTRs > settings['skip']:
+    while not (settings['sync'] in allKeys and receivedTRs > settings['skip']):
         allKeys = event.getKeys()
         if esc_key and esc_key in allKeys:  # pragma: no cover
             core.quit()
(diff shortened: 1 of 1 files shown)


jeremygray:
Erik, thanks for the nudge. I think there's more work to do, maybe the feature just needs a better spec and documentation. With the skip feature, I was aiming to allow people to recreate what happens during a scan but without having to be at the scanner. In my limited experience, doing "skip" or "disdaqs" or "dummy" scans means getting ~10 seconds-worth of EPI volumes, but without receiving a sync pulse from the scanner during that time. The images are not collected or saved. This is what I've seen at 2 MR centers, and of course other places could do things differently. But to emulate what I've seen, there should be scanner noise without sync keys being received. The documentation (which I wrote) is definitely sketchy on this point, http://www.psychopy.org/api/hardware/emulator.html  I can see where I alluded to it, but you might have to know what I was thinking in order to know what I was thinking. So if you think the feature should work a different way, we'll need to differentiate the different use cases.

Also, fyi, tests on travis-ci are now being run with pyglet 1.2.2 (no alpha). I'll take a look at the report and see if that might have contributed.

kastman:
Hey, I actually think you're right - the way I've seen discdacqs is the same as what you describe, where no pulse is sent.

I was helping a friend with a non-standard configuration and got caught up in half-reading the documentation - this should definitely be a doc fix instead of a code fix.

Looks like the Travis Build failed because we didn't catch a Timeout Error. I'm closing this one... Thanks Jeremy!

PR #891, opened 2015-04-22
NF: autosave by default at end of loops

jeremygray:
- intended to mitigate data loss in case of abnormal program exit
- easy to disable (uncheck the box in the loop dialog)
- saves all data (not just the loop's data like the `IsTrials` option will do)
- using `<esc>` to quit experiment will now try to save data first
- changed file-overwrite to use level logging.data instead of logging.warning
  (if you explicitly request overwrite, that's what you want, don't want a warning)

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2841,7 +2841,7 @@ class DlgLoopProperties(_BaseParamsDlg):
         panelSizer = wx.GridBagSizer(5,5)
         panel.SetSizer(panelSizer)
         row=0
-        for fieldName in ['name','loopType','isTrials']:
+        for fieldName in ['name','loopType','isTrials','autosave']:
             try:
                 label = self.currentHandler.params[fieldName].label
             except:
--- a/psychopy/app/builder/experiment.py
+++ b/psychopy/app/builder/experiment.py
@@ -35,7 +35,9 @@ _localized = {
         'N up': _translate('N up'), 'max value': _translate('max value'), 'N down': _translate('N down'),
         'step type': _translate('step type'), 'step sizes': _translate('step sizes'),
     # strings for interleaved Staircases
-        'stairType': _translate('stairType'), 'switchMethod': _translate('switchMethod')
+        'stairType': _translate('stairType'), 'switchMethod': _translate('switchMethod'),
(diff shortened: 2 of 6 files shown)


peircej:
The code you've added here is manually saving the data at the end of the run or on  manual quit, but ExperimentHandler should already be saving data much more safely than this using the __del__ method (including when a script error is encountered). In what cases have you found this not to work?

jeremygray:
I saw your other post too. Some elaboration:

The new version retains the very safe saving upon quit or script error,
using `__del__` in exactly the same way. So no functionality is lost. It
just makes it even safer, and even easier to use.

Ease of use: Currently, calling `__del__` is the only way to do a proper
save of all data as wide text (its possible to save partial data via the
IsTrials feature). This is counter-intuitive, enough so that in the
test_builder tests, the way to trigger a save was to del the experiment.
Calling `_del__` in order to trigger the side-effect of saving the data is
suboptimal IMHO, and there was a comment to explain what was going on in
that test. In the refactored code, del calls a function saveData(). This
refactoring makes it possible to call saveData at any time, and do so with
a self-explanatory name -- both of which are useful. Saving all data into a
single, wide-text file is useful for doing scoring as part of the same
script, e.g., after several loops but before the script quits. I had to
hack around a bit to get the scoring to work as part of the IAT demo and
the mental rotation demo. I wanted to display scores to subjects for mental
rotation (for classroom use), and to save into the data file for IAT (to
avoid needing to run a separate script later).

Safety: The new code makes it possible and easy (but not required) to do
the same kind of autosave that `__del__` does, but also do so at the end of
every loop. The motivation for this, and for doing it by default, is that
there are cases where people have lost data due to an abnormal exit that
does not call `__del__` (see recent post on the users list). The long delay
can be from .xlsx, but people reported it in other cases too, like using
the microphone to record long sounds and then stopping the recording--the
script would not end for a remarkably long time. So there are at least two
known ways that a script might not exit as expected, and could be more that
we don't know about. Currently, the user is just out of luck, all data are
lost for that session. Even if they are warned that hitting the red-cancel
button will lose data, they are still stuck. With the refactored code, only
data since the last call to saveData are lost. With the added flexibility,
that lost data can become none at all in many cases.

PR #893, opened 2015-04-27
NF: allow ratingscale display to support min:sec

jeremygray:
- changes the text displayed during run-time
- value from .getRating() remains as a proportion


jeremygray:
tests are passing. (just a glitch first time around? odd.)

PR #894, opened 2015-04-28
BF: ioHub Linux Mouse events & Closing windows in core.quit()

isolver:
- fixes bug in ioHub Linux Mouse events not being reported
- core.quit() now calls close on any open windows so that it does not need to be done from ___del___ call, which often fails to complete fully.


peircej:
Sol, this one seems to be failing
https://travis-ci.org/psychopy/psychopy/builds/60404409
with the error

>           self.winHandle.dispatch_events()
psychopy/visual/window.py:583: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
        while xlib.XCheckWindowEvent(_x_display, _window,
>                                    0x1ffffff, byref(e)):
E   ArgumentError: argument 2: <type 'exceptions.TypeError'>: wrong type

It's weird because flip must have been called in other tests but failed on this one (test_bitsShaders). I don't understand it. Any ideas at your end?

isolver:
I have no idea why a call to self.winHandle.dispatch_events() from within
the psychopy window class would suddenly be failing for this test because
of the PR I made. The pull request does have a change in core.py so that
windows are closed when core.quit() is called:

https://github.com/isolver/psychopy/commit/50e039119c424d49f080c5d489a6c860c30024a3

I can't see how that could be causing this exception, but I could try
reverting it so that the change to core.py is removed. Should I give that a
shot?

BTW, when I run the test on windows it runs passed the .flip() ok and only
fails at one of the assert points that checks the array adat, which I guess
is because I do not have the needed hardware?

isolver:
K, now I'm really confused. Why is a test failing for me that I have made 0 related lines of code changes to, but passing for others? Help! ;)

isolver:
Also, why is 'continuous-integration/travis-ci/pr ' passing now OK, but the pull is failing because of 'coverage/coveralls - Coverage decreased (-0.0%) to 50.55%'.

Does this mean it is actually passing now?

peircej:
not sure! I'm merging the commit and we'll see what happens next! it could be that not all the slaves of the testing system are identical?

PR #898, opened 2015-05-06
Function to set mouse exclusivity for a window in Pyglet

suddhasourav:
This function binds the mouse to a window (and makes the cursor invisible). The use-case would be a multi-monitor setting where the participants need to click to respond. Without using a function like this, the cursor can be moved out of the region of the experiment.


suddhasourav:
I wonder why coveralls is throwing an error here... should I add a test function?

peircej:
We do get some non-deterministic fails on occasion, but this looks fine. Very minor thing: I wonder if the name might be just:
  mouse.setExclusive(True)
rather than:
  mouse.setMouseExclusivity(True)
?

peircej:
(if you push a change and submit a pull request this one will be updated - no need to close and reopen it)

suddhasourav:
I agree - it's already a method of Mouse. Changed accordingly!

PR #900, opened 2015-05-07
DOCS: updated GestaltReVision PsychoPy course link

qbilius:
Now points to the recent and elaborate version of the PsychoPy crash course.


peircej:
It may take a few days before I get around to rebuilding the website from the source, but this will get wrapped in sometime. 
best wishes, Jon

Issue #909, opened 2015-06-01
Builder loop dialog "Conditions" field truncated contents don't represent actual path


m-macaskill:
In the Builder loop dialog box's "Conditions" field, if the path to the conditions file is long, then the text representation of the path is truncated, with some of the middle content replaced by an ellipsis. Presumably this is so the actual file name is always visible. The problem with this is that the displayed text content of the field does not match the actual stored full path. This causes a number of problems, such as not being able to debug issues with the file not being able to be found (as the details of the path can no longer be read), and not being able to copy and paste a path from one dialog to another. The text representation of the path should match the actual path, even if that means that the file name is not visible without scrolling through the field.

peircej:
Yes, I once thought this was a good idea, but alas it was not. We should ditch the whole abbreviation thing there

PR #913, opened 2015-06-03
ENH: Stop abbreviating condition file path as displayed in loop dialog

m-macaskill:
Addresses this issue:
https://github.com/psychopy/psychopy/issues/909

i.e. Builder loop dialogs should now show the full (absolute or relative) path to a conditions file, rather than displaying an abbreviated version (which was created to ensure the filename would always be visible). i.e. now when a path is long, one might have to move the insertion point through the field to see the filename, but the text representation of the path will match the actual path.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2031,8 +2031,6 @@ class ParamCtrls:
         else:
             #create the full set of ctrls
             val = unicode(param.val)
-            if fieldName == 'conditionsFile':
-                val = getAbbrev(val)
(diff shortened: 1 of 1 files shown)


m-macaskill:
The getAbbrev() function (about line 4856 of builder.py) could probably also be deleted: I just left it there in case it was called elsewhere, but suspect that it isn't.

psychopy/app/builder/builder.py
```diff
@@ -4846,8 +4844,11 @@ def toggleVisible(self, evt=None):
         else:
             self.Show()
 def getAbbrev(longStr, n=30):
```
m-macaskill: Probably should have deleted this function as well.

PR #922, opened 2015-06-17
ENH: Add a parameter to invert an Aperture

thomastweets:
Hello!

I needed a simple means to mask the shape of the Aperture and not everything BUT the shape. This should work (if the changed GL command does not have some terrible side effect that I am not aware of).

Thanks for your great work!


thomastweets:
I just realized that this will probably break .contains().
I still like the easiness of the implementation. Maybe one could throw a warning in this case?

peircej:
Yes, contains() won't work but here are other conditions where the use of a bitmap mask means that doesn't work and I don't think there's anything we can do about that.

I can't easily pull in your current changes though because they clash with your changes from the other branch. I think they're easy enough to fix but it might be better if you did it. Do you know how to merge them with a merge tool?

thomastweets:
I now merged the latest commit from my feature-aperture-image branch to this PR/branch. So #925 is obsolete and I will close it. Still I have problems with the tests I implemented. Maybe you have a suggestion here?

thomastweets:
Great, thanks!

PR #924, opened 2015-06-18
ENH: supporting Portable VLC Media Player

hsogo:
I am trying to install PsychoPy on PortablePython (http://portablepython.com/) for my student who want to learn PsychoPy using a Windows PC in a University computer room. My student doesn't have privileges to install software on them.  Most of demos in psychopy/demos/coder/stimuli work fine simply copying site-packages directory of the StandAlone PsychoPy to Portable Python, but MovieStim2.py doesn't work because VLC Media Player is not installed on the PC.

In order to run MovieStim2.py without installing VLC Media Player to the PC, I installed Porable VLC Media Player (http://portableapps.com/apps/music_video/vlc_portable) and Portable Python together on an USB memory. Then I modified lines 139-140 of psychopy/visual/vlc.py to search for Porable VLC Media Player if VLC Media Player is not found in the PC. I tested this code on several Windows PCs. It worked fine.

Because these lines are executed only when platform is Windows and VLC Media Player is not found in "Program Files" directory, I don't think that this code affect normal behavior of vlc.py.  Please consider to merge this code.

### DIFF
--- a/psychopy/visual/vlc.py
+++ b/psychopy/visual/vlc.py
@@ -136,8 +136,16 @@ def find_lib():
                 dll = ctypes.CDLL('libvlc.dll')
                  # restore cwd after dll has been loaded
                 os.chdir(p)
-            else:  # may fail
-                dll = ctypes.CDLL('libvlc.dll')
+            else:
+                try: # try PortableVLC
+                    drive, path = os.path.splitdrive(os.path.abspath(os.path.dirname(__file__)))
+                    plugin_path = drive+'\\VLCPortable\\App\\vlc'
+                    # libvlccore.dll must be loaded before loading libvlc.dll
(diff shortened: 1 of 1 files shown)


peircej:
Sounds good. By the way, I'm intending to release PsychoPy using 64bit python this summer (on windows I may need to release it with 32 bit version as well) and then the vlc install will need to match the install of psychopy. Not sure how that will affect portable apps

hsogo:
Thank you for letting me know. Currently 64bit Portable Python doesn't seem to be developed. I might have to go to WinPython.

peircej:
Oh, wait, now I think about it I think it won't make any difference to you: if you use protable python it means you aren't using standalone, and that means you aren't affected. It is only users of Standalone that will be affected

hsogo:
I see. Thank you again!

PR #925, opened 2015-06-18
BF/TEST: Fix scaling of ImageStim Aperture and add tests

thomastweets:
I have added a simple test together with a scaled down examle PNG. However, I was not able to run the tests. If I understand correctly I would have to do that from within the GUI, correct?

When I run it from command line it gives me
```bash
INTERNALERROR> SystemExit: This program needs access to the screen.
INTERNALERROR> Please run with a Framework build of python, and only when you are
INTERNALERROR> logged in on the main display of your Mac.
```

I do not understand how to run the psychopy GUI from my checked-out repository.

More than that I changed the scaling of the window during ._reset() to scale ImageStim-type Apertures correctly. Also I disabled the alpha test after the .draw() in ._reset().

As with #922 I realized that the new ImageStim-type Aperture breaks .contains(). How should I solve this?


thomastweets:
This PR is now merged with #922

peircej:
By the way, when you said this 'breaks' contains() did you mean that it breaks it for shape apertures or just that contains wouldn't really work for bitmaps. As I say, the latter isn't an issue in my eyes, but the former would be

thomastweets:
I meant that the .contains() function would not work for bitmaps i.e. that the 'new' ImageStim-type Aperture would not support it. However, I do not see a reason why the 'normal' ShapeStim-type Aperture would not support it anymore. That should still be fine.

PR #926, opened 2015-06-20
ENH: Better handling of Unicode characters in the Preference Dialog

hsogo:
Following problems are fixed.
Please see also https://groups.google.com/forum/#!topic/psychopy-dev/bB_GdHuNjes

* Preference Dialog is not opened when Unicode character is included in userPrefs.cfg (this problem occurs when Unicode character is included in string-type properties).

* Unicode characters are shown as hexadecimal values if they are included in list-type properties.

* It is hard for beginners to notice error message "Output Pane" of the Coder View when ungrammatical string is input to list-type properties in the Preference Dialog.

* Add 'u' to "Data filename $" in the Experiment Settings dialog in case Unicode characters are input to 'set data folder' in the Preference Dialog.


hsogo:
Oh, I didn't know that test fails if default value of "filename" argumet of SettingsComponent.
Error message should be shown if Unicode characters are found in "saved data folder" in the Preference Dialog, rather than allowing Unicode characters?

hsogo:
I found several problems caused by non-ASCII file/directory name. I will close this pull request once and fix those problems.

PR #933, opened 2015-06-30
Use pandas to handle writing in TrialHandler.saveAsWideText()

onesandzeroes:
This is a work in progress, as obviously making changes to any data output stuff is risky and it'd be a good idea to test it as much as possible.

The goal is to let pandas handle the actual file writing in `TrialHandler.saveAsWideText()`, as the existing manual construction is not particularly good at dealing with quoting and escaping. For example, in one of my current experiments I had tuples in one of my output fields, and trying to write a csv with psychopy's builtin methods meant that field was split because the delimiter was not quoted/escaped. pandas is very good at dealing with these kinds of cases.

It seems like there's no real problem with sticking to the existing API, as custom delimiters and encodings are easily passed on to pandas.

That said, I have made one API change: both `saveAsWideText()` and the new `dataToDataFrame()` raise a `ValueError` if there are no trials to save, rather than the existing behaviour of returning -1, as I think it's weird for the function to return a DataFrame most of the time but return an error code in rare cases.

Happy to hear any feedback or suggestions on this.

Quick test script adapted from the `TrialHandler` demo:

```python
#!/usr/bin/env python2
import random
from psychopy import data

random_seed = 333

random.seed(random_seed)

#create stimuli
stimList = []
for ori in range(90,180,30):
    for sf in [0.5, 1.0, 2.0]:
        stimList.append(
            {'sf':sf, 'ori':ori} #this is a python 'dictionary'
            )

#organise them with the trial handler
trials = data.TrialHandler(stimList, nReps=10, extraInfo= {'participant':"Nobody",'session':1})
trials.data.addDataType('choice')#this will help store things with the stimuli
trials.data.addDataType('RT')#add as many types as you like

#run the experiment
nDone=0
for thisTrial in trials: #handler can act like a for loop
    #simulate some data
    thisReactionTime = random.random()+float(thisTrial['sf'])/2.0
    thisChoice = round(random.random())
    trials.data.add('RT', thisReactionTime) #add the data to our set
    trials.data.add('choice', thisChoice)
    nDone += 1  #just for a quick reference

df = trials.saveAsWideText("testDataWide_{s}.csv".format(s=random_seed))
```

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -1231,7 +1231,7 @@ class TrialHandler(_BaseTrialHandler):
         """
         if self.thisTrialN<1 and self.thisRepN<1:#if both are <1 we haven't started
             logging.info('TrialHandler.saveAsWideText called but no trials completed. Nothing saved')
-            return -1
+            raise ValueError("No trials completed, no data to save")
 
         #set default delimiter if none given
         if delim is None:
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -1172,7 +1172,10 @@ class TrialHandler(_BaseTrialHandler):
                 # Add a trial number so the original order of the data can always be recovered if sorted during analysis:
                 trialCount += 1
 
-                entry = self.extraInfo.copy()
+                if self.extraInfo is None:
+                    entry = {}
+                else:
+                    entry = self.extraInfo.copy()
                 for dataType in self.data.dataTypes:
                     entry[dataType] = self.data[dataType][trialTypeIndex][repThisType]
                 entry.update(self.trialList[trialTypeIndex])
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -1132,6 +1132,9 @@ class TrialHandler(_BaseTrialHandler):
         return dataOut, dataAnal, dataHead
 
     def dataToDataFrame(self):
+        """
+        Return the trial data as a pandas DataFrame object for writing/analysis.
+        """
         # If both are < 1 we haven't started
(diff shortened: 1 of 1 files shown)


jeremygray:
I really like this direction. Using pandas for this is very good in general, and being able to get a dataframe of all data so far during an experiment would be very handy in a number of instances. Separating the logic for saving to a file from the logic for conversion to a specific format is also good. I do not think that it makes sense to raise in the case of no data--returning None or an empty dataframe seems better to me. Why should it be an illegal operation? Either with raise or None, you have to take into account that there might be no data.

You might want to just focus on the code for dataToDataFrame(), and not also rewrite saveAsWideText() to use that, at least not yet, not until it has been tried in the real world more. Or you could have a saveAsDataFrame() function for saving to a file. The point is to lessen the risk of introducing a bug that would lead to people losing data.

Finally, it would also be good to have thorough tests for such code. If the dataframe approach can handle everything that the current code does and also handle somethings that are not currently handled properly, it would be a good reason to switch to the new code.

hoechenberger:
Nice job so far, @onesandzeroes! Are you going to continue working on this? :)

peircej:
Hi there. Sorry I didn't pull this in earlier, but the reason was that I had in mind to carry out a more extensive re-write of TrialHandler. See discussion of that on the dev list:
  https://groups.google.com/forum/#!topic/psychopy-dev/GD5bKQYOqg8
and in the relevant commit 533b97482

peircej:
FYI I'm going to close this pull request as I think the new class supercedes it

PR #942, opened 2015-08-05
ENH: many print statements going towards python3 support

peircej:
also made a few lib changes.
Shouldn't break python2 support so far


jeremygray:
could migrate xrange -> range as well for python3. Most instances of xrange seem to involve small values, like 5 or 200, so the impact on memory usage is probably low. Some I can't tell.

peircej:
There are another 500 or so print statements in there (mostly in demos at this point). Also, many of our except... statements are broken for Python3 at the moment (it turns out some were not really working in python2 either but in python3 they get detected as a syntax error)

PR #958, opened 2015-09-21
ENH: Add context manager to visual.Window

effigies:
Allow for a with statement that implicitly closes a `Window` at the end of the block, e.g.

    with visual.Window(...) as w:
        [Perform experiment]

    [w is now closed]

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -383,6 +383,13 @@ class Window(object):
         if self._closed!=False:
             self.close()
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if not self._closed:
+            self.close()
+
     def __str__(self):
         className = 'Window'
         paramStrings = []


psychopy/visual/window.py
```diff
@@ -383,6 +383,13 @@ def __del__(self):
         if self._closed!=False:
             self.close()
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if self._closed != False:
```
peircej: Is that logic the correct way around (confusing double negatives). Close the window if it closed==False. If closed!=False then closed==True and in that case we shouldn't close again, right?

effigies: Good spot. I was actually just copying `__del__`, which I was assuming would need a similar cleanup to `__exit__`. So I think that probably needs fixing, too. Is that okay to put in the same PR?

peircej:
Cool, thanks. I looked into why that code was wrong for Window.__del__ and it turns out I had recently changed it myself. I /thought/ I had written it incorrectly previously and when I changed it a window delete error went away 55fd855 but it turns out that it was going away for the wrong reason. Now fixed correctly here: 3250c28d1c

Double negatives are confusing!

PR #959, opened 2015-09-23
ENH: deprecate web.upload(); use requests.post() instead

jeremygray:


peircej:
You'd like this in the new release (1.83.00)?

jeremygray:
Well, yes if possible. But not if it means a huge hassle. If you rebuild
the StandAlone packages for some other reason, perhaps this could go in
too. Just seems good to start the deprecation clock ticking.

peircej:
There will be a rebuild to do anyway. Just checking whether to merge before or after that. cheers

PR #963, opened 2015-09-30
several small fixes and doc

jeremygray:
an explicitly requested duration for a tone was being ignored, leading to default duration 0.5s
being used, which made 9 or 10 audible clicks during a 5s tone.

(see users list query 9/29/2015)


jeremygray:
The failing travis-ci check looks quirky and not related to the pull request.

jeremygray:
I am not sure this is the 100% right fix. It works in the case I was trying to fix but the -1 infinite duration logic is complicated.

PR #964, opened 2015-09-30
BF: Fix errors, links and a typo in HTML generated by config wizard

philippwiesemann:
The HTML generated by the configuration wizard contained errors (but less strict web browsers tend to ignore them).

Additionally, three links were outdated (the anchors) because the linked documentation was changed some time ago.

And a word was missing in the output and the related unit tests.

Thanks!


jeremygray:
These changes look good -- apparently I have a less-strict browser.

Issue #486, opened 2014-01-22
Python version


danielshub:
Most of the scripts start with `#!/usr/bin/env python` but face.py and plaid.py do not. I think these should be changed. Further, PEP 394 (http://www.python.org/dev/peps/pep-0394/) recommends that "python should be used in the shebang line only for scripts that are source compatible with both Python 2 and 3" since PsychoPy is not compatible with Python 3, the shebang line should use python2. The current shebang lines cause issues with Arch Linux where python refers to python3.

PR #974, opened 2015-10-13
BF: Changed shebang lines from "/usr/bin/env python" to "/usr/bin/env…

danielshub:
… python2". See https://github.com/psychopy/psychopy/issues/486

### DIFF
--- a/psychopy/app/Resources/psychopy.xml
+++ b/psychopy/app/Resources/psychopy.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
+    <mime-type type="text/x-psyexp">
+        <comment>psychopy file</comment>
+        <icon name="psychopy"/>
+        <glob-deleteall/>
+        <glob pattern="*.psyexp"/>
+    </mime-type>
+</mime-info>
--- a/psychopy/demos/coder/hardware/CRS_BitsBox.py
+++ b/psychopy/demos/coder/hardware/CRS_BitsBox.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 #coding=utf-8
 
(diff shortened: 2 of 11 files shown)


jeremygray:
@danielshub @yarikoptic To support python3 (#1061), I think we'll want to move away from hard coding `python2` at the top of every file. What do you think about this SO solution https://stackoverflow.com/questions/18993438/shebang-env-preferred-python-version 

My interpretation is that PsychoPy would only have `#!/usr/bin/env python` at the top. For packaging, people who need it would have scripts to replace that line with these 4 lines (see SO for elaboration):

    #!/bin/sh
    ''''which python2 >/dev/null 2>&1 && exec python2 "$0" "$@" # '''
    ''''which python  >/dev/null 2>&1 && exec python  "$0" "$@" # '''
    ''''exec echo "Error: I can't find python anywhere"         # '''

This seems safe to do as a replacement, and easy to do from a script. I have not tested it out at all, but it looks sensible.

yarikoptic:
using `/usr/bin/env python` sounds sane... for packaged versions we would indeed replace shebang with /usr/bin/python-whatever-is-the-default-to-be-used.

this shell contraption expanded my mental horizons tremendously, thanks for sharing ! ;)

jeremygray:
its a pretty slick little bit of doublespeak :-)

danielshub:
I think we should follow the recommendations of PEP 394 (http://www.python.org/dev/peps/pep-0394/) and only change the shebang line when we change which version of Python is supported. I am not really sure what the plans are for using PsychoPy with Python 3. If PsychoPy at some point supports both Python2 and Python 3, then we should change the shebang lines to be "python" at that time. If at some point PsychoPy only can be used with Python 3, then the shebang should be changed "python3" at that point.

jeremygray:
Thanks both. Daniel I agree, sounds exactly right based on PEP 394. There is has been some movement to support python3, we're not there yet. When we do actually support it, that the proper time to tweak the shebang (and not before).

PR #985, opened 2015-10-21
Restore Pillow support

hoechenberger:
The Builder would not start if the most recent Pillow version (3.0) was installed in my Python environment. Turned out that the method `Image.tostring()` was removed from Pillow, and `Image.tobytes()` should be used instead.

This PR restores support for recent Pillow versions while maintaining PIL compatibility (e.g. for standalone).

### DIFF
--- a/psychopy/app/builder/components/__init__.py
+++ b/psychopy/app/builder/components/__init__.py
@@ -6,10 +6,7 @@
 
 import os, glob, copy
 import wx
-try:
-    from PIL import Image
-except ImportError:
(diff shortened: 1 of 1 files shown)


peircej:
I've pulled this one but I'd generally avoid removing things like the try..except on the import. There's no pressing need to remove it for performance and it doesn't hurt for some people to use (very) old PIL versions

hoechenberger:
Thanks for your suggestion, will try to adhere to it in the future!

PR #987, opened 2015-10-23
workaround to reduce memory leak in movie3

jeremygray:
Doing this PR for discussion purposes. Probably fine to actually pull, as better-than-nothing ( = a workaround rather than proper fix).

- workaround: give sound.SoundPyo() a tmp file, rather than np.array
- not ideal but seems to work fine with small videos, not tested with large ones
- MovieStim3 still leaks a little bit

### DIFF
--- a/psychopy/visual/movie3.py
+++ b/psychopy/visual/movie3.py
@@ -30,12 +30,13 @@ reportNDroppedFrames = 10
 import os
 
 from psychopy import logging
-from psychopy import sound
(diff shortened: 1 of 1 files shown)


jeremygray:
moviepy appears to have specific audio classes -- maybe we can use them instead of a psychopy.sound.Sound?

peircej:
Oh, but we should definitely still try to track down the (albeit small) memory leak.

PR #989, opened 2015-10-27
TESTS / WIP: proof of concept for memory leakage testing

jeremygray:
- useful from the command line
- too unstable at present for travis-ci, so use a filename that pytest will not auto-detect
- some things do not pass, so skip them for now (placeholder tests)
- need pyo compiled with patch for leak (pyo commit, Oct 26 2015) or sound will fail


jeremygray:
refactoring a bit...

jeremygray:
refactored

PR #990, opened 2015-10-29
TESTS: memory leak tests

jeremygray:
- many tests fail at present; not suitable for travis-ci, use command line
- tries to cover all classes used during experiments (Window, visual stim, sounds, mouse), might have missed a couple
- only creates instances of each class, not use them in any way. (Usage could possible result in mem leaks not detected here.)
- try to do twice: with minimal parameters, and with more resource-intensive params

### DIFF
--- a/psychopy/tests/test_misc/memory_usage.py
+++ b/psychopy/tests/test_misc/memory_usage.py
@@ -1,4 +1,3 @@
-
 from psychopy import visual, event, info
 import pytest
 import numpy as np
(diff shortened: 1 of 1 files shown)


peircej:
I've just built the latest pyo (0,7,7) from github on my mac so hopefully the memory leak is better there now. Do you happen to know if that needs rebuilding on win32 or was it just a mac bug?

jeremygray:
>
> I've just built the latest pyo (0,7,7) from github on my mac so hopefully
> the memory leak is better there now.
>
Cool. Sounds should completely pass the memory usage test. I tested with a
rather large sound (200 sec at 44100Hz) and it was fine

> Do you happen to know if that needs rebuilding on win32 or was it just a
> mac bug?
>
I assume its not just mac -- the memory usage test should show this on
other platforms

Issue #936, opened 2015-07-05
ShapeStim should handle complex shapes


lindeloev:
Currently, ShapeStim and it's derivatives does not handle concavities and "holes" - what's called "complex shapes" in OpenGL speak. I myself tried to draw a star the other day using ShapeStim but failed. Holes may be more of a fringe case, I guess. It should be possible to do both by using a tesselation helper, see e.g. http://opengl.czweb.org/ch18/575-580.html. 

Have anyone tried implementing this for PsychoPy and shouldn't it be on the wish list?

peircej:
Yeah, definitely not one of the things I cared to fix anytime soon, given a couple of easy enough workarounds, but if you fancy a little challenge that's cool ;-)

lindeloev:
What are those workarounds? Combining multiple ShapeStims as mentioned in the doc or something more clever? I wanted to control line color and fill color for the star dynamically and separately. Ended up using bitmaps instead which felt very unelegant ;-)

peircej:
Yes, bitmaps are workaround number 2. ;-)

jeremygray:
I read on the internets (so must be true) that this python project has useful code for gl-based tesselation  https://pypi.python.org/pypi/glsvg/0.4. The file svg_path_builder.py in that project has lots of references to tesselation

jeremygray:
This seems even more specific: file tesselate.py (BSD licence) in svgbatch https://pypi.python.org/pypi/svgbatch/0.1.9

PR #992, opened 2015-11-05
ENH: contrib tesselation code for evaluation purposes, re #936

jeremygray:


jeremygray:
Shucks, this can sometimes work (as in the demo) but can sometimes be quite useless

screen shot 2015-11-06 at 7 00 04 am

peircej:
Is the failure due to the order of the points ('winding  rules')?

jeremygray:
Same result with GLU_TESS_WINDING_ODD as GLU_TESS_WINDING_NONZERO

jeremygray:
Ok, my current approach is just wrong wrong wrong. I was fooled because it happened to work for the first example I tried, and greatly simplifies making the new class. Drat.

It is possible to use the vertices returned by tesselate.tesselate() in triplets, to build a python list of multiple separate ShapeStim, each being just a triangle. Then TessStim.draw() is just a loop to call .draw() for each triangles. This means that the right information is in there, which is good to confirm:
screen shot 2015-11-06 at 11 59 54 am

But python lists are not the way to go; much better to something similar via OpenGL in a TessStim.draw() method.

peircej:
Look at the rendering for ElementArrayStim. It uses a list of quads, but 
the idea is the same. You provide a vertex array and tell it what those 
are. You can provide a vertex array of triangles instead. (Oh actually 
RadialStim might do that too. *Should* do that but might have been 
written before I knew about vertex arrays so might use a gl list which 
is less efficient I think).

PR #994, opened 2015-11-07
FF: ShapeStim2 with tesselation using GL

jeremygray:
ShapeStim2 appears to perform well, supports arbitrary fillable shapes including holes. PR includes a visual test and a contains() test.


peircej:
Looks nice Jeremy. It looks like this would be entirely compatible with original ShapeStim? Maybe we just fold it straight in rather than having a separate class? I'm guessing the only place it might have slightly weaker performance is in calculating the vertices; the draw() code looks like it will be as fast as the previous. :-)

jeremygray:
Thanks Jon. I think its fast and pretty close to being a full replacement, but not there yet. Aperture seems to work with Shape2, based on limited testing. Known Issues: 1) Based on limited testing, Shape2 does not handle lines (e.g., RatingScale line fails). The tesselation fails, which is not so surprising. It should be possible to detect and work around this somehow. 2) Drawing borders is not set up to handle multiple-loop stimuli (like a square with a hole = two loops, inner and outer that are not connected). But that could be fixed easily enough.

For me it take 0.3ms to tesselate 14 vertices, 2ms for 140 vertices, and 34ms for 1400 vertices. So looks linear within the range of reasonable values. This is only on initial set-up.

jeremygray:
Also 3) I'm not sure what happens with .contains() for multi-loop stimuli. Probably we could live without it.

peircej:
I think multi-loop stimuli are very much a corner case that we needn't support.

peircej:
But handling a line is obviously essential

jeremygray:
agreed for both. I'll see what I can cook up for lines.

PR #995, opened 2015-11-09
ENH: ShapeStim2 handles lines

jeremygray:
Close to being a drop-in replacement for ShapeStim.
Known backwards-incompatibilities (might be others too):
- no dynamic updating of individual vertices. 
  to see: convert coder demo shapes.py to use ShapeStim2.
- a fix: rewrite the Shape2.vertices property to call tesselate, etc; fast enough with few verts. might be worth doing that regardless, just to pull the setVertices logic out of __init__

### DIFF
--- a/psychopy/contrib/tesselate.py
+++ b/psychopy/contrib/tesselate.py
@@ -62,7 +62,8 @@ class TesselateError(Exception):
 
 tess = gluNewTess()
 gluTessNormal(tess, 0, 0, 1)
-gluTessProperty(tess, GLU_TESS_WINDING_RULE, GLU_TESS_WINDING_ODD)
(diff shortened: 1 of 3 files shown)


jeremygray:
test_all_visual tests pass for ShapeStim if I simply swap in ShapeStim2, so that's a good sign

jeremygray:
Aperture.contains() needs tweaking to work with ShapeStim2, likely easy

peircej:
Yes, if tesselate is called during init then it should also be called after a a change to vertices to allow them to be updated. Using the setter properties of other visual stimuli has been a good way to make that happen automatically (the vertices looks to the user like an attribute but in code it runs a little function).

PR #996, opened 2015-11-10
RF: migrate ShapeStim2 to ShapeStim; closes #934

### LINKED ISSUE #934: creating arrow stimuli
I saw the code in this link (below) that can draw arrows. I'm not the best with python, but is there any way to fill in the arrow stimulus with color (the square is filled in already, but applying the same code doesn't work entirely). Thanks

https://github.com/psychopy/psychopy/blob/master/psychopy/demos/coder/stimuli/shapes.py

~R

jeremygray:
This PR is for discussion on using the tesslation-based code as a replacement for ShapeStim

- try to maintain full backwards compatibility, old ShapeStim becomes BaseShapeStim
- Aperture, Polygon, and Rect use BaseShapeStim; assigning to self.vertices during their init() methods was causing issues (related to the attribSetters decorators for .vertices). Should work fine for these stim.
- dynamic individual vertices are supported; can be slow to assign if there are many vertices, due to re-tesselation. Cannot simply change values in a (mutable) list of vertices; need to explicitly assign that list to s.vertices (to trigger attribSetter code to recompute the tesslation).


peircej:
Looks great Jeremy. Love your work :-)

jeremygray:
Thanks! This one was very satisfying.

PR #998, opened 2015-11-13
DOC: ShapeStim-related comments, changelog, demo

jeremygray:


peircej:
Thanks for doing the changelog - it's going to be really helpful as more and more people add changes that I get some reminder of what happened leading to the release. I myself forget to update this until the release which isn't exactly efficient.

I'll actually change what you've written a bit to make it more aimed at a user rather than developer, but this will be a good start to jog my memory. :-)

PR #1004, opened 2015-11-19
UTIL: csv from psydat utility & demo

jeremygray:
Currently in the Coder > Tools menu. Maybe should be in Builder > Tools instead, but uses the Coder output window for printing file names.


peircej:
This is OK but I think the ideal solution is that when a user "opens" a psydat file in psychopy it brings up a new view, and one of the buttons in that view is "Export to csv" (another button will be "Extract experiment" and we can also gradually add analysis info/methods to the view)

jeremygray:
Interesting. So the idea is that, just like .psyexp file extension triggers "launch Builder", a .psydat extension would trigger "launch DataViewer", i.e., a GUI-based widget for working with .psydat files (inspect, export to whatever, etc). Analysis methods would be most useful if they can be scripted, and applied to multiple input files at once.

peircej:
Exactly. For now, just a little window with very basic info and a couple of buttons.

I have ideas about how the analysis side would work too, using a graphical interface but always maintaining the python code equivalent (a bit like the Builder does). We can use pandas to add filters (e.g. exclude incorrect), collapse across some variables (e.g. stim position) and then perform the analysis on another (eg. RT).
I absolutely want this to evolve very gradually, not suddenly try to create a behemoth do-it-all interface that overwhelms everyone and goes in the wrong direction. I also wonder whether this should be done in QT rather than WX. It probably doesn't need to interact with the rest of the app much (so doesn't *have* to be wx) and QT might be better widgets these days.

jeremygray:
Sounds great. Keeping track of the code-equivalent would be really good. An Ipython notebook might be a good tool for this. It might not require QT / WX at all, and be share-able, archivable, customizable, etc.

peircej:
The code could be sent to a notebook fine but the interface doesn't make sense to be a notebook. It would need to be drag/drop a bit like a spreadsheet view. Think like E-DataAid, but better, obviously! ;-)

Issue #1005, opened 2015-11-20
QUEST creates grid of counter-intuitive dimensions


hoechenberger:
## Motivation

Encouraged by [this discussion](https://groups.google.com/d/msg/psychopy-users/5AsGnvLsPpU/3MA0WJTJRF4J) on the users mailing list, I looked into the QUEST implementation in greater detail. I found that the current API and documentation are counter-intuitive, potentially leading to undesired _("wrong")_ results.

Please bear with me while I'm trying to demonstrate what causes my concerns.
## Problem Description

Currently, `data.QuestHandler.__init__()` has, among others, the following parameters:
`minVal`, `maxVal`, `grain`, and `range`.

They are documented as follows:

```
minVal: *None*, or a number
    The smallest legal value for the staircase, which can be used to prevent it
    reaching impossible contrast values, for instance.

maxVal: *None*, or a number
    The largest legal value for the staircase, which can be used to prevent it
    reaching impossible contrast values, for instance.

grain: *0.01* or a number
    The quantization of the internal table.

range: *None*, or a number
    The intensity difference between the largest and smallest intensity that the
    internal table can store. This interval will be centered on the initial guess
    tGuess. QUEST assumes that intensities outside of this range have zero prior
    probability (i.e., they are impossible).
```

At a first quick glance, this seems pretty straightforward, except for that weird `range` parameter. But since it's optional, the naïve user is just going to omit it for the time being.

However, the other three parameters appear to be useful.

What one should be able to reasonably assume is:
**If I supply a `minVal`, `maxVal`, and `grain` argument, some internal grid of the size `(maxVal - minVal) / grain + 1` should be created.**

(The `+ 1` accounts for the endpoint)

That means, one would assume that the resulting grid has a resolution equal to the `grain` parameter. But this is in fact usually **not** the case with the current QUEST implementation, and should, IMHO, be considered a bug.

The reason for this behavior, which I will demonstrate in the next paragraph, is in fact the dubious `range` parameter. Looking into `contrib.quest.QuestObject`, one [finds](https://github.com/psychopy/psychopy/blob/master/psychopy/contrib/quest.py#L113) that a certain parameter `dim` seems to play a role (unnecessary lines removed):

``` python
if range is None:
    dim = 500
else:
    dim=range/grain
    dim=2*math.ceil(dim/2.0) # round up to even integer
```

[...]

PR #1014, opened 2015-11-21
[WIP] ENH: Improve staircase test coverage, fix QUEST

### LINKED ISSUE #1006: QUEST intensity scaling should be removed
Currently, `data.QuestHandler.__init__()` takes the optional parameter `stepType`, which is later used to transform intensity values, as in:
```python
        startVal = self._intensity2scale(startVal)
        startValSd = self._intensity2scale(startValSd)
```
where `_intensity2scale()` is defined as:
```python
    def _intensity2scale(self, intensity):
        """returns the scaled intensity level based on value of self.stepType"""
        if self.stepType=='db':
            scaled_intensity = numpy.log10(intensity) * 20.0
        elif self.stepType=='log':
            scaled_intensity = numpy.log10(intensity)
        else:
            scaled_intensity = intensity
        return scaled_intensity
```
However, this interferes with the QUEST algorithm, which is implemented in `contrib.quest`, and already fully takes care of the stimulus selection. Jon also [mentioned](https://groups.google.com/d/msg/psychopy-users/5AsGnvLsPpU/wr1EPMnYJIYJ) something similar on the mailing list already.

The `stepType` argument should be removed. If a user wants to specify `startVal` and `startValSd` e.g. in terms of log intensities, they could easily do the calculation manually, i.e.,
```python
QuestHandler(np.log10(startVal), np.log10(startValSd), ...)
```
QUEST will take care of the rest.

Also, currently `stopInterval`, `minVal` and `maxVal` are not transformed, so I assume the current implementation is incorrect for all but linear `stepType`s anyway.

### LINKED ISSUE #1007: Improve test coverage for staircase procedures
Current unit test coverage for the staircase procedures in `psychopy.data` is very low or even inexistent for some staircase types.

hoechenberger:
**THIS IS WORK-IN-PROGRESS, PLEASE READ COMMENTS BELOW AND DO NOT MERGE YET**

QUEST intensity scaling was removed. This is a backward-incompatible change, as it removes the `stepType` keyword argument from `QuestHandler.__init__()`. This closes #1006.

This PR increases the test coverage of all except the Psi staircase methods. Particularly, it
* adds a test for log steps in the `StairHandler`,
* adds a test for the `QuestHandler`,
* introduces a base class `_BaseTestStairHandler` that implements a `simulate()` method for simplified testing,
* completely reworks `makeBasicResponseCycles()` to create simulated response cycles of arbitrary lengths.

Closes #1007 for the time being.

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -2702,7 +2702,6 @@ class QuestHandler(StairHandler):
                  nTrials=None,
                  stopInterval=None,
                  method='quantile',
-                 stepType='log',
                  beta=3.5,
                  delta=0.01,
                  gamma=0.5,
@@ -2751,10 +2750,6 @@ class QuestHandler(StairHandler):
                 The method used to determine the next threshold to test. If you want to get a specific threshold
                 level at the end of your staircasing, please use the quantile, mean, and mode methods directly.
 
-            stepType: *'log'*, 'db', 'lin'
-                The type of steps that should be taken each time. 'db' and 'log' will transform your intensity levels
-                into decibels or log units and will move along the psychometric function with these values.
-
             beta: *3.5* or a number
                 Controls the steepness of the psychometric function.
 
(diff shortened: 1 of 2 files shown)


hoechenberger:
@peircej, if you agree, I would also like to push hoechenberger/psychopy@8deaa24 to this branch.

Eventually, my goal would be adding a solution for #1005 as well, before this PR actually gets pulled. What do you think?

hoechenberger:
I just went ahead and pushed this change (to close #1006). Will now start working on #1005. Please let me know if you don't agree with any of the changes, so I can rebase my branch to remove them.

hoechenberger:
Seems I broke the `MultiStairhandler` tests :) Working on this.

hoechenberger:
Need to fix #1015 first *sigh*

peircej:
This is all great work Richard. Thanks for getting stuck in to such a thorny piece of code! :-)

hoechenberger:
Thanks Jon, doing the best I can. Might take a little while though (1 to 2 weeks), got an urgent thing (or two :)) going on with my PhD project that requires for my attention first.

peircej:
oops, I forgot this was still listed as a work in progress and pulled it! (I'm trying to get the code all into one state so I can run some tests locally but possibly this was on e pull too many)

peircej:
OK I've reverted this pull so you'll need to re-issue the pull request when you're ready
cheers, Jon

hoechenberger:
Okay, thanks for letting me know! I should probably have canceled this PR as soon as I realized more issues needed to be fixed along with this. Would have avoided the confusion. I'll create a new PR as soon as I've fixed all the problems reported. :)

PR #1025, opened 2015-11-24
Update psi.py

JosephGlavan:
Fixed a bug with 2AFC. No longer using the d' based psychometric function. Everything is intensity-based now.


jeremygray:
Thanks for the contribution! Do you have some test-code that you used when developing this fix? It would be handy to include an explicit test as part of the PR. If you can do the Psi part (which I've never done), I can help with constructing a test that will get run automatically.

PR #1033, opened 2015-11-27
Improve staircase test coverage, remove QUEST intensity scaling

### LINKED ISSUE #1006: QUEST intensity scaling should be removed

### LINKED ISSUE #1007: Improve test coverage for staircase procedures

### LINKED ISSUE #1015: Staircases in `MultiStairHandler` execute, but never 'finish'
It appears that, although the staircases called by `MultiStairCase` are recording all responses, they never actually transition into a `finished` state.

```python
from __future__ import print_function
from psychopy.data import MultiStairHandler
import numpy as np

np.random.seed(1000)

if __name__ == '__main__':
    nTrials = 5  # This is going to yield 5 trials per staircase.

    conditions = [
        {'label': 'stim_01', 'startVal': 0.3, 'startValSd': 0.8,
         'minVal': 0, 'maxVal': 1, 'pThreshold': 0.5, 'gamma': 0.01,
         'delta': 0.01, 'grain': 0.01, 'nTrials': nTrials},
        {'label': 'stim_02', 'startVal': 0.3, 'startValSd': 0.8,
         'minVal': 0, 'maxVal': 1, 'pThreshold': 0.5, 'gamma': 0.01,
         'delta': 0.01, 'grain': 0.01, 'nTrials': nTrials},
        {'label': 'stim_03', 'startVal': 0.3, 'startValSd': 0.8,
         'minVal': 0, 'maxVal': 1, 'pThreshold': 0.5, 'gamma': 0.01,
         'delta': 0.01, 'grain': 0.01, 'nTrials': nTrials}
    ]

    stairs = MultiStairHandler(
            stairType='quest', method='random',
            conditions=conditions, nTrials=nTrials
    )

    responses = np.tile([1, 1, 1, 0, 0, 0], 3)[:15]

    for trial, intensity in enumerate(stairs):
        stairs.addResponse(responses[trialN])

    print(stairs.finished)
    print([s.finished for s in stairs.staircases])
    print([s.thisTrialN for s in stairs.staircases])
    print([len(s.data) for s in stairs.staircases])
```
Output:
```python
False
[False, False, True]
[4, 4, 4]
[5, 5, 5]
```

Expected Output:
```python
True
[True, True, True]
[4, 4, 4]
[5, 5, 5]
```

hoechenberger:
QUEST intensity scaling was removed. This is a backward-incompatible change, as it removes the `stepType` keyword argument from `QuestHandler.__init__()`. This closes #1006.

This PR increases the test coverage of all except the Psi staircase methods. Particularly, it
* adds a test for log steps in the `StairHandler`,
* adds a test for the `QuestHandler`,
* introduces a base class `_BaseTestStairHandler` that implements a `simulate()` method for simplified testing,
* completely reworks `makeBasicResponseCycles()` to create simulated response cycles of arbitrary lengths,
* adds tests for `MultiStairHandler`,
* fixes a problem in `QuestHandler` that lead to the staircases not 'finishing' when run through the `MultiStairHandler`. (Closes #1015)

We now also make use of the keyword arguments accepted by `contrib.quest.QuestObject.__init__()`. The parameters were passed via position before. This change is backward-compatible.

`data.StairHandler`'s and `data.QuestHandler`'s ``__init__()` methods now accept additional keyword arguments via `**kwargs`; these will then be discarded. This allows for a simplification of the instantiation of the staircase classes from within `MultiStairHandler`: We don't have to check the `conditions` parameters passed to `MultiStairHandler` for "illegal" keywords first before instantiating the respective staircases. This change is backward-compatible.

Closes #1007 for the time being.


hoechenberger:
I just spotted a small bug there, let me fix it before you pull.

hoechenberger:
Never mind, all is fine! :-)

PR #1034, opened 2015-11-27
PsychoPy'ify PsiHandler API, Make use of PsiObject keyword arguments in PsiHandler

hoechenberger:
`data.PsiHandler.__init__()`'s keyword argument `TwoAFC` does not match the naming schemes chosen in all other parts of PsychoPy. This PR renames the argument to `twoAFC`. This change is not backward-compatible, but given the novelty of the Psi method in PsychoPy, this should affect very few users, if any.

`contrib.psi.PsiObject.__init__()` takes several keyword arguments, which were only passed as positional arguments by `data.PsiHandler` so far. This PR changes `PsiHandler` such that it now passes the respective parameters as keywords. This change is fully backward-compatible.

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -3411,11 +3411,13 @@ class PsiHandler(StairHandler):
             except IOError:
                 logging.warning("The specified pickle file could not be read. Using a uniform prior instead.")
                 prior = None
+
         self._psi = PsiObject(
                 intensRange, alphaRange, betaRange, intensPrecision,
-                alphaPrecision, betaPrecision, delta, stepType,
-                twoAFC, prior
+                alphaPrecision, betaPrecision, delta=delta,
+                stepType=stepType, TwoAFC=twoAFC, prior=prior
         )
+
         self._psi.update(None)
 
     def addResponse(self, result, intensity=None):
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -3346,16 +3346,16 @@ class PsiHandler(StairHandler):
                  intensRange, alphaRange, betaRange,
                  intensPrecision, alphaPrecision, betaPrecision,
                  delta,
-                 extraInfo=None,
                  stepType='lin',
-                 twoAFC=False,
+                 expectedMin=0.5,
                  prior=None,
                  fromFile=False,
+                 extraInfo=None,
                  name=''):
         """
         Initializes the handler and creates an internal Psi Object for grid approximation.
 
-        Parameters:
+        :Parameters:
 
             nTrials (int)
                 The number of trials to run.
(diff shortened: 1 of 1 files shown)


peircej:
Hmm, technically, the difference between 2AFC and yes/no tasks is the expected chance level (and therefore the starting point of the fitted curved. In the curve-fitting routines I added to data this issue is handled by providing an `expectedMin` param. For TAFC (2AFC) this is 0.5 and for Y/N this is 0. The advantage of this is that it generalises: for 4AFC this would be 0.25

If we're going to break code (which is fine right now because nobody is using this) then lI suggest we switch to `expectedMin` as the arg where, right now, we only support an expectedMin of 0 or 0.5

hoechenberger:
Hello Jon, I had already been thinking about such a change: I considered removing the `TwoAFC` argument and introducing `nAFC` instead. The guessing rate could then be calculated automatically as `1/nAFC`, except for `nAFC == 1`, which would be a yes-no task. [Psignifit handles it that way.](http://psignifit.sourceforge.net/DESIGN.html)

However, support for designs other than yes-no and 2-AFC is not yet implemented in `contrib.psi`. Since I didn't want to introduce changes in that module (following my recent inquiry on the mailing list), I thought it would maybe be best to stick with the `twoAFC` argument.

Another approach could be to actually *do* switch to `expectedMin` or `nAFC`, but raise a `NotImplementedError` if an unimplemented design is specified.

My personal preference would definitely be the latter, and I'd highly favor the `nAFC` (or maybe call it `nafc`) approach, because in this case the user could specify their immediate design, and wouldn't have to do any calculations upfront. Also, the name `expectedMin` might be misleading: assuming lapses, the true "expected minimum" would be `guessing rate + lapsing rate`, and therefore be `>0` for yes-no and `>1/nAFC` for n-AFC tasks.

peircej:
I'm OK with us making small changes to the contrib files and just adding notes to them.

I'd still prefer to use expectedMin:
  - better to be consistent within our package than be inconsistent here in order to be consistent with psignifit
  - my preference for this in the first place was that the reason for it is to decide the minimum of a curve (and we could insert a max too). What determines that minimum (based on nAFC, lapse, bias etc) is up to the user

hoechenberger:
You got me convinced. :) Also, re-reading the first paragraph of the Psignifit docs I mentioned reminded me of the great difficulties I had in determining the type of experimental design when I first started using Psignifit:
> However, there are vast differences in naming conventions for experimental designs. The way psignifit approaches the number of alternatives is stimulus based. Note that a large number of labs use a response based nomenclature.

I'll remove the `twoAFC`/`TwoAFC` keywords altogether, and insert an `expectedMin` argument. Supported values will be, for the time being, `[0, 0.5]` (yes-no, 2-AFC). How should the method behave if the user specifies a different value for this parameter? `psychopy.logging.error`? Or `raise NotImplementedError`? This is also a general question I'd wanted to ask for a long time: When to log a warning, an error, and when to raise an exception. Are there some guidelines?

peircej:
In this case Raise NotImplementedError because in this instance I think psychopy can't know what to do next.

In general the question is whether a user would (ever) want to carry on e.g. if we drop a frame that's really bad for some people but others wouldn't want their experiment to abort half-way through on the basis of a dropped frame so I think we need to just log a message and hope the user reads it. I'd love to add a whole system of psychopy standardised errors where the user gets to decide how each is handled so the user can decide whether for them they care enough to raise error and abort. I started once and it's a big job. Never finished

hoechenberger:
Don't know if you receive updates with every push. At any rate: this branch is ready to be pulled. :)

PR #1039, opened 2015-11-28
RF: Add `_eval` and `_inverse` methods to `_baseFunctionFit`

hoechenberger:
The base class `data._baseFunctionFit` did refer to `self._eval()` and
`self._inverse()`, but these methods were thus far only implemented in
 the child classes.

This commit adds both methods to `_baseFunctionFit`, raising a
`NotImplementedError` if not properly overridden by a child class.


hoechenberger:
Landscapebot complains over even *decreased* health. Will look into this in more detail, closing for now!

PR #1044, opened 2015-11-29
RF: `StairHandler._checkArguments()` raises if arguments are invalid

hoechenberger:
`StairHandler._checkArguments()`  used to emit `logging.error`s in case  invalid or insufficient arguments were encountered. However, in these cases we cannot continue, so it's better to raise an exception.

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -20,6 +20,7 @@ import codecs
 import weakref
 import re
 import warnings
+import collections
 
 try:
     #import openpyxl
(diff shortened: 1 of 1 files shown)


peircej:
Switching to Raise instead of error is correct choice here.

In terms of formatting, I know you want the short lines to allow for old/narrow editor windows but I have to say I find the split multi-line code ugly. 
Also the use of breaking characters in the strings. Yuck! ;-)

hoechenberger:
Thanks for sharing your thoughts on this, Jon! I am also very happy to discuss about style ;)) Also, I would like to apologize for continuously stretching your patience with issues like this :)

Regarding the breaking characters: I definitely agree, this is hideously ugly. A better solution in [this particular case](https://github.com/hoechenberger/psychopy/commit/6b6a91ff5f2c438875d7785876adb9b7d5558f64#diff-0883de75eef0cd607cd87fd4c59f7ac8R3599) would have been to use `"` quotation marks, so the `'`s would not need to have been escaped. I don't know why I decided to to otherwise, hence cannot justify it anymore ;)

Re short lines (and I would like to limit this discussion to the particular case of `raise` statements for the time being), I agree that my current approach is disputable. Especially for nested loops, the remaining space (no. of characters per line) decreases quickly, enforcing many line breaks.

How about something like this:
```python
msg = 'This is a rather long exception message, split '
      'over several lines'
raise Exception(msg)
```
Or even:
```python
msg = """This is a rather long exception message, split
over several lines"""
raise Exception(msg)
```
What do you think?

hoechenberger:
Also, I will gladly discuss any such issues even before merging, so don't ever feel urged to merge a PR from me as long as you have concerns of any kind, including style! One thing I like about Github is that it makes discussions of this type very easy and intuitive, since one can even add comments to specific lines of code.

peircej:
The reality is I don't care very much. I guess I'm just having a poke that following a style guide hasn't made the code nicer in my view! But it's a childish poke on my part ;-) For quotes I personally switch to how I define the string according to whether I need single or double quotes inside it.

You're doing a lot of great work improving the experience for the user too, so thanks very much indeed on their behalf!

PR #1058, opened 2015-12-03
FF: use absolute import for visual.TextBox

jeremygray:
implicit was not working, so I followed the same style as the others

### DIFF
--- a/psychopy/tools/wizard.py
+++ b/psychopy/tools/wizard.py
@@ -284,6 +284,7 @@ class ConfigWizard(object):
         report.append(('no dropped frames', '%i / %i' % (nDropped, nTotal), msg, warn))
         win.recordFrameIntervals = False
 
+        """
         msg = _translate('for movies')
         warn = False
         try:
@@ -300,6 +301,7 @@ class ConfigWizard(object):
                 msg = _translate('Warning: version 5 recommended (for movies); Visit <a href="http://code.google.com/p/avbin">download page</a> [google.com]')
                 warn = True
             report.append(('pyglet avbin', unicode(ver), msg, warn))
+        """
 
         if verbose:
             report.append(('openGL max vertices', str(items['openGLmaxVerticesInVertexArray']), '', False))
@@ -461,7 +463,6 @@ class ConfigWizard(object):
         summary = [(check + _translate('video card drivers'), green)]
         ofInterest = ['python version', 'available memory', 'openGL version',
             'visual sync (refresh)', 'refresh stability (SD)', 'no dropped frames',
-            'pyglet avbin', 'microphone latency', 'speakers latency',
             'internet access']
         ofInterest.append('background processes')
         for item in ofInterest:
--- a/psychopy/visual/__init__.py
+++ b/psychopy/visual/__init__.py
@@ -12,15 +12,17 @@ from psychopy import logging
 from psychopy.constants import *
 
 # window, should always be loaded first
-from psychopy.visual.window import Window, getMsPerFrame, openWindows
+from .window import Window, getMsPerFrame, openWindows
 
 # non-private helpers
-from psychopy.visual.helpers import pointInPolygon, polygonsOverlap
+from .helpers import pointInPolygon, polygonsOverlap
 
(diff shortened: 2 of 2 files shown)


peircej:
I think explicit relative import is better than absolute ref. If the user has the lib (psychopy in this case) installed in multiple locations then I don't know which one will be imported from. With an explicit relative import I now that the one right next to me is being used.

peircej:
from .textbox import TextBox

jeremygray:
I have not heard that explicit relative is better than absolute, but have not looked into it much. So I looked just now, and I see what looks a lot like exactly the opposite, including in PEP-8: http://www.gossamer-threads.com/lists/python/dev/1072694 So I think python must keep track of what is going on, and do the right thing, possibly involving python path stuff.

If we do go with explicit relative (readability seems like the only benefit), then we should change all the other imports in the file too.

peircej:
I hadn't heard that as such. I've experienced it as an issue, mostly due to having a developer version and another that has been "installed" but happy to do more reading

jeremygray:
I read up a little more and am still confused. Its clear what PEP-8 says: prefer absolute. But it seems telling that of the three authors of PEP-8, one was asking "Why does PEP 8 advise against explicit relative imports?" and in the same thread GVR (another author) commented "a long time ago, when you couldn't tell a relative import from a top-level import, the ban on relative imports made more sense." So I am fine going with either one, seems a toss up.

jeremygray:
@toddrjen: Do you have any thoughts on the relative merits of explicit relative vs absolute imports?

toddrjen:
My opinion is "it depends".  I think the question needs to be asked on a case-by-case basis.  In my mind, the advantage of absolute imports is that it makes it very explicit where to get the module from, reducing the chance of collisions, while the advantage of explicit relative imports is that it makes it easier to move directories around.  

From these, my conclusion is that imports that are likely to be moved around together, such as a module providing a class and another providing a subclass, it would be better to use explicit relative.  On the other hand, modules that are less tightly linked, such as one providing general convenience or compatibility functions, should use absolute imports.

If we want to provide a consistent rule, perhaps we could say imports in the same directory, in the immediate parent directory (except the top-level directory), or in any subdirectory should use relative imports, anything else should use absolute imports.

jeremygray:
I like this mixed approach. So for example things in visual/ would use explicit relative imports for visual stuff, and use absolute for importing non-visual, like `psychopy.core` or `psychopy.misc`. builder.py could use explicit relative for builder-y things (like experiment) and app-y things (localization), but absolute for preferences, web, tools.*

jeremygray:
So as a dev guide: 
- Never use implicit relative.
- Use explicit relative of the form . or .. when importing things within (including below) or just above the current directory / module. Using ... may be needed in some instances, but do not use it to access the main top level `psychopy.*`.
- Use absolute of the form `from psychopy.XXX import ...` for conceptually "non-local" things, or things in the top level (`web`, `preferences`, etc)

peircej:
Yes, that's my preferred stance.

m-macaskill:
I'm guessing this fixed the issue with the 'textbox_simple.py' demo not working. But the other demo 'textstim_vs_textbox.py' continues to fail, specifically at line 84:

    fm=visual.textbox.getFontManager()
    AttributeError: 'module' object has no attribute 'textbox'

The simple demo doesn't call `.getFontManager()`

The discussion on flavours of imports above passes over my head a bit, so I don't have much to contribute on the issue as it affects this particular demo.

jeremygray:
This issue is due to lazy importing of visual.textbox  Apparently this line

fm=visual.textbox.getFontManager()

is being called before textbox has been imported.

>>> from psychopy import visual
>>> visual.textbox.getFontManager()
AttributeError: 'module' object has no attribute 'textbox'

>>> from psychopy.visual import textbox
>>> textbox.getFontManager()
<psychopy.visual.textbox.fontmanager.FontManager object at 0x10d79fa10>
>>>

--Jeremy

On Sun, Jan 24, 2016 at 5:32 PM, Michael MacAskill <notifications@github.com

PR #1063, opened 2015-12-07
RF better imports

jeremygray:
Summary:
In prep for python 3 and for good style: reduce implicit relative imports, wildcards imports, and deprecated imports; relocate a few files. I did this for the psychopy libraries and `app/`, but not `iohub/` or `contrib/`. 

Main changes:

1. Add `from __future__ import absolute_import` and tweak the import statements accordingly. Remove all implicit relative imports (might have missed some if they are not at the top of a file and only executed conditionally). All psychopy/ level imports should be absolute. Imports within major modules (such as app/ or visual/) can and should be explicit relative (as easier to read, shows that its "local", easier to move around as a group if needed some time later). Imports within lazyImport string seem to need to be absolute.

2. Replace wildcard imports with just the needed explicit imports. Several places had a reasonable use of a wildcard (such as in `__init__.py` files), so those were flagged as such by using a pylint message code (e.g., `# pylint: disable=W0401`).

3. Need more guidance on iohub in order to do similar changes. Some files are using explicit relative imports already, which is great. I have not checked which folders are "contrib"-like inclusions. There are quite a few wildcard imports in iohub/devices/*. Some of these look like 3rd-party code (labjack), so maybe just do `# pylint: disable=W0401` on them.

4. Move some files for tidier organization or thematic fit, and remove some long-deprecated files.

5. Limitations: not everything is changed over yet, notably in `ioHub/` and `contrib/`. Also, did not consider how the coder executes files, and what might need to change there.

Issues:
- test_ports: I could import crs in normal scripts, but could not do so from the test_port test either locally or on travis-ci. So I changed the test to skip if there's a problem. The test set-up is complex, using mock, and I'm not sure what is going on here. Its only for crs, not minolta
- the Builder preference that allows a custom location for components is the same, but how it works has changed. Things are now imported things explicitly using `importlib`, instead of `exec(relative-import)`. `importlib` is supposed to be python3 friendly. `exec` is supposed to be avoided.
- hit an unexpected failure in test_CRS_BitsSharp on travis, use pytest.skip to get around this line: `from serial.tools import list_ports` was failing with `ImportError: No module named tools`

Changes (for changelog):
- Builder components outside of the PsychoPy installation are supported as before, but need a slightly different directory structure: instead of `.../comp/` its now `.../comp/comp/`. This is automatically created as needed, and is almost fully backwards compatible. The same value as before is used in the preference field (`.../comp/`)
- Relocated some files. Some have been deprecated for years, and have been removed; the old way will no longer work. Others are newly deprecated; these still work but will now raise a warning log message. Here's how to update:
  - OLD --> NEW
  - `from psychopy import calib` --> `from psychopy import monitors as calib` (or just use `monitors`)
  - `from psychopy import bits` --> `from psychopy.hardware.crs import bits`
  - `from psychopy import _shadersPyglet` --> `from psychopy.visual import shaders`
  - `from psychopy import log` --> `from psychopy import logging`
  - `from psychopy import gamma` --> `from psychopy.visual import gamma`
  - `from psychopy import filters` --> `from psychopy.visual import filters`


jeremygray:
No merge conflicts with upstream / master as of 9f061e63e9110e9c

jeremygray:
Plan: I hope to go through `iohub/*.py` in January. And to all files, I might still add `print_function`, a standard (new / updated) copyright notice -- all the stuff that goes at the top and is unlikely to cause a merge conflict with anyone. I'll look through contrib/ to see if implicit relative imports are used, will touch as little as possible.

I can the squash all this into a single commit if that seems useful, or leave as multiple commits with their own commit message. I am keeping the overall commit message for the PR updated as I go. 

builder.py, data.py, and sound.py have the potential for merge conflicts, since we know people intend to refactor them into directory-style modules.

PR #1070, opened 2015-12-16
3 small fixes

jeremygray:
The XML fix restores the ability to load .psyexp files having advancedParams (including BART and mental rotation). This might be a FF instead of BF.


peircej:
Thanks. Both BFs are in the 1.83.03 release (currently uploading).

The advancedParams thing was caused partly by those experiment files not being re-saved for a very long time. AdvancedParams was deprecated a long time ago in favour of params each having a category flag. But good to have caught it and fixed it. thanks

Issue #1056, opened 2015-12-03
style-related refactoring


jeremygray:
To reduce technical debt and set the stage for further awesomeness, here's a summary of suggestions after discussion on the dev list. Ideally do them all at once, but that is not so realistic. Need to watch out for merge conflicts. I'll edit and check things off as we go.

Strong consensus, easy to do:
- [x] Revise imports: 
  - Remove all implicit relative imports. Use explicit relative for anything that is conceptually "local", and absolute for anything in (or going through) top-level (psychopy/).
  - Remove all use of wildcard imports, except in `contrib/`, plus some appropriate uses.
  - Remove all unused cruft imports (but not where deliberate). This can be tricky to tell in some cases, so I also added `# pylint: disable` directives on lines where unused imports should remain
- [x] Have a standard heading in each file:
  - `#!/usr/bin/env python2`, changing to `python` when support 2 and 3, or `python3` if/when only support python3
  -  `# -*- coding: utf-8 -*-` in all files, even those that do not need it
  - copyright line(s)
  - docstring
  - `from __future__ import (absolute_imports, print_function, division, unicode_literals)` to all files (but not demos). Handle `division` and `unicode_literals` (see #1061) separately, but might be easy.
- [x] Several smaller things that can be done at once 
  - Update all old-style classes to new style. Find them with `git grep -ne ':$' | grep ':class*' | grep \(\)`. Classes should typically call `super().__init__()`, even when inheriting directly from `object`, unless there's a reason not to.
  - Remove all mutables as default parameter values
  - Replace exec with eval whenever possible, which is most but not all places.
  - Remove all bare except statements. Use `except Exception:` instead when need very general. It allows `sys.exit()` and `KeyboardInterrupt` to raise (= good). Use a more precise Exception when possible, of course
- [x] Remove code that exclusively supports python 2.6.x or earlier (like the `_relpath()` function in builder.py). Need to update documentation to say works with python 2.7 only.
- [x] Fewer long lines; getting to none is not worth it, and not compromising readability is important. PEP-8 suggests 79 chars. Using `autopep8` will do a lot of this, but review by human is important too (otherwise things can look bad). 
- [ ] Add more doc-strings to files, classes, methods, functions
- [ ] Not discussed but seems uncontroversial: 
  - Rename variables that are reserved words in python, including `id`, `file`, `property`, and others. See pylint `redefined-builtin` warnings.

Strong consensus, just need the usual care to avoid merge conflicts and be backwards compatible:
[...]

PR #1074, opened 2015-12-16
RF: reorganize the builder module as multiple files; closes #1068

### LINKED ISSUE #1068: split builder.py into several files

jeremygray:


peircej:
This one scares me a lot. Bear in mind that 
the Builder barely receives any testing

jeremygray:
Yeah, agreed, so many lines. I poked and prodded everything in the Builder that I could by hand, in both English and Japanese locales. The changes were pretty much just moving things around in chunks, not tweaking individual lines. So  Keeping the .py files to less that 2000 lines of code is probably a good idea going forward (PEP8 suggests < 1000 lines).

It is possible to automate GUI testing, just a lot of work. There's a package called pyautogui that can find known visual elements (like an small image of an OK button) wherever they happen to be on the screen, and emulate a mouse click at that location. People have used it to allow AI to play video games, for example. 

My refactoring charge has just one more big thing on the wish list, plus some medium-small but touching many files (#1056). Doing autopep8 for whitespace would touch the most lines of code (using a widely used and battle-tested package to do so). I could do that this week if you just want me to get it over and done with. I know its a pain, and mucks with the history, ... but its worth it IMHO.

peircej:
I don't think we can do it next week. Other developers (psychopy repo currently has 179 forks) might want warning of exactly when this is going to happen so they can commit any changes first.

But i just mailed dev to start the question of when is a good time

PR #1075, opened 2015-12-16
several small fixes: tests, pillow support, configobj files

jeremygray:


jeremygray:
The merge conflict was in CHANGELOG, easily resolved. All changes in these commits are functional, including preparatory work for python3 (not just style preference changes).

peircej:
Thanks for adding notes to the changelog (historically I've had to go back trhough the commit log before a release and try to work out what happened). Note that the changelog is targeted at users not dev though and definitely follows the less-is-more rule. If this has too much info people won't read it at all so no info. Changes that don't impact users shouldn't be mentioned (except to highlight major refactoring where bugs might have arisen) and changes that do impact them need to be written in as non-technical language as we can.
But what you've done is definitely a help - I can rewrite later

PR #1089, opened 2016-01-22
BF OSError: gamma.py can't find Carbon framework + better exception handling

papr:
gamma.py looks for the Carbon framework in the wrong place under OS X 10.11. The framework binary was moved from

/System/Library/Carbon.framework/Carbon
to

/System/Library/Frameworks/Carbon.framework/Carbon
Additionally gamma.py only calls the CGSetDisplayTransferByTable(,,,,). This function was moved to the CoreGraphics framework.

Therefore, my suggestion is that gamma.py should try to load the CoreGraphics framework first and should use the Carbon framework only as fallback.

I implemented better exception handling additionally to pull request #1084


peircej:
Thanks Pablo

Issue #1083, opened 2016-01-14
use atexit standard module


jeremygray:
Can probably replace manual calls to core.quit completely, and also handle clean-up and shutdown processes for Window()s, ExperimentHandler()s, pyo server, etc. See discussion on dev-list Jan 13.

core.quit() would still be called, but called automatically upon sys.exit()

peircej:
On 14/01/2016 17:14, Jeremy R. Gray wrote:
>
> Can probably replace core.quit completely, and also handle clean-up 
> and shutdown processes for ExperimentHandler(s), pyo server, etc. See 
> discussion on dev-list Jan 13.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/psychopy/psychopy/issues/1083>.
>
We don't want to replace core.quit completely because of the scripts 
that would be broken for no reason. Let's just keep core.quit and 
register it as needing to be run atexit

-- 
Jon Peirce
http://www.peirce.org.uk

TEParsons:
I think we do do this - elements which need to do teardown stuff when Python exists do so via `__del__` these days

PR #1096, opened 2016-01-26
RF: replace bare except statements

jeremygray:
Replace with `except Exception:` instead. This is a functional change: a bare except will catch SystemExit. The new way allows using sys.exit() to quit a script, e.g., in conjunction with atexit module; see #1083
Some (many?) of the instances of Exception could be replaced with a more specific exception. ImportError are easy to infer but others are less so.


peircej:
Would you mind doing a pull from upstream and fixing, the merge conflict, and resubmitting the pull request?

jeremygray:
Done! And I'm done for the rest of the week too.

jeremygray:
The merge conflict is fixed, reopening the PR.

peircej:
Should I pull this in? 
Specifying which specific error, as @toddrjen suggests, is fine but we need to be pretty certain that the except isn't catching multiple different errors with a single resolution. I've certainly been lazy on checking what range of errors might occur, but that's also more likely to catch the problem before it leaves the area of interest?

jeremygray:
Yes, please pull it in in its current state. This is an improvement over bare excepts, and does so quite safely. Doing the additional work of more specific exceptions is indeed worth doing in those cases where it is clear what the failures will be, but will take longer.

peircej:
My thoughts exactly. Done :-)

PR #1106, opened 2016-02-09
FF: reposition _relpath function; closes #1098

### LINKED ISSUE #1098: bug: _relpath() in wrong place now
I introduced this when refactoring Builder. The proper fix is to delete _relpath completely, and replace with python 2.7's native handling of relative paths. Will fix when the style dust settles in a few days.

jeremygray:
The proper thing would be to delete this function completely, and rewrite the code. (No need to support python 2.5 anymore.)

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2126,21 +2126,3 @@ def appDataToFrames(prefs):
 def framesToAppData(prefs):
     pass
 
-
-def _relpath(path, start='.'):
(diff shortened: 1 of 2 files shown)


jeremygray:
Will reopen after moving the code to a new branch so I can keep working on master/ more easily.

jeremygray:
Actually, now I'm reluctant to mess with history, just in case. If you could pull this in today that would be cool.

PR #1119, opened 2016-02-18
Rf style contd

jeremygray:
I'm done for now. 
Not done yet (want to do, but no timeline set for this): iohub/ and tests/
Maybe, maybe never: top level psychopy/ (setup stuff, etc.)
Never: contrib/, files with # pylint: skip-file


jeremygray:
Anything I can do to make it easier to pull this one in? The two others were so quick, here I am concerned about merge conflicts with people working in other branches. When this is pulled, I will not do any more whitespace or linelength changes, except for iohub, maybe tests, and maybe top level setup stuff -- all with lots of communication first. If there are particular things you are concerned about for this PR, I can double check or write a new test or test. Or I could review again myself and merge myself if you are too busy.

PR #1127, opened 2016-02-29
RF: convert a few non_camel vars to camelCase

jeremygray:
Quite optional.

This was the only egregious example of multiple lower_underscore var names that I am aware of (having just looked at a lot of app/ and library code). If its worth converting things to camelCase for consistency, maybe this is a good time to do it. If not worth it, no trouble, I do not feel strongly about it.

### DIFF
--- a/psychopy/app/builder/dialogs/dlgsCode.py
+++ b/psychopy/app/builder/dialogs/dlgsCode.py
@@ -42,7 +42,7 @@ class DlgCodeComponentProperties(wx.Dialog):
         self.title = title
         # keep localized title to update dialog's properties later.
         self.localizedTitle = localizedTitle
-        self.code_gui_elements = {}
+        self.codeGuiElements = {}
         if not editing and 'name' in self.params.keys():
             # then we're adding a new component so ensure a valid name:
             makeValid = self.frame.exp.namespace.makeValid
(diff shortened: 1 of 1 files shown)


jeremygray:
Could also do the same for unit_label, mapping_func in:
tools/monitorunittools.py:88:def addUnitTypeConversion(unit_label, mapping_func):

peircej:
Given that these are changes that don't affect users I think it's a good idea to increase the consistency

jeremygray:
Questions:
1. `monitorunittools` seem unlikely to be used by end users -- ok to just change things like `unit_label, mapping_func` in `addUnitTypeConversion(unit_label, mapping_func)`? or make the change but also preserve backwards compatibility with a deprecation warning?
2. What about things like MonitorCenter.py, which are unusual / nonstandard because they start with a capital letter -- change to monitorCenter.py?
3. ./visual/glob_vars.py --> ./visual/globalVars.py (I have never liked glob since there is a package named glob)

peircej:
1 `unit_label` and `mapping_func` can change I think
2 Let's not change file names. People might be running MonitorCenter or have windows shortcuts pointing to it. 
3 I don't caer about glob_vars one way or another. I forgot it existed! ;-)

PR #1129, opened 2016-03-05
BFs: clean up several minor bugs & issues

jeremygray:
As found by @hoechenberger and @jeremygray during whitespace and line-length refactoring; some issues fixed by removing unused code.

CHANGELOG
- fixed 12 or so minor bugs
- core.shellCall now accepts a tuple
- MonitorCenter objects made pre-version 0.63 may not work anymore
- TrialHandler .nextTrial() removed (was deprecated June 2006). Change: use .next() instead
- TrialHandler[2|Ext] objects now correctly report their class in str()
- BaseComponent.writeTimeTestCode() is deprecated (will be removed in 1.85).

### DIFF
--- a/psychopy/app/builder/components/_base.py
+++ b/psychopy/app/builder/components/_base.py
@@ -4,7 +4,7 @@
 
 from __future__ import absolute_import
 
-from ..experiment import Param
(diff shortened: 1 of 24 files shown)


peircej:
good work guys

PR #1134, opened 2016-03-08
BF: code snippet validation; closes #1132

### LINKED ISSUE #1132: Builder: variable naming collisions not apparent to novice user
Hello!

I tried building a test experiment through the builder, and imported a conditions file that has a column named "image". When I added an Image stimulus by clicking on one in the Components, and left the default name as "image" and set the image as $image, this generated a python error: 

`Traceback (most recent call last):
  File "/home/daniel/ownCloud/phdExperiments/maskedVocabTraining/vocabTraining-builder_lastrun.py", line 109, in <module>
    image.setImage(image)
AttributeError: 'unicode' object has no attribute 'setImage'`

The problem was that the stimulus's variable name is "image", and also the variable used in the conditions spreadsheet is "image", so I inadvertently created a naming collision. I was able to quickly figure this out because I program, but I'm posting this here as a bug because to someone who has no programming background this would be a meaningless and intimidating error message, which could give them a negative impression of your wonderful software. I've included a snapshot of the dialogue to hopefully make what I'm saying clearer.
screenshot from 2016-03-07 14 52 20

I suggest that a simple way to avoid this could be that when adding a stimulus from the builder's  dialogues, that if the user tries to click OK and the name of the stimulus and the image (or the Text, for a text stimulus) will be the same, that a clear error or warning be generated, saying something like "The fields for Name and Image cannot have the same value". Just a thought.

Thank you for the great software. I do hope you consider this, it seems like an easily implemented improvement to the user experience.  Take care,

jeremygray:
Detects and warns about syntax errors and self-referential var names in code snippets.
Disables OK button. 
Not finished:
- text field in TextComponent. Guess: text-events are not triggering Validate()
- some field names are not localized in error messages

### DIFF
--- a/psychopy/app/builder/dialogs/__init__.py
+++ b/psychopy/app/builder/dialogs/__init__.py
@@ -667,8 +667,12 @@ class _BaseParamsDlg(wx.Dialog):
         # use monospace font to signal code:
         self.checkCodeWanted(self.startValCtrl)
         self.startValCtrl.Bind(wx.EVT_KEY_UP, self.checkCodeWanted)
+        self.startValCtrl.SetValidator(CodeSnippetValidator('startVal'))
+        self.startValCtrl.Bind(wx.EVT_KEY_UP, self.doValidate)
         self.checkCodeWanted(self.stopValCtrl)
         self.stopValCtrl.Bind(wx.EVT_KEY_UP, self.checkCodeWanted)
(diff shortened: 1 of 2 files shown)


jeremygray:
Will clean up and reopen, a couple other little things to fix

jeremygray:
travis-ci is stalling out, not failing

PR #1136, opened 2016-03-10
Feature renameRoutine

OliJimbo:
Added __setitem__ to get rid of the old name from the component params
list - this does not fix the previous commit error!


OliJimbo:
Thanks Jon and Jeremy - I had played around with pop() a few times but it seemed not to work - so I must have used it incorrectly.

I have now pepped it and the menu version is fine.  I'll work on the context menu later (I'll need to add a little more code so rename only appears when routine is selected rather than the loop as well).

OliJimbo:
Hi Jon and Jeremy,

I have now implemented a rename option in the context menu which will only appear when a routine is selected.  Above, I am told that there are conflicts - but when I run git status nothing shows up so I'm not sure where the problem is!

Also - I am not sure why I edited __init__ so I'd be happy to revert it, if I run git revert(7fc1039fed76883b4e5c27ee24b22ce78aa64bca) will this undo everything, or just the files, like __init__ which haven't been changed since this commit.

Many thanks

Oli

jeremygray:
To revert the `__init__.py` file, I am not sure what git revert will do. (I should look it up sometime!) You will get the desired result if you copy the old file and replace the new one with the old one, and make a new commit.

PR #1141, opened 2016-03-16
DOC: added resource link

djmannion:


peircej:
Thanks Damien. I'll try and get around to rebuilding the website soon (which is when the update will become visible). And I'll have to check out your screen casts too :-)

PR #1183, opened 2016-05-25
Master

OliJimbo:
Added self.redraw() so flow updates itself when routines and loops
deleted


peircej:
Sounds like the right thing to do! ;-)
Thanks

Issue #1203, opened 2016-07-08
"Benchmark wizard" menu opens the configuration wizard before opening the benchmark wizard.


hsogo:
Selecting "Benchmark wizard" from "Tools" menu of the Builder/Coder, the configuration wizard is opened at first. After running or canceling the configuration wizard, the benchmark wizard appears.
I confirmed this issue in PsychoPy 1.84.0rc5 (on Windows10).

I read psychopy/tools/wizard.py and noticed that psychopy.tools.wizard.BenchmarkWizard calls `__init__()` of psychopy.tools.wizard.ConfigWizard ad line 638.

```
class BenchmarkWizard(ConfigWizard):
    """Class to get system info, run benchmarks
    """

    def __init__(self, fullscr=True, interactive=True, log=True):
        super(BenchmarkWizard, self).__init__(interactive=interactive)  # <- line 638
        self.firstrun = False
        self.prefs = prefs
        self.appName = 'PsychoPy2'
        self.name = self.appName + _translate(' Benchmark Wizard')

```
I commented out line 638 and run the Benchmark wizard. It seemed to work fine.
I confirmed that BenchmarkWizard didn't call `__init__()` of ConfigWizard In the previous version (1.82.01).  Why `__init__()` of ConfigWizard is called in the latest version?  If this is necessary, we have to suppress opening the Configuration Wizard when `__init__()` of ConfigWizard is called from BenchmarkWizard class.

hsogo:
In `ConfigWizard.__init__()`, following attributes are defined. All of these attributes are re-defined in `BenchmarkWizard.__init__()`.

- self.firstrun 
- self.prefs
- self.appName
- self.name
- self.reportPath

I think it is safe to delete call of `ConfigWizard.__init__()` from `BenchmarkWizard.__init__()`.

PR #1206, opened 2016-07-15
FF: Configuration Wizard is opened before Benchmark Wizard is opened. (issue #1203)

hsogo:
`BenchmarkWizard.__init__()` calls superclass's (i.e. ConfigWizard's) `__init__()`, which opens Configuration Wizard Window.

### DIFF
--- a/psychopy/tools/wizard.py
+++ b/psychopy/tools/wizard.py
@@ -635,7 +635,8 @@ class BenchmarkWizard(ConfigWizard):
     """
 
     def __init__(self, fullscr=True, interactive=True, log=True):
-        super(BenchmarkWizard, self).__init__(interactive=interactive)
+        # Don't call parent's __init__() to avoid starting
+        # Configuration Wizard.
         self.firstrun = False
         self.prefs = prefs
         self.appName = 'PsychoPy2'
--- a/psychopy/tools/wizard.py
+++ b/psychopy/tools/wizard.py
@@ -61,156 +61,11 @@ _localized = {
 # can't just do the following, or messes up poedit autodiscovery:
 # _localized = {k: _translate(k) for k in _loKeys}
 
-
-class ConfigWizard(object):
-    """Walk through configuration diagnostics & generate report.
+class BaseWizard(object):
+    """Base class of ConfigWizard and BenchmarkWizard.
     """
-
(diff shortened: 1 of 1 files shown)


peircej:
@ jeremygray could you look at this. Was there a reason at some point that this was required? Does this fix the issue of the startup wizard getting stuck open?

hsogo:
Calling ConfigWizard's `__init__()` was added in "RF: reduce long lines in psychopy libs" (1dcc55d10b56dcf361ab2b7a2b3c8fd43e04d1fa). Then, "interactive" parameter was passed to `__init__()` in "FF: fix benchmark wizard runing under tests" (8771684e20b6cea7ef62816a4a56d754ef7957ba) to avoid opening a dialog box in the test mode.
I guess that adding ConfigWizard's `__init__()` was done as a part of refactoring. so it would not be intended to fix bugs or to add new functionality.

peircej:
I'd still like to get @jeremygray to confirm that (he didn't get flagged in last attempt due to space after @ symbol! ;-)

jeremygray:
I'll take a look soon.

jeremygray:
The suggested fix will cause problems because a BenchmarkWizard object's `self.runDiagnostics` method will no longer be defined (for `diagnostics = self.runDiagnostics(win, verbose=True)`).

It will take me a little longer to track this down. At the moment, for me, the BenchmarkWizard is not opening at all, nothing happens.

hsogo:
I am testing PsychoPy 1.84.0rc5 on Japanese Windows7, WinPython 2.7.10 and wxPython 2.8.12.1. On this environment, The BenchmarkWizard can be opened from the Builder/Coder menu with my fix...

By the way, I think that the BenchmarkWizard should not be derived from ConfigWizard.  How about making a base-wizard class and deriving the BenchmarkWizard and ConfigWizard from the base-wizard class?

peircej:
I agree. Actually I'd rather like to revisit the startup wizard at some stage to be simpler, with fewer tests and possibly more *configuration* options, but for now we should at least get to the point that it runs!

jeremygray:
I also agree about refactoring and the revisions.

hsogo:
I added BaseWizard class. It passes "test_app" test on my Windows10 and Ubuntu14.04LTS.

PR #1211, opened 2016-08-05
Fix problem with empty cells when importing a csv conditions file

jonathanoroberts:
empty cells in csv conditions files break the future loading of the
experiment flow in the builder. Empty cells in the
csv are read as type numpy.nan and stored as "nan" in the psyexp file.
This change converts them to python type None which is the same type that
empty cells in xlsx conditions files are read in as...


peircej:
Cool. Thanks Jonathan

PR #1234, opened 2016-08-29
corrected attribute naming

TheChymera:
It seems for MovieStim3() the `_reset` attribute is being replaced by `reset`.

The erroneous call breaks the `stop` atribute:

```
    mov.stop()
  File "/usr/lib64/python2.7/site-packages/psychopy/visual/movie3.py", line 243, in stop
    self._reset()
AttributeError: 'MovieStim3' object has no attribute '_reset'
``` 

This behaviour is fixed by the PR

### DIFF
--- a/psychopy/visual/movie3.py
+++ b/psychopy/visual/movie3.py
@@ -240,7 +240,7 @@ class MovieStim3(BaseVisualStim, ContainerMixin):
         if self.status != STOPPED:
             self.status = STOPPED
             self._unload()
-            self._reset()
+            self.reset()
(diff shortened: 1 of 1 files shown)


TheChymera:
@peircej - I don't think [the CI error](https://travis-ci.org/psychopy/psychopy/builds/156087920#L2292) is because of me.

peircej:
The CI error stems from 7f753f5f1260387894f81108ddc56f960e57ca2e but unfortunately I can't replicate it on my mac so I can't work out what's failing.

PR #1236, opened 2016-08-31
BF: moved GL_STEREO extension check before calling GL.Config

mdcutone:
Check if GL_STEREO not is supported on this system before calling GL.config. Psychopy crashes instead of showing the typical warning. 

Later on, we can create FBOs to substitute quad buffers for systems without the GL_STEREO extension.

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -927,7 +927,7 @@ class Window(object):
             # writes a series of static frames as frame001.tif,
             # frame002.tif etc...
             myWin.saveMovieFrames('frame.tif')
-            
+
             #as of PsychoPy 1.84.1 the following are written with moviepy
             myWin.saveMovieFrames('stimuli.mp4') # codec = 'libx264' or 'mpeg4'
             myWin.saveMovieFrames('stimuli.mov')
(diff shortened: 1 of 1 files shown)


peircej:
This is definitely an improvement but I wonder whether we should be going a stage further and doing raise an exception rather than logging a warning. If someone requests a stereo window and doesn't get one I imagine they want the scrpit to abort rather than quietly continue with no stereo. They might not notice a logging.warning?

ghcmat:
Raising an exception is certainly a good idea at this point. I'm currently using this to detect if driver support for stereo is available, if not use some quasi-stereo mode that will allow the research to code such experiments on a system that does not have support. Since that feature isn't available yet anyways, we should use an exception for the time being. The current exception is raised within Pyglet and is not informative to the nature of the problem.

djmannion:
I'd prefer that it not raise an exception - not yet, at least. My card (Quadro K1200) does not report having GL_STEREO but quad-bufffers work fine. I have been trying to think of a check for this, without finding one so far. 'glxinfo' has 'stereo=1', but I'm not sure how to access that recognition.

ghcmat:
Right, we need to keep in mind "don't break anyone's experiments" when planning these things. ;)

peircej:
OK, if there are times when this works despite the unsupported extension then a warning is the right way to go. Thanks guys

mdcutone:
Hello Jon,

In this case we are completely disabling stereo to avoid the crash and obscure error. On @djmannion's system it throws a warning but still enables stereo. So this patch could break his experiments. Is there another quad-buffer extension out there that we should be checking for? At least on unsupported systems there isn't a crash anymore. 

I'm working on a stereomodes fork trying to resolve this issue. There is an "opengl" option you can specify that will force it regardless if the GL_STEREO extension is detected or not.

PR #1242, opened 2016-09-06
Feature Enhancement: Reflect keyboard modifier flags

jmweir:
This enhancement causes keyboard modifier flags from Pyglet to bubble up to the user. This is an optional feature similar to timestamps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /.ropeproject/history
 /.ropeproject/objectdb
 /.spyderproject
+/.vscode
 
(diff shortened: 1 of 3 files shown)


jmweir:
Contributed to PsychoPy by @jmweir and @Visee-Michael with 5amsolutions.com

peircej:
Sounds good. Does this work on all platforms or is it just windows?

jmweir:
We've verified the new behavior on Mac OS 10.11 and Windows 10.

PR #1249, opened 2016-09-29
BF: Properly handle screenHz argument of StaticPeriod

hoechenberger:
It appears the `StaticPeriod` never actually did what the documentation
promised when specifying the `screenHz` keyword argument. The [docs](http://www.psychopy.org/api/core.html#psychopy.core.StaticPeriod)
say:

```
# the period takes into account the next frame flip
# time should now be at exactly 0.5s later than when ISI.start()
# was called
```

This was indeed never the case (I checked the entire git history via `git log -L :StaticPeriod:core.py`). Instead, the `StaticPeriod` would always wait for the entire specified duration, ignoring the refresh rate and not reserving time for an additional flip. Hence, a user assuming the documented behavior would overshoot by one frame.

I consider this a serious bug we should probably announce somewhere in **bold** letters or something. I don't know if my fix will have any consequences on any Builder components/implementations. At least the `StaticComponent` makes use of the `StaticPeriod`, and supplies the `screenHz` parameter.

**Although the TravisCI test suite already passes, please do not pull until we're sure about possible side effects of this fix.**

**Edit:** Clarified the current (buggy) behavior.

### DIFF
--- a/psychopy/core.py
+++ b/psychopy/core.py
@@ -138,7 +138,7 @@ class StaticPeriod(object):
         be reset and starts again
         """
         self.status = STARTED
-        self.countdown.reset(duration)
+        self.countdown.reset(duration - self.frameTime)
         # turn off recording of frame intervals throughout static period
         if self.win:
             self.win.recordFrameIntervals = False
--- a/psychopy/core.py
+++ b/psychopy/core.py
@@ -136,6 +136,8 @@ class StaticPeriod(object):
     def start(self, duration):
         """Start the period. If this is called a second time, the timer will
         be reset and starts again
+
+        :param duration: The duration of the period, in seconds.
         """
         self.status = STARTED
         self.countdown.reset(duration - self.frameTime)
--- a/psychopy/tests/test_misc/test_core.py
+++ b/psychopy/tests/test_misc/test_core.py
@@ -349,6 +349,26 @@ def testStaticPeriod():
     assert win.recordFrameIntervals == False
     static.complete()
     assert static._winWasRecordingIntervals == win.recordFrameIntervals
+    win.close()
+
+    # Test if screenHz parameter is respected, i.e., if after completion of the
+    # StaticPeriod, 1/screenHz seconds are still remaining, so the period will
+    # complete after the next flip.
+    refresh_rate = 100.0
+    period_duration = 0.1
+    timer = CountdownTimer()
+    win = Window(autoLog=False)
+
+    static = StaticPeriod(screenHz=refresh_rate, win=win)
+    static.start(period_duration)
+    timer.reset(period_duration )
+    static.complete()
+
+    assert np.allclose(timer.getTime(),
+                       1/refresh_rate,
+                       atol=0.001)
+    win.close()
+
 
 @pytest.mark.quit
 def test_quit():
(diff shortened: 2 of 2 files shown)


hoechenberger:
cc @jeremygray

hoechenberger:
Okay, so the code generated by the Builder is the following (irrelevant lines removed):
```python
ISI = core.StaticPeriod(win=win, screenHz=expInfo['frameRate'], name='ISI')

# Create some handy timers
routineTimer = core.CountdownTimer()  # to track time remaining of each (non-slip) routine 

#------Prepare to start Routine "trial"-------
trialComponents = []
trialComponents.append(ISI)

#-------Start Routine "trial"-------
continueRoutine = True
while continueRoutine and routineTimer.getTime() > 0:
    t = trialClock.getTime()
    if t >= 0.0 and ISI.status == NOT_STARTED:
        ISI.tStart = t  # underestimates by a little under one frame
        ISI.start(0.5)
    elif ISI.status == STARTED:
        ISI.complete()
    
    continueRoutine = False  # will revert to True if at least one component still running
    for thisComponent in trialComponents:
        if hasattr(thisComponent, "status") and thisComponent.status != FINISHED:
            continueRoutine = True
            break
    
        win.flip()
```
Since `StaticPeriod.complete()` sets `StaticPeriod.status` to `FINISHED`, builder-generated code would set `continueRoutine` to `False` and consequently would not `flip` after finihsing the `StaticPeriod`. This means that

* the change I introduced fixes the `StaticPeriod` behavior for code users,
* but breaks behavior for Builder users (`StaticComponent` would finish one frame early)

The solution is simple: the `StaticComponent` should drop the `screenHz` parameter for `StaticPeriod`. That is, I suggest the following change to `app.builder.static.StaticPeriod.writeInitCode()`:

current:
```python
    def writeInitCode(self, buff):
        code = ("%(name)s = core.StaticPeriod(win=win, "
                "screenHz=expInfo['frameRate'], name='%(name)s')\n")
        buff.writeIndented(code % self.params)
```
Proposed:
```python
    def writeInitCode(self, buff):
        code = "%(name)s = core.StaticPeriod(win=win, name='%(name)s')\n"
        buff.writeIndented(code % self.params)
```
Please let me know what you think.

hoechenberger:
I was mistaken. The above is only true if the `StaticComponent` is the only component in the routine. If there's additional components, we definitely need to make use of `screenHz` to get the correct timing. Which also implies that timing for all current experiments using the `StaticComponent` in conjunction with additional components finishing after the `StaticComponent` in the same routine was one frame off.

hoechenberger:
Please let me know if more/clearer information is required, or if I maybe should open a ticket on the issue tracker first.

peircej:
Hi Richard. Thanks for this - I'm currently releasing 1.84.2 and I want to hold off until a) that's complete and b) I've had time to look at this carefully. I'll get to it soon

hoechenberger:
Hi @peircej, thanks for merging! But like I mentioned above, this fix is possibly going to alter timing in Builder experiments, as the `StaticComponent` makes use of the `StaticPeriod`. We should definitely run some tests/think about this issue! Thanks

peircej:
Yes, wanted to pull it in to investigate it locally but I'll keep in mind that any immediate (and any 1.84.x) releases will not include this one. thanks

hoechenberger:
Great, thanks.

PR #1256, opened 2016-10-07
Doc: substituted rgb by color

stefanrepplinger:
Use of RGB-parameter has been DEPRECATED since v1.60.05


peircej:
Thanks stefan

Issue #1262, opened 2016-10-09
Improve exception handling


hoechenberger:
As already briefly mentioned in #1056, we should try to use specific exceptions whenever possible (i.e., `except Exception` should be exactly that: an exception). 

In my continued efforts to improve the code base and introduce new and unnecessary bugs, I have grepped the code for `except Exception` blocks, and explicitly named some more specific exceptions whenever there was an obvious way to do so. Anything that wasn't very clear, I left out. However, it is still possible that I broke something, and Travis seems not to be happy anyway (something broken there?? Cannot get any test to pass!)

**At any rate, I invite you all to help me test and improve this code. Thanks!**

**See here: https://github.com/psychopy/psychopy/compare/master...hoechenberger:exceptions?expand=1**

cc @peircej @jeremygray

hoechenberger:
I've opened a pull request to allow us to make use of the new code review features :)

PR #1263, opened 2016-10-09
RF: Explicitly name exceptions to catch

hoechenberger:
If my continuing efforts to improve the code base and introduce new and unncessary bugs, I have grepped the code for `except Exception` blocks, and explicitly named some more specific exceptions whenever there was an obvious way to do so. Anything that wasn't very clear, I left out.

Ref: #1262

### DIFF
--- a/psychopy/colors.py
+++ b/psychopy/colors.py
@@ -26,7 +26,7 @@ def isValidColor(color):
     try:
         color = float(color)
         return True
-    except Exception:
+    except ValueError:
         if isinstance(color, basestring) and len(color):
             return (color.lower() in colors255.keys()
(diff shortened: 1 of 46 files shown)


peircej:
I'm sceptical of this. How absolutely certain are you that there isn't more than one way for the exception to be raised? Are any of these actually causing problems by catching _too many_ exceptions? The last 2 releases have needed several bug fixes due to the line-shortening refactors and I don't want to spend any more time fixing further unnecessary breaks

hoechenberger:
Withdrawing this pull request for the time being, needs further discussion. x-ref #1262

PR #1273, opened 2016-10-14
BF: Filename collisions

### LINKED ISSUE #1207: MultiStairHandler method saveAsExcel creates too many files
Hi,

This is a minor issue but I have a MultiStairhandler with 4 conditions (or 4 interleaved staircases).

When I'm done with the staircases, I call `staircases.saveAsExcel`. This produces **2** `.xlsx` files, with the first staircase being in the first spreadsheet, and the three subsequent staircases being in the second spreadsheet. It looks like this:
image

image

As you can see, the staircase labels are all unique. Only one of the staircases is in a separate `.xlsx` file.

For reference, I am using:
psychopy version 1.82.1
python version 2.7.11 (via anaconda 4.0.0, 32 bit)

hoechenberger:
This PR fixes two bugs:
* `data.MultiStairHandler.saveAsExcel()` would ignore the `fileCollisionMethod` keyword argument
* `tools.fileerrortools.handleFileCollision()` would behave incorrectly if `fileCollisionMethod='rename'` was passed and no existing file could be found.

Closes #1207.

### DIFF
--- a/psychopy/data.py
+++ b/psychopy/data.py
@@ -4334,8 +4334,7 @@ class MultiStairHandler(_BaseTrialHandler):
             label = thisStair.condition['label']
             thisStair.saveAsExcel(
                 fileName, sheetName=label, matrixOnly=matrixOnly,
-                appendFile=append, fileCollisionMethod='rename'
-            )
+                appendFile=append, fileCollisionMethod=fileCollisionMethod)
             append = True
 
     def saveAsText(self, fileName,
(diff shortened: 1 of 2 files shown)


peircej:
These certainly seem correct fixes, but do they address the issue that the original issue reported (some staircase being saved to the same file and some being saved ot different files)? It seems like that must be caused by some race condition that doesn't seem fixed here.

Related note is that, for excel files the issue of overwriting is a bit different because we can have multiple datasets in the same filename as sheets (which is generally more convenient). Does your fix mean that files are now always saved separately?

hoechenberger:
Hi Jon, yes this is not that intuitive, but bea1d058cf425c6a2175096aa1fb265bc2ca27f9 actually fixes #1207 (which only occurred with `fileCollisionMethod='rename'`). We now create only a single excel file with different worksheets.

peircej:
Awesomeness. Thanks for all your recent work on the issue list by the way! :-)

hoechenberger:
You're most welcome, I had promised this, like, a year ago or so already, so I thought now is the time to get started ;)

PR #1276, opened 2016-10-15
BF: Fix import and write issues with openpyxl

### LINKED ISSUE #1274: openpyxl errors when using saveAsExcel methods
As discussed in a previous issue (#1207), there are errors when using saveAsExcel when using openpyxl 2.4.0:

> With PsychoPy 1.84.2 (and openpyxl 2.4.0), I am now getting an ImportError when trying to call saveAsExcel:
> ImportError: openpyxl is required for saving files in Excel (xlsx) format, but was not found.
> 
> I did a little digging, and found that this seems to stem from line 29 in data.py:
> from openpyxl.cell import get_column_letter
> 
> as far as I can tell, get_column_letter sits in openpyxl.utils.cell:
> http://openpyxl.readthedocs.io/en/default/api/openpyxl.utils.cell.html#openpyxl.utils.cell.get_column_letter
> 
> Now, if I change line 29 to import from openpyxl.utils.cell instead, I don't get the ImportError anymore. However, I get a different error instead:
> 
> File "[local python path]/data.py", line 3326, in saveAsExcel:
> ew = ExcelWriter(workbook=wb)
> TypeError: __init__() takes exactly 3 arguments (2 given)

janfreyberg:
- not using ExcelWriter()
- now compatible with latest openpyxl version

This closes #1274


psychopy/data.py
```diff
@@ -663,8 +667,6 @@ def saveAsExcel(self, fileName, sheetName='rawData',
             wb.properties.creator = 'PsychoPy' + psychopy.__version__
             newWorkbook = True
 
-        ew = ExcelWriter(workbook=wb)
```
hoechenberger: Are you sure we don't need the `ExcelWriter` for openpyxl < 2.4?

psychopy/data.py
```diff
@@ -22,11 +22,16 @@
 import re
 import warnings
 import collections
+from distutils.version import StrictVersion
 
 try:
     # import openpyxl
     import openpyxl
-    from openpyxl.cell import get_column_letter
```
hoechenberger: I like this explicit approach, makes it very clear what's going on and why.

psychopy/data.py
janfreyberg: Fairly sure. It worked with openpyxl 2.3.2, which is what I had before - I think workbook.save() has been in the package for a while.

hoechenberger: Okay nice, thanks for checking this!

janfreyberg:
I _think_ I've just done that, but I'm not sure - I used the git commit --amend command, so there's a new commit in my branch - I'm not sure why there's an empty "Merge branch 'master'" commit in there, as well though.

hoechenberger:
Hehe no worries, I think we can just squash these commits.

The proper way to simply rename/change a commit, without creating a new one, is via a rebase, though. But never mind, all is good.

Maybe a little advice for the future: it's always a good idea to create a new branch for fixes and new features. Makes handling easier. My workflow is usually:

- pull in all changes from the psychopy upstream `master` branches into my local `master` branch
- create a new feature branch of this local `master` branch
- work in this branch, commit to that branch
- once I'm done developing, switch to the `master` branch and pull in any changes from upstream `master` that possibly occurred while I'd been busy working on my own stuff
- switch back to my feature branch
- rebase that branch on my local `master`, possibly rewording, re-arranging, and squashing commits
- push to GitHub (or even force-push, if I had previously pushed and rebased afterwards)
- file a PR for that branch
- work on something else (i.e., new feature branch based on `master`) while my PR is awaiting approval
- once the PR gets accepted and merged, delete the feature branch both locally and on GitHub

... and the cycle continues

- pull in upstream changes to local `master`
- rebase new feature branch on `master`
- ...

Maybe there's more elegant ways, but this is how I usually work. I don't use the command line anymore btw, there's great GUI tools available now. And my IDE also does most of the stuff anyway... (I'm using PyCharm)

At any rate, thank you very much for your contribution!

janfreyberg:
OK, thanks for the pointers! That seems like a good workflow, I'll keep that in mind for the future. And I will explore some of the GUI tools... I code in Atom, which does seem to have some GUI extensions that look useful.

hoechenberger:
I hear GitKraken is a neat GUI, maybe you can check it out: https://www.gitkraken.com/features

PR #1279, opened 2016-10-18
NF: SimpleRatingScale

hoechenberger:
The `SimpleRatingScale` is a new type of (continuous) rating scale that can be displayed both horizontally and vertically. It can therefore be used to create visual analog scales (VAS) and labeled magnitude scales (LMS).

The initial mouse cursor position when first displaying the scale can be set via the `mousePos` keyword argument.

This class supports ioHub for response time acquisition. When `iohub=True` is passed to `__init__()`, we will try to connect to an existing ioHub session and  acquire a mouse device. Since ioHub button press events will in some cases  (e.g., when not running in fullscreen mode) return the mouse position in absolute display pixels without relation to the actual PsychoPy window, we  use `event.Mouse.isPressedIn()` to check whether a mouse click within a valid   area was registered, and only subsequently check the ioHub event queue to fetch that event (and its respective response time).

Since `_init__()` tended to become too large and cluttered, many initialization procedures have been placed in separate functions (`_gen_*()`), as commonly done by other stimulus classes in PsychoPy. However, I avoided the "silent" creation of attributes in these functions: they are all side-effect free, and return values that are then explicitly assigned to an attribute inside `__init__()`. So by just looking at `__init__()`, it is already clear which attributes the class has, and when and how they are being created.

Further, all internal implementation details follow, to the best of my knowledge, the PEP-8 style and naming recommendations.

The public API is very minimalistic, as, aside from the provided keyword arguments, only the following attributes and methods are exposed:

`draw()`, `waitForResponse()`, `reset()`, `response`, `rt`

 Currently, only a small number of attributes is user-settable after instantiation:

`mousePos`, `maxTime`, `finishOnResponse`, `resetOnFirstFlip`, `autoLog`

 If `resetOnFirstFlip=True` (the default), the `reset()` method will be called immediately  upon the next `visual.Window.flip()` (via `Window.callOnFlip()`). This ensures that  response times will be precisely time-locked to the onset of the rating scale presentation.

**Help wanted**
~~Tests pass locally, except for PyGame windows:~~ (see [this comment)](https://github.com/psychopy/psychopy/pull/1279#issuecomment-254604822)

Travis tests fail with weird errors:

``` python
        if ((precision is not None) and
>               not np.isclose((limits[1] - limits[0]) % precision, 0)):
E   AttributeError: 'module' object has no attribute 'isclose'
```

I didn't manage to fix these problems, any help on this would be greatly appreciated.


hoechenberger:
Oh I just realized testing PyGame had been [removed](a9cbd65462b2115d4597cb651389c3d5119fca9d) previously, but I accidentally added it back in. Will revert this change of mine, rebase and force-push to this branch. The Travis problems with `numpy` remain to be resolved.

hoechenberger:
Still looking for help re Travis for this one! :)

peircej:
Not sure about the `isclose()` issue. All I can imagine is that travis is installing an old version of numpy (but hard to imagine it's *that* old!). Mayb just replace:
`np.isclose((limits[1] - limits[0]) % precision, 0))`
with
`abs(limits[1] - limits[0])) > precision`
as a workaround?

There's an error with the screenshot comparison. Hard to know why because we don't get to see the travis files that were saved during the fail (the failed screenshot). My guess is that this is an issue with a font not lining up or something so you could probably extend the RMS error limit on that screenshot.

Then the error about 
```
>       assert rs.response == 50
E       assert None == 50
```
is presumably doe to a starting value on the rs.response while waiting?

hoechenberger:
Thanks Jon, will try to fix this!

peircej:
I was going to check with you more generally about the approach you've taken here. I agree that we need a SimpleRating component; @jeremygray would probably agree that the existing one has too much in it which then requires very many params to control.

Yours looks simpler, but maybe it can be even simpler still!?  In particular what I wonder about is the control of when the object "ends". Generally psychopy objects (whether stimulus or response) allow the user to handle that in code. Can we treat SimpleRating like a standard visual stimulus btu one that stores the most recent place someone clicked. They can set the conditions of when it gets presented just using a while loop (which could end either on a response or at a certain time). A Builder component might still have a forceEndRoutine option (as in the mouse or keyboard) but that is handled in the builder code not by the object.

jeremygray:
My 2 cents:

I certainly support the idea of simplicity. The current Ratingscale was my first major code contribution, and there are things I would do differently now. I agree that the current RatingScale pushes the envelope in terms of the number of things that can be customized, and is unpleasantly complex as a result, in turn making it harder to support end users. Nonetheless, the features are almost all things that people have asked for, and presumably use. There are other features that people have asked for that are not included, especially a vertical display and iohub-related control of the scale.

peircej:
Sorry, didn't mean to disrespect RatingScale :-)
Just saying that we could keep features of the SimpleRating more simple and push customization aspects out to other parts of the code

jeremygray:
heh, all good!

--Jeremy

Issue #1305, opened 2016-11-15
moviestim3, OS X 10.11.6, miniconda installation avoiding pyo error in silent movies


ofgulban:
Hi, I was having the following error when using moviestim3 in Mac Os X 10.11.6 through miniconda installation (psychopy release version 1.84.2):

```
1.2897  ERROR  No audio API found. Try installing pyo 0.6.8+, or pygame 1.8+
```

However I do not have sound in any of my movie clips so I had a look into the source code (psychopy/visual/movie3.py) and realized that all I need to do is to comment out this line:

```from psychopy import sound```

moviestim3 works fine afterwards with no errors.

I wanted to let you know about this specific case. Maybe you can handle it differently when there is no sound, like only importing the sound module if noAudio is True.

peircej:
I don't fancy pursuing this one myself but happy to take a pull request if someone else wants to do it. I'd rather get sounds to work on anaconda!

As a side issue, the reason to use anaconda used to be that it was easier to install things than python, but since python wheels have become common I think it might be easier just to use regular python. PsychoPy's installation docs show how easy it is to get everything that way now:
http://psychopy.org/installation.html#manual-install

ofgulban:
Thanks for the response. Moviestim3 works without a problem in windows and linux (with conda installation) but unfortunately in mac this was not the case. I have created a [pull request](https://github.com/psychopy/psychopy/pull/1306). It is more of a workaround for silent movies.

hoechenberger:
Is this issue closed by #1306?

ofgulban:
@hoechenberger yes, I think so.

hoechenberger:
Thanks!

PR #1306, opened 2016-11-16
moviestim3 sound module problem workaround for silent movies

ofgulban:
The related [issue is here](https://github.com/psychopy/psychopy/issues/1305).

I have changed the code to import the sound module depending on the noAudio flag. This removes the current and potentially future sound module dependency problems for movie presentations with no sound. This change does not effect the default behaviour.

### DIFF
--- a/psychopy/visual/movie3.py
+++ b/psychopy/visual/movie3.py
@@ -180,8 +180,10 @@ class MovieStim3(BaseVisualStim, ContainerMixin):
                         self._mov.audio.to_soundarray(),
                         sampleRate=self._mov.audio.fps)
                 except:
-                    # JWE added this as a patch for a moviepy oddity where the duration is inflated in the saved file
-                    # causes the audioclip to be the wrong length, so round down and it should work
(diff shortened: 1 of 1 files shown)


peircej:
Thanks Omer. I guess it was that simple. I was thinking we would need to do testing for whether *any* of the movies was going to require sound etc, but I guess the fact that importing an already-imported module is cheap means that this way works. :-)
thanks

ofgulban:
I am glad to help, thanks.

Just a minor question for my curiosity, I think it is not practical to change it now but I was wondering why the flag is named ```noAudio==false``` instead of ```audio==true```. Isn't it more intuitive to avoid double negatives? I mean when you see things like ```if not noAudio``` later in the code later it takes a little time to understand whether there will be sound or not :)

peircej:
True. I don't recall exactly what the reasoning was. I guess I was originally thinking about it as actively turning off the audio rather than merely not having audio, like the concept of 'mute'. But you're right that audio=False would have made sense.

PR #1308, opened 2016-11-18
BF/ENH:Switched httplib to httplib2 since httplib is outdated.

Infinite-PL:
I don't know if this will be useful but still I did it ;)
Signed-off-by: Michał Oszczak <mimioszczak@gmail.com>

### DIFF
--- a/psychopy/web.py
+++ b/psychopy/web.py
@@ -10,7 +10,7 @@
 import os
 import sys
 import base64
-import httplib
+import httplib2
 import mimetypes  # deprecated; use requests package instead
 import socket
 import re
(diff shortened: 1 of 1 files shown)


peircej:
The thing is that httplib is part of python whereas httplib2 needs to be installed by the user. Are you aware of any particular advantages that we need it for?

Infinite-PL:
Maybe I'm wrong but httplib is no longer supported on python 3.

peircej:
Ah, apparently so, so we will need to change at some point. Oh what about requests though?
http://docs.python-requests.org/en/master/
That's already being packaged in Standalone PsychoPy

Infinite-PL:
So you're suggesting to instead of using httplib , use requests?

peircej:
Yes, exactly. I think it does everything we need and more. I used `requests` for the pyosf lib which allows us to sync projects with the web from the psychopy app. It's already listed as a dependency for that reason and packaged within standalone (httplib2 is not).

There's also a note at the top of web.py saying that `mimetypes` is deprecated and we should use requests lib for that too. I don't know about that aspect.

PR #1311, opened 2016-11-22
Update builderFeedback.rst

KyleMorrissey:
Ok I hope that I am doing this right. I haven't used GitHub before. 
However there was a syntax error with this code component that causes the experiment to crash if a user response is not obtained. The suggestion that I have added resolves this problem. 

The original error that had been received (and appears resolved) was:
if len(key_resp.keys) <1:
TypeError: Object of type 'NoneType' has no len()


peircej:
Thanks Kyle. You're part of the team :-)
Feel free tom improve other parts of the docs too.

FYI I don't update the website daily or on every individual change - just when I get around to it. So changes don't always show up online immediately. On this occasion though I will push it up there in the next few minutes

PR #1319, opened 2016-12-12
change viewOri checking to != from is not

awctomlinson:
ex:

```
0.0 is not 0.  # this works, until assignments
>>> False

viewOri = 0.0
viewOri is not 0.  # this does not return the desired result
>>> True

viewOri != 0  # desired result
>>> False
```


peircej:
Thanks. Unfortunately the branch you'd done this in was quite out-dated and was failing the automated test suite. I tried merging but too much code was breaking so I've reapplied the fix to the current code instead.

PR #1323, opened 2016-12-16
ENH: pandas/xlrd importing of xlsx file for importConditions

peircej:
This is optional. Advantages:
 - more robust at ignoring blank cells
 - handles xls (old excel) files as well as xlsx
Cons:
 - might not be installed (new semi-dependency)


hoechenberger:
Great, thanks! I don't think deps are a problem: panda is part of the "official" scientific Python stack.

peircej:
Pandas was already a strict dependency but xlrd was not. It is easy to install with pip though

PR #1325, opened 2016-12-19
add locale setup to app (OS X)

### LINKED ISSUE #1315: psychopyApp won't start with Matplotlib 1.5 installed
See http://discourse.psychopy.org/t/mac-specific-help/1540/3

We need to figure out
- whether this problem is Anaconda-specific (and would require fixing upstream)
- whether this problem is Mac-specific

cjayb:
On OS X (ca. 10.10 +) in combination with at least iTerm2, the contents of `LC_ALL` is empty by the time `psychopy.compatibility` calls `import psychopy.data`. This leads to

```python
  File "/Users/cjb/miniconda_envs/psychopy/lib/python2.7/site-packages/matplotlib/__init__.py", line 1009, in _open_file_or_url
    with io.open(fname, encoding=encoding) as f:
LookupError: unknown encoding:
```
Due to `encoding` being `None`. It seems that the fix already exists in

`psychopy/locale_setup.py`

and only impacts OS X installations. This PR simply adds an import of it to `psychopyApp.py`, which fixes the problem for me (see [this gist](https://gist.github.com/cjayb/67d8f6ada85f8fcd58ff1d4e1fc20764) for the conda recipe reproducing my environment).

Update: I'm pretty sure this is related to #1315, so I'm going to be brave and say:

Closes #1315

### DIFF
--- a/psychopy/app/psychopyApp.py
+++ b/psychopy/app/psychopyApp.py
@@ -8,6 +8,8 @@ from __future__ import absolute_import, print_function
 
 import sys
 from psychopy.app._psychopyApp import PsychoPyApp, __version__
+# fix OS X locale-bug on startup: sets locale to LC_ALL (must be defined!)
+import psychopy.locale_setup  # noqa
 
 # NB the PsychoPyApp classes moved to _psychopyApp.py as of version 1.78.00
 # to allow for better upgrading possibilities from the mac app bundle. this


peircej:
Thanks. Also thanks for alerting me to the existence of # noqa for those imports to prevent flake getting upset - I must implement that in some files like https://github.com/psychopy/psychopy/blob/master/psychopy/visual/__init__.py which is, otherwise, a big pyflakes mess!

As a minor thing, it helps me if you use commit flags BF, FF or ENH at the beginning of your commits. BF is most important as I use it to find things that need back-porting to release branches (e.g. if I make a 1.84.3 I don't want all the enhancements (ENH)  or new feature fixes (FF) but I do want bug fixes (BF).

cjayb:
Will try to remember the flags, did read about them in your guides...  👍 

The #noqa is from contributing to [mne-python](https://github.com/mne-tools/mne-python) code: their CI-bots are pretty aggressive on flakiness :)

PR #1328, opened 2016-12-19
ENH scaling of window.viewPos according to window units

### LINKED ISSUE #1322: on the scaling of `psychopy.visual.window.viewPos`
I find the use of the `viewPos`-attribute of a `window` slightly unclear. I could manipulate it real time, but the units did not follow anything I specified (`pix` or `deg`). The `flip`-method says:

https://github.com/psychopy/psychopy/blob/89cce557d6daaab4fd7136de22f457322dc96b64/psychopy/visual/window.py#L653-L656

and after some testing I figured out that `viewPos` is in "normalised" coordinates. Presumably because `glTranslate` assumes this.

Am I correct in saying that the handling of `viewPos` should be checked against the current `units` of the window?

All that said, I'm pretty stoked about the ability to manipulate this parameter! /Chris

cjayb:
See discussion in #1322 

__NB__ `tests/test_all_visual/test_winScalePos.py` fails, predictably. Before fixing the test, I'd like to get a quick review on my implementation.

* should the value be (silently) clipped to +/- 1?
* would it be better to write a setter?

I dabbled with using `@attributeSetter` for `viewPos`, but got repeated

```python
TypeError: 'attributeSetter' object does not support indexing
```

Closes #1322

### DIFF
--- a/psychopy/tests/test_all_visual/test_winScalePos.py
+++ b/psychopy/tests/test_all_visual/test_winScalePos.py
@@ -1,51 +1,66 @@
 
 from psychopy import visual
 from psychopy.tests import utils
-import os, pytest
+import os
+import pytest
 
-v = [(1,1),(1,-1),(-1,-1),(-1,1)]  # vertices to use = square
+v = [(1, 1), (1, -1), (-1, -1), (-1, 1)]  # vertices to use = square
 n = 15  # size of the base square
-pimg = ( n, n)  # position for the image
-pgrn = (-n,-n)  # position for green square
+pimg = (n, n)  # position for the image
+pgrn = (-n, -n)  # position for green square
 img_name = os.path.join(utils.TESTS_DATA_PATH, 'filltext.png')
+
 
 class Test_Win_Scale_Pos_Ori(object):
     def setup_class(self):
-        self.win = visual.Window(size=(200,200), units='pix', allowGUI=False, autoLog=False)
+        self.win = visual.Window(size=(200, 200), units='pix',
+                                 allowGUI=False, autoLog=False)
+
     def teardown_class(self):
         self.win.close()
 
     @pytest.mark.scalepos
     def test_winScalePosOri(self):
         """test window.viewScale and .viewPos simultaneous
-        negative-going scale should mirror-reverse, and position should account for that
-        visually, the green square/rect should move clockwise around the text
+        negative-going scale should mirror-reverse, and position should
+        account for that visually, the green square/rect should move clockwise
+        around the text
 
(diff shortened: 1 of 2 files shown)


cjayb:
I managed to implement setting `viewPos` using `@attributeSetter` (code still messy, but functional; test fails for same reason as before)

The main issue I now have is that the setter does not get called when using `__iadd__`. This leads to the following bug:

```python
# assume win in 'pix' units and size (400, 400)
win.viewPos = [0, 100]  # this works: move up by 100 pixels
win.viewPos  # returns correct new normalised coordinate: [0., 0.5] 
win.viewPos[0] += 20  # expects to add 20 pixels in x-direction
win.viewPos  # returns: [20., 0.5] == WRONG, should be [20/400==0.05, 0.5]
```

I haven't found any examples in the code where a list-attribute would be set, nor is `__iadd__` defined anywhere in `psychopy`-proper. 

Any ideas?

peircej:
Yes, we don't currently support incrementing a subset of a property. Rather than doing `win.viewPos[0] += 20` you have to do `win.viewPos += [20,0]`

Right now your addition breaks the test suite. Maybe it doesn't correctly handle the case where `Window(...., viewPos=None)`?
https://travis-ci.org/psychopy/psychopy/builds/185992932

cjayb:
OK. Just to make sure I understand: by "don't currently support" you mean
> "if the the user does this, they're on their own"

Correct? In that case it should be easy enough for me to

- [ ] clean up
- [ ] fix the failing tests
- [ ] create new tests for the new functionality

peircej:
By "doesn't currently support" I mean I've never thought about that or tried that. It's a shame that it doesn't actually error rather than appearing to work but setting the wrong value :-/

cjayb:
I'm surprised (though delighted!) to see if Travis is happy! _Ready for review._

`test_winScalePos.py` passes on my Mac, but most in `test_all_stimuli` fail with stuff like below---also in `master`. Any pointers?

```python
E   ValueError: Monitor __blank__ has no known size in pixels (SEE MONITOR CENTER)
```

or

```python
test_all_visual/test_all_stimuli.py:462: in test_aperture_image
    aperture = visual.Aperture(win, size=1*self.scaleFactor,pos=[0.8*self.scaleFactor,0], shape=fileName)
../visual/aperture.py:124: in __init__
    win=self.win, vertices=vertices, fillColor=1, lineColor=None,
E   UnboundLocalError: local variable 'vertices' referenced before assignment
```

cjayb:
Oh, would it be easier for you to review if I squashed the commits first?

peircej:
Cool. Not sure why those tests are failing on you mac :-/

PR #1329, opened 2016-12-20
ENH: translate new dialog strings

hsogo:
"Paste Components" menu and "Online" tab of the Experiment Settings dialog are translated.


hsogo:
I close my request for now.
I want some English-native members to fix hint message of "Email address (for info emails)" (in app/builder/components/settings/__init__.py). It's the same as the hint message of "Output path".

PR #1331, opened 2016-12-22
Joystick error

ofgulban:
I was getting an error using a joystick and fixed it by following [a previously recommended workaround](https://groups.google.com/forum/#!msg/psychopy-users/CO37d-PMQ9k/iDDVAn0SD6UJ).

I have also realized that the example usage in the [docstring](http://www.psychopy.org/api/hardware/joystick.html) had a variable name error (```currentjoy.getX()
``` should be ```joy.getX()``` because ```currentjoy``` was not defined above). So have fixed it and took the liberty of changing some more docstrings in the same file to comply with PEP 257.

### DIFF
--- a/psychopy/hardware/joystick/__init__.py
+++ b/psychopy/hardware/joystick/__init__.py
@@ -1,4 +1,4 @@
-"""Control joysticks and gamepads from within PsychoPy
+"""Control joysticks and gamepads from within PsychoPy.
 
 You do need a window (and you need to be flipping it) for the joystick to be
 updated.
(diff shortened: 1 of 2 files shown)


peircej:
Thanks for sorting this out. Out of interest, what OS are you using and is this Standalone installation or your own Python? Just wondering what range of systems this fix might work for.
cheers, Jon

ofgulban:
Hi, thanks. This was done on a Win7 (64 bit) computer with standalone Psychopy (1.84.2) for testing a Current Designs fiber optic dial (connected via 932 Interface) in joystick mode.

PR #1332, opened 2016-12-31
Small unicode bug in keyNameFinder demo: 

dantheman39:
Added a u to make a string unicode in a code component, since the names of keys in other locales can be non-ASCII, and as a result the demo was crashing on a Turkish user


peircej:
Thanks Dan. Small thing for the future: could you prefix bug fixes with BF: so that I know to port them over to other branches if, for example, I release a bug-fix version. It's good to know very easily what is s bug fix, what's an enhancement (ENH) and what's a new-feature fix (FF) because ENH and FF shouldn't be copied to a bug-fix release. 
cheers

PR #1334, opened 2017-01-02
BUG: Fixing variable naming errors in sound module

lindeloev:
I encountered these and changed the variable names. I have searched all psychopy files to make sure that no other code use the variable names that I changed before this commit.


peircej:
Thanks, this was due to refactoring (moving sound backends into separate files).
On the plus side, the new sounddevice backend has improved testing that will catch such errors for this backend (travis can't test sound but the new backend only skips a few critical lines whereas the pyo backend skips the entire test)

PR #1336, opened 2017-01-09
Bug fix for dots.py

schofiaj:
The bugs are:

1. Dots that go out of bounds are replaced at a random location. When coherence is high this results in a shift in dot density toward the edge where most dots are going out of bounds. When coherence is low it causes a shift way from all borders. This is because the dots are disappearing at edges to keep the distribution of dots uniform they should be reborn at the opposite edge.

2. When noise dots are in walk mode the dot directions are not updated often enough such that as the coherence changes some noise dots become signal dots but retain the ransom direction they had when they were noise.

3. Dot locations are not refreshed between trials. If dot lifetime is long (esp infinite) one can see clusters of dots consistent from trial to trail. This accentuates the bug in (1) when signal dots reverse direction between trials as the accumulated dots on one side of the field then all charge back across the field together. 

I’ve fixed these by. 

1. When dots go out of bounds they are not set o dead but are rather replaced at –x,-y. So wherever they were going the reappear on the opposite side, corner etc. I step them in by one time sample so they don’t get trapped outside the field. This seems to make sense. A dot that would have gone one stepped beyond the field boundary is put back in an equivalent distance into the field on the opposite side.

2. I’ve added ‘walk’ to the list of noise dot types that require immediate direction updating when coherence is changed.

3. It’s conceivable that some people might prefer the option of keeping the sample of dots the same across trails so to retain this I’ve added a wrapper function to refresh the dots if required.


peircej:
OK the tests were failing for two reasons:

1. The new parameter in the graphical interface needing adding to the template (easy to fix and I've done that)
2. The wrapping around and then moving by one step raises a new problem - the dot can go outside the aperture as a result of the step. In the test suite this was made obvious by the fact that the speed was relatively high (so the step size was high relative to the aperture size) but it's also true in other cases and can be seen around the edges of the circular aperture

I also note a further problem with this method of returning dots to the mirror location, which is that it introduces a certain periodicity to the stimulus: each dot returns to its previous location/path every two sweeps of the stimulus (the points at which it enters/exits the aperture always form a rectangle). Again this is solved by having short lifetime dots but it's another thing to be wary of.

schofiaj:
1) Where can I find the template so that i don't make that mistake again?

2) I see. If a dot is born near the corner of a square or near the edge of a circle and travelling tangential to it and is going quite fast it may step way out of bounds on the first move, then be pulled back to the reflected edge but stepped forward and out of bounds again at which point it is not checked until the next move when the same thing will happen. This gets worse with increased speed. 

The further problem is something a knew about but felt was less of a worry that the drift tin dot density I was trying to fix as the dots need to persist for longer for the repeated cycle to become a problem.

However, I think that in fixing (2) I can fix this other issue also. My proposal would be to deliberately re-birth dots on the opposite edge of the aperture rather than reflecting them with a step of unpredictable size and to place them within a range of positions on that opposite edge so that they don't cycle round the same track if their lifetime is long but the net re-birth location balances the net death location for out of bounds dots. I'll see how that goes at least and do another pull request in due course.

peircej:
Cool. 

1. this is ugly (I'm happy to do it for others because I have to do it more often) but solved by
  ```
cd psychopy/tests/test_app/test_builder/
  python genComponsTemplate.py --out
```
2. yes, exactly. I wonder whether the step in from the edge after reflection just needs to be tiny. Isn't it only to deal with rounding error of the radius calculation, so a tiny step (e.g. fieldSize/100.0) would suffice?
3. I agree that this is a less-bad foul than the previous danger. People just shouldn't use long lifetimes except while debugging

:-)

schofiaj:
Ok I'll have another go at the dot replacement method for out of bounds dots.

PR #1337, opened 2017-01-11
Improvements to SoundDeviceSound and ImageStim

peircej:
SoundDeviceSound now supports MovieStim3

Images now check possible extensions:
  - first check if string is a note name
  - then check if is a file
  - if neither then, rather than raise error, we now check whether
    - removing the extension reveals a file
    - adding other common extensions (jpg, png, bmp etc) reveals file
Although it's possible that this will lead to an incorrect file, it will only do so when no option was available

### DIFF
--- a/psychopy/sound/backend_pyo.py
+++ b/psychopy/sound/backend_pyo.py
@@ -17,7 +17,7 @@ except ImportError as err:
     raise exceptions.DependencyError(repr(err))
 
 import threading
-
+pyoSndServer = None
 
 def init(rate=44100, stereo=True, buffer=128):
     """setup the pyo (sound) server
--- a/psychopy/sound/backend_sounddevice.py
+++ b/psychopy/sound/backend_sounddevice.py
@@ -110,7 +110,7 @@ class _SoundStream(object):
         self.takeTimeStamp = False
         self.frameTimes = range(5)  # DEBUGGING: store the last 5 callbacks
         self.lastFrameTime = time.time()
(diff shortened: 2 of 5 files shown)


lindeloev:
For the filename feature, perhaps also check case insensitive? Something like this:

```python
import os, glob
file_path = 'Desktop/cognitive biases.jpg'

# If file does not exist, search for alternatives
if not os.path.exists(file_path):
    folder, filename = os.path.split(file_path)
    files_in_folder = glob.glob(os.path.join(folder, '*'))  # get all files in file folder
    file_path = None  # The matching file path, if it is found
    
    # Loop through these files
    for file_in_folder in files_in_folder:
        # Loop through possible extensions
        for extension in ('.jpg', '.tif','.bmp'):
            filename_proposal = os.path.split(os.path.splitext(file_in_folder)[0])[1] + extension  # add extensions      
            
            # Compare in same case to make it case-insensitive
            if filename_proposal.lower() == filename.lower():
                file_path = file_in_folder  # yay, we found a match!

# Now do something with this file
if file_path is None:
    print 'Did not find'
else:
    print 'Found', file_path
```

peircej:
I see where you're coming from with that. Not sure I'm keen *if* we have to check each file individually (this could get lengthy because the factorial combination of a large folder and large number of potential extensions) but we might be able to do something with a case-insensitive alternative to glob:
 https://gist.github.com/techtonik/5694830

PR #1344, opened 2017-01-20
Psychojs fixes for usage of units

ilixa:
This commit fixes 2 issues:
- units is no longer ignored in BaseVisualStim constructor
- norm units fixed


peircej:
Simples! Thanks :-)

PR #1348, opened 2017-01-30
BF: TrialHandler2 in method='sequential' used reverse sequence

peircej:
Was using pop from the wrong end of the stack; needed to pop(0)

Also added another option of a null trial list that was apparently
valid in the original TrialHandler `trialList=[None]`


peircej:
Should add a test for this for future though - is an easy one to catch with a test!

PR #1354, opened 2017-02-10
Bugfixing sound

lindeloev:
I encountered these bugs on a Windows computer but fixed them using my linux (don't have a Windows nearby), so I was not able to verify that it works. At least it can't get more broken than it was ;-)

Wondering if *_bestDriver* should be moved from `sound/__init__.py` to `sound/backend_pyo.py` since that's the only place it's used?

### DIFF
--- a/psychopy/sound/backend_pyo.py
+++ b/psychopy/sound/backend_pyo.py
@@ -9,6 +9,7 @@ from psychopy import core, logging
 from psychopy.constants import (STARTED, PLAYING, PAUSED, FINISHED, STOPPED,
                                 NOT_STARTED, FOREVER)
 from ._base import _SoundBase
+from . import _bestDriver
 
 try:
     import pyo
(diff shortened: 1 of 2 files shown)


peircej:
Yes, its only pyo that currently offers a choice (of coreaudio or portaudio) but both are currently a bit broken on mac. Can't remember whether on win32 the setting is useful. Otherwise, I'm tempted to get rid of the setting altogether and hardcode it to use portaudio.

Actually, I'm hoping we can get rid of pyo in psychoypy entirely. It's become my "new avbin"! ;-)

lindeloev:
Your call, sounds good :-) `sounddevice` actually fixed an error on my Ubuntu that I thought originated from portaudio! So I can suddenly play sounds again. So I won't be missing pyo. However, I did encounter a problem with 1.85-rc3 sounddevice on a 4-year old Win7 laptop the other day, where it continuously played a saw-tooth-like loud sound and then the "real" sounds on top of that. I didn't have time to look into it, unfortunately. But it's a sign that `sounddevice` probably needs some testing.

peircej:
Oh for sure it will take us time to get `sounddevice` all correct

PR #1357, opened 2017-02-26
Expose multisampling support through visual.Window class

mdcutone:
This exposes pyglet's mutli-sample AA support to the visual.Window class, adding the multiSample and numSample arguments. Currently, these settings cannot be changed after a window is created. I found multi-sampling useful when drawing complex scenes and others might benefit from this addition.


mdcutone:
Just a note, multi sampling only applies to drawing operation on the back buffer (applied at rasterization). MSAA will not be applied to stimuli drawn to FBOs. It will take a bit of work to get MSAA working within FBOs, I've done it before but it will require some major reworking of existing code to manage.

PR #1362, opened 2017-03-04
NF: Global Keyboard Events

hoechenberger:
Having to check manually for keypresses to abort the experiment is cumbersome and error-prone; and I often forget to add this feature for parts of my code and have to resort to the OS task manager to kill the PsychoPy fullscreen window if I need to quit my experiment prematurely.

This PR adds the ability to specify a _shutdown key_ in the preferences, which will then be checked for during all `event._onPygletKey()` invocations. `event._onPygletKey()` is the pyglet window's keyboard event handler. These events are dispatched on each `flip()` and during `wait()` periods when using `wait(X, hogCPUperiod=X)`. If the shutdown key is pressed, a `CRITICAL` event will be logged, and `core.quit()` will be called.

The PR is to be considered a proposal and is up for discussion. Please let me know what you think of the general idea as such, and of the specific implementation.

### DIFF
--- a/docs/source/coder/globalKeys.rst
+++ b/docs/source/coder/globalKeys.rst
@@ -1,0 +1,314 @@
+Global Event Keys
+=================
+
+Global event keys are single keys (or combinations of a single key and one or
+more "modifier" keys such as Ctrl, Alt, etc.) with an associated Python
+callback function. This function will be executed if the key (or
+key/modifiers combination) was pressed.
+
+.. note::
+
+   Global event keys only work with the `pyglet` backend, which is the default.
+
+PsychoPy fully automatically monitors and processes key presses during most
+portions of the experimental run, for example during
+`core.wait()` periods, or when calling `win.flip()`. If a global
+event key press is detected, the specified function will be run
+immediately. You are not required to manually poll and check for key
+presses. This can be particularly useful to implement a global
+"shutdown" key, or to trigger laboratory equipment on a key press
+when testing your experimental script -- without cluttering the code.
+But of course the application is not limited to these two scenarios.
+In fact, you can associate any Python function with a global event key.
+
+All active global event keys are stored in `event.globalKeys`.
+
+Adding a global event key (simple)
+----------------------------------
+First, let's ensure no global event keys are currently set by calling
+func:`event.globalKeys.clear`.
+::
+    >>> from psychopy import event
+    >>> event.globalKeys.clear()
+
+To add a new global event key, you need to invoke
+func:`event.globalKeys.add`. This function has two required arguments: the
+key name, and the function to associate with that key.
+::
+    >>> key = 'a'
+    >>> def myfunc():
+    ...     pass
+    ...
+    >>> event.globalKeys.add(key=key, func=myfunc)
+
+Look at `event.globalKeys`, we can see that the global event key has indeed
+been created.
+::
+
+    >>> event.globalKeys
+    <_GlobalEventKeys :
+        [A] -> 'myfunc' <function myfunc at 0x10669ba28>
+    >
+
+Your output should look similar. You may happen to spot
+We can take a closer look at the specific global key event we added.
+::
+
+    >>> event.globalKeys['a']
+    _GlobalEvent(func=<function myfunc at 0x10669ba28>, func_args=(), func_kwargs={}, name='myfunc')
+
+This output tells us that
+
+- our key `a` is associated with our function `myfunc`
+
+- `myfunc` will be called without passing any positional or keyword
+  arguments (`func_args` and `func_kwargs`, respectively)
+
+- the event name was automatically set to the name of the function.
+
+.. note::
+
+   Pressing the key won't do anything unless a :class:`psychopy.visual.Window`
+   is created and and its :func:~`psychopy.visual.Window.flip` method or
+   :func:`psychopy.core.wait` are called.
+
+Adding a global event key (advanced)
+------------------------------------
+We are going to associate a function with a more complex calling signature
+(with positional and keyword arguments) with a global event key. First, let's
+create the dummy function:
+::
+
+    >>> def myfunc2(*args, **kwargs):
+    ...     pass
+    ...
+
+Next, compile some positional and keyword arguments and a custom name for this
+event. Positional arguments must be passed as tists or uples, and keyword
+arguments as dictionaries.
+::
+
+    >>> args = (1, 2)
+    >>> kwargs = dict(foo=3, bar=4)
+    >>> name = 'my name'
+
+.. note::
+
+   Even when intending to pass only a single positional argument, `args` must be
+   a list or tuple, e.g., `args = [1]` or `args = (1,)`.
+
+
+Finally, specify the key and a combination of modifiers. While key names are
+just strings, modifiers are lists or tuples of modifier names.
+::
+
+    >>> key = 'b'
+    >>> modifiers = ['ctrl', 'alt']
+
+.. note::
+
+   Even when specifying only a single modifier key, `modifiers` must be a list
+   or tuple, e.g., `modifiers = ['ctrl']` or `modifiers = ('ctrl',)`.
+
+We are now ready to create the global event key.
+::
+
+    >>> event.globalKeys.add(key=key, modifiers=modifiers,
+    ... func=myfunc2, func_args=args, func_kwargs=kwargs,
+    ... name=name)
+
+Check that the global event key was successfully added.
+::
+
+    >>> event.globalKeys
+    <_GlobalEventKeys :
+        [A] -> 'myfunc' <function myfunc at 0x10669ba28>
+        [CTRL] + [ALT] + [B] -> 'my name' <function myfunc2 at 0x112eecb90>
+    >
+
+The key combination `[CTRL] + [ALT] + [B]` is now associated with the function
+`myfunc2`, which will be called in the following way:
+::
+
+    myfunc2(1, 2, foo=2, bar=4)
+
+.. _indexing:
+
+Indexing
+--------
+`event.globalKeys` can be accessed like an ordinary dictionary. The index keys
+are `(key, modifiers)` namedtuples.
+::
+
+    >>> event.globalKeys.keys()
+    [_IndexKey(key='a', modifiers=()), _IndexKey(key='b', modifiers=('ctrl', 'alt'))]
+
+To access the global event associated with the key combination
+`[CTRL] + [ALT] + [B]`, we can do
+
+    >>> event.globalKeys['b', ['ctrl', 'alt']]
+    _GlobalEvent(func=<function myfunc2 at 0x112eecb90>, func_args=(1, 2), func_kwargs={'foo': 3, 'bar': 4}, name='my name')
+
+To make access more convenient, specifying the modifiers is optional in case
+none were passed to :func:`psychopy.event.globalKeys.add` when the global
+event key was added, meaning the following commands are identical.
+::
+
+    >>> event.globalKeys['a', ()]
+    _GlobalEvent(func=<function myfunc at 0x10669ba28>, func_args=(), func_kwargs={}, name='myfunc')
+    >>> event.globalKeys['a']
+    _GlobalEvent(func=<function myfunc at 0x10669ba28>, func_args=(), func_kwargs={}, name='myfunc')
+
+All elements of a global event can be accessed directly.
+::
+
+    >>> index = ('b', ['ctrl', 'alt'])
+    >>> event.globalKeys[index].func
+    <function myfunc2 at 0x112eecb90>
+    >>> event.globalKeys[index].func_args
+    (1, 2)
+    >>> event.globalKeys[index].func_kwargs
+    {'foo': 3, 'bar': 4}
+    >>> event.globalKeys[index].name
+    'my name'
+
+Number of active event keys
+---------------------------
+The number of currently active event keys can be retrieved by passing
+`event.globalKeys` to the `len()` function.
+::
+
+    >>> len(event.globalKeys)
+    2
+
+Removing global event keys
+--------------------------
+There are three ways to remove global event keys:
+
+- using :func:`psychopy.event.globalKeys.remove`,
+- using `del`, and
+- using :func:`psychopy.event.globalKeys.pop`.
+
+:func:`psychopy.event.globalKeys.remove`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+To remove a single key, pass the key name and modifiers (if any) to
+:func:`psychopy.event.globalKeys.remove`.
+::
+
+    >>> event.globalKeys.remove(key='a')
+
+A convenience method to quickly delete *all* global event keys is to pass
+`key='all'`
+::
+
+    >>> event.globalKeys.remove(key='all')
+
+`del`
+~~~~~
+Like with other dictionaries, items can be removed from `event.globalKeys`
+by using the `del` statement. The provided index key must be specified as
+described in :ref:`indexing`.
+::
+
+    >>> index = ('b', ['ctrl', 'alt'])
+    >>> del event.globalKeys[index]
+
+:func:`psychopy.event.globalKeys.pop`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Again, as other dictionaries, `event.globalKeys` provides a `pop` method to
+retrieve an item and remove it from the dict. The first argument to `pop` is the
+index key, specified as described in :ref:`indexing`. The second argument is
+optional. Its value will be returned in case no item with the matching indexing
+key could be found, for example if the item had already been removed previously.
+::
+
+    >>> r = event.globalKeys.pop('a', None)
+    >>> print(r)
+    _GlobalEvent(func=<function myfunc at 0x10669ba28>, func_args=(), func_kwargs={}, name='myfunc')
+    >>> r = event.globalKeys.pop('a', None)
+    >>> print(r)
+    None
+
+Global shutdown key
+-------------------
+The PsychoPy preferences for `shutdownKey` and `shutdownKeyModifiers`
+(both unset by default) will be used to automatically create a global
+shutdown key. To demonstrate this automated behavior, let us first change
+the preferences programmatically (these changes will be lost when quitting the
+current Python session).
+::
+
+    >>> from psychopy.preferences import prefs
+    >>> prefs.general['shutdownKey'] = 'q'
+
+We can now check if a global shutdown key has been automatically created.
+::
+
+    >>> from psychopy import event
+    >>> event.globalKeys
+    <_GlobalEventKeys :
+        [Q] -> 'shutdown (auto-created from prefs)' <function quit at 0x10c171938>
+    >
+
+And indeed, it worked!
+
+What happened behind the scences? When importing the `psychopy.event`
+module, the initialization of `event.globalKeys` checked for valid shutdown key
+preferences and automatically initialized a shutdown key accordingly.
+This key is associated with the :func:~`psychopy.core.quit` function, which will
+shut down PsychoPy.
+::
+
+   >>> from psychopy.core import quit
+   >>> event.globalKeys['q'].func == quit
+   True
+
+Of course you can very easily add a global shutdown key manually, too. You
+simply have to associate a key with :func:~`psychopy.core.quit`.
+::
+
+    >>> from psychopy import core, event
+    >>> event.globalKeys.add(key='q', func=core.quit, name='shutdown')
+
+That's it!
+
+A working example
+-----------------
+In the above code snippets, our global event keys were not actually functional,
+as we didn't create a window, which is required to actually collect the key
+presses. Our working example will thus first create a window and then add
+global event keys to change the window color and quit the experiment,
+respectively.
+::
+
+    #!/usr/bin/env python2
+    # -*- coding: utf-8 -*-
+
+    from __future__ import print_function
+    from psychopy import core, event, visual
+
+
+    def change_color(win, log=False):
+        win.color = 'blue' if win.color == 'gray' else 'gray'
+        if log:
+            print('Changed color to %s' % win.color)
+
+
+    win = visual.Window(color='gray')
+    text = visual.TextStim(win,
+                           text='Press C to change color,\n CTRL + Q to quit.')
+
+    # Global event key to change window background color.
+    event.globalKeys.add(key='c',
+                         func=change_color,
+                         func_args=[win],
+                         func_kwargs=dict(log=True),
+                         name='change window color')
+
+    # Global event key (with modifier) to quit the experiment ("shutdown key").
+    event.globalKeys.add(key='q', modifiers=['ctrl'], func=core.quit)
+
+    while True:
+        text.draw()
+        win.flip()
+
--- a/psychopy/demos/coder/basic/global_event_keys.py
+++ b/psychopy/demos/coder/basic/global_event_keys.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+"""
+This demo shows how to use "global event keys" to interact with a running
+experiment. The keys do not have to be checked manually, but
+rather get handled automatically by PsychoPy.
+
+This works when calling win.flip() repeatedly (as in this example)
+and when using core.wait() (not demonstrated here). If using core.wait(),
+be sure to set the `hogCPUperiod` parameter equal to the entire
+waiting duration, e.g. `core.wait(10, hogCPUperiod=10)`.
+
+"""
+
+from __future__ import print_function
+from psychopy import core, event, visual
+
+
+win = visual.Window()
+rect = visual.Rect(win, fillColor='blue', pos=(0, -0.2))
+text = visual.TextStim(
+    win,
+    pos=(0, 0.5),
+    text=('Press\n\n'
+          'B for blue rectangle,\n'
+          'CTRL + R for red rectangle,\n'
+          'Q or ESC to quit.'))
+
+# Add an event key.
+event.globalKeys.add(key='b', func=setattr,
+                     func_args=(rect, 'fillColor', 'blue'),
+                     name='blue rect')
+
+# Add an event key with a "modifier" (CTRL).
+event.globalKeys.add(key='r', modifiers=['ctrl'], func=setattr,
+                     func_args=(rect, 'fillColor', 'red'),
+                     name='red rect')
+
+# Add multiple shutdown keys "at once".
+for key in ['q', 'escape']:
+    event.globalKeys.add(key, func=core.quit)
+
+# Print all currently defined global event keys.
+print(event.globalKeys)
+print(repr(event.globalKeys))
+
+while True:
+    text.draw()
+    rect.draw()
+    win.flip()
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -12,8 +12,11 @@ See demo_mouse.py and i{demo_joystick.py} for examples
 from __future__ import absolute_import
 
 import sys
+import string
 import copy
 import numpy
+from collections import namedtuple, OrderedDict, MutableMapping
+from psychopy.preferences import prefs
 
 # try to import pyglet & pygame and hope the user has at least one of them!
 try:
@@ -143,6 +146,27 @@ def _onPygletKey(symbol, modifiers, emulated=False):
         keySource = 'Keypress'
     _keyBuffer.append((thisKey, modifiers, keyTime))  # tuple
     logging.data("%s: %s" % (keySource, thisKey))
+    _process_global_event_key(thisKey, modifiers)
+
+
+def _process_global_event_key(key, modifiers):
+    # The statement can be simplified to:
+    # `if modifiers == 0` once PR #1373 is merged.
+    if (modifiers is None) or (modifiers == 0):
+        modifier_keys = ()
+    else:
+        modifier_keys = ['%s' % m.strip('MOD_').lower() for m in
+                         (pyglet.window.key.modifiers_string(modifiers)
+                          .split('|'))]
+
+    index_key = globalKeys._gen_index_key((key, modifier_keys))
+
+    if index_key in globalKeys:
+        event = globalKeys[index_key]
+        logging.exp('Global key event: %s. Calling %s.'
+                    % (event.name, event.func))
+        r = event.func(*event.func_args, **event.func_kwargs)
+        return r
 
 
 def _onPygletMousePress(x, y, button, modifiers, emulated=False):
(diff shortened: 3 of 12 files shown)


psychopy/demos/coder/basic/shutdown_key.py
```diff
@@ -0,0 +1,35 @@
...
+entire waiting duration.
+"""
+
+from psychopy.visual import Window
+from psychopy.clock import CountdownTimer
+from psychopy.visual.text import TextStim
+from psychopy.preferences import prefs
+
+prefs.general['enableShutdownKey'] = True
+prefs.keys['shutdownKey'] = 'escape'
```
lindeloev: Wouldn't it be cleaner to have `prefs.keys['shutdownKeys'] = ['escape', 'f9']` and then an empty list means that no keys will shut it down? Instead of having a separate `enbleShytdownKey` preference that always needs to be followed by setting another variable.

I like the suggestion! It is indeed a pain to check manually.

hoechenberger:
@lindeloev We now support specification of multiple keys. Setting `shutdownKey = ''` disables the functionality. Haven't checked with modifier keys so far, would probably require some additional changes. Would we need this, anyway?

hoechenberger:
Now Travis is unhappy, because `builder.py` traverses through all key bindings, doing
```python
        keys = {k: self.app.keys[k].replace('Ctrl+', ctrlKey)
                for k in self.app.keys}
```
Now since we specify a list, our setting doesn't provide a `replace()` method anymore.

I think there's now three options:
- move the settiong outside of the `[keyBindings]` scope, e.g. to `[general]` (most functional)
- switch back to specifying the shutdown key as a string, which would then limit us to a single key again (seems the cleanest)
- change `builder.py` to ignore the `shutdownKeys` setting (seems kind of ugly/hackish?)

psychopy/demos/coder/basic/shutdown_key.py
lindeloev: Yes, this behavior makes a lot of sense to me. Good work.

lindeloev:
Regarding modifiers: I have no strong opinions but it's probably a very low priority. There are enough obscure keys available that you don't need modifiers to make the action hidden/inaccessible. I usually use F9 out of fear that subjects will press escape out of frustration :-)

hoechenberger:
@lindeloev I think this should be good now, WDYT?

psychopy/tests/test_misc/test_event.py
```diff
@@ -226,6 +227,20 @@ def setup_class(self):
         self.win = Window([128,128], monitor=mon, winType='pyglet', pos=[50,50], autoLog=False)
         assert pygame.display.get_init() == 0
 
+    def test_shutdown_key(self):
+        key = 'escape'
```
lindeloev: Sorry, too busy to clone and test this myself now, but wouldn't this fail? I.e. setting it to a string instead of a list would cause `if thisKey in prefs.general['shutdownKeys']:` in `_onPygletKey` to loop through characters rather than list items. In other words: shouldn't the API here only allow to set `prefs.general['shutdownKeys']` to be a list? Perhaps this should be tested explicitly on assignment. Haven't played with the `psychopy.preferences` before, but looking at it, there's a generic setter which makes such a check convoluted. So we'll probably just leave it for now.

psychopy/demos/coder/basic/shutdown_key.py
```diff
@@ -0,0 +1,35 @@
...
+    win,
+    text='Press ESC or F9 to quit prematurely.',
+    pos=(0, -0.2))
+
+timer.reset(10)
+while timer.getTime() > 0:
+    text_timer.text = 'Auto-quitting in %i sec.' % round(timer.getTime())
+    text_timer.draw()
+    text_esc.draw()
+    win.flip()
```
lindeloev: I wonder if this whole demo could be made shorter to highlight the critical aspect. I'm thinking something like this (untested)

```python
    from psychopy import visual, preferences
    preferences.prefs.general['shudownKeys'] = ['escape', 'f9']  # set quit keys here.

    win = visual.Window()
    stim = visual.TextStim(win)
    for frame in range(600):  # 10 seconds on a 60 Hz monitor
        stim.text = u'Displaying frame %i of 600. Press %s to quit prematurely: %s' %(frame, ' or '.join(prefs.general['shutdownKeys']))
        stim.draw()
        win.flip()
```

psychopy/tests/test_misc/test_event.py
hoechenberger: > wouldn't this fail without a crash? I.e. setting it to a string instead of a list would cause if thisKey in prefs.general['shutdownKeys']: in _onPygletKey to loop through characters rather than list items. In other words: shouldn't the API here only allow to set prefs.general['shutdownKeys'] to be a list?

This is exactly the reason why I'm including this test :) This test will in fact only pass if `SystemExit` is raised, which will only happen if `core.quit()` is called, which in turn only happens (during this test) if our shutdown key was properly handled. See the next lines with the `pytest.raises()` context handler. I was also surprised it worked like this... So either the test is somehow broken or the behavior is actually what we really want! I can check more thoroughly tomorrow or on Tuesday.

psychopy/demos/coder/basic/shutdown_key.py
hoechenberger: You're probably right, I will try to downsize the demo a bit.

psychopy/tests/test_misc/test_event.py
hoechenberger: Ok so there's less magic involved than I hoped:
```python
In [1]: from psychopy.preferences import prefs

In [2]: prefs.general['shutdownKeys'] = 'foo'

In [3]: prefs.general['shutdownKeys']
Out[3]: 'foo'
```
Hence I guess the test is not working properly.

hoechenberger: It's not a mystery at all:
```python
In [4]: keys = 'escape'

In [5]: 'escape' in keys
Out[5]: True
```
But we do get side-effects:
```python
In [6]: 'e' in keys
Out[6]: True
```

Now I wonder – do we _really_ need support for a _list_ of keys? Or wouldn't a single shutdown key suffice? Then we wouldn't need special handling / have to force users to provide even single keys as a list.

lindeloev:
It's just that keys are typed as lists elsewhere in psychopy e.g. in
event.getKeys. I think you're right that the users of this function would
use one key 95% of the time, so my suggestion was mostly one about
consistency. I'd like to lean on the wisdom of @jon here :-) Jon, what do
you think of the idea in general?

Best,
Jonas

______________________________

Jonas Kristoffer Lindeløv, M.Sc., Ph.D.
Assistant Professor in Cognitive Neuroscience and Neuropsychology
Profile at Aalborg University <http://personprofil.aau.dk/117060>, Scientific
blog <http://lindeloev.net/>

hoechenberger:
> It's just that keys are typed as lists elsewhere in psychopy e.g. in
> event.getKeys.

Yes, but certainly not in the _preferences_: https://github.com/psychopy/psychopy/blob/5029631173510fdb7b87fa7eda632dcc1ff4298d/psychopy/preferences/baseNoArch.spec

Eventually, this setting would probably be exposed through a settings dialog, and users wouldn't have to set the key manually in each experimental script.

lindeloev:
Those are key bindings for the PsychoPy interface; experiments could be
different. I actually thought of it as something I'd use on an
experiment-by-experiment basis. But anyway, I really have no opinion on
this; both are very good advancements, I think! (single keys, multiple
keys, generel preference, experiment preference)

______________________________

Jonas Kristoffer Lindeløv, M.Sc., Ph.D.
Assistant Professor in Cognitive Neuroscience and Neuropsychology
Profile at Aalborg University <http://personprofil.aau.dk/117060>, Scientific
blog <http://lindeloev.net/>

hoechenberger:
> I actually thought of it as something I'd use on an
> experiment-by-experiment basis.

We could add a convenience function for that:
```python
# core.py
from psychopy.preferences import prefs


def setShtdownKey(keys=None):
    if keys is None:
        prefs.general['shutdownKeys'] = []
    elif isinstance(keys, (str, unicode):
        prefs.general['shutdownKeys'] = [keys]
    else:
        prefs.general['shutdownKeys'] = keys
```

peircej:
To the specific question of list versus single-key:

- I don't think this needs to be the same as getKeys() and waitKeys() or the same as the preferences options. It seems like a different type of thing to me
- for getKeys/waitKeys I think it's really important that they always return a list even though it's nearly always a list of 1. Otherwise the user has to write code that checks the type of the return before using it. And if they switch from a getKeys() call to a waitKeys() call they might have to change their code, which is annoying. So both should be the same and both should have a constant return type (therefore list)
- for your case, here, my sense is that the user isn't actually going to write the code that cares about it being a list/str/unicode. They won't need to read it so don't care how it's stored? I don't care so much whether the code within the lib has to jump through some extra hoops (type-checking) to make things more intuitive for the user

So I think this one can be flexible (set to a list or string as the user prefers)

peircej:
For the question of whether it should be on by default, for the Builder I made the decision that it should be on by default (in fact, we didn't even have an option to turn it off at one point). So many people accidentally set a trial to last forever and forget to provide a way to quit, that I thought the benefits of an exit key outweighed the disadvantage.

I think for Builder experiments it should be kept as 'escape' by default and we would need to convert the current boolean "Enabled" setting appropriately ('' for False, 'escape' for True)

lindeloev:
I really like the idea of implementing this in the `event` module for experiment-specific implementations! It could be implemented as an attribute setter:

```python
    from psychopy.tools.attributetools import attributeSetter, setAttribute

    @attributeSetter
    def shutdownKeys(self, keys):
        """Docstring here"""
        # stuff here

    def setShutdownKeys(keys, log):
        """As event.shutdownKeys but with explicit control of logging"""
        setAttribute(self, 'shutdownKeys', keys, log)
       
```

Then one could do 

```python
    event.shutdownKeys = 'escape'
    print event.shutdownKeys
```

For later, I think this could be generalized even further! So that any keys could be mapped to any function during the experiment, like `event.setGlobalKeys(['f9'], show_debriefing, 'Thank you for participating!')` where show_debriefing is a function that takes one kwarg. But it's probably overkill - just wanted to share the idea ;-)

hoechenberger:
Hi Jonas, I had written the following comment earlier today, but somehow forgot to actually post it... Just posting now for reference:

Thanks for you feedback, @peircej and @lindeloev!

I have changed the implementation, trying to incorporate your suggestions:
- We have a preferences setting `general['shutdownKeys]`. By default, this is an empty list (meaning shutdown keys are disabled)
- I added a function `event.setShutdownKeys()` to enable / set shutdown keys. This function can handle:
    - lists of keys
    - "individual" keys (strings)
    - a `None` value to disable all shutdown keys
- I've simplified the demo.

> I think for Builder experiments it should be kept as 'escape' by default and we would need to convert the current boolean "Enabled" setting appropriately ('' for False, 'escape' for True)

My current idea was not to interfere with Builder experiments in any way -- shutdown keys are disabled by default -- since they already automatically perform escape key checking, as you pointed out. I would propose to add Builder support at a later time, if required.

hoechenberger:
> I really like the idea of implementing this in the event module for experiment-specific implementations! It could be implemented as an attribute setter:

Very clever proposal, I _love_ that idea! And since we use setters, we wouldn't have to resort to using global variables in the module, but could "hide" the storage details (keeping the shutdown keys setting in the preferences object). Very neat!

> For later, I think this could be generalized even further! So that any keys could be mapped to any function during the experiment, like event.setGlobalKeys(['f9'], show_debriefing, 'Thank you for participating!') where show_debriefing is a function that takes one kwarg. But it's probably overkill - just wanted to share the idea ;-)

Neat! I imagine mapping functions for control of lab equipment, so that for example you could create an EEG trigger on key press, or start or stop a device, ...

hoechenberger:
However, if we implement this, we _must_ think of adding modifier key support, or else we're going to run out of keys quickly ;)

hoechenberger:
> And since we use setters, we wouldn't have to resort to using global variables in the module, but could "hide" the storage details (keeping the shutdown keys setting in the preferences object).

It appears using getters/setters at a module (not class) level is not super straightforward. Need to think some more about this.

hoechenberger:
For reference: http://www.dr-josiah.com/2013/12/properties-on-python-modules.html

lindeloev:
> It appears using getters/setters at a module (not ordinary class) level is not super straightforward. Need to think some more about this.

Oh, maybe then something like `event.globalKeys.shutdown`? It's a bit of an overkill if it only contains one variable...

Per my earlier suggestion, `event.setGlobalKeys('escape', core.quit)` would be equivalent to the `shutdownKeys`, but probably shouldn't replace it since the former requires importing `core` and it's more complicated - not something that would be obvious enough when reading the API or other docs.

peircej:
Interesting idea. We'd end up effectively with an optional event-driven system!

hoechenberger:
What about sth. like this:
```python

_GlobalKeyEvent = namedtuple('_GlobalKeyEvent', ['name', 'keys', 'func'])


class _GlobalKeyEvents(object):
    def __init__(self):
        self._events = []
        self.add('shutdown', prefs.general['shutdownKeys'], core.quit)

    def __repr__(self):
        return '%s' % [event for event in self._events]

    def add(self, name, keys, func):
        if name in [x.name for x in self._events]:
            msg = 'An event with name `%s` already exists.' % name
            raise ValueError(msg)
        elif keys in [x.keys for x in self._events]:
            msg = 'Key(s) `%s` already assigned.' % keys
            raise ValueError(msg)
        elif func in [x.func for x in self._events]:
            msg = 'Function `%s` is already assigned'
            raise ValueError(msg)

        self._events.append(_GlobalKeyEvent(name, keys, func))

    def remove(self, name):
        if name not in [x.name for x in self._events]:
            msg = 'An event with name `%s` does not exist.' % name
            raise ValueError(msg)

        for event_num, event in enumerate(self._events):
            if event.name == name:
                self._events.pop(event_num)
                break  # Names are unique.

    @property
    def events(self):
        return self._events

globalKeys = _GlobalKeyEvents()


def setShutdownKeys(keys, replace=True):
    if 'shutdown' in [x.name for x in globalKeys._events]:
        if replace:
            globalKeys.remove('shutdown')
        else:
            raise ValueError('Shutdown key already assigned.')

    globalKeys.add('shutdown', keys, core.quit)
```

```python
In[72]: setShutdownKeys('foo')
In[73]: globalKeys
Out[73]: [_GlobalKeyEvent(name='shutdown', keys='foo', func=<function quit at 0x103dd6aa0>)]
```

hoechenberger:
Oh, forgot the shutdown key attribute setter, again. Could go to `event.globalKeys.shutdown`, like @lindeloev suggested.

hoechenberger:
Then again, I'm not sure _shutdown keys_ are special enough to receive that much of a special treatment, compared to other key events. Maybe providing `event.setShutdownKeys()` as a convenience function is sufficient.

hoechenberger:
Check out https://github.com/hoechenberger/psychopy/blob/7faf529b6e9d77354c8ada4b77746e5cdbb5e596/psychopy/demos/coder/basic/global_event_keys.py
I like it :)

lindeloev:
Holy moses, Richard, you're on fire! This is awesome! This could greatly simplify code in many situations.

I wonder if some code simplification is possible by having `GlobalKeys._events` be a dictionary where the events.keys() are keyboard keys and event.values() are lists of "event" named tuples.

1. Instead of being the hidden `_keys` it could be a "proper attribute" `event`, avoiding the need for @property and __repr__. This would also allow for more flexibility than if hiding it away.
2. Removing a key would simply be `self.events.pop(key, None)`
3. Checking for keys in event.py would simply be
```python
    if thisKey in globalKeys.events.keys():
        for event in globalKeys[thisKey]:
            #logging and event.func
```

lindeloev:
I guess that `psychopy/demos/coder/basic/global_shutdown_keys.py` is not necessary anymore since that functionality is demoed in psychopy/demos/coder/basic/global_event_keys.py?

hoechenberger:
> Holy moses, Richard, you're on fire! 

Don't let the large number of commits fool you: I'm still quite undecided about which approach(es) would be best, constantly changing my attitude, introducing (and trying to fix) bugs, ... ;)

> This is awesome! This could greatly simplify code in many situations.

Yes, I also believe there's great potential! :)

> I wonder if some code simplification is possible by having GlobalKeys._events be a dictionary where the events.keys() are keyboard keys and event.values() are lists of "event" named tuples.

> Instead of being the hidden _keys it could be a "proper attribute" event, avoiding the need for @property and repr. This would also allow for more flexibility than if hiding it away.
> Removing a key would simply be self.events.pop(key, None)
> Checking for keys in event.py would simply be
>     if thisKey in globalKeys.events.keys():
>         for event in globalKeys[thisKey]:
>             #logging and event.func

I've actually played with different approaches here -- I actually started with a dictionary-like implementation first -- but couldn't find any solution I would really consider satisfactory. Just to get this straight, what you're proposing is sth like:
```python
GlobalEvent = namedtuple('GlobalEvent', ['func', 'args', 'kwargs'])

class GlobalKeyEvents(object):
    def __init__(self):
        self.events = []
        ...

    def add(self, keys, func, *args, **kwargs):
        for key in keys:
            self.events[keys] = GlobalEvent(name, func, args, kwargs)

    def remove(self, key):
        self.events.pop(key, None)
```
Generally I like that, but if an event was assigned to multiple keys, the user would have to remove each of those keys individually, because they could not remove the event by name any longer.

Generally I would love to have sth. like a lookup table that works in _both_ directions (keys -> events, and vice versa). In the most simple case, one could even use lists for that:

```python
keys = ['a', 'b', 'c']
names = ['shutdown', 'shutdown', 'trigger']

shutdown_keys = keys[names.index('shutdown')]
```
but this also comes with serious drawbacks.

hoechenberger:
> having GlobalKeys._events be a dictionary where the events.keys() are keyboard keys and event.values() are lists of "event" named tuples.

`event.values()` cannot be lists, because then each key could be assigned to multiple events/function :) So a key->event mapping must always be one-to-one, whereas a event->key mapping may be one-to-may.

lindeloev:
I could definitely see situations where I want a cascade of events on a key press, e.g. updating a stimulus, sending a trigger, and resetting a clock. But the user could, of course, also just make a function that runs these actions and then bind that function to a key.

I see the problem with multiple keys per action. For example, if you want something to happen on every press on an arrow key. Dictionary keys can be anything, so the dictionary `{['down', 'up']: my_tuple}` is possible, but then the code would get convoluted again.

hoechenberger:
> I could definitely see situations where I want a cascade of events on a key press, e.g. updating a stimulus, sending a trigger, and resetting a clock. But the user could, of course, also just make a function that runs these actions and then bind that function to a key.

I'm afraid the API would lose simplicity and couldn't be used intuitively anymore if we added support for multiple events per key:
```python
globalKeys.add(keys=['a', 'b', 'c'], funcs=[foo, bar, baz])
```
Now would we map each key to the function with the same index, or would we map each key to all functions, ...

Of course we could do:
```python
globalKeys.add(keys=['a', 'b', 'c'], func=foo)
globalKeys.add(keys=['a', 'b', 'c'], func=bar)
globalKeys.add(keys=['a', 'b', 'c'], func=baz)
```
In any case, we should probably offer a way to remove any of those events, and this is where things start to get really messy... I believe users should simply create a function to initiate cascades of actions.

> 
> I see the problem with multiple keys per action. For example, if you want something to happen on every press on an arrow key. Dictionary keys can be anything, so the dictionary {['down', 'up']: my_tuple} is possible, but then the code would get convoluted again.

Yep...

peircej:
So, just to recap, since this has migrated a little. In the original plan, wasn't it going to be a system-wide setting? In the current version you still need to insert the line `event.globalKeys()` or `event.setShutDown()`. I'm good with that - just checking.

I agree with the suggestion made at one point that `event.globalKeys.add('q',core.quit)` makes the need for the `event.setShutdownKeys('q')` less critical. Maybe just have the general one? If we add `event.globalKeys.add('q', core.quit)` at the top of several demos then people will soon get the idea.
 Guido's mantra was to have one way to do things (although that's not very closely followed by him anymore either!)

peircej:
I think a single `add()` call should only add a single option. `keys=['a','b']` is ambiguous as to whether we need to press them all or just one (but it would solve the problem of modifiers). Easy enough to get the many-to-one mapping with:
```
for key in ['a','b','c']:
  globalKeys.add(key, function)
```

lindeloev:
Good points, @hoechenberger. I think that you're right. @peircej's suggestion of one key to one function also makes sense.

@peircej, I guess that it's up for discussion :-) What I'm worried about with the system-wide version is that the experiment then runs differently on different computers, which is not reproducible. That's why I like Richard's implementation here - also with respect to quit keys. Then I can just send a script to a collaborator and say "use f9 to quit" without worrying that they have another setting where an undesired key quits.

hoechenberger:
Hi Jon!

> So, just to recap, since this has migrated a little. In the original plan, wasn't it going to be a system-wide setting? In the current version you still need to insert the line event.globalKeys() or event.setShutDown(). I'm good with that - just checking.

My current idea would be to allow users to specify a _shutdown key_ "globally", in the preferences. On new installations, this would be an empty list, effectively disabling this functionality until explicitly turned on. My reasoning here is that a shutdown key is probably something that is commonly expected to be persistent across experiments. Another reason is that _we know_ how to shutdown PsychoPy gracefully. So, if a user _does_ set `prefs.general['shutdownKeys']`, these key(s) will work with all experiments, automatically, without any further intervention.

The demos call `setShutdownKeys()` explicitly because  by default it is disabled. Improved versions of the demo could:
- try to read the setting from the preferences
- if it is empty (disabled), call `setShutdownKeys()`

The "ordinary" global key events, on the other hand, allow for such a high flexibility that persistent storage doesn't make sense, therefore they are not read from the preferences.

hoechenberger:
> I think a single add() call should only add a single option. keys=['a','b'] is ambiguous as to whether we need to press them all or just one (but it would solve the problem of modifiers). Easy enough to get the many-to-one mapping with:
> 
> for key in ['a','b','c']:
>   globalKeys.add(key, function)

Good point, I like that!

hoechenberger:
I'm currently growing to like the idea of having a simple keyboard->function mapping in a dictionary. This would greatly simplify the implementation. `GlobalKeys.remove()` would then accept a list of keys and remove those from the global event mapping. I think this should work. No need to make things more complicated than necessary!

lindeloev:
`name` doesn't serve a function anymore, but could be preserved as an argument with a default value `'unnamed'` for logging purposes?

Other things I considered, but don't have an opinion on:

1. Maybe always have key be a string in `add` and `remove`, leaving it to the user to add multiple event listeners. That would make the user understand the `events` attribute better.
2. `remove` could be removed and users could just do `globalKeys.events.pop(key, None)`

psychopy/event.py
```diff
@@ -798,3 +807,83 @@ def clearEvents(eventType=None):
...
+
+        for key in keys:
+            self.events[key] = _GlobalKeyEvent(name, keys, func, args, kwargs)
+
+    def remove(self, keys=None):
+        """
+        Remove global event keys.
+
+        keys : string, iterable, or None
+            A single key or list of keys. If None, remove all event keys.
```
lindeloev: This is not intuitive. "Remove None" sounds like removing nothing. Either we should have an `globalKeys.clear()` (as in events.clearEvents()) or `globalKeys.remove('all')`

hoechenberger:
> name doesn't serve a function anymore, but could be preserved as an argument with a default value 'unnamed' for logging purposes?

Currently it's a required argument:
```python
    def add(self, name, keys, func, *args, **kwargs):
```
If we make it optional with a default, I will have to think of a different way to collect the `*args` and `**kwargs`, which will be passed to `func`. (I don't particularly like the current implementation -- I'd personally prefer to have sth like `func_args=(), func_kwargs={}`; but I tried to keep the API consistent with `win.callOnFlip()`).

> Other things I considered, but don't have an opinion on:
> 
> Maybe always have key be a string in add and remove, leaving it to the user to add multiple event listeners. That would make the user understand the events attribute better.

I did consider that, but:

> remove could be removed and users could just do globalKeys.events.pop(key, None)

removing _all_ keys at once would be impractical that way. Therefore, I just modified `remove()` to remove all keys if invoked as `remove(keys=None)`.

psychopy/event.py
hoechenberger: Good point! `clear()` would probably the way to go then!

hoechenberger:
@lindeloev, we now have `clear()` which accepts `keys='all'` to remove all event keys.

hoechenberger:
> If we make it optional with a default, I will have to think of a different way to collect the *args and **kwargs, which will be passed to func. (I don't particularly like the current implementation -- I'd personally prefer to have sth like func_args=(), func_kwargs={}; but I tried to keep the API consistent with win.callOnFlip()).

Sorry, lost in confusion again. Of course `*args, **kwargs` would work just fine. -- See, this is how I work, that's why I produce so many commits. ;)

hoechenberger:
> Sorry, lost in confusion again. Of course *args, **kwargs would work just fine. -- See, this is how I work, that's why I produce so many commits. ;)

Ah no, I was right in the first place! Because with a signature like this:
```python
    def add(self, keys, func, name='', *args, **kwargs):
```
users wouldn't be able to pass `name` to both `add()` and `func`. So my current approach appears to be best:
```python
    def add(self, keys, func, func_args=None, func_kwargs=None, name=''):
```

lindeloev:
... improving to unseen heights with each commit :-D

My impression is that having both `remove` and `clear` is too much - it's not the "one way to do one thing" philosophy. What about having only `remove` which accepts the "all" keyword? "Clear" makes me think of removing everything so it's confusing that I can remove specific keys with it. It could also be that there is a `clear()` which accepts no arguments and simply sets `self.events={}` and then `remove` only accepts keys (probably raising a warning, if the key is not in `events`?)

With respect to `name` and args, I think it's confusing (read: non-standard) if users have to pass dictionaries instead of the classical **kwargs. How about having an obscure enough argument name that users would never need it? E.g. `log_name` or `event_rule_name` or the like.

hoechenberger:
> ... improving to unseen heights with each commit :-D

Hehe, I guess next time I'll propose to meet up in a gitter chat or something, this asynchronous communication overhead is kind of getting out of hand :D But it's still fun!

> My impression is that having both remove and clear is too much - it's not the "one way to do one thing" philosophy. What about having only remove which accepts the "all" keyword? "Clear" makes me think of removing everything so it's confusing that I can remove specific keys with it. It could also be that there is a clear() which accepts no arguments and simply sets self.events={} and then remove only accepts keys (probably raising a warning, if the key is not in events?)

Sounds fair, will do.

> With respect to name and args, I think it's confusing (read: non-standard) if users have to pass dictionaries instead of the classical **kwargs. How about having an obscure enough argument name that users would never need it? E.g. log_name or event_rule_name or the like.

I see exactly what you mean, but a) I don't think it's non-standard (e.g., seaborn uses this a lot, iirc; and I think MNE does this as well; but yes, it may be non-standard w/in PsychoPy, granted!), and b) this comes at a cost. Because essentially you'd be proposing a signature like this:

```python
    def add(self, keys, func, log_name ='', *args, **kwargs):
```
This comes with two problems.

*Problem 1*
The argument order is kind of "disrupted" in the sense that you would specify the function to call, then a `log_name` entirely unrelated to this function, and only _then_ the function arguments.

We could fix that by changing `log_name` to a (required) positional argument:
```python
    def add(self, log_name, keys, func, *args, **kwargs):
```
(which is basically the API I had proposed earlier). But this would still leave us with another issue.

*Problem 2*
Any argument name may only be specifed once in a function call. Now let's take our globalKeys to the extreme and assume we have a power use who wants to set / unset global event keys _by using global event keys_ (key-ception, so to speak!)

```python
globalKeys.add(log_name='key_1', keys=['a'], globalKeys.add, log_name='key_2', keys=['b'])
```

Okayokay, you could argue that `log_name` and `keys` are actually positional arguments, and if passed as such, there wouldn't be a problem. However, we might think of extending the API sometime (e.g., including a `log=True` kwarg or such, and kaboom...

For reference, here is an [example from the seaborn API](http://seaborn.pydata.org/generated/seaborn.lmplot.html#seaborn.lmplot):

```python
seaborn.lmplot(x, y, data, hue=None, col=None, ..., scatter_kws=None, line_kws=None)
```

hoechenberger:
> it may be non-standard w/in PsychoPy, granted!

Except for iohub: the demos typically also first create a kwargs dict and then pass it to `launchServer()`.

hoechenberger:
Todo & To Discuss
==============
- [ ] Decide how to best collect (kw)args for mapped function
- [ ] Is `shutdown` special enough to deserve
    - [ ] its own preferences setting
    - [ ] its own little helper function (`event.setShutdownKeys()`)
- [ ] update tests
- [ ] add documentation and improve docstrings

hoechenberger:
Re @peircej 
> I think a single add() call should only add a single option. keys=['a','b'] is ambiguous as to whether we need to press them all or just one (but it would solve the problem of modifiers).

Oh, I think I misread this previously. So you're suggesting `add()` should not handle lists / iterables, and insist on individual keys (strings)? I had specifically added the iterable handling for better consistency with `getKeys()` and `waitKeys()`. Do you think good documentation could avoid the ambiguity you're concerned about?

lindeloev:
Oh, I see. You clearly know more about **kwargs and dict-argument conventions than I do! :-) I rest my case. Still wondering if the `name` is then worth it at all - it's just so clean without it ;-) Yeah, this discussion is probably too detailed for async; you've shown amazing patience. 

But apart from what we've discussed now, I don't think I have much more. I would trust @peircej's judgement more on the questions about `**kwargs`, `shutdown`, and logging priority. But I can be at gitter tomorrow evening at 20 GMT.

peircej:
On 08/03/2017 19:57, Richard Höchenberger wrote:
>
> Re @peircej <https://github.com/peircej>
>
>     I think a single add() call should only add a single option.
>     keys=['a','b'] is ambiguous as to whether we need to press them
>     all or just one (but it would solve the problem of modifiers).
>
> Oh, I think I misread this previously. So you're suggesting |add()| 
> should not handle lists / iterables, and insist on individual keys 
> (strings)? I had specifically added the iterable handling for better 
> consistency with |getKeys()| and |waitKeys()|. Do you think good 
> documentation could avoid the ambiguity you're concerned about?
>
I don't think it matters to be the same as `get/waitKeys()` since it 
feels like a very different call. I think it *could* be used to mean 
that if all keys are pressed then this call is made (as in 
keys=['a','ctrl'] ) but I don't see the need to assign multiple keys to 
a single function in a single call. In the rare event that this happens 
it can be done in a loop.

For the question of handling the args/kwargs you might want to look at 
`Window.callOnFlip()` which is doing this. It stores args and kwargs and 
the call is almost the same as yours except that the functions and their 
calls are stored in a list instead of a dict.

hoechenberger:
> I don't think it matters to be the same as `get/waitKeys()` since it
> feels like a very different call. I think it *could* be used to mean
> that if all keys are pressed then this call is made (as in
> keys=['a','ctrl'] ) but I don't see the need to assign multiple keys to
> a single function in a single call. In the rare event that this happens
> it can be done in a loop.

I'm fine with either solution :) If @lindeloev has no objections, I would change to code to expect a single key.

> For the question of handling the args/kwargs you might want to look at
> `Window.callOnFlip()` which is doing this. It stores args and kwargs and
> the call is almost the same as yours except that the functions and their
> calls are stored in a list instead of a dict.

The first implementation was in fact inspired by `callOnFlip()`. The issue is that `callOnFlip()`'s signature is much simpler than `globalKeys.add()`'s:

```python
    def callOnFlip(self, function, *args, **kwargs):
```
vs
```python
    def add(self, keys, func, func_args=None, func_kwargs=None, name=''):
    # or sth. like
    def add(self, name, keys, func, *args, **kwargs):
```

Since we're collecting more arguments than just a function name and it's arguments, we get into trouble if a function requires an argument that has the same name as one we're already using to control `add()` behavior. That's why I proposed using `func_(kw)args` to get rid of that ambiguity. This appears to be common practice in other libraries, but may seem a bit strange to some users at first.

hoechenberger:
> I'm fine with either solution :) If @lindeloev has no objections, I would change to code to expect a single key.

Enforcing single keys might be a sane decision anyway, considering I want to add modifier support. This I would do by introducing another kwarg, e.g.:
```python
    def add(self, keys, modifiers=None, func=None, func_args=None, func_kwargs=None, name=''):
```
and `modifiers` _will have to accept interables_, because there can be combinations of them pressed. Then it would be even more confusing as to what to do if `keys` was also a list!

hoechenberger:
> But I can be at gitter tomorrow evening at 20 GMT.

Thanks for the offer, but I think that's not necessary anymore, or is it? :)

hoechenberger:
Signature is now
```python
def add(self, keys, func, func_args=None, func_kwargs=None, name=''):
```

Working _for now_, but as soon as I want to add modifier keys, it will get messy again...
```python
# maybe like this
def add(self, keys, func, modifiers=None, func_args=None, func_kwargs=None, name=''):
# or like this
def add(self, keys, func, func_args=None, func_kwargs=None, modifiers=None, name=''):
```

I'd prefer a signature like:
```python
def add(self, keys=None, modifiers=None, func=None, func_args=None, func_kwargs=None, name=''):
```
with `keys`, and `func` as required arguments, but this is not possible with Python 2. Another reason to move to Python 3 asap. ;)

hoechenberger:
btw just got modifier key support working, needs a bit of tweaking still, but it's on its way...

hoechenberger:
```python
In[1]: event.globalKeys
Out[2]: 
<_GlobalKeyEvents : 
	[CTRL] + [B] (make_rect_blue) -> <built-in function setattr>
	[CTRL] + [R] (make_rect_red) -> <built-in function setattr>
	[Q] (shutdown) -> <function quit at 0x103dec500>
	[ESCAPE] (shutdown) -> <function quit at 0x103dec500>
>
```

lindeloev:
Looks good to me (haven't tested)! In demo, I'd drop the square brackets and remember CTRL, so 

    CTRL+B for blue rectangle, \n
    CTRL+R for red rectangle, \n

hoechenberger:
@lindeloev Done! I'd like to keep the brackets in `__repr__()`, though... I somehow like them ;) Or do you think they should be removed there too?

lindeloev:
Really no strong opinion with respect to `__repr__()`. I just foresaw someone trying to type in square brackets in the demo, that's all ;-)

hoechenberger:
Ok, I
- changed behavior of `add()` to take only one key at a time; modifiers will have to specified as iterables, though (because multiple of them may be pressed at the same time)
- removed the `event.setShutdownKey()` convenience function and updated the demo accordingly (one can simply do `event.globalKeys.add('q', func=core.quit)`)

There's only one thing still up for discussion:
Should there be a preferences setting for a global shutdown key?

Currently we have:
- `prefs.general['shutdownKey']` (string)
- `prefs.general['shutdownKeyModifiers']` (list)

I like the idea of having a prefs setting, because then the shutdown key can be activated automatically, so you cannot "forget" it. This is currently done during [`event._GlobalKeyEvents.__init__()`](https://github.com/hoechenberger/psychopy/blob/78b61c960e6d226d0756dc93be5f3c4f4ae4b98e/psychopy/event.py#L851).

What do you think?

psychopy/demos/coder/basic/global_event_keys.py
```diff
@@ -0,0 +1,47 @@
...
+rect = visual.Rect(win, fillColor='blue', pos=(0, -0.2))
+text = visual.TextStim(
+    win,
+    pos=(0, 0.5),
+    text=('Press\n\n'
+          'CTRL + B for blue rectangle,\n'
+          'CTRL + R for red rectangle,\n'
+          'Q or ESC to quit.'))
+
+event.globalKeys.add(key='b', modifiers=['ctrl'], func=setattr,
```
lindeloev: Maybe we should have this one key be the simplest use case?
    
    event.globalKeys.add(key='b', func=setattr, func_args=(rect, 'fillColor', 'blue'))

... and then the other as is with a name and modifier. Update the text.text accordingly.

psychopy/demos/coder/basic/global_event_keys.py
```diff
@@ -0,0 +1,47 @@
...
+                     func_args=(rect, 'fillColor', 'blue'),
+                     name='blue rect')
+event.globalKeys.add(key='r', modifiers=['ctrl'], func=setattr,
+                     func_args=(rect, 'fillColor', 'red'),
+                     name='red rect')
+
+# Add multiple keys at once.
+[event.globalKeys.add(k, func=core.quit) for k in ('q', 'escape')]
```
lindeloev: `key` instead of `k`? My tiniest comment ever, sorry :-)

lindeloev:
Looks good! I made two minor suggestions to the demo, but other than that, I'm just looking forward to the next release :-) I wondered if we should also add a `core.wait(1, hogCPUperiod=1)` to the demo, but my impression is that it would make the demo too convoluted.

psychopy/demos/coder/basic/global_event_keys.py
hoechenberger: That's a good point anyway :) Will address this.

psychopy/demos/coder/basic/global_event_keys.py
hoechenberger: Can do. I only definitely wanted to have the `quit` option to be accessible w/o modifiers.

hoechenberger:
> I wondered if we should also add a core.wait(1, hogCPUperiod=1) to the demo, but my impression is that it would make the demo too convoluted.

I'll write a chapter of "proper" documentation to explain the two options (`wait()` and `flip()`). This way, the demo can stay slimmer.

hoechenberger:
I'll refactor to inherit from `collections.Sequence` to implement indexing and make the object iterable. Won't happen until the end of March, though.

psychopy/demos/coder/basic/global_event_keys.py
```diff
@@ -0,0 +1,50 @@
...
+                     func_args=(rect, 'fillColor', 'blue'),
+                     name='blue rect')
+
+# Add an event key with a "modifier" (CTRL).
+event.globalKeys.add(key='r', modifiers=['ctrl'], func=setattr,
+                     func_args=(rect, 'fillColor', 'red'),
+                     name='red rect')
+
+# Add multiple keys at once.
+[event.globalKeys.add(key, func=core.quit) for key in ('q', 'escape')]
```
lindeloev: Probably better to write it as

    for key in ('q', 'escape'):
        event.globalKeys.add(key, func=core.quit)

The list comprehension implies that you want to make a list. But add() returns nothing and the list is not saved.

hoechenberger: Addressed in the latest commit, thanks!

hoechenberger:
Sorry for the HUGE amount of commits, but I think this is _almost_ done. Needs tests and documentation still, but other than that, I think it's good. :)

hoechenberger:
@lindeloev, @peircej, if the docs build correctly, this PR should be _essentially_ good to merge. I think will squash the commits into a single large one first, though...

docs/source/coder/globalKeys.rst
```diff
@@ -0,0 +1,281 @@
...
+    ...
+    >>> event.globalKeys.add(key=key, func=myfunc)
+
+Look at `event.globalKeys`, we can see that the global event key has indeed
+been created.
+::
+
+    >>> event.globalKeys
+    <_GlobalEventKeys :
+        [A] -> 'myfunc' <function func at 0x10669ba28>
```
lindeloev: myFunc instead of func?

docs/source/coder/globalKeys.rst
```diff
@@ -0,0 +1,281 @@
...
+    from psychopy import core, event, visual
+
+
+    def change_color(win, log=False):
+        if win.color == 'gray':
+            new_color = 'blue'
+        else:
+            new_color = 'gray'
+
+        win.color = new_color
```
lindeloev: Maybe to convoluted, but this could be simplified to 

    win.color = 'blue' if win.color == 'gray' else 'gray'
    if log:
        print('Changed color to %s' % win.color)

docs/source/coder/globalKeys.rst
```diff
@@ -0,0 +1,281 @@
...
+
+        if log:
+            print('Changed color to %s' % new_color)
+
+
+    def main():
```
lindeloev: I think that the average user will be confused by the `if __name__ == '__main__':` structure. Maybe just make it functional, putting the current contents of `main` in global scope?

docs/source/coder/globalKeys.rst
```diff
@@ -0,0 +1,281 @@
...
+    >>> event.globalKeys[index].name
+    'my name'
+
+Removing global event keys
+--------------------------
+There are three ways to remove global event keys:
+
+- using :func:`psychopy.event.globalKeys.remove`,
+- using `del`, and
+- using :func:`psychopy.event.globalKeys.clear`.
```
lindeloev: I'm really impressed about how thoroughly you implemented this, with indexing and stuff! I'm learning from this. One could argue that we could drop `remove` in favor of `del`, but on balance, I think that the clear analogy between `globalKeys.add('a', func)` and `globalKeys.remove('a')` is a pretty good case for keeping `remove()`. An alternative is to do the same as when removing a dictionary key: `dict.pop(key, None)`, but the word `remove` is much more explanatory.

lindeloev: I guess I'm writing these options because I'm still a bit in doubt though leaning towards `remove`. You judge :-)

docs/source/coder/globalKeys.rst
hoechenberger: Hah, you caught me there. I edited those lines after copy & pasting from the IPython console, and forgot this spot. :)

docs/source/coder/globalKeys.rst
hoechenberger: Good thoughts! About "how thoroughly" this is is implemented: it's just a matter of subclassing from the right object types!

Regarding `remove`:
I had the same rationale: we do have `add`, so probably we should also have `remove`. Another reason to keep `remove` is that it can be used to remove all global event keys at once, via `remove('all')` -- I forgot to mention this in the docs! This would not be easily possible via `pop` or `del`. But I propose we can safely drop `clear` in favor of `remove('all')`. I can add the `pop` approach to the docs as well as another approach to remove individual keys.

docs/source/coder/globalKeys.rst
hoechenberger: Good point, this would probably make the function look more minimalistic.

docs/source/coder/globalKeys.rst
hoechenberger: I did consider this, but the entire document somehow ended up being pretty technical anyway -- it's not really beginner-friendly. So I thought, if the target audience aren't total beginners anyway, might as well demonstrate "best practices" here, using functions and avoiding global variables... Really not sure about this!

lindeloev:
Yes, would be clean with either clear+pop or remove+remove('all'), leaning
towards latter. looks great! I'm a little bit worried about showing many
options to do the same thing, but as you say, this document is probably for
the more technically oriented.

hoechenberger:
@lindeloev I have addressed all of your review comments. `clear()` has been removed from the implementation, and `pop()` has been added to the docs. I have also simplified the code example (putting variables into global scope). See the current document at
https://github.com/hoechenberger/psychopy/blob/a785d4181070e655f44f5d212d2b7dab78e4fd34/docs/source/coder/globalKeys.rst

hoechenberger:
I think I'll stop here -- my changes are starting to become kind of esoteric... @peircej, please let me know if you would like me to squash those commits.

peircej:
Yes, I'm afraid I stopped following this some time ago. Maybe you could squash the commits and resolve the conflict and let me know when you want it pulled in. 

Also, does this stand a chance of damaging any existing behaviour? i.e. should I include it in 1.85 branch or just in master (only for distant 1.86 release)?

hoechenberger:
I won't have time to look into this before April, but I do believe it shouldn't affect current behavior in any way, except for adding (negligible) processing overhead when processing keystrokes.

hoechenberger:
@peircej @lindeloev 
I believe this is good to merge. Should probably not be included in 1.85.x for now, tmailto give us more time for testing.

peircej:
OK, will try and do that later today. I think I will do it as a squash commit to keep the history easier to read. I did that recently for another series of commits and it does actually preserve the individual commits in the message. See e.g 65219f225cb0108a1c485c1255dcea49da16cc0f so I think it's good

hoechenberger:
Great :) I already rebases and squashed the commits, so it's just two commits for now!

peircej:
OK, if you've done the squashing I'll pull immediately :-0)

PR #1364, opened 2017-03-07
BF Get textstim_vs_textbox demo running again

m-macaskill:
As reported by @alexholcombe here: https://discourse.psychopy.org/t/textstim-vs-textbox-py-demo-not-working/2055 this demo had stopped working. Seems to be resolved by explicitly importing the textbox class. 
Perhaps there is a cleaner way to fix this?


peircej:
Not sure if there is a cleaner way but this is certainly useful first step! :-)

PR #1367, opened 2017-03-09
BF: Fix exception handling

hoechenberger:


peircej:
Is this error in the original code (i.e. it's a genuine BF, not FF)?

hoechenberger:
This PR changes behavior that was released with 1.85.0:
https://github.com/psychopy/psychopy/compare/1.85.0...hoechenberger:exception#diff-ac47a0223c02cc101e679607adb18949L432

hoechenberger:
We could also name this "RF", I guess... since it's not fixing a "real" bug?

hoechenberger:
I can also rebase on 1.85.0 if that makes work easier for you!

peircej:
No worries - just wanted to check if it was genuine BF, needing replicating across different release branches

PR #1369, opened 2017-03-15
BF: Allow to open prefs when no audio backend is installed

hoechenberger:
The preferences window imports `psychopy.sound` when compiling
the list of wigdets / settings to display. However, this would fail
if no audio backend is installed.

This commit catches this DependencyError and allows to open the preferences
window nonetheless.

### DIFF
--- a/psychopy/app/preferencesDlg.py
+++ b/psychopy/app/preferencesDlg.py
@@ -367,8 +367,8 @@ class PrefCtrls(object):
             self.valueCtrl.SetValue(value)
         elif spec.startswith('option') or name == 'audioDevice':
             if name == 'audioDevice':
-                value = value[0]
                 options = copy.copy(value)
+                value = value[0]
                 try:
                     from psychopy import sound
                     if hasattr(sound, 'getDevices'):
--- a/psychopy/app/preferencesDlg.py
+++ b/psychopy/app/preferencesDlg.py
@@ -10,6 +10,7 @@ import copy
 
 from . import localization, dialogs
 from psychopy import logging
+from psychopy.exceptions import DependencyError
 from .localization import _translate
 
 # this will be overridden by the size of the scrolled panel making the prefs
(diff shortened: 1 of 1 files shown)


hoechenberger:
Oh, please do not merge yet, I may have introduced a bug here!

hoechenberger:
Should be good now, sorry about that glitch!

peircej:
Right, so back to that thing of the commit tag, I'm pretty sure this IS a BF not a RF, right?
Isn't it a bug that is in the current (1.85) release?

To explain, your new work changing the behaviour of dialogs shouldn't be included in 1.85.1 because it might break things, so it gets included in master. When I release 1.85.1 I will create a 1.85 branch and I'll port across all BF tags (things in master that are appropriate bug fixes only). Things labelled FF or RF will be ignored because they are fixes of things that are not currently released.

hoechenberger:
I do unterstand the rationale, but I find it rather difficult to exclusively label changes like this _either_ as RF _or_ as BF (the FF thing is clear to me). It really depends upon one's definition of a bug... It could have been intended behavior to disallow opening the preferences window when not audio backend can be found; I found this annoying, so I changed it. Whether this classifies as a BF or not, I'm not really certain :) I definitely agree with you that this change should probably not be included in 1.85.1, though! -- Now, should I relabel as BF..?

peircej:
The failure to open prefs when no sound lib is installed is definitely a bug not a feature and I think that *does* need fixing in 1.85.1. (It's the refactor of dialog boxes that I'm scared of including because there are many commits there and something may have become broken)

So this one*should* be a BF. To be honest the difference between the other labels isn't so important. It's just BF because I use that to search for commits to add to a BF-release

hoechenberger:
I renamed to PR accordingly, should I also change the commit summary line?

peircej:
Don't worry - I'll remember for this one. :-)

PR #1382, opened 2017-04-05
RF: moving PsychoJS to a separate git project

peircej:
Work on this will now be conducted at:
  https://github.com/psychopy/psychojs
but the files will still be available (as a git submodule) within the psychopy project


lindeloev:
Should the same be done for the psychopy module, so that there's three: psychopy/psychojs, psychopy/psychopy and psychopy/apps? The latter refers to both of the former. It might not be worth the effort, but it would be elegant :-)

peircej:
Hmm, it might be nice to reinforce the idea that the lib and app are somewhat independent but I'm not sure that they are independent enough for it to work. E.g. the preferences are a mixture of both so if that lived only in the "lib" project it would be referring to things that weren't in the same project

lindeloev:
Yeah, the preferences would need to be split then. There's also a branding problem: as PsychoJS improves it's not just PsychoPY. So the apps would just have to be named "Psycho Builder" and "Psycho Coder" :-D

PR #1386, opened 2017-04-10
Correct error in data.py example

janfreyberg:
Discussed on the mailing list here:

http://discourse.psychopy.org/t/how-to-use-questhandler-to-measure-threshold/2292/1

I set the starting value for contrast at an arbitrary 50%


psychopy/data.py
```diff
@@ -3464,8 +3464,8 @@ class QuestHandler(StairHandler):
         # create staircase object
         # trying to find out the point where subject's response is 50 / 50
         # if wanted to do a 2AFC then the defaults for pThreshold and gamma
-        # are good
-        staircase = data.QuestHandler(staircase._nextIntensity, 0.2,
+        # are good. As start value, we'll use 50% contrast, with SD = 20%
+        staircase = data.QuestHandler(0.5, 0.2,
```
janfreyberg: Good point - I assumed from the naming of the variables 👍 

I'll take a look at it too

hoechenberger: https://github.com/psychopy/psychopy/blob/c564dbd61be91bed6ab47c50f9e9ae8547c2883b/psychopy/contrib/quest.py#L293

hoechenberger: But honestly, this code is absolutely... errm... it's super hard to read. Maybe we should re-implement the entire thing some day.

janfreyberg: Yes, and maybe put in some comments :) I have no clue what is going on in there. For now do you have a suggestion for what to put in the docs for this SD?

psychopy/data.py
```diff
@@ -3464,8 +3464,8 @@ class QuestHandler(StairHandler):
...
             pThreshold=0.63, gamma=0.01,
```
hoechenberger: I'm also wondering where 0.63 comes from. Shouldn't that depict the desired target performance, i.e., typically 50% correct in Yes/No and 75% correct in 2-AFC?

psychopy/data.py
janfreyberg: The docs in quest.py simply say:
```
tGuess is your prior threshold estimate.
tGuessSd is the standard deviation you assign to that guess.
```

hoechenberger: Also, I am pretty certain `startVal` does not necessarily correspond to the first proposed intensity, i.e., stating that `As start value, we'll use 50% contrast` (with `startVal=0.5`) could be misleading. `startVal` specifies our _prior assumption about threshold location_ (together with `startValSd`), but this does not mean that QUEST will actually "start" there, does it?

psychopy/data.py
janfreyberg: I think it can be anything as long as it's `>gamma` and `<1`, but I'm not clear why -.63 was chosen here.

psychopy/data.py
hoechenberger: > I assumed from the naming of the variables

Yes, I believe the parameter names are not very helpful and would love to change them. And the docs need improvement anyway.

hoechenberger: @janfreyberg sorry, didn't see your replies coming in before typing mine. But I think I already addressed your question, no? :)

psychopy/data.py
hoechenberger: It [does say](https://github.com/psychopy/psychopy/blob/c564dbd61be91bed6ab47c50f9e9ae8547c2883b/psychopy/data.py#L3511):
```
        Typical values for pThreshold are:
            * 0.82 which is equivalent to a 3 up 1 down standard staircase
            * 0.63 which is equivalent to a 1 up 1 down standard staircase
                (and might want gamma=0.01)
```
But in the example code, we talk about 2-AFC, so I believe we should also provide a "real-world proof" 2-AFC example, which would typically target the 75% threshold... Just my personal view!

hoechenberger:
> Yes, and maybe put in some comments :) I have no clue what is going on in there.

This code was first implemented in Matlab and then ported to Python and then, later, added to PsychoPy. And indeed, it shows. ;)

psychopy/data.py
hoechenberger: > https://github.com/psychopy/psychopy/blob/c564dbd61be91bed6ab47c50f9e9ae8547c2883b/psychopy/contrib/quest.py#L293
> 
> Definitely not 20%.

This is simply a normal distribution with SD equal to `startValSD` and mean `startVal`.

hoechenberger: So I think I figured it out. The priors are ordinary gaussians (normal distributions). While QUEST uses Weibull functions for its computations.

The normal distributions are normalized "manually"
```python
        self.pdf = num.exp(-0.5*(self.x/self.tGuessSd)**2)
        self.pdf = self.pdf/num.sum(self.pdf)
```
therefore the formula looks a bit different than in text books, i.e. no 1/sqrt(2 pi sigma^2) normalization term; instead, a two-step process to normalization. :)

peircej:
None of it written by me, I must say! :-) And it was too much spaghetti for me to want to touch it!

peircej:
Let me know as/when this needs pulling in

psychopy/data.py
janfreyberg: Yeah, I suspected it had something to do with matching other procedures. I think you're right, but I also think there should be a general explanation of the types of values one can target with the different staircase procedures in psychopy, which I guess would be a more general overhaul of the docs

psychopy/data.py
janfreyberg: Aha, OK! I think you're right on startVal not being the starting point. Maybe it should be renamed prior in the function argument, or tGuess like in the actual quest code. Which is I think what they call it in the matlab QUEST that comes with the psychtoolbox, so that would be nice too!

peircej:
I think the choice of `startVal` being used was because that made it most like the up-down staircase that I had written (by the person that added this to PsychoPy, and I no longer remember who that was). In a QUEST I don't think there is *any* predetermined startValue but the starting estimate of the target value is the closest we have. Not in favour of that changing that term now though

hoechenberger:
I would also not want to change it now. I mean, yes I would, but we better should not.

The implementation is neat in the sense that it's actually working. The coding style unfortunately makes it really hard to maintain :) I never really touched `quest.py` because it's in the contrib directory, whose contents I consider kind of "inert". @janfreyberg would you like to try to improve the docs? I can help with that. We should maybe create a new issue or PR for that and discuss there!

janfreyberg:
Sounds good! I think some good examples of using staircases would be nice, too - I'll see and look if I can use some of my posts on the forum, too. I'll open the issue

hoechenberger:
Awesome!

peircej:
Very much agreed that the docs need a rework on this! :-)

PR #1388, opened 2017-04-10
RF: Make Window's refreshThreshold public

hoechenberger:
`visual.Window._refreshThreshold` must be modified by the users to fine-tune checking for dropped frames when using `Window.recordFrameIntervals = True`. It is used for no other purpose and should therefore be public. This PR renames the attribute to `visual.Window.refreshThreshold`, making it part of the public API.

### DIFF
--- a/docs/source/general/timing/detectingFrameDrops.rst
+++ b/docs/source/general/timing/detectingFrameDrops.rst
@@ -12,23 +12,33 @@ Occasionally you will drop frames if you:
 Things to avoid:
 
 * recreating textures for stimuli
-* building new stimuli from scratch (create them once at the top of your script and then change them using :meth:`stim.setOri(ori)`, `stim.setPos([x,y]...)`
+* building new stimuli from scratch (create them once at the top of your script
+and then change them using :meth:`stim.setOri(ori)`, `stim.setPos([x,y]...)`
 
 Turn on frame time recording
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


docs/source/general/timing/detectingFrameDrops.rst
```diff
@@ -15,37 +15,51 @@ Things to avoid:
...
     logging.console.setLevel(logging.WARNING)
 
+    print('Overall, %i frames were dropped.' % win.nDroppedFrames)
+
 Show me all the frame times that I recorded
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 While recording frame times, these are simply appended, every frame to 
 win.frameIntervals (a list). You can simply plot these at the end of your script using pylab::
```
mmagnuski: maybe "using matplotlib"?

hoechenberger: Good point.

peircej:
I guess the question is whether "normal" users should be modifying that threshold.

hoechenberger:
If users shouldn't normally tinker with this setting, we should remove the related recommendation from the documentation. Currently it explicitly recommends altering the threshold:

http://www.psychopy.org/general/timing/detectingFrameDrops.html

peircej:
OK. Apparently I can't decide on that one then! ;-) I don't mind either way.

hoechenberger:
I'm undecided too ;) @mmagnuski what do you think?

Maybe we should ask: does it harm if we expose this setting?

hoechenberger:
Oh, I see you already merged it, page didn't refresh properly :)

peircej:
Yeah, it's just a question of more info being less info at some point. Very grey area where that "point" is (and varies between individuals). So I went with your gut reaction(s) this time!

PR #1390, opened 2017-04-11
NF: Allow to disable keyboard buffer clearing in waitKeys()

hoechenberger:
`event.waitKeys()` always used to clear the keyboard buffer implicitly. This PR adds a new keyword argument to control this behavior explicitly.

### DIFF
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -376,22 +376,18 @@ def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
             is given then the time will be relative to the `Clock`'s last
             reset.
         clearEvents : **True** or False
-            Whether to clear the keyboard buffer (and discard preceding
-            keypresses) before starting to monitor for keypresses. If
-            `keyList` is specified, only those keys are removed from the
-            keyboard buffer.
+            Whether to clear the keyboard event buffer (and discard preceding
+            keypresses) before starting to monitor for new keypresses.
 
     Returns None if times out.
 
     """
-    if clearEvents:  # Only consider keypresses from here onwards.
-        if not keyList:
-            # We need to invoke clearEvents(), but our keyword argument is
-            # also called clearEvents. We can work around this conflict by
-            # accessing the global scope explicitly.
-            globals()['clearEvents']('keyboard')
-        else:
-            getKeys(keyList=keyList)
+    if clearEvents:
+        # Only consider keypresses from here onwards.
+        # We need to invoke clearEvents(), but our keyword argument is
+        # also called clearEvents. We can work around this conflict by
+        # accessing the global scope explicitly.
+        globals()['clearEvents']('keyboard')
 
     # Check for keypresses until maxWait is exceeded
     #
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -351,7 +351,7 @@ def getKeys(keyList=None, modifiers=False, timeStamped=False):
 
 
 def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
-             timeStamped=False, clearBuffer=True):
+             timeStamped=False, clearEvents=True):
     """Same as `~psychopy.event.getKeys`, but halts everything
     (including drawing) while awaiting input from keyboard.
 
@@ -375,7 +375,7 @@ def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
             keynames. Each tuple has (keyname, time). If a `core.Clock`
             is given then the time will be relative to the `Clock`'s last
             reset.
-        clearBuffer : **True** or False
+        clearEvents : **True** or False
             Whether to clear the keyboard buffer (and discard preceding
             keypresses) before starting to monitor for keypresses. If
             `keyList` is specified, only those keys are removed from the
@@ -384,9 +384,9 @@ def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
     Returns None if times out.
 
     """
-    if clearBuffer:  # Only consider keypresses from here onwards.
+    if clearEvents:  # Only consider keypresses from here onwards.
         if not keyList:
-            clearEvents('keyboard')
+            globals()['clearEvents']('keyboard')
         else:
             getKeys(keyList=keyList)
 
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -397,9 +397,11 @@ def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
     #
     # NB pygame.event does have a wait() function that will
     # do this and maybe leave more cpu idle time?
+
     timer = psychopy.core.Clock()
-    key = None
-    while key is None and timer.getTime() < maxWait:
+    got_keypress = False
+
+    while not got_keypress and timer.getTime() < maxWait:
         # Pump events on pyglet windows if they exist.
         if havePyglet:
             defDisplay = pyglet.window.get_platform().get_default_display()
@@ -409,12 +411,14 @@ def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
         # Get keypresses and return if anything is pressed.
         keys = getKeys(keyList=keyList, modifiers=modifiers,
                        timeStamped=timeStamped)
-        if len(keys):
-            return keys
+        if keys:
+            got_keypress = True
 
-    # If maxWait is exceeded (exits while-loop), return None
-    logging.data('No keypress (maxWait exceeded)')
-    return None
+    if got_keypress:
+        return keys
+    else:
+        logging.data('No keypress (maxWait exceeded)')
+        return None
 
 
 def xydist(p1=(0.0, 0.0), p2=(0.0, 0.0)):
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -794,7 +794,7 @@ class BuilderKeyResponse(object):
         self.clock = psychopy.core.Clock()  # we'll use this to measure the rt
 
 
-def clearEvents(eventType=None, keyList=None, modifiers=None):
+def clearEvents(eventType=None):
     """Clears all events currently in the event buffer.
 
     Optional argument, eventType, specifies only certain types to be
@@ -803,10 +803,6 @@ def clearEvents(eventType=None, keyList=None, modifiers=None):
     :Parameters:
         eventType : **None**, 'mouse', 'joystick', 'keyboard'
             If this is not None then only events of the given type are cleared
-        keyList : **None** or []
-            Allows the user to specify a set of keys to remove from the
-            keyboard buffer. If the keyList is `None`, the buffer will be
-            cleared completely. This argument is unless `eventType
 
     """
     # pyglet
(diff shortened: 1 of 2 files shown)


lindeloev:
Generally a good idea, I think. I haven't tested it, but looking at the code, I have a few ideas for improvement.

1. I would expect `clearBuffer=False` to leave the key in the buffer after having called `event.waitKeys()`. If I understand correctly, the current behavior is that it avoids clearing whatever preceded the `event.waitKeys` call so that if I press 'e' and then call `event.waitKeys(keyList=['e'], clearBuffer=False)`, it will return with that immediately?
2. Maybe the default clearing behavior should be changed to only clear events in `keyList` rather than all events, just as `event.getKeys` does after it has been called. Would be good to get @peircej 's thoughts on this.
3. `event.getKeys` could use a `clearBuffer` argument as well, if you feel like it :-)

hoechenberger:
> I would expect `clearBuffer=False` to leave the key in the buffer after having called event.waitKeys(). If I understand correctly, the current behavior is that it avoids clearing whatever preceded the event.waitKeys call so that if I press 'e' and then call event.waitKeys(keyList=['e'], clearBuffer=False), it will return with that immediately?

This is correct. The idea is that this implementation now allows for code like this:

```python
event.clearEvents()
present_stimulus()
event.waitKeys(keyList=['a', 'b'], clearBuffer=False)
```
Previously, one had to poll using `getKeys()` in this situation, because `waitKeys()` would only consider keypresses after `present_stimulus()` was finished.

> Maybe the default clearing behavior should be changed to only clear events in `keyList` rather than all events, just as `event.getKeys` does after it has been called. Would be good to get @peircej 's thoughts on this.

Good idea, can look into this.

> `event.getKeys` could use a `clearBuffer` argument as well, if you feel like it :-)

Wouldn't make sense, I guess, since `getKeys()` returns immediately. If we cleared the buffer within that function -- well, it would never return any keypresses at all!

hoechenberger:
> I would expect `clearBuffer=False` to leave the key in the buffer after having called `event.waitKeys()`.

We could think about offering two "switches": `clearBufferBefore` and `clearBufferAfter`? Would require modifications to `getKeys()`, though. But should be doable (that's my conclusion after looking at the code for 30 sec, at least!)

lindeloev:
Good replies, Richard, I agree with everything. 

Yes, something like the before-after clearing you suggest, but it looks so ugly/complicated that perhaps we should just leave it to the user to store the returned key and use it later him/herself. In other words, just go with your solution while perhaps adding my suggestion number 2 about only clearing `keyList` by default.

hoechenberger:
Actually the ugliness is caused by `waitKeys()` clearing the buffer by default, which I think was a suboptimal design decision: Users should simply call `clearEvents()` if they intend to work on empty buffers. The only reason to stick to this behavior is ensuring backward-compat. I would be all-in for breaking it, but I know many people would disagree ;)

hoechenberger:
> perhaps adding my suggestion number 2 about only clearing keyList by default.

I just realized this would also break backward-compat.

psychopy/event.py
```diff
@@ -371,27 +377,33 @@ def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
...
     """
-    # NB pygame.event does have a wait() function that will
-    # do this and maybe leave more cpu idle time?
-    key = None
     if clearBuffer:  # Only consider keypresses from here onwards.
-        clearEvents('keyboard')
+        if not keyList:
+            clearEvents('keyboard')
+        else:
+            getKeys(keyList=keyList)
```
hoechenberger: @lindeloev, do you think this is a viable solution? Was the easiest approach I could think of!

peircej:
:-P
I stand by `waitKeys(clearBuffer=True)`. I've never written an experiment where I kept the keys after using waitKeys, and I can't really imagine when I would do so. ;-)

You're welcome to add a `clearAfter=True` (or `keepProcessed=False`) but, yes, I am against you existing users by changing the clearBuffer argument.

hoechenberger:
> I stand by waitKeys(clearBuffer=True). I've never written an experiment where I kept the keys after using waitKeys, and I can't really imagine when I would do so. ;-)
> 
> You're welcome to add a clearAfter=True (or keepProcessed=False) but, yes, I am against you existing users by changing the clearBuffer argument.

I _think_ we all agree on this one, but let me try to recap just to be sure:

## Current behavior in `master`
- the keyboard buffer is always entirely cleared before we start checking for keypresses
- the user can limit the keys to monitor via `keyList`
- key events occurring during the `waitKeys()` run and that are _not_ included in `keyList` are kept in the buffer after `waitKeys()` finishes

## Current behavior of the new implementation
- _(unchanged)_ the user can limit the keys to monitor via `keyList`
- _(new)_ clearing the buffer before we start checking for keypresses is optional
  - _(new)_  if a `keyList` is specified _and_ clearing of the buffer is requested, only the events related to the keys in `keyList` are removed from the buffer before we start monitoring
  - _(new)_ accordingly, keys that were added to the buffer before the invocation of `waitKeys()` will remain in the buffer if they are not included in `keyList`
- _(unchanged)_ - key events occurring during the `waitKeys()` run and that are _not_ included in `keyList` are kept in the buffer after `waitKeys()` finishes

Or in simple words: with the new implementation, currently, the key buffer might end up containing some "old" events (from pre-`waitKeys()` invocation) when suppling a `keyList`; these events are not there with the `master` implementation, because it clears _all_ events first before starting monitoring.

This approach was chosen for consistency with `getKeys()`, which also only removes the keys in `keyList` from the buffer.

However, I am now starting to wonder whether the following would not be a more sane approach:
```python
keys_to_monitor = ('a', 'b', 'c')
clearEvents('keyboard', keys=keys_to_monitor)
waitKeys(keys_to_monitor, clearBuffer=False)
```
i.e. by default, `clearBuffer` would clear the _entire_ keyboard buffer; and for more fine-grained control, users can turn this behavior off and call an extended `clearEvents()` instead.

lindeloev:
The changes look good to me. It would not be backwards compatible if users used `waitKeys` with a particular keyList to clear the entire buffer, but I don't see that use case. On the contra

Regarding the new proposal, I think it highlights why I don't think it's an improvement ;-) It would be the typical use case, I think, and it requires three lines of code instead of one. And it's inconsistent with `getKeys`. 

Adding `keys` to `clearEvents` could make sense. However, it's called "buttons" on the mouse, so maybe the more neutral argument name `events` would be better? `clearEvents(eventType='keyboard', events=['a', 'b', 'c'])`

Now that I think of it, shouldn't we call the argument in `waitKeys` `clearEvents` instead of `clearBuffer`? To be consistent with `event.clearEvents` end not require the user to know what a buffer is.

hoechenberger:
> Regarding the new proposal, I think it highlights why I don't think it's an improvement ;-) [...]
> Adding keys to `clearEvents` could make sense.

Ok so you are suggesting to
- only clear the keys in `keyList` if `clearBuffer=True` in `waitKeys(), while
- additionally add the possibility to drop individual events/button presses from the buffer in `clearEvents`?

Just checking I'm not getting you wrong.

> However, it's called "buttons" on the mouse, so maybe the more neutral argument name events would be better? `clearEvents(eventType='keyboard', events=['a', 'b', 'c'])`

Looking at `event.py`, it's currently only possible to be explicitly drop individual events for keyboards, but not for other input devices (at least when it comes to pyglet). I thought we could just use an "explicit" kwarg and ignore it when it does not apply, e.g.:
```python
keys = ('a', 'b', 'c')  # Heya, I'm a tuple!
clearEvents('mouse', keyList=keyList)
```
would clear the mouse events and ignore `keyList`. And in the future we could add similar kwargs for mouse and joystick, e.g., `mouseButtonList` and `joystickButtonList`.

> 
> Now that I think of it, shouldn't we call the argument in waitKeys clearEvents instead of clearBuffer? To be consistent with event.clearEvents end not require the user to know what a buffer is.

We're currently using both terms interchangeably at numerous places, so maybe we should settle for one. Also, I picked `clearBuffer` because `clearEvents` is already defined as a function at module level ;) Could probably hack around this using the `global` keyword, haven't tried that so far...

hoechenberger:
@lindeloev I think `waitKeys()` is "complete" now. The changes we discussed for `clearEvents()` can probably go to another PR... Also, it just came to my mind that one can already do `getKeys(('a', 'b'))` to remove individual keyboard events from the buffer! So maybe changes to `clearEvents()` aren't needed at all?

peircej:
Wait, no. `clearBuffer` should *not* be changed to `clearEvents`. That isn't a good enough reason to breaks people's code.

hoechenberger:
I think we have some confusion here: the `clearBuffer` vs `clearEvents` thingy was about the best choice of the keyword arg for `waitKeys()` to enable / disable clearing of the keyboard event buffer. At first, I went with `clearBuffer`, but for consistency with the `event.clearEvents()` function later switched to `clearEvents`, i.e. the function signature is now:
```python
def waitKeys(maxWait=float('inf'), keyList=None, modifiers=False,
             timeStamped=False, clearEvents=True):
```
This came with the question of how to invoke `event.clearEvents()` _within that function_, but I solved it:
```python
globals()['clearEvents']('keyboard')
```
But _there is_ currently a backward-incompatible change, specifically:
```
 clearEvents : **True** or False
            Whether to clear the keyboard buffer (and discard preceding
            keypresses) before starting to monitor for keypresses. If
           `keyList` is specified, only those keys are removed from the
            keyboard buffer.
```
The question is if any user will be affected by this, at all: it would only matter in settings like this:
```python
waitKeys(('a', 'b'), clearEvents=True)  # Remove 'a' and 'b' from the buffer, then start monitoring.
k = getKeys()
```
Now `k` could contain key press events from before `waitKeys()` was called. However, I can easily change this behavior if desired to always clear the entire buffer if `clearEvents=True`, regardless of `keyList`. In either case, users could replicate the above behavior by doing something like:
```python
getKeys(('a', 'b'))  # Remove those keys from the buffer.
waitKeys(('a', 'b'), clearEvents=False)
k = getKeys()
```
so maybe we should go with this approach and expand the docs, while retaining full backward-compat?

hoechenberger:
@lindeloev, what do you think? 

I also just realized that the argument

> The question is if any users will be affected by this, at all

actually could also be used to make a point _against_ this change: if we assume users wouldn't be affected, one could conclude this functionality is not important or required. So we might as well leave it out.

hoechenberger:
@peircej I changed the implementation to always clear the entire keyboard buffer by default, even if a `keyList` is specified. This should ensure 100% backward-compatibility. Should be good to merge, if @lindeloev doesn't have any further suggestions.

peircej:
OK, let me know Jonas. 
Sorry, Richard, I misunderstood the change of clearBuffer->clearEvents. The conversation was so long here that I forgot clearBuffer had only been created in this thread!

hoechenberger:
Haha, yes, no problem. My conversations often tend to get too long and too convoluted :(

lindeloev:
Looks good. You're right that what I used as an argument for breaking
backward compatibility, in this case, was actually just as much an argument
for - good catch :-) So thanks and merge away, @peircej :-)

______________________________

Jonas Kristoffer Lindeløv, M.Sc., Ph.D.
Assistant Professor in Cognitive Neuroscience and Neuropsychology
Profile at Aalborg University <http://personprofil.aau.dk/117060>, Scientific
blog <http://lindeloev.net/>

PR #1391, opened 2017-04-12
Move volume calls to callback function (sounddevice backend)

gjcooper:
Following on from the discussion in a previous pull request: https://github.com/psychopy/psychopy/pull/1378 I have finally had a chance to test implementation of the volume for the sounddevice backend within the callback on the `_SoundStream` object.

This involves reverting volume changes introduced in the previously merged pull request and adding an extra line in the callback.

Another option would be to apply it to the block (in `_nextBlock`) **just** before returning it to the callback method - which as far as I can see would have the same effect, and could be argued is a more natural place for it to be.

I have done some testing and have verified that the pull request as it currently stands works locally.


peircej:
Cheers Gavin. This is what I was expecting. :-)
I agree it could go in the `_nextBlock` call instead but I think this is fine

PR #1392, opened 2017-04-12
FF: bug with grating/envelope component params in builder

peircej:
I had accidentally added a second copy of the allowedVals param to the blendMode attribute

Unfortunately which adding this commit I got the history order messed up (with an erroneous use of --rebase). Sorry!


hoechenberger:
Heh, I always make backups of my entire local repository before rebasing. Been destroying my history waaay too often 👍

PR #1395, opened 2017-04-12
RF: Split up data.py into smaller sub-modules

### LINKED ISSUE #1053: Split up data.py
`psychopy.data` has become a fairly huge module over the years. With new trial handlers and staircase methods coming in every now and then, it is going to grow even more in the future, making it harder to understand, navigate, and refactor. It has been suggested the module be split up into smaller sub-modules. I would suggest the following submodules:
- `data.design` or `data.flow`: `ExperimentHandler` and `TrialHandler` derivatives, and the `DataHandler`
- `data.stairs`: all staircase procedures
- `data.fit` or `data.models`: Psychometric function fitting and data bootstrapping
- `data.tools`, `data.utils`, or `data.misc`: Helper functions like `isValidVariableName()`

(Better submodule names very welcome ;))

hoechenberger:
`data.py` has grown huge over the years, making navigation and maintenance troublesome. This commit splits the module into several sub-modules. Should be fully backward-compatible. In a next step, we might consider adding support for lazy loading.

This closes #1053.


hoechenberger:
Ugh, of course now many tests fail because unpickling of some old psydat files fails. We really really really should move away from pickled files for long-term storage, this is discourages _everywhere_. But yeah, that's another issue...

How should I best proceed about this?

hoechenberger:
Ha! Passing now.

peircej:
Looks good. This is certainly a refactor that I think needed to happen sooner or later :-)

For pickling, the thing that's great is that we recreate an exact replica of the ExperimentHandler/TrialHandler that ran the study and it can store any type of object inside. To drop this file type would be a big loss IMO. The alternative would be to create some sort of JSON format that did the same thing ourselves but wouldn't that mean, essentially, rewriting the guts of pickle ourselves? A personal preference for it is that *because* this file can't be easily changed by hand in an editor it means I feel happy that nobody (e.g. an undergrad in the lab) has manipulated the data along the way. I mean, I believe it's possible to change contents but it's beyond my abilities and that seems to have a use.

peircej:
Do you want me to pull or is this a WIP?

hoechenberger:
Hi Jon, thanks for your friendly feedback!

> Looks good. This is certainly a refactor that I think needed to happen sooner or later :-)

Yep, it's been on my agenda for quite a while now ;)

> For pickling, the thing that's great is that we recreate an exact replica of the ExperimentHandler/TrialHandler that ran the study and it can store any type of object inside. 

Yes, absolutely!

> To drop this file type would be a big loss IMO. 

Please excuse my rant above, which might have implied I wanted to entirely get rid of the pickled psyexp files. I don't :) We should definitely keep supporting them! But I believe we should add an additional option, like you pointed out:

> The alternative would be to create some sort of JSON format

This is exactly what I have in mind. JSON is probably one of the most-portable and widely-supported serialization formats ever, and it's even human-readable, how awesome is that?

> that did the same thing ourselves but wouldn't that mean, essentially, rewriting the guts of pickle ourselves? 

Fortunately not! A couple months back I played with a library that would automagically serialize _most_ data into a JSON file -- it works pretty much like a drop-in replacement for `cPickle`. Have to go search through my old code to find out its name again :) Of course an issue arises when you have "binary" stuff like, say, numpy arrays etc; but that library took care of most of this! If only my memory worked better.......

> A personal preference for it is that because this file can't be easily changed by hand in an editor it means I feel happy that nobody (e.g. an undergrad in the lab) has manipulated the data along the way. I mean, I believe it's possible to change contents but it's beyond my abilities and that seems to have a use.

Couldn't one just simply unpickle, alter the data, and pickle again? Never tried that before, though :) Data integrity and consistency is indeed an issue I believe we should _also_ address more thoroughly in the future, e.g. by adding checksums / cryptographic hashes, allow the developer(s) to sign the data, and / or automatically put the data files into a git repository, which will also keep track of any changes and discover corruption and "bitrod" automatically.

> Do you want me to pull or is this a WIP?

I am not sure. I mean, the tests pass, it works on my machine, but I would very much like to get the "Okay, seems good with me and doesn't set my computer on fire -- for now!" from other devs first. Of course, the longer this PR sits here waiting, the more merge conflicts will be incoming :)

Issue #1343, opened 2017-01-17
Coder tutorial 3 uses deprecated function


m2march:
The script returns the following error:

> DeprecationWarning: FitFunction is now fully DEPRECATED: use FitLogistic, FitWeibull etc instead

See: https://github.com/psychopy/psychopy/blob/master/docs/source/coder/tutorial3.py#L37

peircej:
Yes, this needs fixing. Thanks for reporting it. Want to submit a pull request?

peircej:
The fix above wasn't quite right - it changed the fit from a Weibull to a Logistic. I've updated the docs with the correct fix in 671be2d73cf9c47f43e04e9b5b1d2c3016d804d2

PR #1406, opened 2017-04-26
Deprecation Warning Resolved

ShaikAsifullah:
Resolving the issue https://github.com/psychopy/psychopy/issues/1343


peircej:
Thanks for this. I've actually changed it slightly differently to perform a Weibull fit instead of a Logistic (as in the original example).
See 671be2d7

PR #1416, opened 2017-05-18
Fix JS code for checking of correct non-response

jonathanoroberts:
Fix the JS code written to index.html for end of a routine so
that it checks to see if it really was a non-response before
evaluating if the non-response was the correct answer.

This fixes a problem where normal correct responses
are not written to the data file


peircej:
Good catch. This was an error for a while in the Python code. I expect the code was ported to JS before I noticed the error and fixed it in Python!

PR #1423, opened 2017-05-26
To fix previous pull request #1422: Update componsTemplate.txt

schofiaj:


peircej:
@hsogo This change might require an equivalent change in the _translate() lookup?

hsogo:
Yes. I've updated Japanese translation for these changes.

PR #1429, opened 2017-05-31
nReps and sequential/random/fullRandom

jonathanoroberts:
add psychoJS support for nReps and
sequential/random/fullRandom trial sequences.
Add columns from conditions files to data output.
Make loops marked as “isTrials” cause a data line
to be written each iteration. Add an “eventName”
column to the data file to identify whether a data
line was caused by a keyboard event or a loop iteration.
write out experiment data when quitPsychoJS() is called.

This commit goes with the one with the same name in the
psychojs project

### DIFF
--- a/psychopy/app/builder/components/keyboard/__init__.py
+++ b/psychopy/app/builder/components/keyboard/__init__.py
@@ -524,4 +524,3 @@ class KeyboardComponent(BaseComponent):
 
         if currLoop.params['name'].val == self.exp._expHandler.name:
             buff.writeIndented("%s.addData('eventName','keyboard');\n" % self.exp._expHandler.name)
-            buff.writeIndented("%s.nextEntry();\n" % self.exp._expHandler.name)
--- a/psychopy/app/builder/components/keyboard/__init__.py
+++ b/psychopy/app/builder/components/keyboard/__init__.py
@@ -523,4 +523,5 @@ class KeyboardComponent(BaseComponent):
             buff.writeIndentedLines(code)
 
         if currLoop.params['name'].val == self.exp._expHandler.name:
-            buff.writeIndented("%s.nextEntry()\n" % self.exp._expHandler.name)
+            buff.writeIndented("%s.addData('eventName','keyboard');\n" % self.exp._expHandler.name)
+            buff.writeIndented("%s.nextEntry();\n" % self.exp._expHandler.name)
--- a/psychopy/app/builder/components/settings/__init__.py
+++ b/psychopy/app/builder/components/settings/__init__.py
@@ -631,7 +631,17 @@ class SettingsComponent(object):
                 "}\n"
                 )
         buff.writeIndentedLines(abbrevFunc)
+        recordLoopIterationFunc = ("\nfunction recordLoopIteration(name) {\n"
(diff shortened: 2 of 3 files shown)


peircej:
I didn't understand the thing about a keyboard event or a loop iteration. Keyboard entries shouldn't be adding lines to the data file - they should just be on the same line.

jonathanoroberts:
One of the tests I was running had two loops, one presenting a stimulus with a keyboard response, and the other just presenting a series of stimuli. When run in builder, there were data lines for both loops. When I ran the same task in psychojs, there were only data lines for the loop that had keyboard responses. After digging a bunch, I found that the call to thisExp.nextEntry() is responsible for producing a line in the data file, and that the only place that was getting put into the index.html file was in the writeRoutineEndCodeJS function of the keyboard component. When I added a second routine inside the first loop to wait for a second keypress for each looping, it recorded a new data line for each keypress…

The only way I could figure to get the data for the second loop (with no keyboard response) was to add an entry in the loop scheduling to call a function that calls nextEntry() for each iteration of a loop marked as “isTrials”.

Perhaps, the keyboard component end routine code could just call addData() instead of nextEntry(), so that everything for a trial ends up on only one data line?
I’ll give that a try.

-Jon



On Jun 1, 2017, at 3:10 AM, Jon Peirce <notifications@github.com<mailto:notifications@github.com>> wrote:


I didn't understand the thing about a keyboard event or a loop iteration. Keyboard entries shouldn't be adding lines to the data file - they should just be on the same line.

peircej:
Ah, I see. Yes, exactly:

- the keyboard should only call `addData()`.
- `nextEntry()` should be called at the end of a loop if (and only if) it is marked with `isTrials==True`

Nice work! :-)

jonathanoroberts:
Should I withdraw the pull request and work on it? or should I just submit a new one with the changes when I get it worked out?

-Jon



On Jun 1, 2017, at 9:51 AM, Jon Peirce <notifications@github.com<mailto:notifications@github.com>> wrote:


Ah, I see. Yes, exactly:

  *   the keyboard should only call addData().
  *   nextEntry() should be called at the end of a loop if (and only if) it is marked with isTrials==True

Nice work! :-)

jonathanoroberts:
It is a one line fix… There is already a call to addData() in the keyboard component, followed by the call to nextEntry(). I deleted the nextEntry() and it works like a charm with the added loop iteration call to nextEntry()

-Jon



On Jun 1, 2017, at 9:51 AM, Jon Peirce <notifications@github.com<mailto:notifications@github.com>> wrote:


Ah, I see. Yes, exactly:

  *   the keyboard should only call addData().
  *   nextEntry() should be called at the end of a loop if (and only if) it is marked with isTrials==True

Nice work! :-)

peircej:
If you change the files and push back to your copy I think this pull request will automatically update to reflect that.

jonathanoroberts:
done…

Thanks!
-Jon



On Jun 1, 2017, at 10:01 AM, Jon Peirce <notifications@github.com<mailto:notifications@github.com>> wrote:


If you change the files and push back to your copy I think this pull request will automatically update to reflect that.

PR #1438, opened 2017-06-14
Py2and3: conversion fixes and py2 tests

peircej:
This completes futurize up to stage2


peircej:
OK, we're mostly there. All tests passing again on Py2. The app opens in Py3 but I haven't turned on Travis testing for Py3 yet (we should add that as a second test domain in .travis.yml and see what happens).

These changes should be compatible with PY2, PY3, wx3.0, wx4.0, and I think we were already dual-compatible with Qt4, Qt5.

It's done using futurize (which becomes a dependency but we could try to get rid of that later).

To force `myObj` into a string (where we would usually have used `str()`) I've found it's quite convenient to do `"{}".(myObj)` . That happily converts to unicode on Py3 and str on Py2.

peircej:
Any further changes from people fixing bugs in PY3 must not break tests on PY2

peircej:
@cbrnr (& @hoechenberger ) I see the issue you mentioned about monitor calibrations not being reusable. Easy fix:

  - have monitors save themselves into a new folder (./pscyhopy2/monitorsPy3) so that the two versions don't get mixed up

Ideal fix:

 - monitors save themselves as JSON files instead, readable by either system. The critical data in a Monitor object is, essentially, a dict of dicts so it should save easily as JSON. Some values are np.arrays though and I don't know what happens to them.
 - if they do `saveJSON()` this automatically whenever loaded from a .calib file then we would automatically save old files into new ones, prior to people switching over and people wouldn't lose any calibration data!

PR #1442, opened 2017-06-16
ENH: Stimulus class for generating random noise stimuli

schofiaj:
I offer this stimulus class for generating visual random noise stimuli of various kinds. There are 4 main classes. Pixel based, Gabor element based, Filtered noise, and Coloured noise. Pixel based can have grey levels distributed according to binary, normal of uniform distributions. Gabor based can be oriented or isotropic. Filter based used Butterworth filtered white noise. Coloured noise allows gives the noise a skewed spatial frequency spectrum and includes white noise and noise derived from the spectrum of an image - this tool randomises the phase spectrum of the image. I’m not 100% sure about using the term coloured noise. I’ve used it as the technical terms for not flat spectra mostly use colour names: pink-noise being the most common. But is might be confused with noise that actually has different hues. I am working on a builder component to go with this. That component will be quite complex with a lot of parameters. I would like it for my lab but wonder how it might be simplified eg different builder components for different classes of noise or different tabs for the parameter belonging to different classes of noise. Do you have a view about adding such a component? I’m happy to upload what I have for discussion at some point soon.

### DIFF
--- a/psychopy/visual/noise.py
+++ b/psychopy/visual/noise.py
@@ -1,0 +1,553 @@
+#!/usr/bin/env python2
+
+"""Stimulus object for drawing random noise stimuliof various kinds. """
+
+# Part of the PsychoPy library
+# Copyright (C) 2015 Jonathan Peirce
+# some code provided by Andrew Schofield
+# Distributed under the terms of the GNU General Public License (GPL).
+
+from __future__ import absolute_import
+
+import pyglet
+pyglet.options['debug_gl'] = False
+import ctypes
+GL = pyglet.gl
+try:
+    from PIL import Image
+except ImportError:
+    import Image
+import cmath
+import psychopy  # so we can get the __path__
+from psychopy import logging
+from psychopy.visual import filters
+from psychopy.tools.arraytools import val2array
+from psychopy.tools.attributetools import attributeSetter
+from .grating import GratingStim
+import numpy
+from numpy import exp, sin, cos
+from numpy.fft import fft2, ifft2, fftshift, ifftshift
+
+from . import shaders as _shaders
+
+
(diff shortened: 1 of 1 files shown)


peircej:
These look great. I haven't got time right now to go through and look at them in detail but the concept is cool. 
I agree that colored* noise could be a misleading term, but I don't know how to get around it. Maybe there's a way that it doesn't have to be named at all, but is available by setting params accordingly? e.g. filtered noise has params of mean, sd, skew (to achieve pink etc) and if these are None then it becomes a "flat" distribution?
I also agree that we need to think how to add it to builder. It is obviously possible for dialog boxes to present some parameters, dependent on other ones (the loop dialog presents different params according to loop type) but it does currently require hard-coding of the dialog box for that object. Ideally we'd have a system for this so that a builder param can be a "sub-param" of another one, and then depends on the value of its parent. But that will take me more time to implement than I currently have. Tabs for different settings, some of which will be ignored when the stimulus is generated, isn't quite as neat but much easier to generate.

* I'm afraid we need to americanise it for consistency, or at least allow american as well

schofiaj:
I don't think you should re-code builder to accomodte parameter dependency. It would be a lot of work for something required only rarely. I will explore the tabs idea and how to rationaise the main classes of noise and submit a draft for consideration.

As for coloured noise I had thought initialy of the term 'fractal noise' as noise with a 1/f spectrum is sometimes called that. I think the idea of a 'skew' parameter within filtered noise will work (it wont be skew in a staistical sense so needs a better name) but would prefer to keep white noise as type selctions that ignores any parameter settings as it seems somewhat perverse to make people set particular filter settings just to get a flat spectrum.

PR #1456, opened 2017-06-30
[NF] added sketch for running ioSync on Arduino Mega

theodleif:
added sketch for running ioSync on Arduino Mega
enables Arduino Mega to work with ioSync
for those who do not have a Teensy3
tested ioSync <---> Arduino, works
but beware as ioHub <--> mcu does not work correctly (MCU.__init__ never starts up), but that is apart from this commit


peircej:
I should point out that I've accepted this essentially without review. I don't know about arduino/sketch code so I can't assess the quality of what you've done. Then again, I can't see how it could break anything that previously existed so it's in! :-)

Issue #1458, opened 2017-07-02
Adding a labjack U3 in the builder will cause an error due to the lack of status attribute


boazmohar:
Using latest standalone psychopy, I add only a U3 labjack and run.
It tried to add a `NOT_STARTED` status:
```python
trialComponents = [p_port]
for thisComponent in trialComponents:
    if hasattr(thisComponent, 'status'):
        thisComponent.status = NOT_STARTED
```
But [U3](https://github.com/psychopy/psychopy/blob/057736cd951f2d313cb7bb4b11a55829a60a9441/psychopy/hardware/labjacks.py#L19) doesn't have a status.

So later in the code it fails:
```python
while continueRoutine and routineTimer.getTime() > 0:
    # get current time
    t = trialClock.getTime()
    frameN = frameN + 1  # number of completed frames (so 0 is the first frame)
    # update/draw components on each frame
    # *p_port* updates
    if t >= 0.0 and p_port.status == NOT_STARTED:
     ...
```

PR #1462, opened 2017-07-13
[BF] Added status to u3 in builder

boazmohar:
Related to issue #1458.
I think this is the right place to fix this.
Feedback will be appreciated.


peircej:
Looks reasonable. What I'm puzzled by is how this ever worked. It definitely did in my own lab, but I don't think we changed anything recently :-/

Issue #1472, opened 2017-08-07
Window Behavior for Minimized Coder and Builder view


adamkimbler:
Recently I've been running into some issues with the default way in which PsychoPy (1.85.2) handles windows that have been minimized. If the app window is closed from the taskbar while minimized, the application saves the state with X coordinates far from the screen (-32000) and a window size that is nearly too small to select in certain OSs. The issue crops up when you then try to reopen Psychopy, which has stored these traits in the appData.cfg and it opens a window massively outside of the screen space and too small to click in taskbar peek previews.

My question is, is this the desired window behavior? I've had an issue with some less savvy research assistants complaining that they couldn't get the application to open when dealing with this, resulting in lost productivity.

peircej:
No, it clearly isn't desired behaviour. I guess the issue is that it hasn't been pressing enough for any contributor to fix it. Or they haven't been able to reproduce it reliably (on OS X I've just tried and can't reproduce the problem)

- Is the -32000 value always the same? i.e. could we use that as a way to detect the issue when the window opens/closes?
- Is this an issue for Coder, Builder or both?
- You say, "certain OSs". Which have you tried?

adamkimbler:
It's been an issue to both coder and builder. The described behavior appears to be limited to Windows 10, as I can't replicate it for MacOS or Linux (Kubuntu 17.04). The -32000 value appears to be the same across all replications across Windows 10. It's possible that it could be fixed with a conditional statement could deal with the issue.

PR #1474, opened 2017-08-07
fixed issue with coder view minimized in Windows

adamkimbler:
I made a quick workaround for the issue I raised in #1472


peircej:
Just the Coder one in the end?

adamkimbler:
Yeah, I took a look through the builder code and it doesn't seem to have the same issue. The way it sizes the window is completely different.

adamkimbler:
It should be fixed now, I added the BF

peircej:
It's actually in the commit that I need it rather than the pull request. No worries though. Just for future :-)

PR #1479, opened 2017-08-13
Added Some Docstrings, Some linting work

adamkimbler:
added docstrings for various methods that didn't have them, fixed an indirect import in movie component


psychopy/app/builder/builder.py
```diff
@@ -558,9 +559,8 @@ def __init__(self, frame, id=-1):
     def getCurrentRoutine(self):
         routinePage = self.getCurrentPage()
         if routinePage:
-            return routinePage.routine
-        else:  # no routine page
```
hoechenberger: I actually prefer this explicit statement. Matter of taste? What do the other devs think? @peircej?

psychopy/app/builder/builder.py
```diff
@@ -570,16 +570,14 @@ def setCurrentRoutine(self, routine):
     def getCurrentPage(self):
         if self.GetSelection() >= 0:
             return self.GetPage(self.GetSelection())
-        else:  # there are no routine pages
```
hoechenberger: Same here...

psychopy/app/builder/builder.py
```diff
@@ -802,16 +800,18 @@ def getIndexInSizer(self, obj, sizer):
         # for the main sizer we kept track of everything with a list:
         if sizer == self.sizer:
             return self.sizerList.index(obj)
-        else:
```
hoechenberger: … and here.

psychopy/app/builder/builder.py
```diff
@@ -1,14 +1,15 @@
...
+Part of the PsychoPy library
+Copyright (C) 2015 Jonathan Peirce
+Distributed under the terms of the GNU General Public License (GPL).
+"""
 
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import division
 
-from builtins import str
```
hoechenberger: Are you certain about this change?
http://python-future.org/compatible_idioms.html#basestring

adamkimbler: Good catch! That should be fixed in the PR now, I was intending to combine the `builtins` imports to one line, but it looks like `str` got away

psychopy/app/builder/builder.py
adamkimbler: I can agree with you on these.

peircej:
Hi Adam, I'm hesitant about this. We've debated it before (mostly with Richard!) but I' not super-keen on code clean-ups purely for the sake of making things pretty. The issues I have is that:

 - it actually makes it harder to see who last changed the *function* of a line (the blame view on github shows the last change which is annoying if the changes are all cosmetic)
 - it's all too easy to introduce bugs when simply changing a line cosmetically (like the one that Richard caught, above, that wouldn't have broken the tests but would have broken compatibility). Prettier code with more bugs is bad, so I prefer to leave working code alone.
 - making the underlying code look pretty is not the key priority of the users. They prefer us to work on buglets, rather than PEP8 stuff.

The other thing is I'd appreciate it if you follow the other commit tags, again, so we can see what a commit was about (I mentioned BF before, because that's the most important, but if you're going to contribute substantially it helps to use the others):

Sorry, I don't mean to be a pain when you've been kindly volunteering time, but I like people to spend their time on stuff that's useful! :-)

best wishes,
Jon

PR #1490, opened 2017-08-22
RF and compatibility: fixes for more tests to pass under PY3

peircej:


peircej:
Now 37 tests failing and 456 passing on my local OSX machine

PR #1492, opened 2017-08-23
Add == and != operators for DataHandler and _BaseTrialHandler, and JSON export for staircases

hoechenberger:
This PR consists of two improvements:

1. it  implements the `__eq__()` and `__ne__` (`==` and `!=`) methods for `DataHandler` and `_BaseTrialHandler`(and derived classes)
2. it adds JSON serialization for StairHandler and derived classes

### `__eq__()` and `__ne__()` for `DataHandler` and `_BaseTrialHandler`

Example:
```python
In [1]: from psychopy.data import TrialHandler
   ...:
   ...: t1 = TrialHandler([dict(foo=1)], 2)
   ...: t2 = TrialHandler([dict(foo=1)], 2)
   ...:
   ...: print(t1 == t2)
   ...:
True
```
Specializations were required for

- `TrialHandler2`: ignore RNG object
- `QuestHandler` and `PsiHandler`, as they use contrib classes (see below).

`QuestHandle`r and `PsiHandler` rely on classes implemented in contrib modules; these should be left unchanged. Because an `==` or `!=` comparison on two `QuestHandler` / `PsiHandler` instances will eventually descent into their accompanying `QuestObject` / `PsiObject`, `__eq__()` and `__ne__()` have to be implemented for these objects as well. However, I did not want to
modify the contrib code, so I first tried an approach with monkey-patching, but this was ugly and not very intuitive. I now subclass `QuestObject` (as `QuestObject_`) and `PsiObject` (as` PsiObject_`),
respectively, and add the required `__eq__()` and `__ne__()` methods there. Accordingly, `QuestHandler` and `PsiHandler` will now instantiate `QuestObject_` and `PsiObject_`, respectively.

Why all this? Mostly because it will ease writing tests and debugging. For example, when saving a `StairHandler` object to disk and later reading it back in, it can now be compared to an existing in-memory object very easily -- which paves the way for a better testing infrastructure and makes it easier (and safer) to add new serialization implementations, such as JSON export.

### JSON serialization for StairHandler and derived classes

StairHandler and derived objects can now be serialized as JSON. Serialization of TrialHandler* is currently not supported, as json-tricks cannot properly handle the required DataHandler, [because
it is derived from a dict.](https://github.com/mverleg/pyjson_tricks/issues/30)

This is a first step in allowing us and our users to move away from pickle for long-term storage, and makes it easier to interface e.g. with REST APIs.

### DIFF
--- a/psychopy/data/base.py
+++ b/psychopy/data/base.py
@@ -17,6 +17,7 @@ import inspect
 import codecs
 import numpy as np
 import pandas as pd
+import json_tricks
 from distutils.version import StrictVersion
 
 import psychopy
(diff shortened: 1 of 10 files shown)


hoechenberger:
@peircej Let me know what you think, and whether I should rebase -- forgot to do that before opening the PR!

peircej:
Looks good from what I can tell :-)

PR #1493, opened 2017-08-25
BF: Add status attribute in CedrusBox script generation

armanfarhangi:
Attempting to utilize a cedrusButtonBox response in the Builder view results in the error:

**_AttributeError: 'ResponseDevice' object has no attribute 'status'_**

Unlike classes like KeyBoardComponent, cedrusButtonBoxComponent does not have a corresponding response class in which the status attribute is initialized. However, the status attribute is still necessary for the experiment script to run, hence the changes.


peircej:
Your indentation was broken in that change. I've pulled it in anyway and fixed the indentation in 9fb5dd91767b

Try to write/edit python code using an editor that is Python-aware (Spyder and PyCharm CE are both excellent free options). If you use a simpler editor, make sure that it is at least inserting 4 spaces instead of tab characters - that's what nearly all python programmers do. Your commit ended up with mixed tabs and spaces

PR #1496, opened 2017-08-29
FIX: gettext.install() needs unicode=True in Python2.

hsogo:
PsychoPy application crashes on Python 2.7 because translated strings are not output in unicode.
unicode=True is necessary to output translated string in unicode in Python2.

See also [https://discourse.psychopy.org/t/code-name-window-size-on-mac/2939](url)

### DIFF
--- a/psychopy/app/localization/__init__.py
+++ b/psychopy/app/localization/__init__.py
@@ -20,6 +20,7 @@ from builtins import map
 from builtins import range
 import gettext
 import os
+import sys
 import glob
 import codecs
 from psychopy import logging, prefs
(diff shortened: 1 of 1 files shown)


psychopy/app/localization/__init__.py
```diff
@@ -127,7 +128,12 @@ def getID(lang=None):
 except IOError:
     logging.debug("Locale for '%s' not found. Using default." % lang)
     trans = gettext.NullTranslations()
-trans.install()
+
+# gettext.install() needs unicode=True to get unicode output in Python2.
+if sys.version_info[0] >= 3:
```
hoechenberger: Can you please use [`psychopy.constants.PY3`](https://github.com/psychopy/psychopy/blob/9fb5dd91767b3ad112e47a529ac8cc9e1f0254cd/psychopy/constants.py#L11) instead? Thanks!

hsogo:
Would this be OK?

PR #1499, opened 2017-08-30
Update qtgui.py handleLineEditChange fxn

CRiddler:
`isinstance` is used to check if `thisType` is an instance of `basestring` in this case, `thisType` is a class itself, not an instance of a class. If `thisType` is `<type 'unicode'>`, then using `isinstance(thisType, basestring) returns False- which appears to not be the intended function. However using `issubclass(thisType, basestring) returns True because `unicode` is a subclass (not an instance) of basestring.

### DIFF
--- a/psychopy/gui/qtgui.py
+++ b/psychopy/gui/qtgui.py
@@ -181,7 +181,7 @@ class Dlg(QtWidgets.QDialog):
                 thisType = self.inputFieldTypes[ix]
 
                 try:
-                    if isinstance(thisType, basestring):
+                    if issubclass(thisType, basestring):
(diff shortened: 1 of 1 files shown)


peircej:
Thanks - looks like that was introduced by an incorrect auto-convert while updating to Py3 compatibility.
By the way, could you read through the notes on commit messages: http://www.psychopy.org/developers/repository.html#commit-messages
When I have to look back through large numbers of commits, having good messages (with tags about what type of thing this was) is really key. In this case the commit was `FF:` because it relates to a fix in code that hasn't yet been released.

thanks again for helping :-)

PR #1501, opened 2017-08-31
FIX: adds a missing comma to the dictionary

miykael:


peircej:
Thanks. Good catch.

Could you read through the notes on commit messages: 
When I have to look back through large numbers of commits, having good messages (with tags about what type of thing this was) is really key. In this case the commit was FF: because it relates to a fix in code that hasn't yet been released. It means it was a fix that *shouldn't be* in a 1.85.x release, only in 1.86

thanks again for helping :-)

miykael:
Happy to help. PsychoPy is a great tool. Thanks for creating it!

Sorry about the wrong commit prefix, I will use the right one for the next PR.

Issue #1497, opened 2017-08-29
Trial csv not written when 'esc' key pressed


CRiddler:
I'm testing an experiment (very simple set up) that, when run in its entirety is perfectly fine. It logs everything, and my output is written as expected. However whenever I interrupt the experiment via the 'esc' key I get this output message:

`Exception RuntimeError: RuntimeError('sys.meta_path must be a list of import hooks',) in <bound method ExperimentHandler.__del__ of <psychopy.data.experiment.ExperimentHandler object at 0x112387550>> ignored`

Which, to my knowledge, appears as if the ExperimentHandler.__del__ method is being skipped over- resulting in no output `.csv`

This is on a fresh standalone install of psychopy 1.85.3 on a Mac OS X El Capitan 10.11

Edit:
Forgot to mention that this experiment was built completely with the psychopy builder gui- I have not used to coder to create this experiment.

CRiddler:
Did some poking around and found out that this bug has to do with an issue from the pickling the ExperimentHandler class and the python 2 to 3 compatibility changes. 

Prior to the change, python 2x's cPickle module was used- however python 3x only has pickle (which is the equivalent of cPickle to my understanding, pickle was essentially replaced by cPickle). However, to make it 2/3 compatible `import cPickle` was replaced with `import pickle`. This means in python2 the regular `pickle` is being used, while in python3 `cPickle` is being used under the hood. The reason this makes breaking difference is that python2 `pickle.dump()` uses a default protocol of 0- which is "… the original ASCII protocol and is backwards compatible with earlier versions of Python" (see python2.7 pickle module on the docs). This protocol encounters an error when attempting to pickle the ExperimentHandler object- more testing is required to dive into the specifics as to why this is an issue though.

However to remedy this error we can do either of the following inside of `experiment.py`
* replace `import pickle` a version dependent importation of cPickle for py2, or pickle for py3-
    try:
        import cPickle as pickle
    except ModuleNotFoundError:
        import pickle as pickle
* replace line 333 `pickle.dump(self, f)` with `pickle.dump(self, f, pickle.HIGHEST_PROTOCOL)`

I'm more than willing to make a pull request for these, but I figured since I reported this as an issue/bug I'd post it here first.

peircej:
Ah, OK. Thanks for digging around there.

[...]

PR #1504, opened 2017-09-01
BF: Ensure PY2/3 compatibility, remove superfluos code

hoechenberger:
Addresses a part of #1497.

### DIFF
--- a/psychopy/tools/filetools.py
+++ b/psychopy/tools/filetools.py
@@ -12,13 +12,18 @@ from future import standard_library
 standard_library.install_aliases()
 import os
 import shutil
-import pickle
 import sys
 import codecs
 import json_tricks
 
 from psychopy import logging
 from psychopy.tools.fileerrortools import handleFileCollision
+from psychopy.constants import PY3
+
+if PY3:
+    import pickle
+else:
+    import cPickle as pickle
 
 
 def toFile(filename, data):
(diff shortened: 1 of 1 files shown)


hoechenberger:
I tested pickling & unpickling using both Python 2 and 3.

peircej:
Not sure we need to remove the code that disables the saving. The reason it's here as well as the in the save() location is that this can handle files that were created before I realised this was an issue, but probably those are long enough ago now that nobody is still opening them. 

According to the changelog this was an issue up to version 1.76.00. I guess the question is whether we support loading of data from that long ago (indeed, whether it's even possible to load data from that long ago! It won't be in Python3).

hoechenberger:
My reasoning was that unpickling with any version of PsychoPy that was different from the exact same one used for pickling will most likely fail -- or am I mistaken here? Then again, leaving the call to `abort()` in wouldn't hurt, would it..?

peircej:
> My reasoning was that unpickling with any version of PsychoPy that was different from the exact same one used for pickling will most likely fail -- or am I mistaken here? 

Actually yes, mistaken. ;-) The unpickling has mostly been robust to pretty much everything. The only change I know about that caused an issue was when we switched to new-style classes (subclassing `object`) and, for that, we had a period of keeping a copy of old-style classes purely for unpickling files.

I **think** Py3 is not going to be able to unpickle old files though.

hoechenberger:
I guess you're right, closing this one.

PR #1505, opened 2017-09-01
BF: Fix Garbage collector reliance for ExperimentHandler clean up in core.quit()

CRiddler:
I've implemented a clean up system via an ExperimentHandler variable that keeps track of all instances of the class, then calling their `.close()` methods manually prior to shutting down threads.

Fixes the underlying issues discussed in: #1497

### DIFF
--- a/psychopy/data/experiment.py
+++ b/psychopy/data/experiment.py
@@ -9,6 +9,7 @@ from builtins import str
 import sys
 import copy
 import pickle
+import weakref
 
 from psychopy import logging
 from psychopy.tools.filetools import openOutputFile, genDelimiter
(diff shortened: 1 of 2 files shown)


psychopy/data/experiment.py
```diff
@@ -335,6 +334,18 @@ def saveAsPickle(self, fileName, fileCollisionMethod='rename'):
...
+    def close(self):
+        if self.dataFileName not in ['', None]:
+            if self.autoLog:
+                logging.debug(
+                              'Saving data for %s ExperimentHandler' % self.name)
+            if self.savePickle == True:
+                self.saveAsPickle(self.dataFileName)
+            if self.saveWideText == True:
+                self.saveAsWideText(self.dataFileName + '.csv', delim=',')
+        self.abort()
```
hoechenberger: Do we really still need to call `abort()`?

CRiddler: I put the abort call because we kept the '__del__' method. This way if the close method is called explicitly then the ExperimentHandler won't try to rewrite the output files if that instance is later deleted (since deleting it now calls '.close()')

PR #1508, opened 2017-09-07
Update fMRI_launchScan.py

ilkayisik:
Corrected a typo in the comments


peircej:
Thanks. If possible it helps if you could tag the beginning of your commit with:
BF : bug fix
FF : ‘feature’ fix. This is for fixes to code that hasn’t been released
RF : refactoring
NF : new feature
ENH : enhancement (to existing code, but don't worry too much about the diff between this and NF)
DOC: for all kinds of documentation related commits
TEST: for adding or changing tests

In your case it was a DOC commit. These help me to work out which commits need copying to other branches of the project and which are likely/unlikely to be the source of later bugs. thanks!

Issue #1509, opened 2017-09-08
OpenCV playback of movies giving type error float in 1.85.03


frankgasking:
Hi there,  

I'm currently running PsychoPy 1.85.03 on Windows 7, where i'm getting the attached error message each time I run a movie file using the "opencv" playback.

I'll try and dig a bit further next week - but was wondering if anyone has seen this before and know how to resolve?

Many thanks,

Frank
![opencv-error](https://user-images.githubusercontent.com/2048377/30220436-4ed5c4dc-94b7-11e7-91c7-4ef513d93494.PNG)

CRiddler:
It appears that the issue stems from at least one of the inputs into `numpy.zeros()` is a float.

    >>> np.zeros((1.0, 5, 7), dtype=np.uint8)
    Traceback (most recent call last):
        File "<input>", line 1, in <module>
    TypeError: 'float' object cannot be interpreted as an index

The full line in question is:

    `self._numpy_frame = numpy.zeros((self._video_height, 
                                      self._video_width,
                                      self._video_frame_depth),
                                      dtype=numpy.uint8)`

A few lines above this, `self._video_frame_depth` is hard coded to 3, which means this isn't the source of our float issue, which means that `self._video_width` or `self._video_height` which are both gotten via opencv are being returned as floats. 

The `self._video_width` and `self._video_height` are defined as follows:

    self._video_width = self._video_stream.get(
            cv2.CAP_PROP_FRAME_WIDTH)
    self._video_height = self._video_stream.get(
            cv2.CAP_PROP_FRAME_HEIGHT)

On my psychopy if I enter the following commands:
    
    >>> cap = cv2.VideoCapture(0)
    >>> cap.get(cv2.cv.CV_CAP_PROP_FRAME_WIDTH)
    1280.0
    >>> cap.get(cv2.cv.CV_CAP_PROP_FRAME_HEIGHT)
    720.0

They are in fact returned as floats, confirming my suspicions. A fix should be to  wrap both self._video_width and self._video_height inside of `int()` (when we define them) to ensure that they are integers and not floats.

I'll submit these changes as a PR, but if you want to change these inside your copy of psychopy, this should solve the issue you're experiencing.

frankgasking:
Ah, that's brilliant - thanks CRiddler for the explanation and diagnosis.  

[...]

PR #1510, opened 2017-09-08
BF: self._video_height/width should be int, not float

CRiddler:
opencv returns `self._video_stream.get(cv2.CAP_PROP_FRAME_WIDTH)` (and height) as a float. This causes an error a few lines down in
`self._numpy_frame = numpy.zeros((self._video_height, self._video_width, self._video_frame_depth),  dtype=numpy.uint8)`

By changing these variables to integers, this error does not occur.
See: #1509


peircej:
Thanks for sorting that out. Looks like either OpenCV has changed in what it returns or numpy has gotten more strict. Forcing to `int` here is sensible regardless.

PR #1517, opened 2017-09-18
FF: atexit solution in favor for closing ExperimentHandlers

CRiddler:
In correspondence with #1497 I've replaced the old fix for this solution: keeping track of ExperimentHandlers manually with weakref, then manually calling the each instance's `.close()` method in `core.quit()` 

with a new, cleaner fix: use `atexit` to register each instances `self.close` method to be called on interpreter exit before garbage collection. This does fix the same issue that the original weakref patch was working towards.


peircej:
nice one @criddler and @hoechenberger :-)
I think this closes #1497

PR #1527, opened 2017-09-21
BF: Line.{contains,overlaps}

hoechenberger:
The `contains()` and `overlaps()` methods of `visual.Line` had a different signature than their parent (accepting no arguments), essentially being useless and violating [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle). We now consume all passed (keyword) arguments, fixing this problem. We also explicitly return `False` now, as has always been indicated by the docstrings. Added tests for this behavior. While at it, I additionally cleaned up the test scripts a little bit to improve readability.

### DIFF
--- a/psychopy/tests/test_all_visual/test_all_stimuli.py
+++ b/psychopy/tests/test_all_visual/test_all_stimuli.py
@@ -19,14 +19,17 @@ To add a new stimulus test use _base so that it gets tested in all contexts
 """
 _travisTesting = bool("{}".format(os.environ.get('TRAVIS')).lower() == 'true')
 
+
 class Test_Window(object):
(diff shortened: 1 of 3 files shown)


peircej:
I'm pulling this in for now. My thoughts are that, for a line,  it's correct that `contains()` is always False but `overlaps()` seems reasonable to be True (if one of the vertices is inside another shape I would consider that an overlap?)

hoechenberger:
Totally agree, I was just trying to follow the "original" implementation as closely as possible. I will look into the `overlaps` method again.

PS You can certainly bring up such thoughts and concerns during the review process; I don't insist on getting my PRs immediately merged in any way :)

PR #1533, opened 2017-09-26
Multiple improvements to data.py tests, small bugfxies, Py2/3 compat

hoechenberger:
This PR:

- Fixes problems with `StairHandler` tests related to temporary file generation
- Switches `data.py` tests that generate random data to a seeded `numpy.random.RandomState` for reproducibility, and updates test output files accordingly
- Adjusts `Trialhandler{2,Ext}` tests for Py2/3 compatibility
- Fixes a small bug in the `QuestHandler`
- Raises more sensible exceptions in `data.utils.importConditions`
- Switches from an ordinary `dict` to an `OrderedDict` for the conditions generated by `importConditions` for reproducible behavior (i.e., identical behavior in both Py2 and 3)

PR #1534, opened 2017-09-27
Improvements to file saving

hoechenberger:
The PR aims to reduce the `openOutputFile` delimiter and filename extension confusion a bit. Also, we now use a context manager from creating many (but not all) output files. Exception: long "saving procedure" code blocks that I didn't want to wrap in an `with` block (just yet).


peircej:
Am I right in thinking that this rigidly fixes the delimiter to the file extension? I'm not sure that's wise. Although 'csv' implies that ',' is the delim some people use it with other delims (like tab). And some users might want an extension that's explicitly delim-agnostic (like '.dlm').

hoechenberger:
Hi @peircej 

> Am I right in thinking that this rigidly fixes the delimiter to the file extension?

I took great care not to change the behavior of the file saving methods; for that reason, I implemented improvements to the `data.py` tests #1533 before starting to work on the changes proposed here.

My perception was that the existing code was much too convoluted and hard to understand, and although I believe I managed to improve the situation a bit with this PR, your question suggests that it's still not really obvious what's going on when saving a file. The delimiter is not rigidly fixed. Instead, we have to deal with the situations that users may provide 

- a filename without an extension and with no delimiter
- a filename without an extension and with a delimiter
- a filename with an extension and with no delimiter
- a filename with an extension and with a delimiter

- We try to pick a "sane" delimiter (based on the extension) if none was specified.
- We try to pick a "sane" extension if no delimiter was specified.

Overall, saving is still pretty messy in many places, but the aim of this PR is to improve the situation at least a little bit. I have more ideas for future improvements, but I have to move slowly, step-by-step, not to overlook something (or introduce unwanted side-effects), as this is a delicate issue.

hoechenberger:
Maintaining backward-compat with the API is also not an easy task / leads to more complex code here, unfortunately :(

hoechenberger:
These last tests ensure that delimiter & file suffix are handles correctly when either or both are present. Should be good to merge now.

hoechenberger:
Oh well, hang on, I think I can improve the tests even more. Will let you know when I'm done.

hoechenberger:
@peircej This should be good to merge now.

This PR mostly adds lots of tests, makes use of context managers for file handling, and simplifies the `tools.filetools.openOutputFile()` function. It should not change use-visible file saving behavior in any way (except for being more robust in case of exceptions).

PR #1480, opened 2017-08-14
FF: Allow pass on missing configobj/future during install.

aforren1:
I think future and configobj are imported pre-`setuptools.setup`, leading to `ImportError`s because they haven't yet been installed.

Example (applies to python 2 and 3, just switch `python=2.7` to `python=3.6`):

```shell
# create fresh env (I'm on Windows, so get numpy/scipy via conda)
conda create --name py27 python=2.7 numpy scipy -y
activate py27

# first try master (e58f828)
pip install git+https://github.com/psychopy/psychopy

#Collecting git+https://github.com/psychopy/psychopy
#  Cloning https://github.com/psychopy/psychopy to c:\users\aforre~1\appdata\local\temp\pip-efryur-build
#    Complete output from command python setup.py egg_info:
#    Traceback (most recent call last):
#      File "<string>", line 1, in <module>
#      File "c:\users\aforre~1\appdata\local\temp\pip-efryur-build\setup.py", line 62, in <module>
#        exec(vStr)
#      File "<string>", line 42, in <module>
#      File "psychopy\__init__.py", line 42, in <module>
#        from psychopy.preferences import prefs
#      File "psychopy\preferences\__init__.py", line 8, in <module>
#        from . import preferences as prefsLib
#      File "psychopy\preferences\preferences.py", line 4, in <module>
#        from builtins import object
#    ImportError: No module named builtins

# grab my fork (re-add the try/catch for configobj, also include future in the mix)
pip install git+https://github.com/aforren1/psychopy@preinstall
# (success)
```

PR #1540, opened 2017-10-05
RF/BF/FF: Imports during installation, Windows requirements, and loading json calibration files

aforren1:
- Better handling of preferences import during installation.
    - Use install-only variable to determine whether or not to skip preferences import in `psychopy/__init__.py`. The previous way I tried (in #1480) wasn't ideal.
- Added `pypiwin32` as dependency on Windows.
- Open json monitor calibration file with 'r', rather than 'rb'.
    - I get a `TypeError: the JSON object must be str, not 'bytes'` on a Windows 10 machine, current psychopy master, python 3.5, when trying to load the default `testMonitor.json`. However, the problem was fixed in python 3.6 (I think https://bugs.python.org/issue10976?).

### DIFF
--- a/psychopy/monitors/calibTools.py
+++ b/psychopy/monitors/calibTools.py
@@ -500,7 +500,8 @@ class Monitor(object):
         else:
             if ext==".json":
                 with open(thisFileName, 'r') as thisFile:
-                    self.calibs = json_tricks.load(thisFile, ignore_comments=False)
+                    self.calibs = json_tricks.load(thisFile, ignore_comments=False,
+                                                   encoding='utf-8', preserve_order=False)
             else:
                 with open(thisFileName, 'rb') as thisFile:
                     self.calibs = pickle.load(thisFile)
--- a/psychopy/monitors/calibTools.py
+++ b/psychopy/monitors/calibTools.py
@@ -499,13 +499,12 @@ class Monitor(object):
             self.calibNames = []
         else:
             if ext==".json":
-                thisFile = open(thisFileName, 'r')
-                self.calibs = json_tricks.load(thisFile, ignore_comments=False)
+                with open(thisFileName, 'r') as thisFile:
+                    self.calibs = json_tricks.load(thisFile, ignore_comments=False)
             else:
-                thisFile = open(thisFileName, 'rb')
-                self.calibs = pickle.load(thisFile)
+                with open(thisFileName, 'rb') as thisFile:
+                    self.calibs = pickle.load(thisFile)
             self.calibNames = sorted(self.calibs)
-            thisFile.close()
             
             if not constants.PY3:  # saving for future (not needed if we are IN future!)
                 # save JSON copies of our calibrations
@@ -591,9 +590,8 @@ class Monitor(object):
         """
         if not constants.PY3:  # don't ever save pickle files form PY3
             thisFileName = os.path.join(monitorFolder, self.name + ".calib")
-            thisFile = open(thisFileName, 'wb')
-            pickle.dump(self.calibs, thisFile)
-            thisFile.close()
+            with open(thisFileName, 'wb') as thisFile:
+                pickle.dump(self.calibs, thisFile)
         # also save as JSON (at the moment)
         # (When we're sure this works we should ONLY save as JSON)
         self._saveJSON()
--- a/psychopy/monitors/calibTools.py
+++ b/psychopy/monitors/calibTools.py
@@ -498,10 +498,11 @@ class Monitor(object):
         if not os.path.exists(thisFileName):
             self.calibNames = []
         else:
-            thisFile = open(thisFileName, 'rb')
             if ext==".json":
+                thisFile = open(thisFileName, 'r')
                 self.calibs = json_tricks.load(thisFile, ignore_comments=False)
             else:
+                thisFile = open(thisFileName, 'rb')
                 self.calibs = pickle.load(thisFile)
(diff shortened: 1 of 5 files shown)


psychopy/monitors/calibTools.py
```diff
@@ -498,10 +498,11 @@ def _loadAll(self):
         if not os.path.exists(thisFileName):
             self.calibNames = []
         else:
-            thisFile = open(thisFileName, 'rb')
             if ext==".json":
+                thisFile = open(thisFileName, 'r')
```
hoechenberger: While you're at it, would you be willing to switch to a context manager here? i.e.
```python
with open(thisFileName, 'r') as thisFile:
    ...
```
:)

psychopy/monitors/calibTools.py
```diff
@@ -498,10 +498,11 @@ def _loadAll(self):
...
                 self.calibs = json_tricks.load(thisFile, ignore_comments=False)
             else:
+                thisFile = open(thisFileName, 'rb')
```
hoechenberger: same as above

psychopy/monitors/calibTools.py
hoechenberger: Also it would be great if you could add a test so we would catch this kind of problem "automatically" in the future, should a regression occur!

aforren1: Can do. Where do you think the test should live? tests/test_misc? Or should there be a tests/test_monitors?

hoechenberger: > Or should there be a tests/test_monitors?

I believe this would be best, yes. Thanks!

psychopy/tests/test_monitors/test_monitors.py
```diff
@@ -0,0 +1,30 @@
...
+
+@pytest.mark.monitors
+class TestMonitorCalibration(object):
+    def setup(self):
+        pass
+
+    def teardown(self):
+        pass
+
+    def test_reload_monitor(self):
```
hoechenberger: For intelligibility of the tests, it is often best to use simple functions instead of `pytest` classes for simple test cases like this one. This suggestion of mine is a _can_, not a _must_. :)

psychopy/tests/test_monitors/test_monitors.py
```diff
@@ -0,0 +1,30 @@
...
+        self.mon = Monitor('test',
+                           width=40,
+                           distance=57,
+                           gamma=1.0,
+                           notes='Here are notes')
+        self.mon.saveMon()
+        self.mon2 = Monitor('test')
+        assert all([[self.mon.calibs[key][key2] is self.mon2.calibs[key][key2]
```
hoechenberger: Wow, this looks quite complex! Without looking at the code (I'm on my way to bed :)), wouldn't something like `assert mon.calibs == mon2.calibs` suffice?

psychopy/tests/test_monitors/test_monitors.py
```diff
@@ -0,0 +1,30 @@
...
+                     for key2 in self.mon.calibs[key]]
+                    for key in self.mon.calibs])
+
+
+if __name__ == '__main__':
+    cls = TestMonitorCalibration()
```
hoechenberger: I guess you can simply call `pytest.main()` here and remove the other lines.

psychopy/tests/test_monitors/test_monitors.py
aforren1: It would almost work, but there's a numpy array in the `Monitor` object that gums up the works (`gammaGrid`). I've expended the comparison to hopefully make it a bit clearer (and fixed it, the above code didn't work properly). Thank you for looking this over!

psychopy/tests/test_monitors/test_monitors.py
```diff
@@ -0,0 +1,51 @@
...
+    def teardown_class(self):
+        """Remove leftover monitor calibration files (if they exist)"""
+        for f in glob.glob(self.fullname + '.*'):
+            os.remove(f)
+
+    def test_save_monitor(self):
+        """See if the monitor calibration file ended up in the correct location"""
+        self.mon.saveMon()
+        assert os.path.isfile(self.fullname + '.json')
+        if sys.version_info[0] == 2:
```
hoechenberger: Could you use `psychopy.constants.PY3` here instead, please?

psychopy/tests/test_monitors/test_monitors.py
```diff
@@ -1,30 +1,51 @@
...
+            #  additionally, we should have a .calib file in python 2
+            assert os.path.isfile(self.fullname + '.calib')
+
+    def test_reload_monitor(self):
+        """Reload the monitor and verify that all """
+        mon2 = Monitor(self.monitor_name)
+        # test that all values in the reloaded monitor match the original instance
+        for key1 in self.mon.calibs.keys():
```
hoechenberger: This is much cleaner now, thanks!

hoechenberger:
Perfect, thank you! I think this is good to merge now :)

peircej:
Alex, for future could you add the labels FF, BF, etc at the beginning of your commit message. When I look back at changes to be included in a new release I look at what type of thing they are. Some releases are only to include bug fixes so I have to search for the right types and this makes a big difference given a massive `git log`. More info in http://www.psychopy.org/developers/repository.html#commit-messages

thanks for your help!

aforren1:
Can do, sorry about that!

PR #1549, opened 2017-10-11
DOC: README updates

hoechenberger:


peircej:
Slight issue I have here is that, although the root readme file is useful to github, the psychopy/readme file is the one that gets included in the distributed versions (e.g. with pip). Anything outside the psychopy folder is packaged.

PR #1555, opened 2017-10-15
Fixes for shellCall(), and  improvements to psychopyApp.py on macOS with Anaconda

### LINKED ISSUE #1326: GUI install under OS X: handling of framework-python
TL;DR If you want to run the `psychopyApp.py` GUI on an OS X dev-install, call

```bash
python.app `which psychopyApp.py`  # or equivalently use: pythonw
```

## Description

Even on successful OS X compilations, and after this locale-fix #1325, the distutils install leaves the script `psychopyApp.py` with the shebang

```bash
#!/Users/cjayb/miniconda_envs/psychopy/bin/python
```

_i.e._, one that is not compiled against the "Framework" libs and therefore does not get access to the screen:

```bash
This program needs access to the screen.
Please run with a Framework build of python, and only when you are
logged in on the main display of your Mac.
```

Both `pythonw` and `python.app` are [easily installed using conda](https://gist.github.com/cjayb/67d8f6ada85f8fcd58ff1d4e1fc20764), and it's possible to specify the interpreter for `scripts`:

```bash
python setup.py build --executable=`which python.app`
```

I've also tried simply:

```bash
python.app setup.py install --user
```

but I can't get the pieces to work together! Whatever I try, the shebang remains just `python`.

Any ideas on whether this can be fixed in `setup.py`?

hoechenberger:
This PR
- fixes `core.shellCall()` for Python 3, which did not work before (NB: min. required version: 3.6 – allows to re-use most of the existing Python 2 code unchanged, which is not the case for earlier Python 3 releases)
- allows users to pass environment variables to programs executed via `shellCall()`
- ensures `psychopyApp.py` is run through `pythonw` on macOS with Anaconda (closes #1326)

### DIFF
--- a/psychopy/app/psychopyApp.py
+++ b/psychopy/app/psychopyApp.py
@@ -8,14 +8,26 @@
 from __future__ import absolute_import, print_function
 
 import sys
-from psychopy.app._psychopyApp import PsychoPyApp, __version__
 
 # fix macOS locale-bug on startup: sets locale to LC_ALL (must be defined!)
 import psychopy.locale_setup  # noqa
 
+
 # NB the PsychoPyApp classes moved to _psychopyApp.py as of version 1.78.00
 # to allow for better upgrading possibilities from the mac app bundle. this
 # file now used solely as a launcher for the app, not as the app itself.
+
+
+def start_app():
+    from psychopy.app._psychopyApp import PsychoPyApp
+
+    showSplash = True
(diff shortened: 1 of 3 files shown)


hoechenberger:
I can resolve the merge conflicts later today, @peircej

hoechenberger:
Conflicts revolved, please wait for Travis to finish before merging :)

PR #1557, opened 2017-10-15
ENH: Prepend correct shebang and encoding lines

hoechenberger:
Many `.py` files came with missing, incorrect or incomplete shebang lines and/or encoding specifications. This commit aims to fix this.


hoechenberger:
This is annoying: even after restarting the Travis build, it will use the old `.travis.yml`, which causes errors :(

hoechenberger:
I can try rebasing on `master`, though?

peircej:
Don't worry. Once it's pulled in it will then pass I'm sure. I'm trying to pull in each one today and I'll do some hurried fixing if anything breaks after a pull (I've been tied up with other things the last 2 weeks and haven't had time to update things, but should be better now

PR #1563, opened 2017-10-16
DOC: Add AUTHORS.md, and a script to autogenerate it from git shortlog

hoechenberger:


hoechenberger:
Just rebased on `master`, so the tests will hopefully pass now.

peircej:
Sounds good. I wonder if we should sort names by nCommits?

FYI, I've been planning to create a new Gource animation to reflect the more recent additions and your mailmap might be useful there too :-)

hoechenberger:
> Sounds good. I wonder if we should sort names by nCommits?

I'd been thinking about this too, but then I felt one could get the impression this might kind of segregate contributors into "more" and "less" important people, and could also impair motivation as newbies would always find their names at the very, very bottom. (ok one could claim this might actually work as a _motivation_ to work yourself up the list, but… idk :))

Maybe a "reconciling" approach could be to list authors alphabetically, but to also include the number of commits after their name, like
```
Bill Gates (1337 commits)
```

> FYI, I've been planning to create a new Gource animation to reflect the more recent additions and your mailmap might be useful there too :-)

Sounds great, looking forward :)

PR #1568, opened 2017-10-19
DOC: Updating authors

hoechenberger:
Finally! I've tracked down all real names belonging to the "anonymous" email addresses in the git log.


peircej:
looks like travis is building again! ;-)

PR #1572, opened 2017-10-20
Some improvements for Py2/3 compatibility

hoechenberger:


hoechenberger:
I have no clue why tests are failing :(

hoechenberger:
I believe I figured why the test failed when trying to import `constants.PY3` from `psychopy`: Our current Travis setup doesn't seem to be properly installing PsychoPy! It just sits there in its own directory, but neither `pip` not `distutils` are touching it before we start running our tests. Will look into this some more over the weekend.

PR #1574, opened 2017-10-20
ENH: Improve Py2/3-compatibility by always importing absolute_import

hoechenberger:
We now import `absolute_import` and `print_function` from `__future__` in every PsychoPy file (except iohub).

In a later step, we should also import `division` in all modules, but this requires careful testing.


peircej:
Yes, if nothing else this should flush any problems

hoechenberger:
x-ref #1061

Issue #1571, opened 2017-10-19
Reproducible testing and builds


hoechenberger:
_This is all just some work-in-progress thoughts and be no means complete._

The Problems
==========
We cannot reproduce tests.
--------------------------
We quite frequently have to deal with issues related to differences in the specific environments in which PsychoPy is installed and used. The most obvious scenario is that tests pass locally, but fail on the CI server for no apparent reason. Sometimes, tests that passed previously start failing out of the blue and can be fixed by [installing a different version](1554) of a package that used to work just fine before.

We don't know what we're testing.
--------------------------------
Out current [CI testing configuration](https://github.com/psychopy/psychopy/blob/9995ca7c99bd068482766cf92665cab357d8e518/.travis.yml) for Travis installs a wild mixture of Ubuntu and `pip` Python packages. The directive
```
virtualenv:
  system_site_packages: true
```
means that we can never be entirely sure which set of packages is actually installed, since `apt` –which we use to install the system-wide packages – resolves dependencies automatically. Further, we don't know if the distro's package maintainers may have applied certain changes / fixes to these packages in order to better fit in with the rest of the system.

We do not test on all supported platforms.
------------------------------------------
While we've been working towards Python 3 support for a while now, Travis still runs tests on a Python 2.7 environment only.

Even more, the [vast majority](http://www.psychopy.org/usage.php) PsychoPy users is using Windows and macOS, yet our CI setup is limited to Linux.

_(I'll have to admit that the stats presented on this website probably don't accurately represent _actual_ usage because they don't track `pip` or Linux package manager installations; but the core of my statement remains valid nevertheless.)_

Development and testing are not conducted on the end-user product.
----------------------------------------------------------------------
The standalone bundles of PsychoPy are an extremely convenient way for users to install and explore the software. In theory, we can guarantee that _everything_ will work out-of-the box – which, granted, appears to be the case most of the time. 

There is however a very important and serious issue with our current approach, which is that _development and testing_ are not conducted with the exact same set of packages that are finally bundled and shipped. Therefore, users may experience issues that did not show up during (local and CI) testing. This would still be the case even if we expanded our current CI setup to include Windows and macOS. 

Developers cannot determine which PsychoPy version they're using at runtime.
-------------------------------------------------------------------------------
[...]

PR #1575, opened 2017-10-23
TESTS: Add Anaconda builds & tests for Python 2.7 and 3.6

hoechenberger:
I expanded our testing (Travis) setup to use Anaconda for setting up Python 2.7 and 3.6 environments, and tuned some tiny bits and pieces here and there.

This PR:
- Now also tries to build & install PsychoPy, instead of simply running the tests on the downloaded source tree
- Adds Anaconda testing environments for Python 2.7 and 3.6, while preserving our previous test setup. That is, we now have a testing matrix that includes non-Anaconda 2.7, Anaconda 2.7, and Anaconda 3.6
- Fixes bugs in `iohub` and the `MovieStim2` demo

I don't know if all dependencies for properly testing on Python 3 are actually installed, because obviously `pytest` fails quite early. We'll now just have try and fix the Python3 show stoppers one by one, and add additional required packages whenever needed.

As a next step, I would like to add `macOS` to the test matrix, and set up AppVeyor after that.

Relates to #1571.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_install:
   - echo "Installing PsychoPy dependencies via apt..."
   - travis_retry sudo apt-get install -qq python-xlib python-pygame python-opengl
   - travis_retry sudo apt-get install -qq python-numpy python-scipy python-matplotlib python-pandas
-  - travis_retry sudo apt-get install -qq python-yaml python-lxml python-configobj python-openpyxl
+  - travis_retry sudo apt-get install -qq python-yaml python-lxml python-configobj
   - travis_retry sudo apt-get install -qq python-imaging python-mock
   - travis_retry sudo apt-get install -qq python-qt4 python-wxgtk2.8
   - travis_retry sudo apt-get install -qq python-pyo python-opencv
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,8 +102,7 @@ before_script:
        /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1024x768x24 -ac +extension GLX +extension RANDR +render -noreset;
     fi
 
-# This might come in handy once we switch to Trusty, as Xvfb on precise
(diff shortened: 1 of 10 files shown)


peircej:
Looks amazing Richard (and I know how much effort this takes)! Cool that you got it to test on the built-in *and* on the conda method. What are your thoughts about keeping them both? It seems it won't substantially change test duration because they run parallel.

My guess is that it will take a little while to get the py3.6 tests passing. Do you happen to know if we can change the Travis button so that it reflects a "passing" state if the 2.7 installations pass?

Anyway, great work! :-) Do you want me to hold off merging for now? I'm happy to merge, given that the 2.7 tests pass (on two separated sets of dependent libs!)

hoechenberger:
> Looks amazing Richard (and I know how much effort this takes)! Cool that you got it to test on the built-in and on the conda method. What are your thoughts about keeping them both? It seems it won't substantially change test duration because they run parallel.

Thanks Jon :) I wanted to keep them both for the time being -- like you said, parallel builds don't extend testing time, so it doesn't hurt!

> > My guess is that it will take a little while to get the py3.6 tests passing. Do you happen to know if we can change the Travis button so that it reflects a "passing" state if the 2.7 installations pass?

I've added an `allow_failures` section to `.travis.yml`, which does just that!

> Anyway, great work! :-) Do you want me to hold off merging for now? I'm happy to merge, given that the 2.7 tests pass (on two separated sets of dependent libs!)

I'm not sure about the coverage information that is reported here; I don't know if it's the combined coverage of all tests (including the failing one)?! Maybe we should invoke `coverage` only for one of the builds?

hoechenberger:
Mhh ok, according to `https://coveralls.io/builds/13843338` only the successful (Python 2.7) builds are considered by coveralls. Makes me wonder why the coverage is so massively decreased?! Did I break anything?

peircej:
Don't know. Mysterious :-/

peircej:
I think excluding iohub is probably reasonable. As you say, most of the hardware can't be accessed anyway. Was it being excluded before? Is that why the coverage was where it is?

hoechenberger:
@peircej Should be good to merge now!

Issue #1548, opened 2017-10-11
iohub : datastore errors due to outdated calls to tables API


dantheman39:
Hello all!

I've been having a lot of trouble with ioHub, and when I saw the same type of errors between operating systems (Ubuntu and Mac) I tried to investigate. I've been down a real rabbit hole so I'm trying to keep this constructive, and will only address one issue here. 

iohub.datastore uses pytables (imported simply as "tables"). Since these are manual installs, pip on Mac (via homebrew) and Ubuntu installs a newer version of pytables. The issue here is that PyTables has done a lot of reorganizing and renaming in their API (see <a href="http://www.pytables.org/MIGRATING_TO_3.x.html">here</a>), so this essentially breaks parts of ioHub. 

In my case (on the Mac) I had tried to install an older version of PyTables, but it created some dependency hell with hdf5, and I just couldn't go down that road. PyTables 3 automatically installs a script that can quickly convert a lot of those API calls, and I ran it on datastore.\_\_init__.py, it seemed to do the job, but now on the Mac it's hanging on a call to subprocess (client/\_\_init__.py, about line 960, r=self._server_process.stdout.readline(), which is (probably?) a separate issue ... 

So back to the topic at hand, I have seen a number of issues in discourse and here in the issues page that could be related to this. Particularly import errors, name errors, conspicuously errors where the datastore is not being created or method calls are coming back as not existing: it's because if the user has a more recent version of PyTables, ioHub won't work (and all of the calls to subprocess make this a real pain to debug). 

Of course any decisions will have to include backwards compatibility... I'm curious to see what you all think the way forward on this will be. Thanks, hope you're all well.

peircej:
Yes, it looks like the pytables people read most of PEP8 but not the section https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds
where it says very clearly "do not break backwards compatibility just to comply with this PEP!" ;-)

My *guess* is that there are relatively few of these calls to pytables and we could insert some if... statements to execute the appropriate version (i.e. try to maintain dual compatibility). When I search for `openFile` across all files it only crops up once (in datastore.py).

dantheman39:
I saw that @Schubisu has fixed most of the calls, but I believe two were missed. Check the pull request here: https://github.com/psychopy/psychopy/pull/1582 
[...]

PR #1577, opened 2017-10-26
Fix: updated PyTables API in ioHub/datastore

Schubisu:
As per this issue #1548 ioHub/datastore needed an update to be compatible with the PyTables API version > 3.x

* I've removed the `from tables import *` statements and added explicit imports, to make the files easier to debug.
* Rather than checking PyTables version on every API call, I do the check once after the import statements and use `getattr` for all API calls, as I found that more elegant - not sure what you think.
* I haven't included any tests yet, as this may take a little longer, I only tested with my local set-up and serial device (successfully)
* I've also added some white space around operators and fixed some other Pep8 formatting issues, not sure if this is wanted (do you follow another code style guide?)

In terms of testing, please let me know how to proceed. I plan on writing some dummies for external devices to make ioHub testable at all, but it will take a while to work through the code base first.

### DIFF
--- a/psychopy/iohub/datastore/__init__.py
+++ b/psychopy/iohub/datastore/__init__.py
@@ -2,8 +2,26 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import division
+import os
+import atexit
+import numpy as np
 from builtins import str
 from builtins import object
+from pkg_resources import parse_version
+from psychopy.iohub import printExceptionDetailsToStdErr, print2err, ioHubError, DeviceEvent, EventConstants
+
+import tables
+if parse_version(tables.__version__) < parse_version('3'):
+    from tables import openFile as open_file
+    create_table = "createTable"
+    create_group = "createGroup"
+else:
+    from tables import open_file
+    create_table = "create_table"
+    create_group = "create_group"
+from tables import parameters
+
+
 """
 ioHub
 .. file: ioHub/datastore/__init__.py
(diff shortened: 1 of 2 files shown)


peircej:
Looks good. We know that testing probably doesn't have the coverage to catch all bugs that this might have introduced, so adding tests in this domain will help.

I think @hoechenberger would prefer you use `StrictVersion` rather than `distutils.parse_version` (I don't imagine it makes any difference to be honest, but we are now using `StrictVersion` elsewhere e.g. 8e64b7dc65823595361ddfcfa5ae13307136742e

And I'd prefer that you use BF: or FF: in your commit message (pull request message doesn't matter). When I create a bug-fix only branch I cherry-pick commits with BF: in their commit so if there's a 1.85.5 these 2 commits will get missed. FF (feature fix) is for fixes to code that hasn't been released already; we've added a new feature then found a bug in it but this shouldn't be ported to older branches in maintenance.

But, those things asside, thanks and welcome to the contributors team! :-)

Schubisu:
Thanks, that sounds reasonable, I'll use the BF/FF terminology in future.

I've chosen `distutils.parse_version` over `StrictVersion` because PyTables versioning isn't strict. It would have worked with `LooseVersion` but after reading some more posts, I learned that use of `distutils.parse_version` is more common. I don't know if there are any implications, but will be happy to change that for consistency if desired.

PR #1580, opened 2017-11-02
BF: #1578 wx Phoenix showStartupTips

Schubisu:
I have scanned the psychopy code for old wx API calls just using a grep loop. I checked for all occurrences; apparently all calls have been wrapped in version query if-statements already, except the one that caused my segfault.

Nevertheless, I have replaced all occurrences of `if wx.version() > "some_version_string"` with `StrictVersion` comparisons, the used version strings have not been altered in most cases. Here's the list of occurrences that I have checked:

``` {org}
#+seq_todo: TODO | DONE NOT-APPLICABLE

* wx.AboutBox [1/1]
** DONE psychopy/psychopy/app/_psychopyApp.py:            showAbout = wx.AboutBox
   CLOSED: [2017-11-01 Wed 21:53]

* wx.AboutDialogInfo [1/1]
** DONE psychopy/psychopy/app/_psychopyApp.py:            info = wx.AboutDialogInfo()
   CLOSED: [2017-11-01 Wed 21:53]

* wx.CreateFileTipProvider [1/1]
** DONE psychopy/psychopy/app/_psychopyApp.py:            tp = wx.CreateFileTipProvider(tipFile, tipIndex)
   CLOSED: [2017-11-01 Wed 21:53]

* wx.ShowTip [1/1]
** DONE psychopy/psychopy/app/_psychopyApp.py:            showTip = wx.ShowTip(None, tp)
   CLOSED: [2017-11-01 Wed 21:53]

* wx.EmptyImage [1/1]
** DONE psychopy/psychopy/app/builder/components/__init__.py:        image = wx.EmptyImage(pil.size[0], pil.size[1])
   CLOSED: [2017-11-01 Wed 21:57]

* wx.GetDefaultPyEncoding [1/1]
** DONE psychopy/psychopy/app/coder/introspect.py:            command = command.encode(wx.GetDefaultPyEncoding())
   CLOSED: [2017-11-01 Wed 22:06]

* wx.PySimpleApp [10/10]
** DONE psychopy/psychopy/tests/test_app/test_builder/test_components.py:                tmpApp = wx.PySimpleApp()
   CLOSED: [2017-11-01 Wed 22:10]
** DONE psychopy/psychopy/tests/test_app/test_builder/genComponsTemplate.py:    tmpApp = wx.PySimpleApp()
   CLOSED: [2017-11-01 Wed 22:11]
[...]

Issue #1578, opened 2017-10-27
Segmentation fault on startup when showStartupTips = True


Schubisu:
I'm getting a segmentation fault (core dumped) when starting psychopyApp.py **only with `showStartupTips = True`** in my `.psychopy2/userPrefs.cfg`. I'm not sure if this is gtk related, before removing my .psychopy2 folder, I also received the message
`Gtk-CRITICAL **: gtk_box_gadget_distribute: assertion 'size >= 0'`

I'm using the latest version of the git master branch on Manjaro Linux (gtk 3.22.24) installed in a virtual environment with pip.

This is the backtrace from gdb:

```
Error in sys.exitfunc:

Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x0000000000000030 in ?? ()
(gdb) backtrace
#0  0x0000000000000030 in  ()
#1  0x00007ffff0dc8a45 in wxEvtHandler::Unlink() () at /usr/lib/libwx_baseu-3.0.so.0
#2  0x00007ffff0dc9a23 in wxEvtHandler::~wxEvtHandler() () at /usr/lib/libwx_baseu-3.0.so.0
#3  0x00007fffb968686f in sipwxAuiManager::~sipwxAuiManager() (this=0x2e59230, __in_chrg=<optimized out>) at ../../../../sip/cpp/sip_auiwxAuiManager.cpp:67
#4  0x00007fffb968688a in sipwxAuiManager::~sipwxAuiManager() (this=0x2e59230, __in_chrg=<optimized out>) at ../../../../sip/cpp/sip_auiwxAuiManager.cpp:70
#5  0x00007fffb9689688 in release_wxAuiManager(void*, int) (sipCppV=0x2e59230, sipIsDerived=2) at ../../../../sip/cpp/sip_auiwxAuiManager.cpp:1312
#6  0x00007fffb968976f in dealloc_wxAuiManager(sipSimpleWrapper*) (sipSelf=0x7fffb81d80e8)
    at ../../../../sip/cpp/sip_auiwxAuiManager.cpp:1328
#7  0x00007fffe70a45ea in forgetObject (sw=0x7fffb81d80e8)
    at ../../../../sip/siplib/siplib.c:11711
#8  0x00007fffe70a3e96 in sipWrapper_dealloc (self=0x7fffb81d80e8)
    at ../../../../sip/siplib/siplib.c:11274
[...]

PR #1581, opened 2017-11-02
BF/RF: #1578 wx version checks

Schubisu:
I have refactored [my previous PR](https://github.com/psychopy/psychopy/pull/1580) and dropped the `StrictVersion` version checks in favour of `pkg_resources.parse_version` comparisons.

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -22,9 +22,9 @@ if not hasattr(sys, 'frozen'):
     if haveWxVersion:
         wxversion.ensureMinimal('2.8')  # because this version has agw
 import wx
-if parse_version(wx.__version__) < parse_version('4.0a1'):
+try:
     from agw import advancedsplash as AS
-else:
+except ImportError:  # if it's not there locally, try the wxPython lib.
     import wx.lib.agw.advancedsplash as AS
 
 """ Aug 2017: for now we want to turn off warning for AddSimpleTool
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -316,7 +316,7 @@ class PsychoPyApp(wx.App):
             tipFile = os.path.join(
                 self.prefs.paths['resources'], _translate("tips.txt"))
             tipIndex = self.prefs.appData['tipIndex']
-            if parse_version(wx.__version__) >= parse_version('4.0a1'):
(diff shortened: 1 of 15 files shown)


peircej:
I think this must be an issue with line 81 in `psychopy/app/builder/flow.py`
I think it must be incorrectly treating the test version (wx2.8gtk) as being wx 4, and therefore the call to the rounded rect is incorrect, but I don't know why your new use of __version__ isn't working to show that this is less than 4

peircej:
I *wonder* whether wx in version 2.8 required us to use `version()` instead of `__version__`. I'm sure that was true at one point or we would have used `__version__` all along. But I don't know when they started allowing __version__ as well

hoechenberger:
> I wonder whether wx in version 2.8 required us to use `version()` instead of `__version__`.

Good point, I do remember something similar, too!

Schubisu:
Ah, okay. I will try to find out in detail and correct that. I think a version check with `parse_version` is still safer than string comparison, if done correctly ;)
But this doesn't explain why the Python 3 test fails, does it? To me, the error seems completely unrelated to the edited files.

hoechenberger:
> But this doesn't explain why the Python 3 test fails, does it? To me, the error seems completely unrelated to the edited files.

They Py3 tests never passed so far, no need to worry about them just now.

hoechenberger:
>> I wonder whether wx in version 2.8 required us to use version() instead of __version__.

> Good point, I do remember something similar, too!

Sorry, I was mistaken. Confused that with `pyglet`. Our current Travis setup actually does:
```
$ python -c 'import wx; print(wx.__version__)'
2.8.12.1
```
So using `__version__` should be the correct approach.

hoechenberger:
FWIW, I've added wxPython 4 on Python 2.7 to our test environment: #1585

peircej:
Thanks for your perseverance on this and for getting it ultimately to work. It looks like the problem was the use of "4." rather than a longer (valid) version. "4.0" is apparently OK:

```
>>> p = parse_version
>>> p('2.8.1.0')<p('4.')
False
>>> p('2.8.1.0')<p('4.0')
True
```

Schubisu:
Yes, I realized that too late. I've choosen the `'4.'` notation because `parse_version('4.0') > parse_version('4.0.0a1') is True` and I thought it would cover all these cases. But apparently it doesn't work at all so the explicit way should be the safest.

PR #1582, opened 2017-11-02
BF for ioHub's calls to tables. Most calls were already fixed, but tw…

dantheman39:
BF for ioHub's calls to tables. Most calls were already fixed, but two were missed: _f_get_child in __init__.py and read_where in util.py

Please note that I can't get ioHub working on my machine for unrelated reasons, so I would appreciate someone checking my work. The changes are few.

### DIFF
--- a/psychopy/iohub/datastore/__init__.py
+++ b/psychopy/iohub/datastore/__init__.py
@@ -16,12 +16,12 @@ if parse_version(tables.__version__) < parse_version('3'):
     from tables import openFile as open_file
     create_table = "createTable"
     create_group = "createGroup"
-	_f_get_child = "_f_getChild"
+    _f_get_child = "_f_getChild"
 else:
     from tables import open_file
     create_table = "create_table"
     create_group = "create_group"
-	_f_get_child = "_f_get_child"
+    _f_get_child = "_f_get_child"
 
 
 """
--- a/psychopy/iohub/datastore/__init__.py
+++ b/psychopy/iohub/datastore/__init__.py
@@ -16,10 +16,12 @@ if parse_version(tables.__version__) < parse_version('3'):
     from tables import openFile as open_file
     create_table = "createTable"
     create_group = "createGroup"
+	_f_get_child = "_f_getChild"
 else:
     from tables import open_file
     create_table = "create_table"
     create_group = "create_group"
+	_f_get_child = "_f_get_child"
 
 
 """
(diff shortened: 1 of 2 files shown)


hoechenberger:
Thinking more about this (and #1577), I just figured that PyTables 2.4 (the latest "pre-3.0" release) was released [in 2012](https://github.com/PyTables/PyTables/releases?after=v.3.0.0rc3), alsmost 5.5 years ago. It won't install (lest work) correctly on most modern systems anymore, so I suggest we just drop support for it instead of trying to support both the old- and new-style API. I cannot say that I particularly like the current approach of resolving the correct method names of either version via [string mappings](https://github.com/psychopy/psychopy/blob/0f943e5bcbbd269687be5dede7a8b09fd9b0547c/psychopy/iohub/datastore/__init__.py#L15) and use of `getattr()` calls very much…

Schubisu:
Thanks @dantheman39, indeed I missed those. I double-checked again for further PyTables API calls but this seems to fix the last occurrences.

peircej:
I agree with Richard (probably with everyone!) that the string mappings to function names are a bit ugly. I guess they have the advantage that the line 
` if parse_version(tables.__version__) < parse_version('3'):`
only has to occur once, at the top of the file. But in the future, the lines where we do `getattr()` are probably going to confuse me! Not sure what the preferred solution is. Sounds like Richard would like us to drop `tables<3.0`. Then again it looks like the default version on ubuntu/travis is 2.3.0 (ubuntu is often a long way behind the world)

Schubisu:
I agree that both solutions are all but beautiful. Dropping `tables < 3.0` would of course yield the cleanest code, but we will face the problem with other packages anyway.
I was thinking about writing a wrapper that deals with the version checks in a separate module, but I'm not sure how much effort this would probably take. 
I'm however happy with refactoring the `getattr()` calls to standard `if ... else` statements once this PR is merged; To leave this PR un-merged would leave the bug fix only half done.

hoechenberger:
>  Sounds like Richard would like us to drop tables<3.0. Then again it looks like the default version on ubuntu/travis is 2.3.0 (ubuntu is often a long way behind the world)

We're currently stuck on Ubuntu Precise, which was released more than 5 years ago. I tried moving Travis to Trusty (also to take advantage of the container-based infrastructure), but Bits++ tests started failing there…

We could try to install a newer PyTables on Travis via `pip`, maybe?

> I was thinking about writing a wrapper that deals with the version checks in a separate module, but I'm not sure how much effort this would probably take.

I'm not a big fan of inserting too many abstraction layers, because it tends to make the code more convoluted and less readable. Also, I'm not sure if it's really worth the effort?

> I'm however happy with refactoring the getattr() calls to standard if ... else statements once this PR is merged; To leave this PR un-merged would leave the bug fix only half done.

I would also suggest to merge this PR and maybe continue the discussion in a new issue?

hoechenberger:
> My point was more that the use of old versions of libraries is very widespread, even in developer communities. The fact that it couldn't now be installed doesn't mean that people aren't using it in existing setups.

My reasoning was that we manage to install it on Travis (i.e., Precise) via `pip`, so could everyone else :)

hoechenberger:
But I don't really care, I'm fine with either solution, I guess.

hstojic:
Thanks a lot for the fix, this was causing a headache for me. 

I have tested it (not thoroughly) on Ubuntu 17.04 and EyeLink 1000 and it seems to work.

PR #1587, opened 2017-11-05
ENH: Allow users to disable the "initial rule" of StairHandler

hoechenberger:
The "initial rule"  disregards user-specified n-up/n-down settings and performs a 1-up/1-down staircase until the first reversal point is found. This commit introduces a new keyword argument `applyInitialRule` that can be used to alter this behavior.

### DIFF
--- a/psychopy/data/staircase.py
+++ b/psychopy/data/staircase.py
@@ -69,6 +69,7 @@ class StairHandler(_BaseTrialHandler):
                  nTrials=0,
                  nUp=1,
                  nDown=3,  # correct responses before stim goes down
+                 applyInitialRule=True,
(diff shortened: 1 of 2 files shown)


peircej:
I'm realising we're going to have a BIG problem with the next major release. Currently I manually construct the changelog before a release. The problem is that there are SO many fixes and changes in the prospective 1.86 release. And if we want to wait until it's fully py3 compatible then it will be even longer

In future we'll need people to add changelog entries themselves (or work out a system for doing it between us). One problem I've found in the past is that people writing their own changelog entries made them way too detailed for the average user to care. That means it doesn't get read! :-/

PR #1594, opened 2017-11-09
BF: sys.platform comparison for linux/python3 in some files

Schubisu:
I stumbled over this when running an ioHub example on linux/pthon3 and realized, that many `sys.platform` checks are already conducted with `sys.platform.startwith('linux')` while some still tested for `linux2` only.

I did not update the occurrence in `psychopy/tests/test_hardware/test_ports.py`, so that test will mock the version to `linux2` which should be okay for testing, if I understand correctly.

### DIFF
--- a/psychopy/iohub/constants.py
+++ b/psychopy/iohub/constants.py
@@ -692,7 +692,7 @@ try:
 
         VirtualKeyCodes.initialize()
 
-    elif sys.platform == 'linux2':
+    elif sys.platform.startswith('linux'):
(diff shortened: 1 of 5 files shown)


peircej:
I think the test_ports one is probably largely irrelevant but I can't think of a reason that we actually need it to be `linux2`. I expect these have crept in by contributions from different authors, for whom there might be a preference for `==` comparisons

peircej:
"probably largely irrelevant" because the testing of hardware ports using the test suite is hard to do in an automated setup

PR #1596, opened 2017-11-15
BF: Replaced Hamming window with Hanning window

noahmotion:
Replaced Hamming window with Hanning window in sound/_base.py and backend_*.py where hamming was referenced.

Hamming windows are non-zero at the edges, so they do not remove artifacts effectively.

Hanning windows are zero at the edges, so they do a better job removing artifacts.

### DIFF
--- a/psychopy/sound/_base.py
+++ b/psychopy/sound/_base.py
@@ -38,18 +38,17 @@ knownNoteNames = sorted(stepsFromA.keys())
 
 
 def apodize(soundArray, sampleRate):
-    """Apply a Hamming window (5ms) to reduce a sound's 'click' onset / offset
+    """Apply a Hanning window (5ms) to reduce a sound's 'click' onset / offset
     """
     hwSize = int(min(sampleRate // 200, len(soundArray) // 15))
-    hammingWindow = numpy.hamming(2 * hwSize + 1)
+    hanningWindow = numpy.hanning(2 * hwSize + 1)
     soundArray = copy.copy(soundArray)
-    soundArray[:hwSize] *= hammingWindow[:hwSize]
-    for i in range(2):
-        soundArray[-hwSize:] *= hammingWindow[hwSize + 1:]
+    soundArray[:hwSize] *= hanningWindow[:hwSize]
+    soundArray[-hwSize:] *= hanningWindow[hwSize + 1:]
     return soundArray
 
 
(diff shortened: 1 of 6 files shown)


peircej:
Hmmm, OK, that explains my previous confusion. I remember when this was first added (by @jeremygray I think) that I was puzzled by "Hamming" because I'd heard of "Hanning" but then found that Hamming windows did exist and thought nothing more. Now it turns out they both existed all along! ;-)

The issue now is that Hamming has been in the code for a while and it will break people's scripts to change it. I don't like breaking people's scripts unless it's really necessary, so the question is, "How bad are the effects of the non-zero artefacts?" My guess is that they exist mathematically but would never be noticed by an ear. In that case maybe we should leave things as-is and accept that Hanning might have been a better choice?

noahmotion:
I have heard clicks, which is why I started digging around to try to see what was going on. I think it is worth fixing.

After some of the checks failed, I was reading through the logs and trying to figure out what had gone wrong. As it happens, I think the fix could be implemented without breaking anyone's code. The apodize function could just use a Hanning widow instead of a Hamming window, and we can leave all of the keyword arguments and whatnot as referring to Hamming or hamming.

I was trying to make things consistent throughout, making everything refer to Hanning windows. But the acoustic problem can almost certainly be fixed without breaking any code.

noahmotion:
I just pushed some changes to my fork (restoring references to hamming rather than hanning), but keeping `numpy.hanning()` in the apodize function. I don't know if I need to make a new pull request or not.

peircej:
OK, if we were getting clicks still with some sounds then I agree this is better. :-)
Then the question is how to handle it. With your new code (the pull request is auto-updated with any more recent commits to the same branch) we don't break code but the documentation looks odd because the argument is Hamming and the effect is Hanning. I think this is preferable, and only occurs in the 2 places (Sound __init__ functions). I suggest you change the documentation where appropriate to make it clear that "the arg `hamming` actually applies a Hanning window but was previously a Hamming window and kept the name for compatibility reasons"

Do you agree?

noahmotion:
Yes, that seems reasonable to me. I will try to take care of it as soon as possible (I have meetings for the next few hours...)

peircej:
So I've merged this one for now but create a new pull request if you get around to improving the documentation

PR #1598, opened 2017-11-20
BF: Cleaned up documentation

noahmotion:
Revised the documentation to explain that arguments named 'hamming' are preserved even though the apodize function now uses a Hanning window instead.


peircej:
Thanks for your work on this! :-)

noahmotion:
Glad I could help. Thanks for PsychoPy!

PR #1609, opened 2017-12-05
RF: getting test suite to pass in PY3

peircej:


peircej:
Current status: all the tests themselves actually now pass on all the environments!! :-)

Unfortunately the app then seg faults during its shutdown! :-(

It only seems to happen if a script is run (when running locally). I think it _might_ be to do with the way we extract the stdout/stderror at the end of an experiment.

hoechenberger:
Amazing work, Jon!

Issue #1608, opened 2017-12-05
Issues on non-ASCII characters in Python2


hsogo:
I found that the latest version of Builder in the repository have two bugs in handling non-ASCII characters on Python2.

At first, experiment can't be saved when non-ASCII characters are included in 'Text' field of TextComponent.  Following is an example of error message. This error occurs only on Python2.

```
Traceback (most recent call last):
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\builder.py", line 1604, in fileSaveAs
    self.fileSave(event=None, filename=newPath)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\builder.py", line 1550, in fileSave
    filename = self.exp.saveToXML(filename)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\experiment.py", line 313, in saveToXML
    parent=componentNode, param=param, name=name)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\experiment.py", line 363, in _setXMLparam
    thisChild.set('val', "{}".format(param.val).replace("\n", "&#10;"))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-7: ordinal not in range(128)
```

Next, condition file can't be opened when non-ASCII characters are included in the file path. Following is an example of error message. Again, this error occurs only on Python2.

```
Traceback (most recent call last):
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\dialogs\__init__.py", line 1514, in onBrowseTrialsFile
    returnFieldNames=True)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\data\utils.py", line 276, in importConditions
    logging.debug("Read excel file with pandas: {}".format(fileName))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 26-27: ordinal not in range(128)
```

I confirmed that adding 'u' to strings resolve these issues.  For example, the first error can be fixed as follows. 

```
diff --git a/psychopy/app/builder/experiment.py b/psychopy/app/builder/experiment.py
index 79182ce..97c545a 100644
--- a/psychopy/app/builder/experiment.py
+++ b/psychopy/app/builder/experiment.py
@@ -360,7 +360,7 @@ class Experiment(object):
[...]

PR #1611, opened 2017-12-06
BF: Builder crashes due to non-ASCII characters in path of condition file and TextComponent in Python2

hsogo:
Following bugs are fixed:

1. Builder cannot open condition file when non-ASCII characters are included in the file path in Python2.
2. Builder cannot save experiment when non-ASCII charcteers are included in "Text" field of TextComponent in Python2.

Please see #1608 for detail of error messages.


peircej:
thanks @hsogo 
I suspect this will be a few of many such issues following the conversion of code to Py3 compatibility

PR #1613, opened 2017-12-06
Fixed problem launching builder files on win with wx4

peircej:


peircej:
Fixes issue brought up by @hsogo on discourse:
https://discourse.psychopy.org/t/how-is-the-state-of-progress-on-migration-to-wxpython-4/3570

PR #1625, opened 2017-12-15
BF: Redundant assignment

boazmohar:
Another static code analysis PR.

### DIFF
--- a/psychopy/data/staircase.py
+++ b/psychopy/data/staircase.py
@@ -857,6 +857,8 @@ class QuestHandler(StairHandler):
             method=method, stepType='lin', minVal=minVal,
             maxVal=maxVal, name=name, autoLog=autoLog)
 
+        self.startVal = startVal
+        self.startValSd = startValSd
         self.stopInterval = stopInterval
         self._questNextIntensity = startVal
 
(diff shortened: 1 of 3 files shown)


psychopy/data/staircase.py
```diff
@@ -858,9 +858,6 @@ def __init__(self,
             maxVal=maxVal, name=name, autoLog=autoLog)
 
         self.stopInterval = stopInterval
-
-        startVal = startVal
-        startValSd = startValSd
```
hoechenberger: These lines should _not_ be removed; instead, the values should be assigned to the attributes `self.startVal` and `self.startValSd`, as it probably was meant to be all along.

We should also add a unit test for this.

boazmohar: I have corrected that and added two assert statements to TestQuestHandler, is that ok?

boazmohar:
Weird, that passes on my [Travis](https://travis-ci.org/boazmohar/psychopy/builds/317387838)

hoechenberger:
@boazmohar Thanks for addressing to points I raised! I currently don't have time to check why the Travis tests are failing, though… Looks like a problem related to visual testing, not to any of the changes you've made.

peircej:
FYI, I think either solution with the `startVal` and `startValSD` would have been fine - the value is actually used any further once it's passed to the underlying Quest object but I guess it's nice to keep it there anyway

Issue #1618, opened 2017-12-09
gamma on linux


djmannion:
After [this report on discourse](https://discourse.psychopy.org/t/setting-the-gamma-ramp-manually/3543), I realised that setting the gamma does not work on linux.

Having investigated a bit, I think the problem comes down to the LUT size.

On one machine ("[AMD/ATI] Turks PRO [Radeon HD 6570/7570/8550]"), the LUT size seems to be 1024 rather than 256. Adjusting the code in ``gamma.py`` and passing an appropriate ramp then allows the gamma to be set.

On another machine ("Intel Corporation HD Graphics 520 (rev 07)"), the LUT size was being returned as 0. Changing the xorg driver fixed the LUT size to 256 and gamma functionality was restored.

Moving forward, it seems that we should adjust ``psychopy.visual.gamma`` to create and set the ramps based on a variable ramp size (with a warning or bail if the ramp size is 0).

Here's a draft/mockup of a ramp size getter:

```python
def getGammaRampSize(pygletWindow):
    
    rampSize = ctypes.c_int()

    success = xf86vm.XF86VidModeGetGammaRampSize(
        pygletWindow._x_display,
        pygletWindow._x_screen_id,
        ctypes.byref(rampSize)
    )

    assert success, "XF86VidModeGetGammaRampSize failed"

    return rampSize.value
```

It would also be good to get a sense from other linux users about whether gamma is working for them, given the apparent driver variability.

djmannion:
The ramp size is 2048 (!) on my nvidia machine ("NVIDIA Corporation GK107 [GeForce GTX 650]").

peircej:
Mario Kleiner might know how we can interrogate/set the size of the LUT

djmannion:
Would you be happy for me to put together a draft PR that adapts the ramp to the LUT size? Might be useful even if we do end up being able to set the LUT size.

peircej:
Most definitely! I assume this is a key step in allowing the higher-than-8-bit framebuffers (but maybe not all of it). eg. I assume your 2048 LUT is indicating a 10-bit output (or at least the ability to hold one)

djmannion:
Closed via #1626.

djmannion:
It looks like the gamma functionality is also broken under pygame. A look at the source code suggests that there are a few hardcoded 256s in there - presumably it can't cope with > 8-bit LUTs.

Should we use the gamma approach we use for pyglet windows for pygame also?

[...]

PR #1626, opened 2017-12-16
BF: restore gamma functionality with non 8-bit LUTs 

djmannion:
This is a request for feedback on this approach to restoring gamma functionality on non 8-bit LUT systems.

This relates to issue #1618, where gamma setting silently fails on linux systems with LUT sizes other than 256.

* I've added a test that checks that the requested LUT matches the LUT after setting the gamma. This was failing prior to the other changes.
* I've added a new function in ``gamma.py`` to return the system LUT size. This is only implemented on Linux currently.
* I've used this function to replace the '256' values with the system LUT size.
* The new test passes after these changes.
* While making the changes, I noticed that the gamma table is not reset on window closure if the new gamma is set after window creation. I think we should leave it in the gamma state it was when the window was opened? I've made a couple of edits to do this.

Thanks - let me know what you think of this direction, and I can work on adding Windows and Mac support.


peircej:
The approach looks fine and, yes, I agree that the window should return to initial gamma whenever the user updates (not just on `__init__` update)

djmannion:
I think this is ready to take a look at. I've tested it on each platform and it seems to work OK on each.

One potential issue is what to do when the LUT size is reported as 0. Here, I'm raising an error - but I could see an argument for a warning instead.

peircej:
Thanks @djmannion. This looks good to me, but

- @mdcutone might have an opinion too
- I might need to do some slight refactoring to fit with the changes I made to Window class last week. It's showing merge conflicts right now and I don't know how bad those are, but probably nothing major

djmannion:
I've had a go at resolving the merge conflicts and updating the rest of the code to match the new backend approach. I'll test on Windows and Mac tomorrow.

Also, windows weren't closing - there was an AttributeError caused by looking for ```_hw_handle``` in the backend rather than the window (I think).

djmannion:
There is also some gamma code in ```bits.py``` that I will need to update.

psychopy/visual/window.py
```diff
@@ -1102,8 +1102,12 @@ def close(self):
             openWindows.remove(self)
         except Exception:
             pass
-        if not self.useNativeGamma and self.origGammaRamp is not None:
-            self.backend.gammaRamp = self.origGammaRamp
```
peircej: This change means that we're always using this mechanism whereas, before, we were using pygame's own gamma correction for pygame windows and only using our function for pyglet. Then again, I guess it will work everywhere and we might have more confidence in our own function than one in another external API. So I guess this is fine, but I'm just noting that it's a change

djmannion: I think this is a side effect of being able to restore the system gamma on window closure when the gamma is set after window creation. I had wanted to stop the check on ```useNativeGamma``` (because that is not updated if gamma is set after window creation), but I hadn't picked up on the ```backend``` change.

I think your way of using the backend to restore the gamma is better - presumably the definition of ```self.origGammaRamp``` should use it too.

djmannion: New issue in #1618 might be relevant - perhaps pygame windows should also use the custom gamma methods.

PR #1628, opened 2017-12-18
BF: Movie3 try/except on frame update

jfkominsky:
Further upgrading the resiliency of movie3 from my last fix, I've added a try/except structure to ._updateFrameTexture, such that if it tries to load a frame that does not exist for whatever reason, rather than crashing outright it raises a warning and then moves to the next frame and tries again. I've recreated most previous failure conditions that I encountered and it's robust against all of them so far. There's a theoretical risk of it fast-forwarding to the end of the movie if it somehow tripped on a frame near the end, but I haven't been able to create a situation that does that, and even under those circumstances the worst-case scenario would be it skipping the movie rather than crashing the program.


peircej:
I'm a little concerned about sweeping it under the carpet and moving on. 

Things to improve if possible:

  1. ideally we'd identify the particular errors that you've seen rather than just a blank except. Right you, might be _aiming_ to avoid problems with the file or frame sync, but the blanket  except statement will ignore _all_ errors. That then makes the problem hard to track down. If you found different types of errors in your investigation, we can separate them
  2. many scientists would want to know that something had gone wrong during their study (some _would_ want to abort the run if the movie hadn't been perfectly presented). If nothing else we should provide a warning to let people know that something was missed? Unless the sort of errors you're seeing are genuinely no problem (e.g. because frame rate mismatched make this normal/expected)
  3. ideally we'd also try to detect the root source of the problem, whereas right now we won't notice when it happens and so we'll forget to look any more

jfkominsky:
These are all reasonable concerns. 

1) I have only ever gotten one type of error related to this line, and the error message is a little unhelpful. See screenshot. 
screen shot 2017-12-07 at 3 00 41 pm
The reason I put the try/except here is that as far as I can tell, this is the only level at which I can reasonably interact with the process. The underlying issue, from what I've been able to figure out about moviepy, is that it's looking for a frame at an invalid timestamp. I've only managed to get this to happen when attempting to seek to timestamp 0, which seems to have some kind of rounding error associated with it or something that sometimes (but frustratingly inconsistently) leads it to say that timestamp 0 does not exist (this is part of why I told it to check if the value is less than 0 in my previous bug fix). I haven't been able to figure out why moviepy ends up doing that, or how to get it to fail gracefully with a blank frame or something if it does. 

2. I concur, which is why I have it raise a specific warning, which should (in theory) appear in the console. The warning says "Could not find frame at timestamp [timestamp], advancing to next frame and trying again." to the degree that we trust psychopy's warnings to be informative, it's the best I could do, and it is an accurate description of an issue: if it fails at that line, it has tried to locate a particular frame in the stream and failed. The only risky part is what it does to try to solve the problem, which is advance the timestamp by on frame interval. If it somehow ended up doing that repeatedly it could theoretically advance by frame intervals until it hit the end of the movie. That's a purely theoretical concern, I haven't been able to devise any circumstances where that actually occurs, and I tried (weird timestamps near the end etc.). 

3. The root source is, as far as I can tell, that psychopy can ask for a frame that moviepy does not think exists. More specifically the one failure condition I've noticed is that moviepy is treating timestamp 0 as something other than an int or somehow treating it as -0, because I've never gotten the error to occur except when trying to seek to the beginning of a movie. This was the only way I could think to address that, but I'm open to recommendations. As far as I can from the discourse forum tell this is the only error that arises with MovieStim3, other than the issue of memory overloading (which I think we will ultimately solve with sounddevice, but that's a later problem).

psychopy/visual/movie3.py
```diff
@@ -312,7 +312,13 @@ def _updateFrameTexture(self):
             if self._nextFrameT > (self._videoClock.getTime() -
                                    self._retraceInterval/2.0):
                 return None
-        self._numpyFrame = self._mov.get_frame(self._nextFrameT)
+        try:
+            self._numpyFrame = self._mov.get_frame(self._nextFrameT) 
+        except:
```
peircej: Rather than catching all errors here we need to catch only the one that's relevant (otherwise, when other errors get swallowed here it's hard to debug why). So line 317 should be:
```
        except OSError:
```

psychopy/visual/movie3.py
```diff
@@ -312,7 +312,13 @@ def _updateFrameTexture(self):
...
+            if self.autoLog:
+                self.win.logOnFlip("Frame %s not found, moving one frame & trying again" % (str(self._nextFrameT)), level=logging.WARNING, obj=self)
```
peircej: A warning message like this doesn't need to be time-stamped with the screen refresh. Easier just to issue a warning immediately:
```
logging.warning("Frame {} not found, moving one frame and trying again" 
                .format(self._nextFrameT), obj=self)
```
I also tend to use `{}` and `format(entries)` for string formatting these days - it makes for neater auto-conversion of entries

jfkominsky:
Updated based on requested changes. Let me know if that looks good.

peircej:
Looks good. Will just wait to confirm the tests pass, but I'm sure they will. 
Thanks :-)

jfkominsky:
Addendum: I mentioned the small possibility of it tripping the exception repeatedly and the movie skipping to the end as a result. I finally encountered repeated exceptions, but the behavior was not what I expected. Rather than skipping to the end, it only skipped to the timestamp 1.666667, which in the case of the specific movie I was using is where the sound starts (I created the movie and there is no audio track at all for the first 1.6 seconds). Oddly it is still inconsistent: it doesn't do this every time I seek to the beginning of the video, only sometimes. I have no idea why. It seems specific to this movie file, and probably won't happen with movie files that have audio tracks all the way through (and it's easy enough to put in a "dummy" audio track when creating the movie). I think there's something going on with how moviepy reloads the audio stream when I pause the video partway through rather than at the end, but even then it's not consistent.

I don't think this is a problem for the code. This is a very niche edge case, but we might want to update the docs at some point to give a warning about the behavior of moviestim and intermittent audio.

peircej:
OK, thanks for the info

PR #1630, opened 2017-12-20
Fix a few lgtm.com alerts

sjvs:
This PR fixes some alerts reported by lgtm.com: https://lgtm.com/projects/g/psychopy/psychopy/alerts/

It'd be good if someone could verify whether my change to builder.py was indeed the right thing to do.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -1551,7 +1551,7 @@ class BuilderFrame(wx.Frame):
         self.setIsModified(False)
         # if export on save then we should have an html file to update
         if self.htmlPath:
-            self.fileExport(filePath=self.htmlPath)
+            self.fileExport(htmlPath=self.htmlPath)
         return True
 
     def fileSaveAs(self, event=None, filename=None):
--- a/psychopy/iohub/datastore/util.py
+++ b/psychopy/iohub/datastore/util.py
@@ -405,7 +405,6 @@ class ExperimentDataAccessUtility(object):
                         raise ExperimentDataAccessException(
                             'getEventAttributeValues: {0} is not a valid attribute name in {1}'.format(
                                 v, cvNames))
-                        return None
                 elif isinstance(value, basestring):
                     resolvedValues.append(value)
             return resolvedValues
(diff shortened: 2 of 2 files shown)


peircej:
Those are fine. Thanks. FYI the reason that the bug in the builder code had not been picked up is that it's a feature I've not yet turned on! But nicer catch it beforehand! :-)

peircej:
This pull request fixes 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-1fa649ede6a19f4fb33197103c7c85c0836cab46)

**fixed alerts:**

* 1 for Wrong name for an argument in a call

---

peircej:
Thanks @sjvs 
Couple of things:
  - it seems to take a long time to perform the analysis - longer than our unit tests and they take up to 10 mins!
  - I doubt you'll make nay more commits to this project but, in case you do, could you use commit messages that have prefixes mentioned here (so that I know whether the changes need porting to other git branches): http://www.psychopy.org/developers/repository.html#commit-messages

thanks

sjvs:
Thanks for the feedback @peircej. I had a quick look at what the time is spent on.

About 50 minutes is spent on setting up the right build environment (installing and building your dependencies; e.g. wxPython takes over 30 minutes). We need to replicate the build environment so we can do things like resolving types from dependencies, and analysing data flow through dependencies. If you can see anything in the [log files](https://lgtm.com/projects/g/psychopy/psychopy/logs/rev/pr-1fa649ede6a19f4fb33197103c7c85c0836cab46) that we don't need to do: please let me know!

After the build environment was set up, it took about 10 minutes to extract the relevant information from the source code for analysis. The analysis itself then took about 45 minutes, which is similar to the Travis CI build time.

Any suggestions for improvements are more than welcome!

peircej:
Ah, OK. I think that's too long for me to use with CI. It means I either end up accepting commits where the tests haven't finished (which gives the impression to others that I'm being sloppy) or I have to wait nearly 2hrs before merging a pull request (gulp). I think I'll turn of CI and just run the test periodically when I have time to do some optimisation.

thanks

PR #1637, opened 2017-12-22
Fixes to shellCall (for py3 and also py2!)

peircej:


psychopy/core.py
```diff
@@ -114,9 +115,9 @@ def shellCall(shellCmd, stdin='', stderr=False, env=None, encoding='utf-8'):
     when running Python 2.7.
 
     """
-    if env is None:
-        env = dict()
-
+    if encoding == None:
```
hoechenberger: We should probably use `is` instead of `==` for this singleton comparison. :)

psychopy/tests/test_misc/test_core.py
```diff
@@ -442,14 +440,6 @@ def test_encoding(self):
             with pytest.raises(TypeError):
                 shellCall([self.cmd, self.msg], stdin=self.msg, encoding=None)
 
-    def test_env(self):
-        echo = shellCall(self.env_cmd)
-        assert echo == ''
-
-        echo = shellCall(self.env_cmd, env=dict(FOO='1'))
```
hoechenberger: But shouldn't we leave this test in? Seems like a useful/important functionality to me! In fact, I _think_ we even use that somewhere -- or I'm using it at some places in my code, idk…

psychopy/core.py
peircej: True

psychopy/tests/test_misc/test_core.py
peircej: The command 'env' doesn't exist on windows (the command 'set' was here but it isn't at all equivalent). So, right now, the test only introduces problems and doesn't achieve anything more than the other shellCall tests. Put another way, this test should make sure that the function itself (`shellCall()`) works, not that the call it makes (`env`) works and we're already achieving that.
Happy for it to be added back if you have a smart way to do it but I don't see the need

hoechenberger: Ok, thanks for the clarification! I'm okay with removing it for the time being.

PR #1646, opened 2018-01-04
NF: New feature includes comment toggle, using ctrl/cmd + / for WinLi…

dvbridges:
…n and Mac, respectively, used for toggling comments on and off with a single command. Also, commented text remains selected.

### DIFF
--- a/psychopy/app/coder/coder.py
+++ b/psychopy/app/coder/coder.py
@@ -1057,7 +1057,7 @@ class CodeEditor(wx.stc.StyledTextCtrl):
                 newText = newText + lineText
             else:
                 newText = newText + "#" + lineText
-        self._ReplaceSelectedLines(newText)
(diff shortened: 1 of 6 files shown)


peircej:
The addition of the deselect argument should be removed (it ended up not being used) but otherwise all good
Actually I'm also not sure about the need for two functions, with v similar names, where one simply calls the other. I think that might get confusing later

peircej:
The test fails are not related to the code change. At least part of the problem is a newly broken release of json_tricks that is failing to build on travis
https://github.com/mverleg/pyjson_tricks/pull/40

I'll try and set this to be a fixed version and see if the problem then goes away

peircej:
Nice one @dvbridges Your first contribution complete!
I had to do some work on the test suite to make it work (one of the packages we depend on was broken and it took me a while to work out how to circumvent that)

Issue #1525, opened 2017-09-21
Unicode mess


EruIluvatar:
OS: Windows 10
Psychopy: v1.85.3 fresh install

I get a couple of Unicode{En,De}codeErrors using this version.
For example when starting the "Preferences" menu, it crashes:

> pyo version 0.8.5 (uses single precision)
> Traceback (most recent call last):
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\app\_psychopyApp.py", line 611, in showPrefs
>     prefsDlg = PreferencesDlg(app=self)
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\app\preferencesDlg.py", line 159, in __init__
>     specSection=self.prefsSpec[section])
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\app\preferencesDlg.py", line 264, in makePrefPage
>     parent=panel, name=pLabel, value=thisPref, spec=thisSpec)
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\app\preferencesDlg.py", line 375, in __init__
>     devs = sound.getDevices('output')
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\sound\backend_pyo.py", line 68, in getDevices
>     devName = dev['name'].decode(osEncoding) # convert to unicode
>   File "C:\Program Files (x86)\PsychoPy2\lib\encodings\mbcs.py", line 21, in decode
>     return mbcs_decode(input, errors, True)
> UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 48: ordinal not in range(128)

Similarly the demo "soundStimuli.py" crashes:

> pyo version 0.8.5 (uses single precision)
> Using pyo(with None) for sounds
> Traceback (most recent call last):
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\demos\coder\stimuli\soundStimuli.py", line 25, in <module>
>     highA = sound.Sound('A', octave=3, sampleRate=44100, secs=0.8, stereo=True)
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\sound\backend_pyo.py", line 279, in __init__
>     init(rate=sampleRate)
>   File "C:\Program Files (x86)\PsychoPy2\lib\site-packages\psychopy\sound\backend_pyo.py", line 128, in init
[...]

PR #1656, opened 2018-01-10
FF: support for pyo in python2 and 3

hsogo:
This fix enables to run Coder'S soundStim demo and Builder's voiceCapture demo with pyo 0.8.8 in Python 2 and 3 under non-English environment.  sounddevice package is used to resolve Unicode error of recent versions of Pyo.  Please see [this thread](https://discourse.psychopy.org/t/python3-compatibility-issues/3581/12) for detail.

I think issue #1525 is also resolved by this fix.

Tested setups:
- Japanese Windows 10, python 3.6 (x86), pyo 0.8.8
- Japanese Windows 10, python 2.7, pyo 0.8.8
- Japanese Ubuntu 16.04 LTS, python 3.6, pyo 0.8.8
- Japanese Ubuntu 16.04 LTS, python 2.7 pyo 0.7.8


peircej:
Thanks. I might adapt your `get_devices_infos` to try using pyo first (which will work for many people on english systems and means they don't need souddevice) and then fall back to sounddevice if it fails

hsogo:
I tried to use original `pyo.pa_get_devices_infos()` like following:

```
            try:
                devNames, devIDs = pyo.pa_get_output_devices()
            except UnicodeDecodeError:
                devNames, devIDs = get_output_devices()
```

But unfortunately it doesn't work because UnicodeDecodeError occurs after a while.  Please see [this comment](https://github.com/psychopy/psychopy/issues/1525#issuecomment-332114654) on Issue #1525.  We have to wait for a while to catch UnicodeDecodeError.

peircej:
Ah, OK. Never mind then

PR #1661, opened 2018-01-12
FF: Coder can't open/paste text including non-ASCII characters

hsogo:
**1.** Coder can't open file which contains non-ASCII characters in Python 3 on OS whose default encoding is not UTF-8 (e.g. Windows).  Encoding (utf-8) must be specified to open such file in Python 3.

**2.** Coder can't paste text which contains non-ASCII characters in Python2. The variable `txt` in line 1126 of coder.py is unicode. So we have to  use `encode()`, not `decode()`.

**3.** Script that imports `psychopy.sound` causes RuntimeError when the script is finished when audio library is pyo and running on Windows10.  I experienced this error with pyo 0.8.8 in both of python 2 and 
Following single line is enough to replicate this phenomenon.

```
import psychopy.sound
```

This error doesn't occur when pyo 0.8.8 is directly imported (i.e. `import pyo`).  I investigated this issue and noticed that importing `psychopy.sound._base._SoundBase` before importing pyo causes this error.  Moving `from ._base import _SoundBase` below `import pyo`, the error doesn't occur.  I tested this fix on Ubuntu 16.04LTS and found no problem.

### DIFF
--- a/psychopy/app/coder/coder.py
+++ b/psychopy/app/coder/coder.py
@@ -1128,8 +1128,6 @@ class CodeEditor(wx.stc.StyledTextCtrl):
                 except:
                     # if not then wx conversion broke so get raw data instead
                     txt = dataObj.GetDataHere()
-            else:
-                print("pass!")
             self.ReplaceSelection(txt)
 
     def _GetSelectedLineNumbers(self):
--- a/psychopy/app/coder/coder.py
+++ b/psychopy/app/coder/coder.py
@@ -33,6 +33,7 @@ import threading
 import bdb
 import pickle
 import py_compile
+import locale
 
 from . import psychoParser, introspect
 from .. import stdOutRich, dialogs
@@ -1123,7 +1124,7 @@ class CodeEditor(wx.stc.StyledTextCtrl):
             if not PY3:
                 try:
                     # if we can decode/encode to utf-8 then all is good
-                    txt.decode('utf-8')
+                    txt.encode('utf-8')
                 except:
                     # if not then wx conversion broke so get raw data instead
                     txt = dataObj.GetDataHere()
--- a/psychopy/app/coder/coder.py
+++ b/psychopy/app/coder/coder.py
@@ -2230,12 +2230,21 @@ class CoderFrame(wx.Frame):
                                              readonly=readonly)
             # load text from document
             if os.path.isfile(filename):
-                with open(filename, 'rU') as f:
+                try:
                     if PY3:
-                        self.currentDoc.SetText(f.read())
+                        with open(filename, 'rU', encoding='utf8') as f:
+                            self.currentDoc.SetText(f.read())
+                            self.currentDoc.newlines = f.newlines
                     else:
-                        self.currentDoc.SetText(f.read().decode('utf8'))
-                    self.currentDoc.newlines = f.newlines
+                        with open(filename, 'rU') as f:
+                            self.currentDoc.SetText(f.read().decode('utf8'))
+                            self.currentDoc.newlines = f.newlines
+                except UnicodeDecodeError:
+                    dlg = dialogs.MessageDialog(self, message=_translate(
+                        'Failed to open {}. Make sure that encoding of '
+                        'the file is utf-8.').format(filename), type='Info')
+                    dlg.ShowModal()
+                    dlg.Destroy()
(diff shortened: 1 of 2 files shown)


hsogo:
Oh, sorry, bug fix for soundPyo is also included in this pull request.  This fix hasn't been tested on ubuntu yet.

hsogo:
I tested the fix of soundPyo on Japanese Ubuntu 16.04LTS and found no problem.  I've edited the first comment to explain the problem which is resolved by this commit.

hsogo:
I tested my code only on wx4.  I guess this issue depends on version of wx...  What version of wx do you use?  In addition, what is preferred encoding of your system (i.e. return value of `locale.getpreferredencoding()`)?

hsogo:
Oh, sorry, version of my wx is 3.0.2 for python 2 and 4.0.0b1 for python 3.  Preferred encoding is 'cp932' both in Python 2 and 3.

hsogo:
I read [documentation of wx.TextDataObject](https://wxpython.org/Phoenix/docs/html/wx.TextDataObject.html). It says:

> 
>  **GetFormat**(self)
>     Returns the preferred format supported by this object.
>     This is DF_TEXT or DF_UNICODETEXT depending on the platform and from the build mode (i.e. from USE_UNICODE ).

> **GetFormatCount**(self, dir=DataObject.Get)
>     Returns 2 under Mac and wxGTK, where text data coming from the clipboard may be provided as ANSI ( DF_TEXT ) or as Unicode text ( DF_UNICODETEXT , but only when USE_UNICODE==1 ).
>     Returns 1 under other platforms (e.g. wxMSW) or when building in ANSI mode ( USE_UNICODE==0 ).
> 

I tested these functions in my systems.

- Win, Py2.7, wx3.0.2     GetFormatCount() = **1**,  GetFormat().GetType() = **DF_UNICODETEXT**
- Win. Py3.6, wx4.0.0     GetFormatCount() = **1**,  GetFormat().GetType() = **DF_UNICODETEXT**
- Ubuntu, Py 2.7, wx3.0.2     GetFormatCount() = **2**,  GetFormat().GetType() = **DF_UNICODETEXT**
- Ubuntu, Py3.6, wx4.0.0     GetFormatCount() = **2**,  GetFormat().GetType() = **DF_UNICODETEXT**

Data types of clipboard are DF_UNICODETEXT in all of my systems.  In these systems. `ReplaceSelection()` accepts return value of `dataObje.GetText()` without any problem.

I guess that, in your system, `dataObje.GetText()` may return ANSI text.  If so, following code may resolve this issue.

```
        if success:
            txt = dataObj.GetText()
            if not PY3 and dataObj.GetFormat().GetType() != wx.DF_UNICODETEXT:
                try:
                    # if we can decode/encode to utf-8 then all is good
                    txt.encode('utf-8')
                except:
                    # if not then wx conversion broke so get raw data instead
                    txt = dataObj.GetDataHere()
            self.ReplaceSelection(txt)
```

peircej:
> I guess that, in your system, dataObject.GetText() may return ANSI text. If so, following code may resolve this issue.

No, it does return a unicode object from GetText. The problem is that the unicode object is **incorrect** (nonsense characters). As I said, it really looks like a bug in the conversion performed and the way I've been detecting this is to try and convert back to ANSI and detect if that fails (in the code below I've comment in that code to reflect this).

As far as I can tell, the issue only crops up on Mac with wx3 so maybe we should make that the test:

```
if wx.__version__[0] == '3' and sys.platform == 'darwin'::
    try:
        # if we can decode the utf-8 back to ASCII then all is good?
        txt.decode('utf-8')
    except UnicodeEncodeError:
        # if not then wx conversion broke so get raw data instead
        txt = dataObj.GetDataHere()
```

Notes:
  - it really does need to be `decode('utf-8')` here to work on my machine (with MacOS and wx3)
  - I've now only trapped the specific error that is raised to catch the bug: `UnicodeEncodeError`

It might also be that this is only in the EN locale (in which case there might *still* be a problem on Mac with wx3 in Japan) but we've captured a lot of the cases with the code as above. As we move to wx4 this won't be a problem anyway - there the unicode conversion is being done correctly by wx and we can simply use GetText()

hsogo:
I'm sorry that I've forgotten to revert `decode()` back.  Your code works fine in all of my systems (Windows and Ubuntu).

PR #1666, opened 2018-01-15
[BF] Replace registerPygletWindowHandles with registerWindowHandles

ariekahn:
The ioHub server was previously changed renaming `registerPygletWindowHandles` to `registerWindowHandles` and `registerPygletWindowHandles` to `registerWindowHandles`: https://github.com/psychopy/psychopy/commit/55bbce6cab50d2ec1cc7661240cb8a8f26b74f32

This fixes a remaining reference to the old function names that otherwise causes a crash.

Issue appears to be present in 1.85.6

### DIFF
--- a/docs/source/api/iohub/starting.rst
+++ b/docs/source/api/iohub/starting.rst
@@ -22,7 +22,7 @@ function provides methods for controlling the iohub process and
 accessing iohub devices and events.
 
 .. autoclass:: ioHubConnection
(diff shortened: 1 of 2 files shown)


peircej:
Thanks - good catch :-)
As it happens I'm currently refactoring some of this stuff (so that the pyglet/pygame backend code is better separated from the rest of the Window logic) and I'll need to apply your fix there too.

PR #1669, opened 2018-01-16
ENH: Adds nTrials parameter to TrialHandler. Fixes #407

### LINKED ISSUE #407: Specifying number of trials as parameter in Trialhandler
It would be very nice if you could specify a specific number of trials in the Trialhandler instead of having to implement a counter and an if loop with the body trials.finished = True.

E.g. if one has a design with 50 possible trials but only want to randomly loop through 25 of them. 

Something like:

```
trials = data.TrialHandler(design, nReps=1, nTrials=25, method='random', extraInfo='info', name='Trial')
```

instead of:

```
if count > 25:
    trials.finished = True
```

dvbridges:
This new enhancement provides a parameter to TrialHandler that determines the n number of trials to be presented. The nTrials parameter controls sequence length returned by the _createSequence method and by controlling when the __next__method should iterate over the next repetition of trials i.e., when nTrials per repetition have been presented. Fixes #407

### DIFF
--- a/psychopy/data/trial.py
+++ b/psychopy/data/trial.py
@@ -77,9 +77,7 @@ class TrialHandler(_BaseTrialHandler):
 
             nReps: number of repeats for all conditions
 
-            nTrials: Number of trials to present. If picking fewer trials than
-                    given in conditions list of stimuli, randomisation selects the
-                    n number of trials given.
+            nTrials: Number of trials to present per repetition.
 
             method: *'random',* 'sequential', or 'fullRandom'
                 'sequential' obviously presents the conditions in the order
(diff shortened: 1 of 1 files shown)


dvbridges:
Hi @peircej , just closing the pull request following discussions that this additional nTrials parameter is not useful. It is rare that one would use a random subset of trials from an experiment as this may affect any counterbalancing in place. Also, if subsets are required for practice blocks, then it would make more sense to control the list of practice trials through other methods, e.g., by presenting a single block (using nReps) of main task stimuli, or by creating a practice list of counterbalanced stimuli.

PR #1670, opened 2018-01-16
Simplification of iohub.datastore - remove exception handling.

### LINKED ISSUE #1266: Small iohub.datastore weirdness
So there is this weird block in `iohub.datastore.__init__.py` (https://github.com/psychopy/psychopy/blob/88a359df45ce453d4fd95986cf902e66cf2fe480/psychopy/iohub/datastore/__init__.py#L581):

```python
try:
    global registered_close_open_data_files
    if registered_close_open_data_files is True:
        pass
except Exception:
    registered_close_open_data_files = True
    atexit.register(close_open_data_files, False)
```

As far as I can tell, `registered_close_open_data_files` is never defined in the top-level scope, meaning, the exception will always be raised. Therefore it appears to me this code could be simplified to:
```python
registered_close_open_data_files = True
atexit.register(close_open_data_files, False)
```
@isolver, do you feel like quickly looking into this? Not urgent, though ;) Thanks!

dvbridges:
Fixes #1266

### DIFF
--- a/psychopy/iohub/datastore/__init__.py
+++ b/psychopy/iohub/datastore/__init__.py
@@ -477,13 +477,8 @@ def close_open_data_files(verbose):
             if verbose:
                 print2err('done')
 
-try:
-    global registered_close_open_data_files
-    if registered_close_open_data_files is True:
-        pass
-except Exception:
-    registered_close_open_data_files = True
-    atexit.register(close_open_data_files, False)
+registered_close_open_data_files = True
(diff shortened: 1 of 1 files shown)


peircej:
David, try and remember to use FF: or BF: in your commit tags so that I know which commits need porting to other branches. This one is a bit odd because it technically could be ported but isn't needed - so it doesn't really matter - but it's good practice

PR #1676, opened 2018-01-17
Refactored Window backends into separate files

peircej:
The backend functions for Window (like creating the OpenGL context and providing keys) are now separated out from other plain Window logic (which doesn't care about the underlying window lib). This makes it easier in future to add further backend options.

We now have 
```
Window.backend = PygletBackend(win)
```
And things like
```
Window.flip()
```
actually makes a call to
```
Window.backend.swap_buffers()
```


peircej:
This pull request introduces 3 alerts - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-19915eef5bc4ecfa91f2bc40191278e8c88944b7)

**new alerts:**

* 1 for Unnecessary pass
* 1 for Mismatch between signature and use of an overridden method
* 1 for Potentially uninitialized local variable

---

PR #1690, opened 2018-01-26
Feature improved LS100 linux

tomwallis:
Getting access to the LS100 photometer via the `pyserial` package needed only a simple change (to also check ports on Linux, as done for the PR650 etc). This probably didn't need its own feature branch but I made one before I realised how simple the fix was.

This appears to work on my setup. I wasn't sure what other tests I should add, if any.

Sorry about all the extra "merge" commits; I'm not yet used to Git-with-others. The merge commits happen as a result of using the Github GUI. I'm not sure if I can get rid of them.


peircej:
Hi Tom, for future could you use a prefix on all commit messages (the pull-request message is less important to me) so that I can see whether this is a bug/enhancement etc. It means I know what commits need including in bug-fix releases and what goes into the changelog. Thanks!

peircej:
(that isn't stopping me merging this pul request - I'm just waiting on the tests completing!)

peircej:
FYI the merge commits come about when you merge changes from upstream after you begin work. they aren't a big deal. I will probably 'squash' these into a single commit when I pull it in (and that means I can alter the commit message too).
Is all good :-)

tomwallis:
> for future could you use a prefix on all commit messages (the pull-request message is less important to me) so that I can see whether this is a bug/enhancement etc. It means I know what commits need including in bug-fix releases and what goes into the changelog. 

Sorry! I misread the [developer's guide](http://www.psychopy.org/developers/repository.html) as being about branches, not commits. Will fix in future.

PR #1694, opened 2018-01-28
DOC: corrected PyPi name, pysounddevice to sounddevice

hstojic:
pysounddevice does not exist on PyPi, corrected the documentation

I'm not sure if it is by design but sounddevice does not appear in "...if you want **everything** available..." part?


peircej:
Thanks, yes, I think the "install everything" should include this too although it's a slightly moot point because in the next release we're actually going to provide pip install of dependencies as part of the `pip install psychopy`

PR #1703, opened 2018-02-02
EN: GLFW backend

mdcutone:
# 1. GLFW Backend

This is the initial commit of the GLFW (http://www.glfw.org/) backend for testing purposes. The backend uses a ctypes wrapper so you don't need anything other than the GLFW dynamic libs installed on your PC. You can get them from here: http://www.glfw.org/download.html (or install them from a software repo). You _still_ need Pyglet installed to use this backend.

Changes are quite extensive, with many kludges and hacks to piggy-back on existing Pyglet code to maintain compatibility. _It would be best to refrain from merging until things settle down._ 

## 2. What works
- Opening a GLFW window, full screen and windowed (winType='glfw').
- Keyboard and mouse events (presses and clicks, needs refinement).
- Joystick support (event dispatching does not), use "joystick.backend='glfw'"
- OpenGL setup (useFBO and multisampling)
- Multi-display (fullscreen)
- Screen timing.
- **NEW 04/02/18** Video mode configuration, size, bits per channel, refresh rate. See 2.1.
- **NEW 05/02/18** Setting the gamma ramp.

### 2.1. Video modes
Allows forcing the display into a video mode supported by the graphics card. Only works with fullscreen windows. Some displays need this feature to enter special modes. The display mode to use is specified using the 'size' and new arguments 'refreshRate' and 'bpc' (bits per color channel, specified as tuple (red_bits, green_bits, blue_bits)) when creating a Window. Check your video card's settings panel for supported video modes. If an invalid video mode is specified, the current mode of the display is used (emulates PsychoPy's current behavior).

### 2.2. Gamma Ramps
This backend should support hardware CLUT's >8-bits if there is a supported video mode. 

## 3. What doesn't
- Some stimuli, such as the TextStim.
- Mouse movement handlers.
- Joystick event dispatching (GLFW handles events differently than Pyglet).
- Retina display (removed for now).
- Window positioning (just need to figure out a nice way to do this).

## 4. Not tested
- iohub

## 5. New Features
- Joystick template for the Xbox 360 controller (GLFW only, might work with Pyglet with a bit of work). See 5.1.
- Standard cursor selection (GLFW only).

### 5.1 Joystick Templates
Joystick templates are classes that provide a simplified interface to common joystick/gamepad models. One can access button and axis states though properties which are mapped by the template. For instance, if you want to access the state of a controller's "X" button, you access the property "joy.x". It work the same for axes (ex. "joy.left_thumbstick_axis")


peircej:
This pull request introduces 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-e35f447ce2b9ce873c07a76464a78d199267af98)

**new alerts:**

* 1 for Variable defined multiple times

---

peircej:
This pull request introduces 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-6b804d781e67b9291063a945ec5a0893fdeeace8)

**new alerts:**

* 1 for Raising a tuple

---

peircej:
To make retina work add this immediate after creating win.winHandle:
``` python
win.size = np.array(glfw.get_framebuffer_size(self.winHandle))
```
so that `win.size` correctly now reflects the true size of the window. If the user requests a window of [800,600] they will actually get a resolution of [1600,1200] and this will be reflected correctly in the win.size and subsequent scalings will work correctly (but the user may be surprised that the stimulus appears smaller than the window!)

mdcutone:
I added in the Retina code back into the backend class, I can't test it here since I don't have access to a newer Mac. Also the lgtm errors are going to keep happening since the issue is in the pyGLFW library. The module needs to ship with PsychoPy since its not up on pip yet.

peircej:
This pull request introduces 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-24bd37c12567db3ba49153109b56ba23c575aea3)

**new alerts:**

* 1 for Raising a tuple

---

peircej:
OK, let me know when you think you'd like it pulled in. My feeling is that, as long as it doesn't break any existing code, it's fine to include. I'm not too worried about lgtm warnings (given that we have so many that I won't be fixing!) but it looks like this glfw binding IS available on pip:
https://pypi.python.org/pypi/glfw
Actually there are apparently other Cython and CFFI bindings too (although I'd stick with what you've done)! 
https://pypi.python.org/pypi?%3Aaction=search&term=glfw&submit=search

mdcutone:
Yeah, it seem like it's on there now (or I completely missed it in my searches :/). I'll switch over to using the packaged version on PIP. The Cython bindings are too thin to make a difference here. If we are going to use Cython, we could offload more logic to C.

psychopy/event.py
```diff
@@ -78,6 +89,55 @@
     # eventThread = _EventDispatchThread()
     # eventThread.start()
 
+    # GLFW keycodes for special characters
```
peircej: This needs preceding with 
```
if haveGLFW:
```
otherwise people without glfw installed get a `NameError: name 'glfw' is not defined` here

mdcutone:
I made the suggested fix and now we're using the pip version of the GLFW interface. You might be able to merge without breaking the Pyglet side of things. There are still some issues with gamma not returning to it's original and errors when the GLFW window is closed.

peircej:
Yeah, I think this is OK to bring in, to allow people to kick its tyres a little.
I think existing scripts (using pyglet) won't be altered from what I can tell

PR #1707, opened 2018-02-06
Docs and packaging

peircej:
Avoid wheels (or we need to build one on each platform to get the dependencies right)
pip install will happily use a zip so why not do that?

Also got rid of all uses of StrictVersion because we've been bitten again by a package 
releasing a not-so-strict version and breaking PsychoPy!


peircej:
This pull request fixes 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-9cb0fdeb950c05fa218d69c3d849fd8bb0844683)

**fixed alerts:**

* 1 for Variable defined multiple times

---

PR #1710, opened 2018-02-07
BF: Removes Sound.stop() code for short duration (2<secs).

dvbridges:
This fix attempts to stop spectral splatter for short duration tones.


dvbridges:
@peircej although this is not resolving the issues, we at least have sound for short durations e.g., 100ms.

PR #1714, opened 2018-02-11
RFC: move gamma responsibility to each backend

djmannion:
This is a request for comment on proposed changes to gamma handling.

Each backend is responsible for providing ```gamma``` and ```gammaRamp``` setters, which are called from the window.

As part of this change, the code in ```backends/gamma.py``` has been incorporated into ```pygletbackend.py```, since it was all pyglet-specific.

I can't find any way to establish the gamma or gamma ramp active at window opening for each of the backends, so I have removed the code that resets any gamma changes on window closure. Happy to hear other thoughts on how to handle it.


peircej:
I feel like moving gamma to be within pyglet might be a step too far. Although, right now, these calls are only being used by pyglet they are quite general in design. The only thing they need is a screen ID and, potentially, an xDisplay. It seems like other backends could provide this and want access to these functions.

IMO: 

- Window.setGamma() should indeed call backend.setGamma() (or whichever backend is used)
- pyglets Backend.setGamma() shoudl then call gamma.setGamma()

mdcutone:
It would be helpful to have general gamma functions to generate ramps and backend specific code to use them. The GLFW backend has no method to generate a linear ramp but accepts the data the pyglet gamma code yields. GLFW ctypes library does not expose platform specific data (i.e. X11 screen IDs), so requiring that information prevents those functions from being used.

djmannion:
That all makes sense.

How about if ```createLinearRamp``` is changed so that instead of:
```python
createLinearRamp(screenID, rampType=None, xDisplay=None)
```
it is something like:
```python
createLinearRamp(rampType=None, rampSize=256, driver=None)
```

That should allow it to be independent of backend, but each backend would need to identify the LUT size (if it can) and the driver itself.

peircej:
Sounds good guys! Matthew, it looks like glfw does have calls to get the window ID, but I guess these aren't in the ctypes wrapper: http://www.glfw.org/docs/latest/group__native.html

mdcutone:
It appears so. I used native GLFW functions in C and you need some Win32 and X11 API calls to use them.

djmannion:
Revised and submitted as #1732.

PR #1718, opened 2018-02-12
 BF: Fixes hanning error wherein EOS was called too early

dvbridges:


peircej:
You know you extended the width of the hanning window before, to try and get rid of clicks? I wonder if that step can be undone again now that this bug fix has been applied?

PR #1719, opened 2018-02-13
BF: Various GLFW gamma ramp fixes

mdcutone:
Fixed an "access violation" error when getting the gamma ramp of a non-fullscreen window. Modified 'setupGamma' in Window so the original gamma ramp is restored when closing a GLFW window. There are no longer errors when closing a GLFW window. 

Presently, you cannot set the gamma table of a display when using non-fullscreen windows. Requesting a gamma table will return None.

There also seems to be an issue that you must call each open window's close method explicitly when exiting your experiment or you'll get errors. I fixed it by adding a 'atexit' handler to the main window class. This is increases the likelihood a window's close() method is called, even if an error occurs, restoring the original gamma ramp. If `del `is explicitly called on the window before exit, the close function will not be called twice on exit.


peircej:
This pull request introduces 29 alerts and fixes 16 - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-05e8bbe3ef01e2fa64b9c96f827d8d5f29d721f7)

**new alerts:**

* 14 for Explicit export is not defined
* 8 for Variable defined multiple times
* 4 for __iter__ method returns a non-iterator
* 2 for Unreachable code
* 1 for Non-iterable used in for loop

**fixed alerts:**

* 4 for Missing call to __init__ during object initialization
* 2 for Wrong name for an argument in a call
* 2 for Mismatch between signature and use of an overridden method
* 2 for Inconsistent equality and hashing
* 1 for __eq__ not overridden when adding attributes
* 1 for Multiple calls to __init__ during object initialization
* 1 for Missing call to __del__ during object destruction
* 1 for Iterable can be either a string or a sequence
* 1 for Module-level cyclic import
* 1 for __init__ method calls overridden method

---

peircej:
This pull request introduces 29 alerts and fixes 16 - [view on lgtm.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-4b2b230997c984eeaa2868304f7944a05312d7a6)

**new alerts:**

* 14 for Explicit export is not defined
* 8 for Variable defined multiple times
* 4 for __iter__ method returns a non-iterator
* 2 for Unreachable code
* 1 for Non-iterable used in for loop

**fixed alerts:**

* 4 for Missing call to __init__ during object initialization
* 2 for Wrong name for an argument in a call
* 2 for Mismatch between signature and use of an overridden method
* 2 for Inconsistent equality and hashing
* 1 for __eq__ not overridden when adding attributes
* 1 for Multiple calls to __init__ during object initialization
* 1 for Missing call to __del__ during object destruction
* 1 for Iterable can be either a string or a sequence
* 1 for Module-level cyclic import
* 1 for __init__ method calls overridden method

---

PR #1733, opened 2018-02-16
EN: Misc GLFW window features and docs

mdcutone:
Added code to set the window position and swap interval. Added some documentation for how to use swapInterval correctly for good timing across multiple monitors.


mdcutone:
Small issue, right now you need to use 'swapInterval' (either 0 or 1) to enable vertical sync. Using waitBlanking does nothing but enable the timing code/buffer lock code. So one needs to make a window with both waitBlanking and swapInterval to get the expected behavior.

PR #1737, opened 2018-02-19
NF: Added CIEL*a*b* color space conversion.

mdcutone:
New function to convert CIEL*a*b* to RGB values. By default the conversion is between CIELAB and sRGB (BT. 709). You can specify your own conversion matrices and white point if needed (CIE XYZ 1931 working space). Any reasonable array (Nx3, NxNx3) can be converted. Output is in PsychoPy signed RGB values. 

Someone once requested this feature on the message boards, https://discourse.psychopy.org/t/rgb-to-cielab-conversion/3662/6

### DIFF
--- a/psychopy/tools/colorspacetools.py
+++ b/psychopy/tools/colorspacetools.py
@@ -15,6 +15,104 @@ import numpy
 from psychopy import logging
 from psychopy.tools.coordinatetools import sph2cart
 
+def cielab2rgb(lab, whiteXYZ=None, conversionMatrix=None, clip=False):
+    """Transform CIEL*a*b* (1976) color space coordinates to RGB tristimulus
+    values.
+
+    CIEL*a*b* are first transformed into CIE XYZ (1931) color space, then the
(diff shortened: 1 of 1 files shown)


mdcutone:
I have loads of these colour space conversion functions in a C library I wrote. They all convert to and from CIE-XYZ. It might be a good idea to keep them that way instead of going directly to sRGB. Here is a list of those functions people may want ported to PsychoPy.

- xyY_to_XYZ
- xy_to_XYZ
- XYZ_to_Lab (CIELAB)
- Lab_to_XYZ
- XYZ_to_Luv (no forward transform yet)
- color_temp_ptb  (blackbody colour using PTB's locus approximation)
- color_temp_kang (another backbody color approximation based on another method I forgot about)
- XYZ_to_rec709 (sRGB)
- XYZ_to_bt2020 (HDTV)
- Lab_to_rec709 (sRGB)

There are a also couple of esoteric colour spaces too but I doubt anyone will need those.

PR #1743, opened 2018-02-22
NF: Adds a new component called 'Variable' to Builder.

dvbridges:
The variable component allows users to define variables in builder. A variable can be defined and updated at the beginning of the experiment, routine, or on every frame. Values are stored in a dict, and if requested the values are written on every trial using trial handler. The value of the variable can be saved as either a constant, a variable from every routine, as a list of every value from each frame, or the final routine or frame value for each trial.


dvbridges:
No problem, I will work on the documentation in a separate pull request.

PR #1744, opened 2018-02-23
BF: Add importing pyglet.media.sources for pyglet 1.3

uozias:
I added importing pyglet.media.sources for pyglet 1.3.

"Pyglet.media import avbin" causes import error,
 because pyglet 1.3 has pyglet.media.sources.avbin instead of pyglet.media.avbin in pyglet 1.2 and 1.1.

The source of pyglet.media.sources.avbin of pyglet 1.3 
https://bitbucket.org/pyglet/pyglet/src/ab98c6b7478422acbf9b893f01c0e7778bc1ad9d/pyglet/media/sources/avbin.py?at=pyglet-1.3-maintenance&fileviewer=file-view-default

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -55,6 +55,17 @@ if sys.platform == 'win32':
         # AttributeError if using avbin5 from pyglet 1.2?
         haveAvbin = False
 
+    # for pyglet 1.3
+    if not haveAvbin:
+        try:
+            from pyglet.media.sources import avbin
+            haveAvbin = True
+        except ImportError:
+            haveAvbin = False
+        except AttributeError:
+            haveAvbin = False
+        except Exception:
+            haveAvbin = False
 
 import psychopy  # so we can get the __path__
 from psychopy import core, platform_specific, logging, prefs, monitors


hoechenberger:
Wouldn't make it sense to check for the pyglet version and `import` accordingly, sth. like (untested):

```python
from pkg_resources import parse_version
...

     try:
        if parse_version(pyglet.__version__) >= parse_version('1.3'):
             from pyglet.media.sources import avbin
        else:
            from pyglet.media import avbin

         haveAvbin = True
    ....
```
This would avoid most of the redundancy your current PR introduces.

uozias:
I tested your suggestion, then I noticed my misunderstanding.
Location of teh avbin module can not be distinguished by version,
because in my environment, pyglet.version is 1.2.2, but avbin is in pyglet.media.sources.
I am confused. 
This is different from current state of  pyglet's 1.2-maintenance branch.

I think my code works fine in this confusing situation, 
but I also agree with you about redundancy.
I respect your decision.

peircej:
I also find the pair of these exception testing blocks a bit strange in that there are 3 tests that result in the same result, whatever the exception is (AttributeError, ImportError and Exception)

PR #1751, opened 2018-02-27
BF: Resets ratingscale label color if using hover marker 

dvbridges:
Rating Scale leaves the text label selected in the following trials when using hover as a marker style. This fix resets label text to original text color in the reset function, thus resetting label color on every trial.


peircej:
The key commit in this pull request ( 11f0e6f ) is now included (as 
4c16bf6 ) in PR #1751

The other commits I don't want to include for now, with a release imminent. They'll be pulled in as part of the PR #1748 once 1.90.0 is released. Closing

PR #1752, opened 2018-02-28
BF: was passing function rather than returned value

djmannion:
```getResponse``` in ```serialdevice.py``` was passing the function rather than the function's result, if requested to get all the buffer contents.


peircej:
Doh! I wonder how this hasn't broken all sorts of things in the past! :-/
Anywa, thanks for catching!

PR #1765, opened 2018-03-08
BF: Save recordFrameIntervals in StaticPeriod before disabling.

cgc:
When a `StaticPeriod` starts, the window's `recordFrameIntervals` is set to `False`. After completion, the previous value of `recordFrameIntervals` is restored. However, the order of statements in `StaticPeriod.start()` means that the initial value of `recordFrameIntervals` is always set to `False`, and _then_ saved. By saving before disabling, this initial value will be successfully restored in the future.


peircej:
Ha! Good catch, and thanks for following all the guidelines for a perfectly-constructed pull request :-)

PR #1770, opened 2018-03-14
BF: Fixes ioHub compatiblity issues with Python 3.

dvbridges:
The fixes are mainly specific to keyboard devices.

### DIFF
--- a/psychopy/iohub/client/__init__.py
+++ b/psychopy/iohub/client/__init__.py
@@ -29,6 +29,7 @@ from ..devices import DeviceEvent, import_device
 from ..devices.computer import Computer
 from ..devices.experiment import MessageEvent, LogEvent
 from ..constants import DeviceConstants, EventConstants
+from psychopy import constants
 
 getTime = Computer.getTime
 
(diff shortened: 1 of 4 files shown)


psychopy/iohub/client/__init__.py
```diff
@@ -610,7 +611,7 @@ def createTrialHandlerRecordTable(self, trials, cv_order=None):
 
         for cond_name in self._cv_order:
             cond_val = trial[cond_name]
-            if isinstance(cond_val, basestring):
```
peircej: If we switch this line then unicode strings passed from Py2 will raise errors. I *think* you can leave it as basestring but then do thisi import at top: `from past.builtins import basestring`
see http://python-future.org/compatible_idioms.html

psychopy/iohub/client/keyboard.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, ioe_array):
 
     @property
     def key(self):
-        return self._key
+        return str(self._key, 'utf-8')
```
peircej: This will raise an error on Py2.7 (because can't convert to a bytes string with unicode encoding)
You could do:
```
from past.builtins import unicode
```
then
```
return unicode(self._key, 'utf-8')
```

PR #1775, opened 2018-03-21
BF: iohub python 3 compatibility

ariekahn:
A few minor issues cropping up with iohub on Python 3:

1) trailing commas were causing a few of the keycodes to be represented as tuples, which was ignored in python 2 but in 3 gives:
```
TypeError: '>' not supported between instances of 'tuple' and 'int'
```

2) darwinkey needed to be a relative import

3) locals() returns a view in python 3, which means we can't change it when iterating over it. Explicitly creating a copy should preserve the old behavior.

### DIFF
--- a/psychopy/iohub/constants.py
+++ b/psychopy/iohub/constants.py
@@ -761,9 +761,9 @@ elif sys.platform == 'darwin':
         F4 = 131  # Keycode on Apple wireless kb
         VK_ISO_SECTION = 0x0A
         VK_JIS_YEN = 0x5D
-        VK_JIS_UNDERSCORE = 0x5E,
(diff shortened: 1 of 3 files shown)


peircej:
Great. Well done. I think we're going to find quite a few of these issues, especially in iohub, so it's good to have someone jumping in there on it. I'll be trying to push out a few quick bug-fix releases.

Oh, one thing though, could you annotate your commits in future with BF: for bug-fixes FF: for fixes to not-yet-released features and ENH for new features/enhancements. The labels in the oull request doesn't matter but the commit message is important. When I create a bug fix release I typically do that in a separate git branch and only commits tagged with BF: get included

ariekahn:
My bad! Will annotate any commits themselves in the future.

PR #1789, opened 2018-04-04
ENH: add bidirectionality and Arabic reshaping to the TextStim

m-macaskill:
- Added new parameters `bidirectional` and  `arabicReshape` to the TextStim initialisation.
- These rely on the `bidi` and `arabic_reshaper` packages.
- The first fixes the left-to-right display of what should be right-to-left text. This should be sufficient to display Hebrew properly.
- Arabic (and other languages like Farsi that use much the same script) have an additional need. The characters need to flow into each other, so each character's shape depends on its neighbours. Without applying this re-shaping, characters are instead shown in their 'isolated' form.
- Examples of raw and corrected text have been added to the Coder `textStimuli.py` demo.
- It wasn't feasible to allow these settings to be adjusted after the TextStim has been initialised. Both operations change the text in the text stimulus, which created a circular situation when needing to monitor whether the text had been altered in the `.text()` function. Hence, they are set only once, at TextStim initialisation.
- This limits the usability of these features in Builder. Until the Builder text component dialog has been modified to allow these new parameters to be set graphically, to access them, we need to use TextStims created in code rather than via a Text component dialog.


m-macaskill:
Failing tests as Travis doesn't have the arabic_reshaper library installed, so that new import doesn't work.

m-macaskill:
Thanks Richard, had no idea how that worked. Have fumbled around amending the commit to include those requirements and the tests seem to be running again automatically. Fingers crossed.

m-macaskill:
I'll put this on hold for a moment and restructure it to just have just one parameter (e.g. `languageStyle`) which can take one of several values (e.g. `regular`, `bidirectional`, `Arabic`). `regular` would take no action, `bidirectional` would allow RTL support, with bidirectional content; and `Arabic` would apply the bidirectional algorithm as well as Arabic reshaping. I'll also try to get Builder Text component support at the same time.

m-macaskill:
Closing this in favour of the updated PR #1843

Issue #1809, opened 2018-04-23
Psychopy2_PY3 cannot load inpout32.dll


stfnrpplngr:
Hi,

trying to run my experiment using the `Psychopy2_PY3` version, it fails importing `psychopy.parallel`. It seems like it cannot load the inpout32.dll giving the following error message:

`OSError: [WinError 193] %1 is not a valid Win32 application`

The error occurs running  [`__init__.py`](https://github.com/psychopy/psychopy/blob/master/psychopy/parallel/__init__.py) of the parallel module at lines 38-42:
```
elif sys.platform == 'win32':
    from ctypes import windll
    if hasattr(windll, 'inpout32'):
        from ._inpout32 import PParallelInpOut32
        ParallelPort = PParallelInpOut32
```

As is being described [here](https://discourse.psychopy.org/t/cannot-send-a-signal-via-parallelport-using-psychopy-on-windows7-64bit/969), I saved the `inpout32.dll` in the directory of the executable file.
I tried loading the file with `ctypes.cdll.LoadLibrary()` referring directly to the `inpout32.dll` getting the same error as above.

What do you think?

PS: I found [this comment](https://hynek.me/articles/hasattr/) advising against the use of `hasattr()`.
____
Python 3.6.3 |Anaconda custom (64-bit)| (default, Oct 15 2017, 03:27:45) [MSC v.1900 64 bit (AMD64)]
IPython 6.1.0 -- An enhanced Interactive Python.
Spyder 3.2.8
PsychoPy-1.90.1-py3.6

peircej:
I'm not sure if this is a problem with Python3 per se or with the fact that you're using a 64bit python. To use 64bit python you will need the 64bit version of inpout32. (Using 32bit libs on 64bit platforms is fine, but all the libs need to match so you can't have 64bit python with a 32bit dll).

Could you:

 - try using the Standalone PsychoP3 (which is the 32bit version) instead of Anaconda?
 - try installing a 64bit parallel port driver?

> PS: I found this comment advising against the use of hasattr().

`hasattr` is just easier in this line of code and we know the object that we're getting so it's fine.

stfnrpplngr:
Yes, you are right. I could start the experiment in the standalalone version.
Now, I try to get make it run with the `inpoutx64.dll`.

> hasattr is just easier in this line of code and we know the object that we're getting so it's fine.

The whole if construction does not work with `hasattr` because it raises an OSError.

peircej:
[...]

PR #1812, opened 2018-04-25
NF: Adding support for 64bit driver

stfnrpplngr:
Adding support for 64bit parallel port driver as described in #1809


psychopy/parallel/__init__.py
```diff
@@ -37,16 +37,24 @@
...
+        try:
+            hasattr(windll, 'inpoutx64')
+            from ._inpoutx64 import PParallelInpOutx64
+            ParallelPort = PParallelInpOutx64
+        except OSError:
+            try:
+                hasattr(windll, 'dlportio')
+                from ._dlportio import PParallelDLPortIO
+                ParallelPort = PParallelDLPortIO
+            except OSError:
```
hoechenberger: This is super convoluted now. If it stays that way, we should at least add a comment above all those try/except blocks to explain what's going on.

psychopy/parallel/_inpoutx64.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python
```
hoechenberger: So you're essentially duplication `_inpout32.py` here? If the contents of `PParallelInpOut32 ` and `PParallelInpOutx64` are identical, there should be no need to duplicate them just for the sake of different names.

psychopy/parallel/_inpout.py
```diff
@@ -35,7 +35,12 @@ def __init__(self, address=0x0378):
             self.base = int(address, 16)
         else:
             self.base = address
-        self.port = windll.inpout32
+        try:
+            hasattr(windll, 'inpout32')
```
hoechenberger: Since we now have the `hasattr()` calls both in `_inpout.py` and `parallel.py`, I am wondering whether it could be wise to change the signature of `_inpout.PParallelInpOut.__init__()`, and if `inpout` is `None` (default), we check for the "bitness" of the running Python install and choose `inpout32` or `inpoutx64` accordingly; `inpout` can also be a string (`inpout32` or `inpoutx64`) to explicitly select a driver.

```python
def __init__(self, address=0x0378, inpout=None):
    ...
    if inpout is None:
        # Check for bitness and select driver accordingly
        ...
    elif inpout == 'inpout32':
        self.port = windll.inpout32
    elif inpout == 'inpoutx64':
        self.port = windll. inpoutx64
    else:
        raise ValueError('Invalid driver specified')
    ...
```


In `parallel.py` we could then do:

```python
     try:
        from ._inpout import PParallelInpOut
        ParallelPort = PParallelInpOut
      except OSError:
          try:
              hasattr(windll, 'dlportio')
    ...
```

What do you think?

stfnrpplngr:
@hoechenberger What's your opinion on the changes?

hoechenberger:
Sorry, was on the road, will look at it tonight!

psychopy/parallel/__init__.py
```diff
@@ -31,22 +31,32 @@
...
-    elif hasattr(windll, 'dlportio'):
-        from ._dlportio import PParallelDLPortIO
-        ParallelPort = PParallelDLPortIO
-    else:
+    from importlib import import_module
+    for key, val in drivers.items():
+        try:
+            hasattr(windll, key)
+            import_module('.'+val[0], __name__)
+            ParallelPort = getattr(locals()[val[0]], val[1])
```
hoechenberger: For clarity, I would suggest we do sth. like:

```python
       try:
            hasattr(windll, key)
            driver_name, class_name = val

            import_module('.' + driver_name, __name__)
            ParallelPort = getattr(driver_name, class_name)
```

I also removed the reference to `locals()` -- is this really necessary? (Haven't tested the code)

psychopy/parallel/__init__.py
```diff
@@ -31,22 +31,32 @@
...
+# Because hasattr() in connection to windll ends up in an OSError trying to
+# load 32bit drivers in a 64bit environment, this passage is written as an
+# inconvenient try-except-construction.
 if sys.platform.startswith('linux'):
     from ._linux import PParallelLinux
     ParallelPort = PParallelLinux
 elif sys.platform == 'win32':
+    drivers = dict(inpout32=['_inpout', 'PParallelInpOut'],
+                   inpoutx64=['_inpout', 'PParallelInpOut'],
+                   dlportio=['_dlportio', 'PParallelDLPortIO'])
```
hoechenberger: I wonder whether we should use a dict of tuples instead of a dict of lists, since the values ought to be immutable?

psychopy/parallel/__init__.py
stfnrpplngr: First argument for `getattr()` is the object, and the second is only the name of an object. Therefore, `locals()` is necessary. Or is there another way of calling the object using its name?

psychopy/parallel/__init__.py
stfnrpplngr: Yes, good idea.

psychopy/parallel/__init__.py
```diff
@@ -31,22 +31,32 @@
...
+    for key, val in drivers.items():
+        try:
+            hasattr(windll, key)
+            import_module('.'+val[0], __name__)
+            ParallelPort = getattr(locals()[val[0]], val[1])
+            break
+        except OSError:
+            ParallelPort = None
+            continue
+    if ParallelPort == None:
```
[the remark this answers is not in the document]
stfnrpplngr: No problem, changed.

psychopy/parallel/__init__.py
```diff
@@ -39,21 +39,22 @@
...
+        driver_name, class_name = val
         try:
             hasattr(windll, key)
-            import_module('.'+val[0], __name__)
-            ParallelPort = getattr(locals()[val[0]], val[1])
+            import_module('.'+driver_name, __name__)
+            ParallelPort = getattr(eval(driver_name), class_name)
             break
-        except OSError:
+        except (OSError, KeyError, NameError):
```
stfnrpplngr: I am not proud of simply including more exceptions, but now it works with Python 2.7

hoechenberger: Great!

psychopy/parallel/__init__.py
```diff
@@ -39,21 +39,22 @@
...
             ParallelPort = None
             continue
-    if ParallelPort == None:
+    if ParallelPort is None:
```
hoechenberger: Nice, thanks.

psychopy/parallel/__init__.py
```diff
@@ -39,21 +39,22 @@
...
     from ctypes import windll
     from importlib import import_module
     for key, val in drivers.items():
+        driver_name, class_name = val
         try:
             hasattr(windll, key)
-            import_module('.'+val[0], __name__)
-            ParallelPort = getattr(locals()[val[0]], val[1])
+            import_module('.'+driver_name, __name__)
+            ParallelPort = getattr(eval(driver_name), class_name)
```
hoechenberger: Just checked the `importlib` docs; `import_module()` returns the loaded module, so we should be able to do:
```python
ParallelPort = getattr(import_module('.' + driver_name, __name__),
                       class_name)
```
if I'm not mistaken? WDYT?

psychopy/parallel/__init__.py
```diff
@@ -39,21 +39,22 @@
     from ._linux import PParallelLinux
     ParallelPort = PParallelLinux
 elif sys.platform == 'win32':
-    drivers = dict(inpout32=['_inpout', 'PParallelInpOut'],
-                   inpoutx64=['_inpout', 'PParallelInpOut'],
-                   dlportio=['_dlportio', 'PParallelDLPortIO'])
+    drivers = dict(inpout32=('_inpout', 'PParallelInpOut'),
+                   inpoutx64=('_inpout', 'PParallelInpOut'),
+                   dlportio=('_dlportio', 'PParallelDLPortIO'))
```
hoechenberger: Good, thanks.

stfnrpplngr:
Thank you for the suggestion. It works like a charm. 
I think it is done now, or do you have anything else in mind?

hoechenberger:
LGTM! Good job! And thanks for your patience and persistence ;)

peircej:
Well done guys!

PR #1822, opened 2018-05-15
fixed color specification of shapes using hex

dibgerge:
When specifying the color of shapes (circle/rectangle) using hex such as '#000000', this is converted to RGB tuple correctly using the `colors.hex2rgb()` function, but in the helper code, 1 is subtracted from the values, giving the wrong colors...


peircej:
Ah, this was caused by a mistake in #1696 
Your fix (combined with that commit) is actually going to break hex colors in stimuli instead of the colors in the window. I'll follow this up with a commit to fix both (and tests to go with them)

Issue #1821, opened 2018-05-10
Issues with PsychoPy 1.90.1 (Python 2 edition - Windows 10)


frankgasking:
Hi there,

We recently tested out 1.90.1 (Python 2 edition) with a number of our scripts, and found a number of issues.  

This was using a Dell Optiplex 7050 (http://www.dell.com/en-uk/work/shop/desktop-and-all-in-one-pcs/optiplex-7050-tower-small-form-factor/spd/optiplex-7050-desktop) with Windows 10 installation.

Here are the issues spotted:

- [x] **1)** Hex colour encoding not working correctly.   Seems to create a speckled effect when used for a background colour.   My colleague has reported also at https://discourse.psychopy.org/t/hexadecimal-color-strings-not-working/4471

- [x] **2)** There is a "Unknown stereo type -1" error when using Sound class in Coder and using the "Sounddevice" codebase.   Workaround is to have to set "stereo=True" on a sound() call, though documentation doesn't seem to mention that you have to include this.

- [x] **3)** On the initial run, PsychoPy tries to download ffmpeg.  This fails.   When manually trying to kick off again, you see an error reporting why:
`"Error while fetching file: <urlopen error [SSL:CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c.590)>"` Internet connection is confirmed as fine, and the same issue occured on another test machine.

- [x] **4)** When using a sound component, when pulling duration from an Excel file,  it causes the following error: `"Could not convert string to float: VARIABLE NAME"` Seems like the autoconversion from Excel is broken, specficially at \psychopy\experiments\components\sound\__init__.py line 65 in writeInitCode.

- [ ] **5)** If playing audio files with different encodings,  it fails when using "sounddevice" with a `"psychopy.exceptions.SoundFormatError: Tried to create audio stream 48000_1_128 but 44100_1_128 already exists and win32 doesn't support multiple portaudio streams"` error

- [x] **6)** PYO sound was broken on the Dell Optiplex 7050 (at least).   Segmentation fault on desktop PC,  "Server must be booted" error. A sound driver update initially worked, but then the issue would still occur on other scripts - due to initialization of the portaudio server I think?

- [x] **7)** Moviestim2 demo does not work.  Complains about VLC module not being present, even though VLC has been installed/reinstalled.  This actually affects using OpenCV though as a movie playback type.

[...]

Issue #1820, opened 2018-05-08
Psychopy with sounddevice backend can't play sounds that sounddevice can


mwaskom:
I have some experiment code that plays some sounds. I have been using the pygame backend, but am aiming to move over to sounddevice.

I've found that some of my .wav files could be played by PsychoPy with the pygame backend but cannot be played by PsychoPy with the sounddevice backend. Surprisingly, I have no problem paying these files directly using soundfile and sounddevice.

Here's an example small wav file that I have trouble with: [secalert.zip](https://github.com/psychopy/psychopy/files/1985852/secalert.wav.zip)

(I compressed it so that github will let me upload it; the underlying file is only 12Kb).

This won't play using a simple PsychoPy example

```python
from psychopy import sound
sound.Sound("secalert.wav").play()
```

Sometimes an exception makes its way to stderr (but is not raised in the Python interpreter):

```python-traceback
Traceback (most recent call last):
  File "/Users/mwaskom/anaconda/envs/psychopy/lib/python2.7/site-packages/sounddevice.py", line 733, in callback_ptr
    return _wrap_callback(callback, data, frames, time, status)
  File "/Users/mwaskom/anaconda/envs/psychopy/lib/python2.7/site-packages/sounddevice.py", line 2508, in _wrap_callback
    callback(*args)
  File "/Users/mwaskom/anaconda/envs/psychopy/lib/python2.7/site-packages/psychopy/sound/backend_sounddevice.py", line 186, in callback
    dat = thisSound._nextBlock()  # fetch the next block of data
  File "/Users/mwaskom/anaconda/envs/psychopy/lib/python2.7/site-packages/psychopy/sound/backend_sounddevice.py", line 489, in _nextBlock
    .format(self.stereo))
```

But I have no problem playing the file using the underlying libraries without any bells or even whistles:

```python
import soundfile
import sounddevice
array, bitrate = soundfile.read("secalert.wav")
sounddevice.play(array, bitrate)
```

On the small sample of files I have used in my experiments, it seems like stereo files work fine but mono ones break.

If I force loading in mono (`sound.Sound("secalert.wav", stereo=False)`), then I get a distorted version of the sound, and *all other* sounds from my computer (including from other apps) are distorted until I exit the Python session.

Versions used:
- psychopy: 1.90.1 (also present on older versions)
- sounddevice: 0.3.10
- soundfile: 0.9.0

mwaskom:
[...]

PR #1824, opened 2018-05-16
FIX: souddevice stereo detection

CRiddler:
Code for determining a sound array was either mono or stereo in the sounddevice backend wasn't being run due to being indented too far.

See point 2 in #1821 and #1820


peircej:
This is interesting. I'm a bit concerned there's a deeper problem, given the presence of `self.channels` and `self.stereo`, which should be providing identical information for the class :-/

But if your fix works then that's a good start!

Please use BF or FF to indicate BugFixes longstanding bugs that need porting to release branches to be included in minor releases) or FeatureFixes (fixes to new features, that should NOT be cherry-picked because the feature won't exist)

thanks!

peircej:
Yes, I think this might be wise, but I'll include the band-aid in release 1.90.2

PR #1833, opened 2018-05-23
New Slider to simplify and extend RatingScale

peircej:
supports
- vertical and horizontal (using just `size` setting)
- categorical, continuous (VAS) and Likert (using just `labels` and `ticks` settings)
- live-update of other objects
- styles and simple code-customization

So far there is no documentation page, no test suite entry for this and only a Builder demo (not Coder)


peircej:
@hoechenberger @jeremygray I've got a working scale now. I've called it Slider (because we can change the look very easily to  be a slider or more like Jeremy's scale). It doesn't have an "Accept" button, or a Description, or built-in keyboard controls - users will need to control those themselves, but I think it supports most of the other features of RatinsScale.

PR #1845, opened 2018-05-31
BF: Fixes Py3 compatibility problem with genComponsTemplate.

### LINKED ISSUE #1844: Python 3 incompatibility in genComponsTemplate.py
Running 
    python psychopy/psychopy/tests/test_app/test_buildergenComponsTemplate.py --out

to generate a new `componsTemplate.txt` file for automated Builder tests fails when using Python 3.6, with this error:

    File "genComponsTemplate.py", line 82, in <module>
        outfile.write(order.encode('utf8')+'\n')
    TypeError: can't concat str to bytes

It runs fine under Python 2.7

dvbridges:
In Python 3, the encode function returns type bytes and the write function requires str not bytes. In Python 2 the encode function returns type str, thus no error is raised. Removing the encode function resolves issue. Fixes #1844


peircej:
@dvbridges the test problem looks like it was indeed an issue with neurodebian. I still think it's worth hadnling the `next` attribute better but this looks like it does at least pass

PR #1846, opened 2018-06-01
installation with plain ana/conda packages

rphlypo:
this pull requests documents on how to use plain ana/conda and pip install for `psychopy` no longer using Erik's binstar channel (which today seems to have some flaws not allowing installation to end properly). This is possible since `wxpython` and other dependencies can now readily be installed with `conda install` not using any additional channels.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -120,12 +120,12 @@ Anaconda and Miniconda
 
 The following should allow you to get PsychoPy working using Ana/MiniConda::
 
-  conda config --add channels https://conda.binstar.org/erik
-  conda install -c erik psychopy
-  conda create -n psychopyenv psychopy
-  source activate psychopyenv
-
-but the recipe may be out of date and `pygame` was not available in the past (now?)
+  conda create -n psypy
+  conda activate psypy
+  conda install python=2.7
+  conda install numpy scipy matplotlib pandas pyopengl pyglet pillow moviepy lxml openpyxl xlrd configobj pyyaml gevent greenlet msgpack-python psutil tables requests[security] pyosf cffi pysoundcard pysoundfile seaborn psychopy_ext python-bidi
+  conda install wxpython
+  pip install psychopy
 
 .. _macports_install:
 


kastman:
Hi @rphlypo - Thanks for updating this! 

We're actually in the process of finishing up a real conda-forge package over at [conda-forge #5739](https://github.com/conda-forge/staged-recipes/pull/5739), but this is a good stop gap and probably worth updating. Specifically, we're missing `pysoundfile` / `libsndfile` for windows, are having trouble with the CI for `wxpython` and are having trouble with the headless CI for `pyglet`. Would you be interested in helping us with any of those?

As an alternative to the package list you suggest, I wonder if it might be slightly better to distribute a frozen, version-locked enviornment.yaml with instructions on how to re-create that environment (based on OS), i.e. something like the instructions here: https://docs.qiime2.org/2018.4/install/native/#install-qiime-2-within-a-conda-environment

That would help us unify the versions everyone is using and possibly head off odd bugs that would arise between versions?

kastman:
For the record, I'm absolutely onboard with removing the `erik` binstar channel. I haven't updated it in literally years, and it's probably just adding confusion.

m-macaskill:
Definitely support updating this documentation, but would suggest orienting it towards working under Python 3 rather than 2.7. Eventually Python 2 support by PsychoPy will be deprecated and we should be encouraging people to use Python 3 for their custom installations.

peircej:
OK, I'm pulling it in for now but, yes, Py3 is a good way to go at this point

rphlypo:
Looking into the Py3 install under miniconda, but struggling with the dependency `pysoundcard`. Will keep you posted.

rphlypo:
@kastman my knowledge today on the CI for `wxPython` or `pyglet` does not allow me to contribute much to these. Also not running Windows, so will pass on the Windows dependencies **for now**.

kastman:
@rphlypo Totally understand - it's a pain and I don't fully understand it myself. I'll keep you in the loop as we learn more in case you're interested, but no pressure. Thanks!

PR #1847, opened 2018-06-02
Don't try to be clever about filling the output buffer with zeros

mgeier:
This fixes the issue described at https://discourse.psychopy.org/t/sounddevice-issue-with-stream-started-at-beginning/

### DIFF
--- a/psychopy/sound/backend_sounddevice.py
+++ b/psychopy/sound/backend_sounddevice.py
@@ -181,7 +181,7 @@ class _SoundStream(object):
                 (time.time() - self._tSoundRequestPlay) * 1000))
         t0 = time.time()
         self.frameN += 1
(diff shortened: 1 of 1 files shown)


peircej:
I didn't know about `ndarray.fill()` and it should be more efficient than multiplying by zero, but I'm not clear why it would cause any functional difference to the result. Any idea why that would fix this bug?

mgeier:
The output array (sometimes) contains garbage bytes which turn out as NaN or +/- infinity when interpreted as `float32` values. Those values don't become zero when multiplied by zero, they rather become NaNs which then end up being audible (they are probably turned into +/- 1 at some point on their way to the sound card?).

This is also the reason why the message `invalid value encountered in multiply` was appearing.

peircej:
Good to know. Thanks Matthias!

PR #1851, opened 2018-06-08
conda install for python 3.5

rphlypo:
Updated python 2.7 for more explicit splitting between plain `conda`, `conda-forge` and `pypi`. Included the install for python 3.5.
Careful ! Did not test `psychopy`'s full functionality after install.


peircej:
Actually, I *think* we are 3.5+ compatible at the moment but possibly not 3.4

PR #1852, opened 2018-06-12
BF: Builder variables should not be used as the column name in condit…

hsogo:
Builder variables can be used as the column name in conditions file, which can cause annoying problems.  For example, using `filename` in the condition file as following, name of output files are overwritten.

I think that Builder must check whether Builder variables are included in the column names of conditions file.


peircej:
The build appeared to fail only on the badly outdated system python installation. (It's time we shut that one down I think.)

PR #1854, opened 2018-06-14
 BF: Amendments to JP fix to getRegionOfFrame - now working

dvbridges:
Note, getpixel from the returned image object does not work if back buffer is used, or if full screen is used.

### DIFF
--- a/psychopy/tests/test_hardware/test_CRS_bitsShaders.py
+++ b/psychopy/tests/test_hardware/test_CRS_bitsShaders.py
@@ -88,9 +88,9 @@ def test_bitsShaders():
             stim.image = np.resize(intended,[256,256])*2-1 #NB psychopy uses -1:1
 
             stim.draw()
-            #fr = np.array(win._getFrame('back').transpose(Image.ROTATE_270))
+            #fr = np.array(win._getFrame(buffer='back').transpose(Image.ROTATE_270))
             win.flip()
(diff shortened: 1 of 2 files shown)


peircej:
But it looks like you're saying that accepting this change means that we're no longer able to fetch from the back buffer? We do need to be able to do that. We can't just turn off uses of the back-buffer; there are times when it's important to be able to exam pixels before a flip command.

dvbridges:
I dont think I have made any changes that stop you from fetching from the back buffer. I was just saying that if you do try to use the back  buffer, the buffered image does not seem to hold all the components, only the background image.

peircej:
Ah, you're meaning from a Builder experiment? That's fine, yes the issue is that all stimuli that have been rendered (with .draw() ) should show up in the back buffer, but stimuli rendered using `autoDraw=True` (as in a Builder experiment) are only rendered when win.flip() is called. So if you grab the backbuffer before `win.flip()` then yes, visual builder components won't show up.  Then again, if you grab it with `buffer='front'` then you're effectively getting the contents of the previous frame, not the current one.

dvbridges:
Great thanks. So, should we draw to back buffer if _getFrame is requested with ```buffer='back'```?

peircej:
> should we draw to back buffer if _getFrame is requested with buffer='back'?

I don't follow. Can chat tomorrow though

dvbridges:
Ok, as this request is ok for now, would you mind merging with master?

PR #1855, opened 2018-06-15
FF: Builder cannot compile script

hsogo:
Two issues are fixed.

1) Builder cannot generate script in Python2 if non-ASCII characters are included in path.  This was fixed by encoding path string into file system coding in Python2.

2) Builder uses Python2 to generate script even when Builder is running on Python3 in Ubuntu (16.04 LTS).  This is because compile is done by "**python** -m psychopy.scripts.psyexpCompile ...".  In Ubuntu commad line should be "**python3** -m ...".  My fix is based on `_runFileAsProcess()` in psychopy/app/coder/coder.py.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2237,12 +2237,12 @@ class BuilderFrame(wx.Frame):
         # Compile script from command line using version
         compiler = 'psychopy.scripts.psyexpCompile'
         if not constants.PY3:
-            print(experimentPath, type(experimentPath))
-            experimentPath.decode(sys.getfilesystemencoding())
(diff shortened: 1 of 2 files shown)


hsogo:
Sorry, this issue is not included in 1.90.2. So the label should be FF, not BF.

hsogo:
In Python2, `fullPath` should be encoded into file-system encoding to manage non-ASCII characters. However, because type `fullPath` is unicode and `type(fullPath) == bytes` is `False`, `fullPath` is not encoded.

peircej:
Thanks @hsogo 
Yes, this fixes some changes recently made by @dvbridges for the the 3.0 release so we shouldn't include the bug-fix 1.90.3

PR #1856, opened 2018-06-18
 BF: Fixes TypeError converting 'Param' object to bytes in PY3

dvbridges:


dvbridges:
Hi @peircej , looks like #1854 will pass now. If you pull in that request, I can merge that branch with this branch, fixing the issues which are stopping this branch from passing.

PR #1858, opened 2018-06-18
BF: Allows sound duration defined from cond. file. Fixes 4) from #1821

dvbridges:


dvbridges:
@peircej , would you like this adding to the change log? It has not been included in the current list as the pull request is awaiting approval.

dvbridges:
@peircej , I have included it in the changelog now, as I am guessing you do want it included. Can be changed if required.

PR #1869, opened 2018-07-02
 BF: Fixes bug where default sound not overwritten when filename used.

dvbridges:


dvbridges:
@peircej , this will need pulling in before the next release as it relates to one of the bug fixes listed

PR #1880, opened 2018-07-10
DOC: Update builderTerminateLoops.rst

LJGz:
DOC: To prevent error :

if len(key_resp.keys) <1:
TypeError: Object of type 'NoneType' has no len()

As discussed in https://discourse.psychopy.org/t/error-feedback-has-syntax-error/1509 and https://groups.google.com/forum/#!msg/psychopy-users/jZDTXg8oXgY/8v94pS3YDQAJ


peircej:
👍

PR #1890, opened 2018-07-17
FF: Fixes to mouse component

dvbridges:


psychopy/experiment/components/mouse/__init__.py
```diff
@@ -597,7 +598,7 @@ def writeRoutineEndCodeJS(self, buff):
             # use that set of properties to create set of addData commands
             for property in mouseDataProps:
                 if store == 'every frame' or forceEnd == "never":
-                    code = ("psychoJS.experiment.addData('%s.%s', %s.%s);\n" %
+                    code = ("psychoJS.experiment.addData('%s.%s', %s.%s[0]);\n" %
```
peircej: This _looks_ like you're only storing the first click now, no? But the code section suggests you should be storing every click

psychopy/experiment/components/mouse/__init__.py
```diff
@@ -508,7 +508,7 @@ def writeRoutineEndCode(self, buff):
             # use that set of properties to create set of addData commands
             for property in mouseDataProps:
                 if store == 'every frame' or forceEnd == "never":
-                    code = ("%s.addData('%s.%s', %s.%s)\n" %
+                    code = ("%s.addData('%s.%s', %s.%s[0])\n" %
```
peircej: This _looks_ like you're only storing the first click now, no? But the code section suggests you should be storing every frame (i.e. many)

psychopy/experiment/components/mouse/__init__.py
dvbridges: Cheers Jon, sorted in most recent push.

PR #1900, opened 2018-07-23
BF: Fixes staticText wrap in online search projects dlg

dvbridges:


psychopy/app/pavlovia_ui/project.py
```diff
@@ -179,11 +179,11 @@ def __init__(self, parent, noTitle=False,
             self.title.SetFont(font)
 
         # if we've synced before we should know the local location
-        self.localFolderCtrl = wx.StaticText(
+        self.localFolderCtrl = AutoWrapStaticText(
             parent=self, id=wx.ID_ANY,
             label="Local root: ")
         self.browseLocalBtn = wx.Button(parent=self, id=wx.ID_ANY,
-                                        label="Browse...")
+                                        label="Browse...", size=(88,25))
```
peircej: maybe we could set this to be a minSize setting rather than size (so it can be stretched bigger with the dialog but can't go smaller?)

PR #1906, opened 2018-07-24
NF: Oculus Rift Support

mdcutone:
This is the initial commit for Oculus Rift support for PsychoPy. See `demos/coder/hardware/Rift*` for basic examples on how to use the library. Requires PsychXR (https://github.com/mdcutone/psychxr) for hardware access and currently only supports Windows. PsychXR is limited to 64-bit Python 3.6, I'll build packages for 3.7 soon. For 32-bit support, people will need to build PsychXr themselves using the guide on its project page. 

@peircej Is there any objection to where the modules are placed in the PsychoPy file structure? Furthermore, are there issues around lazy loading the library with 'visual'? I added code which keeps the module from loading if PsychXR is not installed.

### DIFF
--- a/psychopy/demos/coder/hardware/RiftHeadTrackingExample.py
+++ b/psychopy/demos/coder/hardware/RiftHeadTrackingExample.py
@@ -1,0 +1,84 @@
+# Oculus Rift head-mounted display example for rendering 3D with head tracking.
+# Press the 'q' key or use the application GUI to exit. Press 'r' to recenter
+# the HMD's view. Requires PsychXR to be installed.
+#
+# This file is public domain.
+#
+from __future__ import division
+
+from psychopy import visual, event, core
+import pyglet.gl as GL
+import math
+
+# Create a VR session, treat the returned object just like a regular window.
+# Change headLocked to True to disable head tracking, increase the number of
+# samples for anti-aliasing, could be 2, 4, 6, 8, 16 or 32 depending on your
+# hardware.
+hmd = visual.Rift(headLocked=False, samples=1)
+
+# expose the math libraries
+vrmath = visual.rift.math
+
+# loop until the user quits the app through the GUI menu
+stopApp = False
+while not stopApp:
+    for i in ('left', 'right'):
+        hmd.setBuffer(i)  # select the eye buffer to draw to
+
+        # Setup the viewing parameters for the current buffer, this needs to be
+        # called every time the buffer changes.
+        #
+        # Use setRiftView to setup the projection and view matrices
+        # automatically from data provided by the API. Take note of which eye
+        # buffer is active when rendering.
+        #
+        hmd.setRiftView()
+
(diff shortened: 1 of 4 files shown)


peircej:
This pull request **introduces 1 alert** when merging 793cf6d18d90bbcb8e8d83ce0411ea87d861af6a into 7ca7f6062429b24c027176dda9c73d0175a203f0 - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-9f782b4991d0386752c25c2f75a43c54283d5eea)

**new alerts:**

* 1 for Syntax error

---

peircej:
Looks fantastic Matthew! Well done!
I'll be including this in the new release, although for now I haven't got time to create the 64bit-win standalone package so it will require people to install using pip to their own Python setups. 

Your use of lazy import and putting rift.py inside visual are all correct.

thanks so much!

mdcutone:
How do I generate *.rst files for documentation?

peircej:
How do you generate them? They're just text files. Maybe you mean how to generate/test the html from it? This all uses sphinx (and a few sphinx extensions) and the final compile/upload is currently handled by me.

The simplest example is like polygon.rst which simply fetches the docstrings from the python code and auto-generates the documentation file. 
Something like moviestim.rst shows you how to customize what is generated (e.g. nice summary table of attributes at the top). 
Oh, and to get it included into the main docs, take a look at the visual.rst which has the list of objects that it reads from (otherwise there won't be links/reference to your new page).
The sphinx documentation will tell you lots more.

peircej:
Update: I've added stubs for your documentation in this commit to show you how it could work (and I've added Rift into the table of contents as well as the other projector classes like warped projections). I've already done a build and pushed online so you can see how it looks right now with these stubs in place.

One note is that you'll see in the attributes (at the top) that some attributes provide a link (e.g. Rift.firmwareVersion) but not others (Rift.fovType). I've just done this deliberately to show you that. Python property objects have docstrings and therefore they show up this way (one of the cools things about them!) whereas simple attributes don't.

Also, right now inherited-members are being shown but maybe they don't make sense for the Rift (e.g. getMovieFrame() and saveMovieFrames())?

peircej:
Oh, and i had to tweak *args in Rift.__init__(). In Py2 that has to be after the actual named args and I think Py3 is ambivalent, so I've moved it to the Py2-safe version so I could compile the docs

mdcutone:
The order I had that forces those arguments to be keyword only. I suppose Python 2.7 would break in that case. Thanks for clarifying the documentation, I've had issues with that in the past.

mdcutone:
I noticed that the current config is not parsing Numpy style doc strings. I think the fix would be to add `sphinxcontrib.napoleon` to `extensions` in `conf.py` and include `napoleon_numpy_docstring = True` some other place in that file.

peircej:
sure thing

PR #1917, opened 2018-07-29
add self._move.close()to MovieStim3._uload,

tjwilli58:
add self._mov.close() to MovieStim3._uload, move self._unload() in MovieStim3.stop after logging code.

Calling self._mov.close() seems to prevent invalid handle error in call to
ffmpeg_reader in moviepy module.

See
https://stackoverflow.com/questions/43966523/getting-oserror-winerror-6-the-handle-is-invalid-in-videofileclip-function

### DIFF
--- a/psychopy/visual/movie3.py
+++ b/psychopy/visual/movie3.py
@@ -257,11 +257,11 @@ class MovieStim3(BaseVisualStim, ContainerMixin):
         """
         if self.status != STOPPED:
             self.status = STOPPED
-            self._unload()
             self.reset()
             if log and self.autoLog:
                 self.win.logOnFlip("Set %s stopped" % (self.name),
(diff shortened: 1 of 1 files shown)


peircej:
Thanks. For future commit messages (the pull request text doen'st matter) could you try and make sure you follow the information here regarding BF, FF... 
In your case the commit should have been BF: to indicate that this commit would need copying to other (release) branches as a fixed bug. Thanks.

PS I don't *think* the reordering aspects are probably important but well done for tracking down the `_move.close()` call :-)

tjwilli58:
Sorry about that.  I was intending to, but after reading through the github tutorial,  creating a fork, and getting to the point where I actually could make a pull request,  I simply forgot. (My first PR)

peircej:
No worries. Thanks for the PR!

PR #1928, opened 2018-08-07
BF: Fixes file export on run when exp new and not saved

dvbridges:


psychopy/app/builder/builder.py
```diff
@@ -1592,7 +1592,8 @@ def fileSave(self, event=None, filename=None):
             self.fileHistory.AddFileToHistory(filename)
         self.setIsModified(False)
         # if export on save then we should have an html file to update
-        if self.htmlPath or self.exp.settings.params['exportHTML'] == 'on Save':
+        if self.exp.settings.params['exportHTML'] == 'on Save' and os.path.split(filename)[0]:
```
peircej: Is this right with a single capital "on Save"? Would it be wise to do something like
```
exportHtml = str(self.exp.settings.params['exportHTML']).lower()
if exportHTML == 'on save'........
```

dvbridges: Yes this matches the list of allowedVals from the exportHTML parameter, its an example of my habitual camel case use. I think your suggestion is an improvement, so will change 👍

peircej:
Sorry I'd forgotten about the isModified attribute before! That would have saved some time! ;-)

dvbridges:
No worries, and I much prefer this fix

PR #1941, opened 2018-08-14
BF: Amendments to JS compiler

dvbridges:


psychopy/experiment/components/mouse/__init__.py
```diff
@@ -149,7 +149,7 @@ def _writeClickableObjectsCodeJS(self, buff):
         code = (
             "// check if the mouse was inside our 'clickable' objects\n"
             "my.gotValidClick = false;\n"
-            "for (const obj of [{clickable}]) {{\n"
+            "for (const obj of {clickable}) {{\n"
             "  if (obj.contains({name})) {{\n"
             "    my.gotValidClick = true;\n"
             "    {name}.clicked_name.push(obj.name);\n"
```
peircej: This is the line that I was saying before should say something like `clicked_{param}` 
Is there a reason that you've diverged from the Python equivalent function?

dvbridges: No, but the fix is on its way.

dvbridges: There is one last fix required, where there is only 1 clickable object, no list is returned and the ```for loop``` in JS requires a list.

PR #1945, opened 2018-08-15
ENH: Allow Slider dragging outside of "valid" area

hoechenberger:
With a slider I commonly expect that, once I have started dragging, I can move the mouse cursor away from the slider while dragging continues (and the slider updates accordingly). This commit adds this feature.


peircej:
That's good point. And what an easy fix! :-)

PR #1949, opened 2018-08-16
ENH: Allow adjustment of Slider's marker size

hoechenberger:
Colleagues suggested that the default marker for `rating` was too big and that a larger marker for `slider` would look nicer. This PR adds the ability to adjust the marker size via a keyword argument.

Also included:

- `size` is now a readonly prop
- small style & doc cleanups, removed references to unused params


hoechenberger:
Thanks for merging. I am now actually wondering whether `markerScalingFactor` is a good name, or if it could simply be `markerSize`. But then users might assume that one could pass an _actual_ size (in window  or stimulus units), which might be confusing?

PR #1950, opened 2018-08-16
RF: fixed incorrect code styling in #1949

peircej:
PEP8 says consistency within a project is the #1 priority :-)

For info, the fact that in tests we do use methods like`test_marker_scaling_factor` is because pytest/nose didn't used to support methods written in camelCase.

### DIFF
--- a/psychopy/tests/test_all_visual/test_slider.py
+++ b/psychopy/tests/test_all_visual/test_slider.py
@@ -49,11 +49,11 @@ class Test_Slider(object):
             s.size = (1.5, 0.5)
 
     def test_marker_scaling_factor(self):
-        marker_scaling_factors = [1, 1.0]
+        markerScalingFactors = [1, 1.0]
 
-        for marker_scaling_factor in marker_scaling_factors:
-            s = Slider(self.win, markerScalingFactor=marker_scaling_factor)
-            assert s.markerScalingFactor == marker_scaling_factor
+        for thisScalingFactor in markerScalingFactors:
+            s = Slider(self.win, markerScalingFactor=thisScalingFactor)
+            assert s.markerScalingFactor == thisScalingFactor
 
     def test_change_marker_scaling_factor(self):
         s = Slider(self.win, markerScalingFactor=1.0)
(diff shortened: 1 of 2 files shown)


hoechenberger:
> Oh, regarding the markerScaleFactor I think I'm OK with it. My major doubt was whether this is part of feature-creep because you could do the following, which is a bit more in-keeping with my preferred approach (show people how to modify directly, rather than giving params for every possible modification)
> ```pytnon
> scale = visual.Slider(...)
> scale.marker.size *= 1.5
> ```
> but if we include at all then that name is OK with me. 

Dang, I didn't even think of that! I actually prefer that one to my approach in this PR… Should we maybe revert it?? And improve the docs instead?

> And maybe the documentation should make it clear what the "default size" is that the scale factor operates on (this is 2 x tick length)

But is that really the case? From looking at the code I first thought that's true, but in this example I have the impression that the diameter of the marker is larger than the sum of 2 ticks. Could it be bc the circle border has a thickness on its own which are currently do not considered in the calculations?

```python
from psychopy.visual.window import Window
from psychopy.visual.slider import Slider

win = Window()
vas = Slider(win,
             ticks=(1, 100),
             labels=('Not at all confident', 'Extremely confident'),
             granularity=1,
             color='white')

while not vas.rating:
    vas.draw()
    win.flip()

print(f'Rating: {vas.rating}, RT: {vas.rt}')
```

peircej:
> Dang, I didn't even think of that! I actually prefer that one to my approach in this PR… Should we maybe revert it?? And improve the docs instead?

Yes, I think it's better. I was willing to include if you had a strong opinion though. Ha, now it turns out I **should** have used a code review instead of pulling in!

> But is that really the case? From looking at the code I first thought that's true, but in this example I have the impression that the diameter of the marker is larger than the sum of 2 ticks. 

You're right and I don't know why without deeper inspection. Also, different styles might change the marker, so it couldn't be relied on anyway. And now we're thinking we don't need that parameter so the point becomes irrelevant. Let's get rid of the whole thing!

peircej:
Wasn't sure what to do about this one either. Have done the merge, and then you can remove the necessary lines at your end. Sorry for the messy way of working. I should discuss first and merge second!

PR #1953, opened 2018-08-16
DOCS: Initial commit of Oculus Rift docs.

mdcutone:
Added directives to document most of the attributes of the Rift class.


peircej:
Live now at http://www.psychopy.org/api/visual/rift.html

PR #1959, opened 2018-08-18
PKG, RF, ENH: Improvements to setup.py

hoechenberger:
- We now make use of environment markers to conditionally install dependencies on specific platforms
- Anaconda special handling reworked
- Metadata classifiers updated


peircej:
The environment markers are neat, but I wonder if that means we need to force a certain version of setuptools at the start and warn people if they don't have sufficient?

Also curious what happens with building wheels in this scenario. Does the wheel respect the rule? In the previous system the wheel was broken because it had the depends of the platform on which it was built, but labelled itself as a general-purpose wheel, so I stopped building wheels and released only as a zip. Potentially with this system we can release as a wheel again (although not sure what advantage except slightly quicker?)

hoechenberger:
> The environment markers are neat, but I wonder if that means we need to force a certain version of setuptools at the start and warn people if they don't have sufficient?

Support for environment markers was added in Setuptools v20.6.8, which was released in May 2016. I don't believe we'll _want_ to support users who feel brave enough to run `python setup.py` on the command line, but don't bother to  keep their environment up to date ;) Nevertheless, I can add suck a check if need be, no problem.

> Also curious what happens with building wheels in this scenario. Does the wheel respect the rule? 

Yes, in fact that was one motivation for me to go that path (actually this was the second motivation; the primary motivation was to switch to the "standard" way of specifying these kinds of dependencies, as this is formalized in a PEP). _But_ I have to admit I haven't tried building and installing those wheels yet!

Wheel support for environment markers [was added](https://github.com/pypa/wheel/issues/181) in Setuptools 36.2.0.

peircej:
Given the packaging battles I've been having recently I'm scared but... [gulp]

PR #1964, opened 2018-08-20
BF: importConditions() should return list of fieldNames

hoechenberger:
`data.utils.importConditions()` returned a tuple of `fieldNames`, but `TrialHandler2` expects a list (so it can `.append()` to it).


peircej:
Agreed. I can't see any downsides

PR #1967, opened 2018-08-20
BF/RF: GLFW textstim and stencil buffer fixes

mdcutone:
Made changes which allow the Pyglet text renderer to be used with GLFW. Seems to work fine after running a bunch of demos. Also fixed a bug where the depth of the stencil buffer was incorrectly set when opening a GLFW window. That made the 'allowStencil' argument produce unexpected results.


mdcutone:
Also fixed a bug where the best settings are not used if the user specifies an unsupported video mode.

Issue #1859, opened 2018-06-18
windows gamma table bias


djmannion:
It looks like windows does not use the full uint16 range when setting the gamma ramp, causing repeated get/set calls to reduce the screen brightness (see [this discourse report](https://discourse.psychopy.org/t/monitor-gamma-or-something-else-brightness-contrast-light-related-decreases-each-time-i-run-my-experiment/4706)).

For example, the following returns ```65280```:
```python

import ctypes
import numpy as np
import psychopy.visual

win = psychopy.visual.Window()

ramp = np.tile(np.linspace(0, 1, 256)[:, None], (1, 3))
win.gammaRamp = ramp

hw_ramp = np.empty((3, 256), dtype=np.uint16)
ctypes.windll.gdi32.GetDeviceGammaRamp(win.backend.screenID, hw_ramp.ctypes)

print(hw_ramp.max())

win.close()
```

I think the problem is the following lines in ```gamma.py```:

```python
newRamp = (255.0 * newRamp).astype(numpy.uint16)
# necessary, according to pyglet post from Martin Spacek
newRamp.byteswap(True)
```
This can be seen by ```np.array(255.0).astype(np.uint16).byteswap(True)``` giving ```65280```.

Could we replace these two lines with:
```python
newRamp = (65535.0 * newRamp).astype(numpy.uint16)
```

This seems to resolve the issue for me, but perhaps there is another issue that it raises by not having the ```byteswap``` in there?

peircej:
I don't know why the byteswap was previously necessary and now not! :-/ I wonder if this is about a) 64 bit windows or b) Win10 or c) pyglet 1.3. Actually I don't see how it could be (3) because we're making calls directly to the windows functions, but either (1) or (2) seem plausible still

djmannion:
It was with Windows 7 for me.

Actually, I think the ```byteswap``` might still be necessary - but there might also need to be one when *getting* the gamma table.

(It sounds like Windows takes the first 8 bits as the LUT index and ignores the second 8 bits. So '1111111100000000' = 65280 would be right).

So rather than:

```python
origramps = numpy.empty((3, rampSize), dtype=numpy.uint16)
# ...
origramps = origramps/65535.0 # rescale to 0:1
```

something like:
```python
origramps = numpy.empty((3, rampSize), dtype=numpy.uint16)
# ...
origramps = origramps.byteswap(True)  # back to 0:255
origramps = origramps / 255.0
```

djmannion:
[...]

PR #1972, opened 2018-08-22
BF: gamma table issues

djmannion:
This relates to issue #1859.

The first problem was that, on windows, the gamma table that was read was not the same as the gamma table that was set. The set operation used a ```byteswap```, whereas the get operation did not.

To expand:

* It seems like windows takes a 16 bit unsigned integer for each entry in the LUT. The first 8 bits are taken as the LUT index and the second 8 bits are ignored/undefined.
* The ```byteswap``` function swaps the first and last bytes, so one can make a LUT between 0 and 255 and then use ```byteswap``` to move that into the first byte rather than the second.
* When reading the LUT, the first 8 bits will where the LUT is defined. That means that we can do a ```byteswap``` to shift them into the second 8 bits and then normalise by dividing by 255. This should remove the get/set discrepancy.

The second problem was that a linear gamma table, defined between 0 and 1, was not producing evenly spaced integers with an 8-bit LUT between 0 and 255. The problem seems to be that the integer conversion was truncating, so a rounding operation was added prior to type conversion.

### DIFF
--- a/psychopy/tests/test_all_visual/test_gamma.py
+++ b/psychopy/tests/test_all_visual/test_gamma.py
@@ -109,5 +109,32 @@ def test_setGammaRamp():
     assert numpy.allclose(desiredRamp, setRamp, atol=1.0 / desiredRamp.shape[1])
 
 
+def test_gammaSetGetMatch():
+    """test that repeatedly getting and setting the gamma table has no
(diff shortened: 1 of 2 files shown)


mdcutone:
Thanks for fixing this!

PR #1970, opened 2018-08-21
 PKG, ENH, RF: Add versioneer

hoechenberger:
As proposed in GH-1571, I've finally found the time to add [`versioneer`](https://github.com/warner/python-versioneer).

No more out-of-sync release versions & tags!

```
$ python -c "import psychopy; print(psychopy.__version__)"
3.0.0b6+78.g13d282c3
$ git tag 3.0.0                                           
$ python -c "import psychopy; print(psychopy.__version__)"
3.0.0
```
Uncommitted changes:
```
$ touch foo; git add foo                                  
$ python -c "import psychopy; print(psychopy.__version__)"
3.0.0+0.g13d282c3.dirty
```
Committed changes:
```
$ git commit -m 'Test'
[test 0e6384a8] Test
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo
$ python -c "import psychopy; print(psychopy.__version__)"
3.0.0+1.g0e6384a8
```

This PR also moves most of the metadata previously found in `setup.py` to `setup.cfg`. This cleans things up, avoids duplications, and was necessary to get rid of the need to `exec()` `__init__.py`, which was making things difficult.

**Currently, Travis doesn't finish successfully, although `setup.py` can build the package, and the tests pass. Any idea what's going on and if that's even related to this PR?**

PR #1973, opened 2018-08-22
PKG: Remove version file

hoechenberger:
The version file is not required anymore since we moved to versioneer
via GH-1970.


peircej:
Ah, that wasn't too bad (maybe more natural actually):
```bash
$version = python -c 'import psychopy; print(psychopy.__version__)'
```

PR #1986, opened 2018-08-28
BF: Make colorPicker return RGB rather than RGBA

### LINKED ISSUE #1985: ColorPicker returns RGBA rather than RGB as expected
The `ColorPicker` class (available from both the Builder and Coder interfaces) returns a list of 4 values (R, G, B, A) whereas generally an RGB triplet is expected. Hence, right-clicking in a colour field in Builder returns a value that will cause an error when trying to run the experiment (`ValueError: Invalid parameter. Should be length 3 but got length 4.`)

https://github.com/psychopy/psychopy/blob/e6d804a487d02daefd2d5c84ca317b531c9a052a/psychopy/app/_psychopyApp.py#L582

m-macaskill:
Fixes #1985.
Ensure `colorPicker` returns an RBG triplet rather than RGB + alpha. Currently causes errors when assigning colors, as 3 rather than 4 channel values are expected.

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -589,6 +589,7 @@ class PsychoPyApp(wx.App):
                 if dlg.ShowModal() == wx.ID_OK:
                     data = dlg.GetColourData()
                     rgb = data.GetColour().Get()
+                    rgb = rgb[0:3] # drop alpha value if present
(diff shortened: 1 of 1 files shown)


m-macaskill:
Closing as have found a wx API-based way to do this explicitly. Will re-open with that approach.

PR #1992, opened 2018-08-29
NF/EN: Added OBJ/MTL importer. General improvements to VAO support.

mdcutone:
This PR adds functions for importing Wavefront *.OBJ models with materials, mainly for use in VR experiments. Vertex data is loaded into vertex buffer objects and indexed VAOs are created for each material. VAO support has been improved to support index buffers which reduces redundant vertex data. These functions will form the basis of the Mesh3dStim class which is planned for release soon, which hides all the OpenGL stuff from the user. I'll soon release a guide on how to export models from Blender for presentation in PsychoPy.

### DIFF
--- a/docs/source/api/tools/gltools.rst
+++ b/docs/source/api/tools/gltools.rst
@@ -1,5 +1,3 @@
-
-
 :mod:`psychopy.tools.gltools`
 ----------------------------------------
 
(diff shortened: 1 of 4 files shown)


mdcutone:
Also just updated the Rift class documentation.

PR #2003, opened 2018-09-03
ENH: Highlights errors in component dialogs

dvbridges:
When variables should be set to update, the component dialog now warns the user. Also, when correct answer is improperly defined i.e. does not prepend code with '$', the user is warned of the error.


mirh:
To be fair, I set my polygon size to a constant `$(win.size[0],win.size[1])`.. and it's not really like I require reminding that it "should update".

PR #2005, opened 2018-09-04
Extend class "DlgFromDict"

BlackyFuchsberger:
Extended class "DlgFromDict" with the following features:

(1) The "sort_keys" parameter now also accepts a list defining the order of the displayed fields. This allows for a user defined order of input fields in the dialog.

(2) A "labels" parameter can be used to define labels which are displayed instead of the keys. This allows for comprehensible captions for input fields in the dialog while maintaining brief key strings.

(3) A "preserve_type" parameter can be used to enforce preservation of the variable types provided in the dictionary, if at all possible. This remedies the following problem currently occurring under Windows. Say, the call to DlgFromDict() is:

```
dictionary = {'Parameter1', 99}   # the 99 is an int
DlgFromDict(dictionary)
```

The dialog will display an input field labeled "Parameter1" with a pre-entered value of 99. If the user then deletes the latter "9" and inputs "7", the returned dictionary will be:

```
dictionary = {'Parameter1', 97}   # the 97 is still an int
```

However, if the user deletes the "99" - thus having the field be empty at some point during user input - and then re-enters "97", the dictionary will be:

```
dictionary = {'Parameter1', '97'}   # the 97 has become a string
```

So, in the former case, the provided value will remain an int. In the latter case, since an empty input field cannot be converted to int, the variable type will switch to string and remain so. That's not nice and fixed in this edit.

### DIFF
--- a/psychopy/gui/qtgui.py
+++ b/psychopy/gui/qtgui.py
@@ -332,9 +332,30 @@ class DlgFromDict(Dlg):
     Parameters
     ----------
 
-    sort_keys : bool
-        Whether the dictionary keys should be ordered alphabetically
-        for displaying.
+    dictionary : dict
+        A dictionary defining the input fields (keys) and pre-filled values
+        (values) for the user dialog
+        
+    title : str
+        The title of the dialog window
+
+    labels : dict
+        A dictionary defining labels (values) to be displayed instead of
+        key strings (keys) defined in ``dictionary´´. Not all keys in
+        ``dictionary´´ need to be contained in labels.
+
+    fixed : list
+        A list of keys for which the values shall be displayed in non-editable
+        fields
+        
+    order : bool | list
+        Either a list of keys defining the display order of keys in
+        ``dictionary´´. If not all keys in ``dictionary´´ are contained in
+        ``order´´, those will appear in random order after all ordered keys.
+        Or a boolean flag indicating that keys are to be sorted alphabetically.
+
+    tip : list
+        A dictionary assigning tooltips to the keys
 
     copy_dict : bool
         If False, modify ``dictionary`` in-place. If True, a copy of
@@ -342,6 +363,10 @@ class DlgFromDict(Dlg):
         user interaction) can be retrieved from
         :attr:~`psychopy.gui.DlgFromDict.dictionary`.
 
+    preserve_types : bool
+        If False, the return type of the values in ``dictionary´´ is
+        undetermined and will depend on the value format provided by the user
+        
     show : bool
         Whether to immediately display the dialog upon instantiation.
          If False, it can be displayed at a later time by calling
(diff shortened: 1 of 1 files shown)


peircej:
This pull request **introduces 2 alerts** when merging 41fa704350b2ec6a46c5b350ec2e51fdecb530f3 into cf7d17aa33bfb3c98d12e19c160a0e5c3c2ee52e - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-e2e61de47d0601dd1b234c8c5bf4ba5bb87136f1)

**new alerts:**

* 2 for Wrong name for an argument in a class instantiation

---

peircej:
Thanks for the PR and apologies for the delay. 

The slight issues that had me confused:
1. I'm puzzled that this was needed. The preserve_type behaviour has been there for years and doesn't have a means to turn it off (it doesn't have an arg, but was doing this always). I think we can just assume that preserve_type should always be `True`
2. I was also not happy with the fact that you'd used snake_case instead of camelCase, but I now realise that was caused by @hoechenberger last year and I failed to spot it going in. So now the question is what to do about it (revert it to the main PsychoPy style, hoping that now many people have used it, or leave it as-is)

For (1) I'm pretty sure you can simply drop the `preserve_type` arg and make this standard behaviour (I can't think of a time when the user would *want* types to convert to strings) but for (2) I'm open to suggestions of which is the lesser evil.

BlackyFuchsberger:
(1) Totally agree that preserve_type=True should not be changeable. I, too, cannot see any use case where the desired behavior would be a change of type. However, I did not want to intervene in the existing code/behavior too much. **If it's ok with you, I remove the parameter and make preserveType default behavior for the next PR.**

(2) Thanks for clarifying the snake_case vs camelCase issue. I had been wondering why camelCase was substituted by snake_case for only this portion of qtgui back in 2017. As for a resolution, I'd suggest the following:

```
def __init__(self, dictionary, title='', fixed=None, order=None,
                 tip=None, screen=-1, sortKeys=True, copyDict=False,
                 show=True, **kwargs):

[...]

    if 'sort_keys' in kwargs:
        sortKeys = kwargs['sort_keys']
        logging.warning("Parameter 'sort_keys' is deprecated. Use 'sortKeys' instead.")

    if 'copy_dict' in kwargs:
        copyDict = kwargs['copy_dict']
        logging.warning("Parameter 'copy_dict' is deprecated. Use 'copyDict' instead.")

```
I'd also provide property setter methods for 'sort_keys' and 'copy_dict' which set the according camelType variables and also issue the deprecation warning. Then, in a future release, you can remove the kwargs portion of the code and revert to camelCase. **Ok for you for my next PR?**

peircej:
That sounds perfect, yes. :-)

BlackyFuchsberger:
> That sounds perfect, yes. :-)

Ok, on it ;-)

BlackyFuchsberger:
Ok, changes are committed. HTH!

peircej:
Great! :-)

peircej:
Ah, I didn't mention but in future it would be great if you could you prefixes on your commit messages as outlined in http://www.psychopy.org/developers/repository.html#commit-messages
When I build new releases I go through the git commit history to work out what commits need including and this helps a lot (especially the subtle difference between BF and FF)

BlackyFuchsberger:
Ah, darn, I did that on my first bug fix forgot it this time. I'll make
sure to remember next time. Thanks for reminding.

Jon Peirce <notifications@github.com> schrieb am Fr., 14. Sep. 2018, 21:40:

PR #2009, opened 2018-09-10
BF: Prevent save into psychopy folder when exporting an unsaved exp.

dvbridges:


peircej:
Could you provide a bit more info about in your commit messages (in the other lines) about what exactly is going wrong, so I can tell what problem is being solved?

PR #2020, opened 2018-09-14
BF: Allows user to set sound volume dynamically

dvbridges:
In the second commit for the pyo backend:

When updating the sound volume dynamically using Pyo, a pyolib._core.PyoArgumentTypeError is raised because self.volume is a numpy float64, which is not expected by pyolib. This fix converts the self.volume variable to a native python float that is compatible with the Pyo library.

### DIFF
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -88,9 +88,11 @@ class SoundComponent(BaseComponent):
 
     def writeRoutineStartCode(self, buff):
         if self.params['stopVal'].val in [None, 'None', '']:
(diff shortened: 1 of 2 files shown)


peircej:
One issue here is that this will create a line in the log file announcing the new volume every time we start a new trial. If the user hasn't chosen to set volume "on every repeat" this might be confusing and unnecessary?

dvbridges:
Ok good point, perhaps I set the code so that setVolume is only called if volume has a variable vs. a float?

peircej:
Because you're creating a new sound you probably do need to call it each time, but most of these setXXX() commands support a log=False option to prevent swamping the log files with things the user didn't set to change.

PR #2028, opened 2018-09-17
BF: Proposed fix to allow continuous mouse tracking for every frame

### LINKED ISSUE #2019: Mouse component settings contradict
With the Builder mouse component, choosing "every frame" for the "save mouse state" setting does not always work as expected (i.e. with the mouse position and button values being recorded on every screen refresh). 

Continuous tracking only works if the "End routine on press" is set to "never". If the routine is set to end on a valid mouse click, then only the final mouse state is recorded in the data (i.e. the mouse position at the moment the click was made), even though the user has requested continuous recording.

Whether or not the mouse button is used to end the trial should be independent of whether the mouse position is sampled continuously throughout the trial. The saved mouse state should be restricted to a single final instantaneous value only if the user has explicitly chosen that option under the separate "save mouse state" control.

Currently mouse trajectories can only be recorded if some other non-mouse means is used to end the trial, but there should be no reason why continuous tracking and mouse-click responding shouldn't both be able to applied simultaneously (and the interface implies to the user that they can be).

dvbridges:
Fixes #2019
If accepted, the JS script will be fixed to match the python script

### DIFF
--- a/psychopy/experiment/components/mouse/__init__.py
+++ b/psychopy/experiment/components/mouse/__init__.py
@@ -186,7 +186,6 @@ class MouseComponent(BaseComponent):
         # we need more than one
         code = ("# setup some python lists for storing info about the "
                 "%(name)s\n")
-
         if self.params['saveMouseState'].val in ['every frame', 'on click']:
             code += ("%(name)s.x = []\n"
                      "%(name)s.y = []\n"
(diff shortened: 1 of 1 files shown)


peircej:
This pull request **fixes 1 alert** when merging 92bd013ca3496105b458c7fcb9c8d1c7df0c8419 into 36ddcd8b3b0a06fc1f01ad9e88c52dc72563dc0a - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-0cf6dbd5ab39a5686be59be68ac522ef8635a3ca)

**fixed alerts:**

* 1 for Unused local variable

---

PR #2033, opened 2018-09-19
ENH: ExperimentHandler.saveAsWideText() has a new parameter "sortedColumns"…

BlackyFuchsberger:
… which has header names sorted alphabetically when storing data in a text file.

This guarantees a predictable order of columns in the stored csv/tsv files.


hoechenberger:
Mh, we should add some tests for that too…

PR #2037, opened 2018-09-26
BF: Fix event dispath for piglet joystick.

chrisgatwin:
This change fixes the long standing issue with polling of joystick events using the pyglet API. It has been successfully tested on Windows/Mac Py2 and Py3 Pyschopy2 and Psychopy3. The work has been carried out to enable support of generic joystick/game-pad button responses as an alternative to keyboard responses - in particular to support easy access to MRI triggers/Button boxes.


peircej:
Thanks for working on this. Very useful that this is finally fixed! :-)

PR #2052, opened 2018-10-08
Testing Travis precise->trusty

peircej:


.travis.yml
```diff
@@ -115,10 +115,10 @@ before_script:
        /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1024x768x24 -ac +extension GLX +extension RANDR +render -noreset;
     fi
 
-# This might come in handy once we switch to Trusty
-#  - if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
-#      xpra --xvfb="Xorg +extension GLX +extension RANDR +extension RENDER -config `pwd`/dummy_xorg.conf -logfile ${HOME}/.xpra/xorg.log"  start :99;
-#    fi;
+  # This might come in handy once we switch to Trusty
+  - if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
+      xpra --xvfb="Xorg +extension GLX +extension RANDR +extension RENDER -config `pwd`/dummy_xorg.conf -logfile ${HOME}/.xpra/xorg.log"  start :99;
```
peircej: @yarikoptic I'm trying to get tests to work under travis trusty. Initially we're getting an error that glReadBuffer(GL_BACK) isn't valid. 
Then I've seen these lines that I think you added a long time, which I figured might be related. But when I've inserted them the travis tests don't even launch. I suspect that means there's a syntax error in this. I can randomly cut bits out (like the back-tick \`pwd\`) and see what causes things to run again, but I thought I'd check if you had more knowledge. cheers

hoechenberger:
Hi @peircej, I added that a year ago, at least that's what `git blame` is telling ;) I don't remember where I got this from, probably Stack Overflow or a blog post. I do remember that it was intended to work around certain limitations on Trusty (?), but forgot about all the details.

.travis.yml
hoechenberger: Hi @peircej, I added that a year ago, at least that's what `git blame` is telling ;) I don't remember where I got this from, probably Stack Overflow or a blog post. I do remember that it was intended to work around certain limitations on Trusty (?), but forgot about all the details.

yarikoptic: Yeah, those aren't familiar to me yet. But now I will be interested to see them final recipe... so far I didn't bother with xpra and just used xvfb for X apps testing. Would be interested to know what  specific in trusty requiring this

peircej:
OK, then I guess I'm resorting to random changes and see what happens!

The error I'm trying to fix is
```
psychopy/visual/window.py:958: in _getFrame
    GL.glReadBuffer(GL.GL_BACK)
../../../miniconda/envs/psychopy-conda/lib/python2.7/site-packages/pyglet/gl/lib.py:105: in errcheck
    raise GLException(msg)
E   GLException: invalid operation
```
which is most likely caused by the lack of a back buffer(!) so what I'm trying to work out is why the double-buffered windows aren't working correctly and my best guess was this code about lack xvfb

.travis.yml
peircej: Never mind. the issue we were having was not to do with this after all. OpenGL tests passing fine with Trusty virtual machine without needing this step so I'll probably remove all mention of it

PR #2053, opened 2018-10-08
ENH: Adds redirectURL for psychoJS.

dvbridges:
Note, studyID is not a parameter and at this point is always undefined. One option is to remove studyID and allow user to complete this in experiment settings.


peircej:
I'm not sure we need studyID to exist. Why not just make that a part of the URL that the participant has to create? I don't think it will be used anywhere else, and what if one of the services puts the ID somewhere in the middle not at the end?

dvbridges:
Yes, no problem, I had thought about removing it before the commit, but wanted to check first.

PR #2055, opened 2018-10-09
BF: New loop 'Selected rows' - should be created as type 'str' not 'c…

chrisgatwin:
In Builder new loop elements default the 'Selected rows' parameter to type 'code' but the XML reader enforces that this field is 'str'. The reader text implies that this should be the default for new loops since 1.81.00.

### DIFF
--- a/psychopy/experiment/loops.py
+++ b/psychopy/experiment/loops.py
@@ -81,7 +81,7 @@ class TrialHandler(object):
             hint=_translate("The start and end of the loop (see flow "
                             "timeline)"))
         self.params['Selected rows'] = Param(
-            selectedRows, valType='code',
+            selectedRows, valType='str',
             updates=None, allowedUpdates=None,
             label=_localized['Selected rows'],
             hint=_translate("Select just a subset of rows from your condition"


peircej:
What if the user wants to insert a variable `myRows`, changing selection each time? "myRows" is then code (not a string). Maybe it came from a blocks.xlsx sheet.

Either way, both text and code can be achieved. If SelectedRows type is code then:
-  "0:5"
- myRows

are both valid entries. On the other hand, if type if str then:

-  0:5
-  $myRows 

are what the user wants.

The question is which is *more often* the right answer. My sense for SelectedRows is that it will usually be used to extract variable subsections, so code is generally the right thing to default to.

chrisgatwin:
So my query then is that the XML loader does not honer the 'code' property but forces the field to be a 'str' regardless of what it was saved as. The source file loops.py has a note saying that it should be forced to 'str' since 1.81.00. Currently if I save the field as code when reloading the file it turns into a string and needs to be edited into the alternative form. Is this the intended behavior?

peircej:
You shouldn't change the type. Leave it as code and set your value to be "0:5"

I supposed we could avoid writing it out and giving the impression that this is editable (but it's much easier to write the entire objects, ncluding their read-only properties). Actually, once upon a time I was planning to give the option to change entry types in the dialog boxes, but I decided that using $ and "" to switch manually within the box was just as easy, and maybe more visually clear

peircej:
Ah, OK, I've understood you now. I thought you were changing the type yourself and wanted the change to be the new default.

PR #2058, opened 2018-10-10
BF: Switch to AddTool for Builder Frame as AddSimpleTool deprecated...

dvbridges:
However, having found msg in _psychopyApp.py that AddTool breaks in version 3.0.2, holding off on further work. If it is time to switch, will replace remaining AddSimpleTool calls.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -1155,68 +1155,71 @@ class BuilderFrame(wx.Frame):
         keys = {k: self.app.keys[k].replace('Ctrl+', ctrlKey)
                 for k in self.app.keys}
 
-        item = tb.AddSimpleTool(wx.ID_ANY, newBmp,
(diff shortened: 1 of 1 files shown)


dvbridges:
@peircej do you know if AddTool still breaks in version 3.0.2?

dvbridges:
@peircej , this still breaks with wx3, so I am closing this pull request.

peircej:
Yes, I added something at the start of psychopyApp to turn off the warning:

https://github.com/psychopy/psychopy/blob/b6f17b791170e33d839df1a4f21b0bd91b96406b/psychopy/app/_psychopyApp.py#L33

I think we need to wait until people aren't using Py2 anymore (when they switch to Py3 they'll switch to wx4)

peircej:
I mean, when we stop supporting Py2 with new releases. They can still use older versions, with Py2 and the change won't be a problem.

PR #2066, opened 2018-10-16
TESTS: Split out system and Python setup into separate shell scripts

hoechenberger:
This should make the test setup procedures much more readable as we can get rid of all those repetitive `if` statements.

The individual steps were not modified.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,70 +30,18 @@ matrix:
 #    - python: "2.7_with_system_site_packages"
 
 before_install:
-  # System setup
-  - travis_retry sudo apt-get update -qq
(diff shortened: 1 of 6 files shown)


peircej:
Big thanks for this. I was actually trying something similar a few weeks ago, and failed to get it all working, so I'm glad you made this happen! (I should say that @dvbridges is now working on a Jenkins setup too but that will probably take a while to get going right!)

hoechenberger:
Thanks Jon! It also took me a while to get the conda environment stuff and `travis_retry` right :)

The Jenkins project is surely exciting, looking forward!

PR #2068, opened 2018-10-18
NF: New Joystick Builder component - based on mouse component.

chrisgatwin:
New candidate builder Joystick component.

Designed to emulate the current mouse component functionality and workflow using a joystick.
If no joystick is present one is emulated using the mouse and possibly keyboard for additional buttons. Coordinates are accessed with object.getX() and object.getY().

Documentation (based on mouse) will follow if accepted for inclusion into mainline.
Tested on PsychoPy2 & PsychoPy3. Alternative icons may be preferred.


peircej:
Hi Chris. I don't think I realised that the previous one was only about joystick *buttons*. Sorry I hadn't given it my full attention. I think it's a bit weird to have one for joystick and another for joystick buttons, no?

chrisgatwin:
Hi Jon. My initial starting point, with no experience with Psychopy, was to fully emulate the semantics of the keyboard control, so as to cover as many use cases as possible without my fully understanding how users design experiments in Psychopy (My only use case is for button responses and fMRI scanner triggers). Having done that work it seemed that other users would probably want to support a joystick more fully so I implemented the Mouse semantics again to hopefully cover as many use cases as possible. The main difference is that joyButtons has the keyboard store: last, first, all, nothing options, and the joystick has save joystick state: final, on click, every frame, never. You are in a much better position than I am to judge the relative importance of these distinctions and whether the button version is redundant. Either way the gamepad icon would probably best suit the joystick control. Please feel free to include or not which ever combination seem best to you. Regards Chris.

peircej:
Right, OK, so with this one (sorry for the delay) I think we just need the new option, mimicking the mouse. I think it makes the previous one (joystick as a keyboard) a bit redundant. Do you want to update it:
- remove the previous
- copy the fancy joystick icon (or gamepad, or a combination icon of them both) to this new component

cheers,
Jon

peircej:
Great. I've had to pull this in from another branch due to a conflict, but it will be there in the next release (3.0.0b12 on Monday I think)

PanBartosz:
My problem with this change is that does not work on my hardware/OS. Now you can only add one joystick component per experiment. For example now you cannot have two routines with joystick component (no joystick found error).  I cannot figure out what is responsible for this change of behaviour.

m-macaskill:
Can the issue @PanBartosz raises re being able to instantiate only one joystick component be related to the same issues being addressed by @hoechenberger with his new syringe component? i.e. some components should only be initialised once, no matter how many times they appear in routines.

hoechenberger:
There's two potential issues here, as far as I can see. As @m-macaskill pointed out, I encountered similar problems when working on my pump component.

First potential issue is that each instance of the component adds their own `import` statements, so if you have two joystick components, all imports will be duplicated; for three components, imports will be triplicated (?!) etc. This can be easily addressed once #2145 is merged.

Second potential issue is that if you have multiple components trying to access the same device, there might naturally be resource conflicts. My solution with the pumps was to introduce a module-global list of initialized devices, which could then be re-used.

This then leads to a third potential issue, which is that if we simply re-use the initialized devices across multiple components, we will experience race conditions as the the same `status` property gets updated by those different components. My workaround was to create [a thin "wrapper class"](https://github.com/psychopy/psychopy/blob/77ab604c31303d79c9aaa7ab23ae2d6f6818d780/psychopy/hardware/qmix.py#L290) with its very own "status", and make the Builder component instantiate this wrapper, not the "actual" device class. So now, multiple components can use the same hardware device, but maintain separate `status`es.

But all of this should actually go into the issue tracker, as it's not related to this PR ;)

hoechenberger:
Created an issue on the issue tracker, please continue discussion there.

PR #2069, opened 2018-10-19
Don't mutate list while iterating

mgeier:
A classic Python bug.

There may be a better solution than what I'm suggesting here, but in any case the list `self.sounds` should not be mutated while being iterated.

This fixes the crackling problem described in https://discourse.psychopy.org/t/sound-crackling-under-specific-conditions-demo-included/5480.


peircej:
Oh yeah, well spotted. thanks

PR #2070, opened 2018-10-20
FIX: changing sound duration in builder causes the routine with the sound to not be played due to indentation bug

mmagnuski:
Hi there,

this is a small PR to fix the problem with setting sound duration.

### How to reproduce the problem
1. Open psychopy 1.90.3 builder. (the problem seems to have been introduced in this version)
2. Add sound component to the routine. At this point if you play the procedure the sound would still play.
3. Change the sound duration. At this point the routine will be ommitted and thus there would be no sound.

### This fix
The problem comes from missing ` buff.setIndentLevel(-1, relative=True)` in one of the code branches. I just changed the indentation of that code which fixes the indentation of the generated code. :)

### DIFF
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -143,8 +143,8 @@ class SoundComponent(BaseComponent):
                 self.writeStopTestCode(buff)
                 code = "%s.stop()  # stop the sound (if longer than duration)\n"
                 buff.writeIndented(code % self.params['name'])
-                # because of the 'if' statement of the time test
-                buff.setIndentLevel(-1, relative=True)
+            # because of the 'if' statement of the time test
+            buff.setIndentLevel(-1, relative=True)
 
     def writeFrameCodeJS(self, buff):
         """Write the code that will be called every frame


dvbridges:
Thanks @mmagnuski for the detailed description for recreating the error in 1.90.3. However, the indentation in this (current) version of the code submitted does not need correction. The conditional on line 138 indents the code within the string, and thus does not need the indentation changing using ```setIndentLevel```. The code only needs the indentation changed after the ```writeStopTestCode``` function has been called following the conditional on line 142.

dvbridges:
Thanks @mmagnuski for the detailed description for recreating the error in 1.90.3. However, the indentation in this version of the code submitted does not need correction. The conditional on line 138 indents the code within the string, and thus does not need the indentation changing using ```setIndentLevel```. The code only needs the indentation changed after the ```writeStopTestCode``` function has been called following the conditional on line 142.

mmagnuski:
@dvbridges Right, I stand corrected.  
However the root of the problem I mention is in bad indentation in the generated code - but as you suggest the code that leads to this must be somewhere else. The indentation of the generated code changes after one edits sound duration (there is one unindent missing) - and this indentation problem occurs just after the line with comment `# stop the sound (if longer than duration)`.

dvbridges:
@mmagnuski, that is correct :), and thank you bringing this up. I can trace the original error to commit 476a8fd10aa7601cf3a00ec45a0ff2d344d647de. This happened when fixing the sound component so variable could be used for sound duration. If @peircej plans to release a 1.90.4, the fix will be included, but for now the fix is already in place in versions >= version 3.0.0.

peircej:
There are no plans to include a 1.90.4 release, so @mmagnuski unless you find this recurring in 3.0+ I'll close this

mmagnuski:
@peircej I thought that 3.0.0 is in beta and it is not adviced to use it in the classrom yet. On versions `1.85.6` I have other sound-related problems (fixed later in psychopy) and on `1.9.1` the proceure I need in the classroom doesn't run (without any messages or errors shown) - so I am a bit puzzled on which version it is safe to use.

PR #2074, opened 2018-10-24
ENH: Adds logging messages to Form class.

dvbridges:


peircej:
I wonder if some of these should be set to info() rather than exp(). e.g. updating the layout probably isn't key part of the data most studies. It's just info for the curious?

dvbridges:
> I wonder if some of these should be set to info() rather than exp(). e.g. updating the layout probably isn't key part of the data most studies. It's just info for the curious?

Sounds good, anything unrelated to stimuli or the exp I will change to info.

PR #2083, opened 2018-11-02
BF: Allow earlier destruction of GLFW window

aforren1:
Currently, the GLFW window sticks around until the python script terminates. This allows destruction of the window + context before that point.

Example:
```python
from psychopy import core, visual

win = visual.Window(winType='glfw', allowGUI=False, pos=(0, 0), size=(400, 400))
core.wait(2.0)
win.close()
core.wait(2.0)
```


mdcutone:
Nice fix!

And yes, accessing GLFW purely through the binary interface makes some functionality inaccessible.

PR #2088, opened 2018-11-12
NF: Tools for computing view projections.

mdcutone:
Initial commit of 'viewtools', a collection of functions to compute projection and view matrices. These will facilitate accurate 3D and stereoscopic rendering. I'm going to add some methods to the windows class to easily use the data these functions generate. 

I re-formulated the frustum calculation to support off-axis views too, so I removed the functions from (undocumented) stereotools to remove the redundancy.


mdcutone:
Added a bunch of methods to the 'window' class to easily use viewtools. This is a WIP so features may change without notice. Here are some examples:

Setup for rendering with projection, the scene will be translated so the origin falls on the centre of the screen. All rendering will be correct (i.e. rendered objects will subtend the same visual angles they would have in reality at some distance) assuming monitor configurations are correct and PsychoPy is running fullscreen.
```
win.setPerspectiveView()
# draw objects here 
win.flip()
```

Specifying your own view and projection matrices:
```
win.viewMatrix = np.asarray([[1.0, 0.0, 0.0, 0.0],
                             [0.0, 1.0, 0.0, 0.0],
                             [0.0, 0.0, 1.0, 0.0],
                             [0.0, 0.0, 0.0, 1.0]])

# computed using viewtools
win.projectionMatrix = viewtools.orthoProjectionMatrix(-1, 1, -1, 1, -1, 1)

# then you can apply your custom view as follows
win.applyEyeTransform()
# draw objects here 
win.flip()

```
Note 'applyEyeTransform' is called automatically when you call 'setPerspectiveView'.

mdcutone:
Added 'pointToNDC' which computes the position in normalized coordinates a point in the 3D scene will appear given a view and projection matrix. This can also be used to determine if a point falls within a viewing cone (frustum) for visibility calculations or to calculate point disparities between two different viewpoints.

peircej:
This is all very cool. I don't understand all of it, but it's very cool! ;-)
I'll just keep pulling it in as you tell me it's ready (but warn me if anything happens that might affect the operation of the rest of PsychoPy). thanks for all your hard work!

mdcutone:
My only concern is with retina support breaking since the projection is updated. It should be fine, but I can't test it since we've moved off from Macs here.

peircej:
I have a retina machine in the office. Can test if you tell me exactly what I need (you mentioned MeshStim not being in there yet)

mdcutone:
> 
> 
> This is all very cool. I don't understand all of it, but it's very cool! ;-)

Defining view (and model) transformations this way will allow us to use more modern OpenGL rendering techniques (or at least give people the option to). We can also involve Numpy and SciPy directly in computing these transformations, taking advantage of the massive amount of libraries out there. In addition, we can use data pulled from devices such as the Opti-Track and Leap Motion directly.

dvbridges:
Hi @mdcutone , looks cool! Just had a question asking whether your new viewtools collection can/will support HTC Vive? Thanks.

mdcutone:
> 
> 
> Hi @mdcutone , looks cool! Just had a question asking whether your new viewtools collection can/will support HTC Vive? Thanks.

Viewtools is intended for rendering 3D scenes on conventional displays. For HMDs, you need to pull these transformation matrices from the HMD's runtime since they use additional (sometimes secret) parameters. The Oculus Rift support in PsychoPy provided by PsychXR can obtain these transformations from the runtime. In the future, PsychXR may support the Vive too (possibly through OpenHMD).

The modifications I made to the window class will simplify using transformations from various sources in a standardized way, so 3D stimuli will render correctly on all sorts of displays.

Issue #2089, opened 2018-11-13
The GLFW library is not initialized


mdcutone:
In PsychoPy 3.0.0b11, exiting the application with an open GLFW window causes the following error...

```
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "C:\Users\mdc\AppData\Local\Programs\Python\Python36\lib\site-packages\glfw\__init__.py", line 611, in callback_wrapper
    return func(*args, **kwargs)
  File "C:\Users\mdc\AppData\Local\Programs\Python\Python36\lib\site-packages\glfw\__init__.py", line 785, in _raise_glfw_errors_as_exceptions
    raise GLFWError(message)
glfw.GLFWError: (65537) b'The GLFW library is not initialized'
```

when running ...

```
...
while 1:
    grating.draw()
    win.flip()

    if event.getKeys(['q']):
        break

core.quit()
```

It seems to be caused by recent changes to how the GLFW backend handles closing windows. We may need to modify the registered 'atexit' function 'close_on_exit' associated with the window class to fix this problem.

gitouni:
I met this problem when use open3d in remote ssh mode
`from open3d.web_visualizer import draw`
comment this line will tackle this problem.

PR #2091, opened 2018-11-14
BF: Run `glfw.terminate()` later during cleanup.

aforren1:
Addresses #2089.

Currently, if `core.quit()` is called while a GLFW Window is open, `glfw.terminate()` (in core.py) runs before the window's `close_on_exit()` (via atexit). This PR registers `terminate()` with atexit when core is imported, meaning it ought to run *after* `close_on_exit()` (which is registered later).

Example:
```python
from psychopy import core, visual

win = visual.Window(winType='glfw')

core.quit()
```

@mdcutone, what do you think?

### DIFF
--- a/psychopy/core.py
+++ b/psychopy/core.py
@@ -9,6 +9,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import atexit
 from builtins import object
 import sys
 import threading
@@ -39,6 +40,7 @@ except ImportError:
 try:
     import glfw
     haveGLFW = True
+    atexit.register(glfw.terminate)
 except ImportError:
     haveGLFW = False
 
@@ -72,17 +74,13 @@ def quit():
     """
     # pygame.quit()  # safe even if pygame was never initialised
     logging.flush()
-    
+
     for thisThread in threading.enumerate():
         if hasattr(thisThread, 'stop') and hasattr(thisThread, 'running'):
             # this is one of our event threads - kill it and wait for success
             thisThread.stop()
             while thisThread.running == 0:
                 pass  # wait until it has properly finished polling
-
-    # call terminate() on GLFW if available
-    if haveGLFW:
-        glfw.terminate()
 
     sys.exit(0)  # quits the python session entirely
 
@@ -126,7 +124,7 @@ def shellCall(shellCmd, stdin='', stderr=False, env=None, encoding=None):
     """
     if encoding is None:
         encoding = locale.getpreferredencoding()
-    
+
     if type(shellCmd) == str:
         # safely split into cmd+list-of-args, no pipes here
         shellCmdList = shlex.split(shellCmd)
--- a/psychopy/core.py
+++ b/psychopy/core.py
@@ -72,17 +72,13 @@ def quit():
     """
     # pygame.quit()  # safe even if pygame was never initialised
     logging.flush()
-    
+
     for thisThread in threading.enumerate():
         if hasattr(thisThread, 'stop') and hasattr(thisThread, 'running'):
             # this is one of our event threads - kill it and wait for success
             thisThread.stop()
             while thisThread.running == 0:
                 pass  # wait until it has properly finished polling
-
-    # call terminate() on GLFW if available
-    if haveGLFW:
-        glfw.terminate()
 
     sys.exit(0)  # quits the python session entirely
 
@@ -126,7 +122,7 @@ def shellCall(shellCmd, stdin='', stderr=False, env=None, encoding=None):
     """
     if encoding is None:
         encoding = locale.getpreferredencoding()
-    
+
     if type(shellCmd) == str:
         # safely split into cmd+list-of-args, no pipes here
         shellCmdList = shlex.split(shellCmd)
--- a/psychopy/core.py
+++ b/psychopy/core.py
@@ -9,7 +9,6 @@
 
 from __future__ import absolute_import, division, print_function
 
-import atexit
 from builtins import object
 import sys
 import threading
@@ -40,7 +39,6 @@ except ImportError:
 try:
     import glfw
     haveGLFW = True
-    atexit.register(glfw.terminate)
 except ImportError:
     haveGLFW = False
 
@@ -74,13 +72,17 @@ def quit():
     """
     # pygame.quit()  # safe even if pygame was never initialised
     logging.flush()
-
+    
     for thisThread in threading.enumerate():
         if hasattr(thisThread, 'stop') and hasattr(thisThread, 'running'):
             # this is one of our event threads - kill it and wait for success
             thisThread.stop()
             while thisThread.running == 0:
                 pass  # wait until it has properly finished polling
+
+    # call terminate() on GLFW if available
+    if haveGLFW:
+        glfw.terminate()
 
     sys.exit(0)  # quits the python session entirely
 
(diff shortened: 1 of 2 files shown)


mdcutone:
We may need a more complicated solution. Right now the window handles the gamma table and display configuration. I think the better solution would be to have the application configure the displays independent of the windows.

mdcutone:
Furthermore, I really believe we need some sort of global setting where a single backend is used from the start to avoid the whole business with `haveGLFW` or `havePyglet`. There may be cases where the user has GLFW and might not be using it. However, we're still doing GFLW specific things. It also cleans up the conditional checks for a particular backend. For instance, we can check a single global variable for which backend is being used, avoiding the need to write code to resolve conflicts when multiple backends are loaded.

peircej:
I'm fine with this. We could set the backend in prefs and just always use that (deprecating `winType` arg in Window settings)? If people want to change that then they do
```
from psychopy import prefs, visual
prefs['window backend'] = 'glfw'
# can we only use backend pref on instantiation (not during import)?
win = visual.Window()
```
Alternatively it could be just in psychopy (but populated from prefs?):
```
import psychopy, visual
psychopy.winBackend = 'glfw'
win = visual.Window()

Related to the backend question, I've been having a little play with the psychtoolbox->python port of the usbHID interface and this should make a backend-agnostic keyboard/mouse/joystick system that is superior to any of the backend-specific engines:

- timestamping at (close to) source
- faster than standard keyboard monitoring (going straight to usb events)
- global events (not requiring a window)
- not requiring Mac Accessibility settings
- device-specific (can specify *which* keyboard/buttonbox you want events from, which isn't possible with standard OS keyboard libs)

**Comment by mdcutone** [comment]
> I'm fine with this. We could set the backend in prefs and just always use that (deprecating `winType` arg in Window settings)? If people want to change that then they do
> 
> ```
> from psychopy import prefs, visual
> prefs['window backend'] = 'glfw'
> # can we only use backend pref on instantiation (not during import)?
> win = visual.Window()
> ```
> 
> Alternatively it could be just in psychopy (but populated from prefs?):
> ```
> import psychopy, visual
> psychopy.winBackend = 'glfw'
> win = visual.Window()
> ```

We need to call GLFW's initialization routines while psychopy's top-level module is being loaded, so I think having it as a preference stored in a config file is the best route. Finally, when the module is being unloaded we can safely terminate the GLFW session and restore gamma tables.

> Related to the backend question, I've been having a little play with the psychtoolbox->python port of the usbHID interface and this should make a backend-agnostic keyboard/mouse/joystick system that is superior to any of the backend-specific engines:
> 
>     * timestamping at (close to) source
> 
>     * faster than standard keyboard monitoring (going straight to usb events)
> 
>     * global events (not requiring a window)
> 
>     * not requiring Mac Accessibility settings
> 
>     * device-specific (can specify _which_ keyboard/buttonbox you want events from, which isn't possible with standard OS keyboard libs)

The above is concerned only with the display backend. For input and sound, using the PTB port is a no-brainer.

**Comment by aforren1** [comment]
I moved `glfw.terminate()` out of core.py to the backend file for cleanliness, though it ought to have the same phenotype. Maybe I don't understand the issue, but isn't this cleanup independent of how the display is configured?

I'm definitely on board with having a preference for the window backend (maybe closer to the first option, for consistency with how sound does it?).

**Comment by peircej** [comment]
> We need to call GLFW's initialization routines while psychopy's top-level module is being loaded

OK, well then I still suggest we use prefs but it means that if someone wants to override for a specific script then they can't import prefs and visual in the same line. They'd need to do this instead:
```
from psychopy import prefs
prefs['window backend'] = 'glfw'
from psychopy import visual
win = visual.Window()
```

That isn't awful, just not *quite* as neat

**Comment by mdcutone** [comment]
> > We need to call GLFW's initialization routines while psychopy's top-level module is being loaded
> 
> OK, well then I still suggest we use prefs but it means that if someone wants to override for a specific script then they can't import prefs and visual in the same line. They'd need to do this instead:
> 
> ```
> from psychopy import prefs
> prefs['window backend'] = 'glfw'
> from psychopy import visual
> win = visual.Window()
> ```
> 
> That isn't awful, just not _quite_ as neat

I see your point. We could move the GLFW initialization elsewhere, as long as it is called prior to any window creation. 

My overall goal here is to eventually decouple the displays from the window objects, where the windows are only concerned about which display they are presented on. The current paradigm works well in a single window per display setup, however, I've been having issues with multi-display systems.

I would imagine it would work something like this...

```
>>> import psychopy.visual as visual
>>> visual.getNumDisplays()
2
>>> visual.getNumScreen()  # extended mode, only one desktop
1
>>> visual.getDisplays()
[<Monitor object @ blah>, <Monitor object @ blah>]
>>> visual.getDisplays()[0].setGamma( ... )  # copies original gamma table, saves it
>>> visual.getDisplays()[1].setGamma( ... )
>>> visual.getDisplays()[0].widthCM
52.0
>>> win0 = visual.Window(screen=0)  # displays are extended, only one screen technically
>>> core.exit()  # saved gamma tables are restored here
```
This is a pretty contrived example, but we wouldn't need to hack the Window class to support such a configuration.

**Comment by aforren1** [comment]
If the worry is getting the gamma back to the original state, it looks like [glfwTerminate](http://www.glfw.org/docs/latest/group__init.html#gaaae48c0a18607ea4a4ba951d939f0901) automatically handles it?

**Comment by peircej** [comment]
This pull request **introduces 1 alert** when merging 0658016fa04dc8b2c230ba79ff415d4327f0c3b1 into c7a33b01244c8afd9a0d5f724eced919b33b49d7 - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-5ae462fe35e225a9eb01939bf9377b1303b9c220)

**new alerts:**

* 1 for Unused import

---

**Comment by mdcutone** [comment]
> 
> 
> If the worry is getting the gamma back to the original state, it looks like [glfwTerminate](http://www.glfw.org/docs/latest/group__init.html#gaaae48c0a18607ea4a4ba951d939f0901) automatically handles it?

This is more of a general thing about handling displays in the future. However, terminate isn't called on error which messes up people's desktops.

**Comment by peircej** [comment]
Tell me, what's the deal with this actual PR? ;-) Is the discussion about the refactoring unrelated and the PR is fine? Sorry, I wasn't sure about it not having tested the code myself

**Comment by mdcutone** [comment]
I'm assuming it will work. Though terminate may be called once for each window. Moving the creation of the atexit handler to the top-level visual module will make sure it's only called once. I think that's what @aforren1 did, but I haven't tested it.

**Comment by aforren1** [comment]
Right, it should only call terminate once. Here's a snippet (python 2, atexit lost `_exithandlers` in python 3):

```python
from psychopy import core, visual
import atexit

win = visual.Window(winType='glfw', pos=(0,0))
win2 = visual.Window(winType='glfw', pos=(800,0))

for f in atexit._exithandlers:
    print(f[0])

win2.close()

core.wait(2)

core.quit()
```
which gives me something like:

```python
<function shutdown at 0x7f3e3b9cd230>
<function flush at 0x7f3e3addd668>
<built-in function clear_cache>
<function _exit_function at 0x7f3e29b55ed8>
<built-in function quit>
<function terminate at 0x7f3e3ab55f50> # registered when glfwbackend is imported
<function close_on_exit at 0x7f3e168d1f50> # for win
<function close_on_exit at 0x7f3e168d1b90> # for win2
```

mdcutone:
Looks okay, but we should probably remove the close_on_exit handler from the window class, or move it to the pyglet backend if it still needs that routine.

Issue #2095, opened 2018-11-16
Can't install 3.0.0b8


tjwilli58:
I wasn't sure if I should just reopen #2049 or start a new issue. I'm basically having the same problem.
I'm have psychopy-1.90.3 installed in an anaconda environment, and wanted to try out PsychoPy3. I did a 
`pip  uninstall psychopy` , downloaded the latest PsychoPy3 release, and tried to install it with pip.  I got the error

`FileNotFoundError: [Errno 2] No such file or directory: 'version'`

I then saw that 3.0.0b11 was not a stable release and found #2049 . I then download 3.0.0b8 and updated my setuptools to v40.6.2

I'm still getting the error when I try to do a pip install.

> (perceptionlab) L:\Downloads\Anaconda\PsychoPy\psychopyWheels>pip install --no-index -f . psychopy==3.0.0b8
> Looking in links: .
> Collecting psychopy==3.0.0b8
>     Complete output from command python setup.py egg_info:
>     Traceback (most recent call last):
>       File "<string>", line 1, in <module>
>       File "C:\Users\TIMOTH~1.WIL\AppData\Local\Temp\4\pip-install-ha2vxtkb\psychopy\setup.py", line 22, in <module>
>         with open('version') as f:
>     FileNotFoundError: [Errno 2] No such file or directory: 'version'
> 
>     ----------------------------------------
> Command "python setup.py egg_info" failed with error code 1 in C:\Users\TIMOTH~1.WIL\AppData\Local\Temp\4\pip-install-ha2vxtkb\psychopy\
> 

Thanks for any help.

peircej:
As in the issue you linked to, you're installing a version that is neither cutting edge nor a bug-free stable version. Recommended options would be:

- 1.90.3 is the latest stable release (fewer bugs)
- use `pip install -U psychopy-dev` to get the latest release from the development branch
- install from the github repository to get the latest cutting-edge code (including not yet released features and possibly bugs)

Installing 3.0.0b8 is using the recent branch but before we fixed several bugs

Issue #2049, opened 2018-10-07
Can't install 3.0.0b10


Kodiologist:
$ sudo -H pip3.6 install PsychoPy-3.0.0b10.zip 
    Processing ./PsychoPy-3.0.0b10.zip
        Complete output from command python setup.py egg_info:
        Traceback (most recent call last):
          File "<string>", line 1, in <module>
          File "/tmp/pip-req-build-ne5qm4gw/setup.py", line 22, in <module>
            with open('version') as f:
        FileNotFoundError: [Errno 2] No such file or directory: 'version'
    $ sudo -H pip2.7 install PsychoPy-3.0.0b10.zip 
    Processing ./PsychoPy-3.0.0b10.zip
        Complete output from command python setup.py egg_info:
        Traceback (most recent call last):
          File "<string>", line 1, in <module>
          File "/tmp/pip-req-build-VSxdQo/setup.py", line 15, in <module>
            from setuptools.config import read_configuration
        ImportError: No module named config

I'm running Ubuntu 18.04 and pip 18.1.

hoechenberger:
Hi @Kodiologist 

```python
$ sudo -H pip3.6 install PsychoPy-3.0.0b10.zip 
Processing ./PsychoPy-3.0.0b10.zip
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-ne5qm4gw/setup.py", line 22, in <module>
        with open('version') as f:
    FileNotFoundError: [Errno 2] No such file or directory: 'version'
```

I suspect this is caused by an incomplete revert of our (my) attempt to use `versioneer` for versioning.
Introduction of `versioneer`: #1970, #1973, #1975
Removal: #1981

```python
$ sudo -H pip2.7 install PsychoPy-3.0.0b10.zip 
Processing ./PsychoPy-3.0.0b10.zip
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-VSxdQo/setup.py", line 15, in <module>
        from setuptools.config import read_configuration
    ImportError: No module named config
```

What is the output of 
```
python2.7 -c "import setuptools; print setuptools.__version__"
```

I suspect Ubuntu's `setuptools` for Python 2.7 is horribly outdated.

Kodiologist:
[...]

PR #2096, opened 2018-11-17
PKG, BF: added 'include version' to MANIFEST.in

tjwilli58:
This is to have the file 'version' in the sdist, and fix the problem with
'pip install' not finding the version file.
See issue #2049, #2095

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -35,3 +35,4 @@ prune docs/source/nonSphinx
 prune windlls
 include versioneer.py
 include psychopy/_version.py
+include version


peircej:
Thanks, while playing with the versioneer system (designed to handle versioning better, bt which we haven't got to work properly yet) we obviously ended up removing this

tjwilli58:
I'm glad it was a simple fix. When I could do a
`python setup.py bdist_wheel`
from my github clone, I was wondering why I couldn't pip install the tarball. When I saw that the version file was in the repository, but not in the tarball, it didn't take long to find out why.

PR #2100, opened 2018-11-20
BF: Adds howler library to JS HTML file thus allowing sound resources.

dvbridges:


peircej:
Cool. Thanks. And I worked out the problem with the other updates (you had put them into master I had just failed to pull them through because of a merge conflict)

PR #2103, opened 2018-11-21
TESTS: Enhancements to Travis testing

hoechenberger:
- Update the `conda` environments to match the current set of PsychoPy dependencies
- Actually try to build and install `sdist` and `bdist_wheel` distributions on Travis to catch any (simple) bugs in the first-stage packaging scripts (`setup.py` and such)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,18 @@ before_install:
 
 install:
   - python setup.py build
-  - python setup.py install
+
+  # Build & install sdist.
+  - python setup.py sdist
+  - pip install --no-deps dist/PsychoPy-*.tar.gz
+  - pip uninstall --yes psychopy
+
+  # Build & install wheel.
+  - python setup.py bdist_wheel
+  - pip install --no-deps dist/PsychoPy-*.whl
+  - pip uninstall --yes psychopy
+
+  - rm -rf build/ dist/
 
 before_script:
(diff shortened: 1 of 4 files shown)


peircej:
This is good. Yes, now that we have many places to update the dependencies ;-) we do need some way to test whether they all got done.

FYI gitpython has now been removed from the dependencies. I'm now using dulwich for most things and a simple call to `git` directly using subprocess when this is needed (merge pull/push operations).

So just dulwich and gitlab are needed now.

hoechenberger:
@peircej The `pip install --no-deps` in the updated `.travis.yml` is intended to reduce the opportunities for us to miss out on a dependency :) I also hope that, once we manage to produce a proper `conda` package, we can finally consolidate the `conda` environment files.

> FYI gitpython has now been removed from the dependencies. I'm now using dulwich for most things and a simple call to `git` directly using subprocess when this is needed (merge pull/push operations).

Oh, I only _added_ `dulwich`, but didn't remove `gitpython`. And probably should add `gitlab` too… Will do another PR. – This probably also means that we could use some more testing coverage here ;))

peircej:
Yes, the pushing and pulling of experiments to pavlovia doesn't feature in the test suite at all yet!! 😱
I know!! Too busy fixing known bugs to go looking for unknown ones!

PR #2107, opened 2018-11-22
WIP: Fix versioneer

hoechenberger:
To enable `versioneer` to work properly and to report correct version numbers when building a package (sdist, bdist_wheel), I had to disable updating `__git_sha__` and `__build_platform__` in `__init__.py`.

These values are not being used anywhere else, so I wonder if it's safe to remove them entirely? (I currently set them to `'n/a'`)

If the removal is okay, one could even simplify `createInitFile.createInitFile()` by removing special handling for `sdist` (currently writes `__git_sha__`) and `bidst` (currently writes (`__git_sha__` and `__build_platform__`).

WDYT?

----

During the Travis build, for some reason, we don't get the correct version number (or, rather, no version number at all -- it's sth like `0+unknown...`) However this _should_ work -- it does work for me with other projects. So there's definitely something off here.


peircej:
This pull request **introduces 3 alerts** and **fixes 30** when merging 79d1b21f6afec8ab311b55526ef34ee1677f6caa into e6e02d0a3e7083c139fae7fde74321a958068d0c - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-9f5667bd41cc05dec3e314c479282670d4398b79)

**new alerts:**

* 3 for \_\_iter\_\_ method returns a non\-iterator

**fixed alerts:**

* 19 for Module is imported with &#39;import&#39; and &#39;import from&#39;
* 4 for Missing call to \_\_init\_\_ during object initialization
* 3 for Non\-iterable used in for loop
* 3 for Inconsistent equality and hashing
* 1 for Missing call to \_\_del\_\_ during object destruction

---

peircej:
The reason to have a `__git_sha__` value was that when a version had been created _as a distribution_ we could always then detect the exact commit that it was based on.

I feel like, for a distribution, a fixed version (and a git sha) is actually the right thing to do. Maybe we should set it up so that versioneer is used *during setup* but that fixes the version to something, rather than supplying versioneer as part of the (dist) package and using it. 

I suggest something similar to what I've already been doing. setup.py calls createInitFile which updates the `__init__` file differently according to whether it's an sdist or bdist. At the end of setup.py we rewrites a vanilla `__init__`, so that devs have something empty. 

So let's say if we call `createInit(dist='bdist')` we bake in the versioneer number, and if we call `createInit(dist=None)` then we write it with `__version__=versioneer.get_version()`

hoechenberger:
Closing until I have time to look into this again.

PR #2133, opened 2018-12-10
BF: jsMouse was not saving data when save state was "final".

dvbridges:


dvbridges:
A ```test_radial``` fail for Py 3.6, unrelated to pull request.

PR #2141, opened 2018-12-14
NF: Add support for syringe pumps

hoechenberger:
This PR adds support for Cetoni neMESYS syringe pumps, as used in some labs (and, hopefully: _many more_ labs in the future!): https://www.cetoni.com/products/pumps/

Pump interfacing is enabled through [`pyqmix`](https://github.com/psyfood/pyqmix), which now consequently is a dependency (available via PyPI and conda-forge). The library is called `pyqmix` because it wraps Cetoni's `Qmix SDK`, which allows access to pumps, digital I/O boards, and many more devices. Currently, only support for the pump interface is implemented in PsychoPy.

In detail, this PR:

- Implements a new `Pump` class in `psychopy.hardware.qmix`. This module has the generic `qmix` name because we might want to add further pieces of hardware interfaces offered through `pyqmix` at some later time.
- Implements a new `QmixPumpComponent` for use in Builder.
- Adds a new preferences tab `Misc.`, which allows to set the name of the Qmix configuration. (When setting up the hardware, one has to create a certain device configuration using tools provided by the manufacturer. Later, this configuration can be consumed by `pyqmix`).

In `hardware.qmix`, we also implement a `_PumpWrapperForBuilderComponent` class. This class essentially wraps `hardware.qmix.Pump` and is only used by the respective Builder component. It exposes a `status` property to allow for re-use of the component. We keep track of all instantiated pumps in `psychopy.hardware.qmix.pumps`.

Please also note that each component will add a `from psychopy.hardware import qmix` import. The _proper_ way to avoid this, I guess,  would be to do sth. like `self.exp.requirePsychopyLibs(['qmix'])` in the component's `__init__()`, but since the `qmix` module is located in `psychopy.hardware`, this doesn't work (and specifying `self.exp.requirePsychopyLibs(['hardware.qmix'])` doesn't work either).

screenshot 2018-12-14 at 11 15 13
screenshot 2018-12-14 at 11 15 30
screenshot 2018-12-14 at 11 18 10

cc @shirllim @ArndalAndersen


peircej:
Love the icon. :-)
Some thoughts:

- Maybe this could be in the "custom" category, rather than "stimuli", or maybe we need a "hardware" category? Most users are going to find it strange to have a syringe in stimuli (this really highlights the need for plugins!)
- I wonder whether the preferences tab should also be to "hardware"
- Yes, I don't think you'll be able to achieve much with tests here (unless we buy a qmix for a new jenkins setup!)
- For the import, `psychopyLibs` was obviously simple for a narrow purpose but I guess we should add something like `requireImports` that operates in a similar way but writes the entire string rather than combining it with `from psychopy import ...` so you would do:
```
        self.exp.requireImports(['from psychopy.hardware import qmix'])
```

hoechenberger:
> ```
>         self.exp.requireImports(['from psychopy.hardware import qmix'])
> ```

Now this is becoming just an arbitrary line of Python code :) – Which might be good! But then, maybe, we could also rename the method that invokes it to something more generic? Like, `exp.executeOnce()` or so?

hoechenberger:
> Love the icon. :-)

Thanks, this is actually courtesy of my colleague, @ArndalAndersen :)

peircej:
> maybe, we could also rename the method that invokes it to something more generic? Like, exp.executeOnce() or so?

It is an arbitrary line of code that will be executed *very* near the top of the script, where we normally only have import statements. But yes, it could be `executeOnce()`

hoechenberger:
> * Maybe this could be in the "custom" category, rather than "stimuli", or maybe we need a "hardware" category? Most users are going to find it strange to have a syringe in stimuli (this really highlights the need for plugins!)

I had actually placed it in "Custom" originally, then moved it to "I/O", only to settle for "Stimuli" eventually. Don't really know where to put it! And adding a "Hardware" category – would that really help? Because we would also have to put all input devices, sound, and I/O there, no? :S

hoechenberger:
I've opened a separate PR (#2143) for the changes in the preferences dialog, and will rebase this branch once that PR gets merged.

peircej:
Is this ready to pull in?

hoechenberger:
I actually cherry-picked some of the changes proposed here and filed them as separate PRs; namely, #2143 and #2145, thereby separating more general “infrastructure improvements” from the addition of the pump support itself. Was hoping that, after these PRs have been merged, I could rebase the branch of _this_ PR before it gets pulled in.

hoechenberger:
@peircej I think this is good to be merged! Pump Component in Builder is now in the "Custom" category.

PR #2143, opened 2018-12-15
ENH, RF: New "hardware" prefs tab

hoechenberger:
As proposed in https://github.com/psychopy/psychopy/pull/2141#issuecomment-447287872, this PR introduces a new "hardware" section in the preferences and, likewise, a "hardware" tab in the preferences dialog. I've moved the following settings from "general" to "hardware":

   - audioLib
   - audioDriver
   - audioDevice
   - parallelPorts

screenshot 2018-12-15 at 15 19 36

### DIFF
--- a/psychopy/app/preferencesDlg.py
+++ b/psychopy/app/preferencesDlg.py
@@ -20,7 +20,7 @@ from psychopy.localization import _translate
 from pkg_resources import parse_version
 
 # this will be overridden by the size of the scrolled panel making the prefs
-dlgSize = (520, 600)
+dlgSize = (600, 600)
 
 # labels mappings for display:
 _localized = {
@@ -29,6 +29,7 @@ _localized = {
     'app': _translate('App'),
     'builder': "Builder",  # not localized
     'coder': "Coder",  # not localized
+    'hardware': _translate('Hardware'),
     'connections': _translate('Connections'),
     'keyBindings': _translate('Key bindings'),
     # pref labels:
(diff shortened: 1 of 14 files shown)


peircej:
I can't decide. Is choosing your audio library a "hardware" thing? Mostly it's an onboard chip, with a jack, built-in to all machines. I think it should stay as general

hoechenberger:
Hi Jon,

> I can't decide. Is choosing your audio library a "hardware" thing? Mostly it's an onboard chip, with a jack, built-in to all machines. I think it should stay as general

heh, so when _I_ think of hardware, the first two things that come to my mind are graphics and sound cards! I think sound interfaces fit much better into the "hardware" than into the "general" category.

But if you'd rather keep sound under "general", I could also rename "hardware" to "external hardware"..? But I thought it could be nice to have all hardware-related settings grouped in a single tab.

hoechenberger:
@m-macaskill Any _preferences_ (hurr hurr) on that one?

m-macaskill:
I think the new tab would work if there are other hardware-related items that would be added there. What would they be? Otherwise, the current items could perhaps stay where they are.

hoechenberger:
@m-macaskill Currently it would be the audio settings (3 of them) and parallel port, as shown in the screenshot. Syringe pump support would add one more entry.

peircej:
Yeah, I don't have strong opinions here. Ithink ideally we'd need to ask someone less computer-savvy to see what they think but I'll pull it in for now

PR #2145, opened 2018-12-15
NF: Add Experiment.requireImport() method

hoechenberger:
This is a more generic way to add import statements to a Builder-generated experiment. (See https://github.com/psychopy/psychopy/pull/2141#issuecomment-447287872 for a first short discussion)
Previously, only `requirePsychopyLibs()` was available, which would only `from psychopy import xxx`; whereas the new `requireImport()` can import `from` and `as` arbitrary packages/modules/names.

`requirePsychopyLibs()` was modified such that it now acts as a convenience method for `requireImport()`.

Example:
```python
Experiment.requireImport(importName='matplotlib.pyplot', importAs='plt')
# Would generate
import matplotlib.pyplot as plt


Experiment.requireImport(importFrom='psychopy.preferences', importName='prefs', importAs='p')
# Would generate
from psychopy.preferences import prefs as p
```

I first wanted to create an "execute arbitrary code here"-like method, but then thought it would be cleaner to provide a dedicated way to do generic imports. Might add the arbitrary-code thingy later though?? (as a separate method)

### DIFF
--- a/psychopy/experiment/_experiment.py
+++ b/psychopy/experiment/_experiment.py
@@ -40,7 +40,12 @@ from psychopy.localization import _translate
 import locale
 
 # standard_library.install_aliases()
-from collections import OrderedDict
+
+from collections import OrderedDict, namedtuple
+RequiredImport = namedtuple('RequiredImport',
+                            field_names=('importName',
+                                         'importFrom',
+                                         'importAs'))
 
 
 class Experiment(object):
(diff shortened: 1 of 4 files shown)


peircej:
This pull request **introduces 1 alert** and **fixes 1** when merging 8a40d900b8a57cb9bb001827f65f2e76f2989328 into 874ba1e98ec812f967d4343d89a4b7da6b416e5d - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-2d2f107557571d50ce896aec83136c232ca199d7)

**new alerts:**

* 1 for Unused import

**fixed alerts:**

* 1 for Unused import

---

Issue #1817, opened 2018-05-01
Support pathlib


sbor23:
Since Python 3.4 there's a new standard library which makes dealing with paths a lot easier and more platform agnostic.
A little introduction is [here](https://medium.com/@ageitgey/python-3-quick-tip-the-easy-way-to-deal-with-file-paths-on-windows-mac-and-linux-11a072b58d5f), docs are [here](https://docs.python.org/3/library/pathlib.html)

For example it would be neat to create an ImageStim this way:
```Python
instruction = ImageStim(win=window, image=Path(INSTR_PATH) / some / subdirectory / 'instruction.png')
```

I do realize this might remove feature parity between the Python 2 and 3 versions and is by far not a pressing issue.
For now we resort to wrapping it into a `str()` which is fine but somehow pointless.
What do you think?

peircej:
At first glance I'm not personally a fan. I usually want this:

```
instruction = ImageStim(win=window, image='images/instructions.png')
```
and it does work on all platforms

Under the new scheme I would need
```
from pathlib import Path
imagesPath = Path('images')
instruction = ImageStim(win=window, image= imagesPath/'instructions.png')
```
or
```
from pathlib import Path
instruction = ImageStim(win=window, image= Path('.')/'images'/'instructions.png')
```

I'm imagining leading a workshop and trying to explain to attendees:
  - why we need a new import
  - why we need a path instance of Path class
  - why parts of the image path are a  string, but then we break out and use `/` not as a string

I guess I don't care that the feature is there for people that like it, but I wouldn't want it to be prominent any tutorial materials etc, and I definitely wouldn't personally stop using `'images/image.png'`

hoechenberger:
@peircej I am also not a big fan of pathlib, but I don't think @EruIluvatar asks to _replace_ the current functionality of PsychoPy, but rather would like to be able to simply pass a `Path` object and see it handled correctly, while retaining current support for path strings.

So I believe the only change that would be required would be to check whether the passed filename is an instance of `os.PathLike`, and if it is, convert it to as string and proceed as usual. Which should be straightforward to implement?

EruIluvatar:
Thanks for your input.
Sounds reasonable, this might be a chance for me to submit a PR at some point.

hoechenberger:
[...]

PR #2149, opened 2018-12-17
NF: psychopy pathlib support

CRiddler:
* Added `path_to_string` function under tools/filetools which will coerce pathlib objects (or any object that utilizes the `__fspath__` dunder method) to their corresponding filesystem representation (string or bytes). Any object that does not have an `__fspath__` dunder method will be returned as is.

* Added support for pathlib to:
  * microphone
  * data/utils
  * data/base
  * visual/movie*.py
  * visual/helpers.py
  * sound/_base

Wasn't sure if tests for the new `path_to_string` function were needed. I think I covered most of the important areas where users will supply their own paths, but if you think of any others please let me know.

see #1817

### DIFF
--- a/psychopy/tools/filetools.py
+++ b/psychopy/tools/filetools.py
@@ -42,7 +42,7 @@ def toFile(filename, data):
 def fromFile(filename):
     """Load data from a pickle or JSON file.
     """
-    filename = path_to_string(filename)
+    filename = pathToString(filename)
     if filename.endswith('.psydat'):
         with open(filename, 'rb') as f:
             contents = pickle.load(f)
@@ -122,7 +122,7 @@ def openOutputFile(fileName=None, append=False, fileCollisionMethod='rename',
         A writable file handle.
 
     """
-    fileName = path_to_string(fileName)
+    fileName = pathToString(fileName)
     if (fileName is None) or (fileName == 'stdout'):
         return sys.stdout
 
@@ -170,7 +170,7 @@ def genDelimiter(fileName):
         character otherwise.
 
     """
-    fileName = path_to_string(fileName)
+    fileName = pathToString(fileName)
     if fileName.endswith(('.csv', '.CSV')):
         delim = ','
     else:
@@ -182,7 +182,7 @@ def genDelimiter(fileName):
 def genFilenameFromDelimiter(filename, delim):
     # If no known filename extension was specified, derive a one from the
     # delimiter.
-    filename = path_to_string(filename)
+    filename = pathToString(filename)
     if not filename.endswith(('.dlm', '.DLM', '.tsv', '.TSV', '.txt',
                               '.TXT', '.csv', '.CSV', '.psydat', '.npy',
                               '.json')):
@@ -245,7 +245,7 @@ class DictStorage(dict):
             self.save()
         self._deleted = True
 
-def path_to_string(filepath):
+def pathToString(filepath):
     """
     Coerces pathlib Path objects to a string (only python version 3.6+)
     any other objects passed to this function will be returned as is.
--- a/psychopy/tools/filetools.py
+++ b/psychopy/tools/filetools.py
@@ -42,6 +42,7 @@ def toFile(filename, data):
 def fromFile(filename):
     """Load data from a pickle or JSON file.
     """
+    filename = path_to_string(filename)
     if filename.endswith('.psydat'):
         with open(filename, 'rb') as f:
             contents = pickle.load(f)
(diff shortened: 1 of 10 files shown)


psychopy/tools/filetools.py
```diff
@@ -241,3 +241,22 @@ def __del__(self):
...
+        file system path that needs to be coerced into a string to
+        use by Psychopy's internals
+     :Returns:
+    
+    filepath : string or same as passed object
+        file system path coerced into a string type
+    """
+    if hasattr(filepath, "__fspath__"):
+        return filepath.__fspath__()
+    return filepath
```
hoechenberger: There's a newline missing here :)

psychopy/tools/filetools.py
```diff
@@ -241,3 +241,22 @@ def __del__(self):
             self.save()
         self._deleted = True
 
+def path_to_string(filepath):
```
hoechenberger: The function name should probably be in `mixedCase`, not `snake_case`, to fit in with other parts of PsychoPy.

psychopy/tools/filetools.py
CRiddler: Ah yes it should! I'll add another commit adding a new line and make sure that `path_to_string` is in the rest of the filetools functions.

For `tools.filetools.genFilenameFromDelimiter` should we work on ensuring that the function returns a pathlib object if a pathlib object is passed in as `filename` or should we always coerce it to return a string?

psychopy/tools/filetools.py
CRiddler: I'll add in another commit changing snake_case to camelCase, may not be able to do it tonight- but I'll have time to work locally tomorrow.

psychopy/tools/filetools.py
hoechenberger: I think ideally that would be the case; but then again it's only used internally anyway, I believe… So shouldn't really matter? I'm undecided!

hoechenberger: No strong opinion on that one. Yes, maybe simply return a string.

psychopy/tools/filetools.py
```diff
@@ -248,15 +252,18 @@ def path_to_string(filepath):
     This WILL NOT work with on Python 3.4, 3.5 since the __fspath__ dunder
     method did not exist in those verisions, however psychopy does not support
     these versions of python anyways.
-     :Parameters:
-     filepath : string or pathlib Path object
+
+    :Parameters:
+
+    filepath : string or pathlib Path object
```
hoechenberger: Could you please reference to `pathlib.Path` via `:class:pathlib.Path` please, so we get correct markup via Sphinx?

hoechenberger: Oh, and `string` should be `str`, ideally :)

hoechenberger: To cut it short, the line should probably be:
```python
filepath : str or pathlib.Path
```

CRiddler: Hmm, I'm not 100% certain either. Should we give it a go and just add in ```:class:`pathlib.Path` ``` so the parameters appear as follows:

    :Parameters:

     filepath : string or :class:`pathlib.Path`

Or do you think we should wait for Jon to weigh in to make sure its done correctly?

hoechenberger: We could also simply build the docs ourselves and check ;) But I'm currently a little busy. I think I would try without the `:class:` markup for now… :)

peircej: The change I made to confy.py in a4de786574ea3155681b6cb4db48107bbe6b619d makes it possible to use :class:`pathlib.Path` (needed the addition of intersphinx to the config file and then an instruction for how to find the docs of pathlib

peircej:
3 years later... I'm wondering if we needed `pathToString(filename)` rather than just using `str(filename)` 🤔 but in the spirit of "if it ain't broke don't fix it" I'll leave it for now

camriddell:
Can't say I disagree there- my only thought is that by checking for `__fspath__` is that we can avoid accidentally altering any non `os.pathlike` object that is passed in. (e.g. if the function receives `None`, it returns `None` instead of `"None"`. Though at this point, most dependencies probably have pathlib support now adays.

I do agree with the "if it ain't broke" approach, though I am happy to discuss if there's a push to change!

CRiddler:
Can't say I disagree there- my only thought is that by checking for `__fspath__` is that we can avoid accidentally altering any non `os.pathlike` object that is passed in. (e.g. if the function receives `None`, it returns `None` instead of `"None"`. Though at this point, most dependencies probably have pathlib support now adays.

I do agree with the "if it ain't broke" approach, though I am happy to discuss if there's a push to change!

PR #2154, opened 2018-12-17
RF/TESTS: Changes py2js conversion of params and adds py2js tests

dvbridges:
Numbers in lists with unary operators were converted to tuples, numbers in tuples with unary operators were converted to  lists. Now, elements of list/tuple with unary operators are converted to nested list, unless a binary operator exists, in which case parenthesis are used to determine order of operations.
Also, some tuples were not converted to lists, and now py2js checks for tuple in params, and converts to list.
py2js tests are added to check conversions of parameters.

### DIFF
--- a/psychopy/tests/test_psychojs/test_py2js.py
+++ b/psychopy/tests/test_psychojs/test_py2js.py
@@ -46,5 +46,8 @@ class Test_PY2JS_Compile(object):
                   '[(- 0.7), ((- 0.7) * 7)]']
 
         for idx, expr in enumerate(input):
-            assert py2js.expression2js(expr) == output[idx]
-            
+            # check whether direct match or at least a match when spaces removed
+
+            assert (py2js.expression2js(expr) == output[idx] or
+            py2js.expression2js(expr).replace(" ", "") == output[idx].replace(" ", ""))
+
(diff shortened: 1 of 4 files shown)


psychopy/tests/test_psychojs/test_py2js.py
```diff
@@ -0,0 +1,50 @@
...
+                  '[1, [2, 3]]',
+                  '[2 * [2, 3]]',
+                  '[1, (2 * 2)]',
+                  '[0.7, 0.7]',
+                  '[[- 0.7], 0.7]',
+                  '[[- 0.7], [- 0.7]]',
+                  '[(- 0.7), ((- 0.7) * 7)]']
+
+        for idx, expr in enumerate(input):
+            assert py2js.expression2js(expr) == output[idx]
```
peircej: It looks like this fails because of a simple difference in number of spaces (output in Python2.7 is `'(5 * (-2))'` whereas in Py3.6 it's `'(5 * (- 2))'`. I'd suggest:
```
# check whether direct match or at least a match when spaces removed
assert (py2js.expression2js(expr) == output[idx] or 
        py2js.expression2js(expr).replace(" ", "") == output[idx].replace(" ", ""))
```

dvbridges: Cheers. There is also a problem where ```ast.UnaryOp``` do not seem to be recognised in Python 2, so the decision of whether a node is a unary op is always false. Just trying to figure out why.

dvbridges:
*test not text 👍

Issue #2113, opened 2018-11-23
App always asks twice to overwrite a file


hoechenberger:
If I choose to save a file in Coder or Builder with a filename that already exists, I get asked twice whether I would like to overwrite the existing file: once (it seems) by the OS file save dialog, and after that through PsychoPy:

![builder 1](https://user-images.githubusercontent.com/2046265/48957841-9d9a5600-ef5b-11e8-8036-3454f9a2627b.png)
![builder 2](https://user-images.githubusercontent.com/2046265/48957842-9d9a5600-ef5b-11e8-912a-9a10ec684dde.png)
![builder 3](https://user-images.githubusercontent.com/2046265/48957843-9e32ec80-ef5b-11e8-8d03-9bf09b300dc6.png)

This is with `master` on macOS Mojave.

PR #2158, opened 2018-12-20
ENH: Use WX's built-in prompt-to-overwrite-file feature

hoechenberger:
By adding the window style `wx.FD_OVERWRITE_PROMPT` to the `wx.FileDialog`, we don't have to manually handle the "file already exists, should we overwrite?" case anymore.

This also closes GH-2113.


peircej:
Ha! Happy days! I wonder if this didn't used to exist, or whether it always existed and I just didn't know!

PR #2168, opened 2018-12-29
 ENH: Use utf-8-sig instead of utf-8 encoding

### LINKED ISSUE #2166: Potentially add UTF-8 BOM header to data files
Hi Jon:
I find this problem when I'm running a quite simple program, and I've found a way to fix it. Just change 'utf-8' to 'utf-8-sig' in experiment Handler when saving. Maybe this experience can do some help.

 [problem] In this study, I created a condition.xlsx to make Instructions change in every repeat. And I have to input the instructions in Chinese. When I run the program, everything is quite fine.Of course, Chinese characters can be displayed quit well. However, the `.csv `file in the output data folder was in a mess, especially the columns including Chinese instructions. 

[fix] And I track to psychopy.data.experiment.saveAsWideText() function. In this function, the parameter `encoding = 'utf-8'` may not be suitable for windows system, so I change it to `utf-8-sig.` Finally, the output file is correct.

Therefore, I'm wondering if this experience could help us to make psychopy3 more compatible to windows PC? Especially in Non-English countries.
p.s.
Thank you for your excellent efforts for psychology researches :+1: . And if you are interested in seeking cooperation in China, don't hesitate to contact me. And I own a Master's Degree in Applied Psychology in East China Normal University, ShangHai. :)

[some information]
Psychopy version: v3.0.0.b13
OS name: Microsoft Windows 7 enterprise
OS version: 6.1.7601 Service Pack 1 Build 7601

hoechenberger:
This allows for better support of "special" characters and improves e.g.
support of Chinese. Closes #2166.

### DIFF
--- a/psychopy/data/base.py
+++ b/psychopy/data/base.py
@@ -369,7 +369,7 @@ class _BaseTrialHandler(_ComparisonMixin):
 
     def saveAsJson(self,
                    fileName=None,
-                   encoding='utf-8-sig',
+                   encoding='utf-8',
                    fileCollisionMethod='rename'):
         """
         Serialize the object to the JSON format.
--- a/psychopy/data/base.py
+++ b/psychopy/data/base.py
@@ -151,7 +151,7 @@ class _BaseTrialHandler(_ComparisonMixin):
                    appendFile=True,
                    summarised=True,
                    fileCollisionMethod='rename',
-                   encoding='utf-8'):
+                   encoding='utf-8-sig'):
         """
         Write a text file with the data and various chosen stimulus attributes
 
@@ -190,7 +190,7 @@ class _BaseTrialHandler(_ComparisonMixin):
             :func:`~psychopy.tools.fileerrortools.handleFileCollision`
 
         encoding:
-            The encoding to use when saving a the file. Defaults to `utf-8`.
+            The encoding to use when saving a the file. Defaults to `utf-8-sig`.
 
         """
         fileName = pathToString(fileName)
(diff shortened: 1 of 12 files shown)


hoechenberger:
Needs tests (CSV condition file with non-ASCII characters in both header and body, with BOM).

hoechenberger:
Tests pass now. Just rebased and force-pushed.

hoechenberger:
Just read a part of RFC 7159, which clearly [states](https://tools.ietf.org/html/rfc7159#section-8.1) that

>    Implementations MUST NOT add a byte order mark to the beginning of a
>    JSON text.  In the interests of interoperability, implementations
>    that parse JSON texts MAY ignore the presence of a byte order mark
>    rather than treating it as an error.

So will have to undo the JSON changes.

peircej:
Looks good. I wonder whether this should only be during write operations? I'm wondering what happens with a file written using utf-8 and then decoded with utf-8-sig. Maybe it doesn't matter. Maybe if there is no BOM then utf-8-sig effectively reverts to being the same as utf-8. 

Basically, I'm just thinking aloud, with minimal understanding of sig and BOM and their impacts

hoechenberger:
> I'm wondering what happens with a file written using utf-8 and then decoded with utf-8-sig. Maybe it doesn't matter. Maybe if there is no BOM then utf-8-sig effectively reverts to being the same as utf-8.

Oh I believe I tried that and didn't make a difference. But I don't have results from systematic tests to share. We should better do that before merging.

hoechenberger:
We should check whether the CSV's we currently have in `tests/data` are BOM'd or not :)

hoechenberger:
Behavior when reading a file with ˋencoding=utf-8-sigˋ is not conditional on the actual presence of the BOM in the file:  https://stackoverflow.com/a/44573867/1944216

So should be safe to merge!

PR #2170, opened 2018-12-30
ENH/BF: New Experiment Setting – Flip after Routine is finished

### LINKED ISSUE #2169: Last Component in a Routine might have unexpected duration
The Builder creates Routine code like the following (here: with a `TextComponent` in the Flow that is to be displayed for 10 frames):

```python
while continueRoutine:
    ...
    # *text* updates
    if text.status == STARTED and frameN >= (text.frameNStart + 10):
        text.setAutoDraw(False)
    
    # check if all components have finished
    if not continueRoutine:  # a component has requested a forced-end of Routine
        break
    continueRoutine = False  # will revert to True if at least one component still running
    for thisComponent in trialComponents:
        if hasattr(thisComponent, "status") and thisComponent.status != FINISHED:
            continueRoutine = True
            break  # at least one component has not yet finished
    
    ...
    
    # refresh the screen
        win.flip()
```

What is interesting here is the last 3 lines:
```python
    # refresh the screen
        win.flip()
```
They cause the stimulus to remain on screen **even after the user-requested duration has passed**. This is **always** the case for the _last_ stimulus in _any_ Routine.

This behavior is commonly _hidden_ because as soon as the _next Routine starts_ or _the experiment ends_, we will flip again so the behavior is _not obvious_.

Now if we were to assume my experiment consisted of two Routines: the one shown above, and another one that does some preparatory calculations before it actually starts. Let's further assume these preprations of the second Routine take longer than one screen refresh. In that case, we will "skip" or "drop" one frame, and the duration of my `TextComponent` would be extended by this one frame.

It may also lead to `win.callOnFlip()` calls getting delayed or never executed (at the end of an experiment), as is the case e.g. with the `ParallelOutComponent`.

This was already briefly dscussed with @m-macaskill and @peircej a couple weeks back, and while I don't remember the details, I _do_ remember that we agreed this behavior is typically not what a user expects. (Personally, I consider it a bug, because stimulus duration might deviate from what the user requests.)

I have been working on a fix, for which I will open a PR.

hoechenberger:
As described in #2169, Builder Components might have unexpected durations if they occur at the end of a Routine (or at the end of the experiment) as we currently skip the last screen flip.

This PR addresses the issue by introducing a new Experiment Setting (a checkbox `Screen -> Flip at Routine end`) which enables flipping after the last Component in a Routine has finished. The setting is **ON** by default, meaning this is a "breaking" change, if you will. Users can, however, uncheck the setting to restore previous behavior.

Closes #2169.

### DIFF
--- a/psychopy/experiment/components/settings/__init__.py
+++ b/psychopy/experiment/components/settings/__init__.py
@@ -49,6 +49,7 @@ _localized = {'expName': _translate("Experiment name"),
               'colorSpace':  _translate("Color space"),
               'Units':  _translate("Units"),
               'blendMode':   _translate("Blend mode"),
+              'flipAfterRoutine': _translate("Flip at Routine end"),
               'Show mouse':  _translate("Show mouse"),
               'Save log file':  _translate("Save log file"),
               'Save wide csv file':
@@ -85,6 +86,7 @@ thisFolder = os.path.split(__file__)[0]
 #     def allowedVals(self, allowed):
 #         pass
 
+
 class SettingsComponent(object):
     """This component stores general info about how to run the experiment
     """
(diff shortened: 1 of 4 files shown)


hoechenberger:
Closing for now, this needs further discussion in #2169.

PR #2182, opened 2019-01-04
PKG: Use entry_points to create PsychoPy executable

hoechenberger:
This automatically creates an executable with the right interpreter, in the right place, on all platforms. It's essentially [the recommended modern-day replacement for the currently used `scripts` directive](https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts). 

This is also going to help us provide `conda` packages, as [`conda-build` also supports a similar
directive](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#python-entry-points), even for platform-independent (`noarch`) packages, and ensures the script is run via `python.app` on macOS if desired.

This PR
- gets rid of the outdated `scripts` directive (it's a disutils thing) for creating the PsychoPy executable
- renames the created executable from previously `psychopyApp` to `psychopy`, which, I believe, makes more sense. So now on Linux and macOS, one can start PsychoPy simply by typing `psychopy` after installation. (I checked the docs and believe they do not need updating)

There's still that [`psychopy_post_inst.py` script](https://github.com/psychopy/psychopy/blob/4e980da72a7307dc288c9d3b2922b36350f39b08/psychopy_post_inst.py), but I believe it's not being used anymore? Can we not just drop it?


hoechenberger:
> The post-install was useful in the days of `setup.py install` to add a start-menu item but has never been supported by pip, sadly, so may as well delete.

This is something that can apparently be solved via `conda`, too :)
https://github.com/conda/constructor/issues/38#issuecomment-427161932

peircej:
This pull request **fixes 1 alert** when merging a3bb53d71ab2f58080c28fbb478337956810058a into b75b813eb8b18ccf6083cfdb1426bdf9f12e9226 - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-d85084573d90e68c8a5c5bcb62a4db8c3dfbfa87)

**fixed alerts:**

* 1 for Except block handles &#39;BaseException&#39;

---

Issue #2189, opened 2019-01-09
pytest_namespace not available in latest pytest


hoechenberger:
`pytest` 4.1.0 [removed](https://docs.pytest.org/en/latest/deprecations.html#pytest-namespace) the previously-deprecated `pytest_namespace`, which means our tests won't run with the newest `pytest` versions as `tests/conftest.py` [makes use of the removed feature](https://github.com/psychopy/psychopy/blob/b75b813eb8b18ccf6083cfdb1426bdf9f12e9226/psychopy/tests/conftest.py#L8).

[More info on `conftest.py`](https://docs.pytest.org/en/latest/writing_plugins.html#conftest-py-local-per-directory-plugins)

Interim solution: pin `pytest` version to `<4.1.0`.

peircej:
Yes, let's fix to <4.1.0 for now

hoechenberger:
Just wanted to bump this one back into awareness :)

peircej:
Filling me with dread! ;-)

danielshub:
I am using the pandemic as a chance to update my experiments to a newer version. Is pytest still pinned at <4.1?

peircej:
I'm afraid so - I spent several days trying to update the test suite to the new mechanisms and couldn't get it to work - from my memory the issue is that the app starts up but then doesn't exit cleanly

peircej:
This is definitely fixed now. It was a world of pain but we are now using the pytest fixtures of >4.1

PR #2190, opened 2019-01-09
Add time logging to components

njvack:
Adds .startTime and .stopTime to all components. Times are relative to the start of the experiment (using core.monotonicClock).

This is *only* for the python component writers for now, and does not include any UI controls to enable/disable the audit data.


njvack:
Travis is failing on this and I don't know why; it doesn't look like it's something in my changes?

hoechenberger:
It’s not your fault, see #2189

peircej:
OK, closing this in favour of #2213
Thanks Nate. That was a fair first-approximation but the key issue is that, for visual stimuli or responses that are tied to visual stimuli, we need the timestamp to be the time of the screen refresh (which occurs _after_ the call to log it!)
The deeper solution, providing the option to sync with screen, and extended to all Components, is in #2213

PR #2195, opened 2019-01-11
TESTS: Pin pytest version to <4.1

hoechenberger:
Temporary "fix" for GH-2189.

### DIFF
--- a/conda/environment-2.7.yml
+++ b/conda/environment-2.7.yml
@@ -58,7 +58,7 @@ dependencies:
 - yaml
 - lzo
 - pytables
-- pytest
+- pytest<4.1
 - pytest-cov
 - python-sounddevice
 - wxpython!=4.0.2,!=4.0.3
--- a/conda/environment-3.6.yml
+++ b/conda/environment-3.6.yml
@@ -36,7 +36,7 @@ dependencies:
 - pyqt
 - pyserial
 - pysoundfile
-- pytest
+- pytest<4.1
 - pytest-cov
 - python-bidi
 - python-sounddevice
--- a/conda/environment-3.7.yml
+++ b/conda/environment-3.7.yml
@@ -36,7 +36,7 @@ dependencies:
 - pyqt
 - pyserial
 - pysoundfile
-- pytest
+- pytest<4.1
 - pytest-cov
 - python-bidi
 - python-sounddevice
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,5 +1,5 @@
 # these are only needed for for development/testing
 codecov
-pytest
+pytest <4.1
 sphinx
 cython


hoechenberger:
Once this has been merged into `master`, the open PRs should probably be rebased on `upstream/master` and force-pushed so that the test suit will run.

peircej:
I'll try first just doing a test restart on travis. I can't remember if it tests on the post-merge result (in which case the tests will start passing again) or whether it only tests within the branch. I think it tests the post-merge result (that seems the "right" thing to do)

hoechenberger:
> I can't remember if it tests on the post-merge result (in which case the tests will start passing again) or whether it only tests within the branch. I think it tests the post-merge result (that seems the "right" thing to do)

It tests the branch you intend to merge, it doesn't merge `master` first.

peircej:
Maybe that's changed - the tests are passing with no further merge/push on my part. In fact they restarted themselves automatically too! :-)

hoechenberger:
Wow, interesting! Good to know! But wonder _how_ that even worked… I mean, was there some magical merge performed, silently??

peircej:
I image travis does:
```
clone orig_repo
merge pull_request
execute tests
```
It would be pretty annoying if the tests pass in the branch but fail once merged with other commits (which could happen if there were no merge step), so I think this is the right thing. If the merge fails due to a conflict then github doesn't call travis at all anyway

hoechenberger:
> It would be pretty annoying if the tests pass in the branch but fail once merged with other commits (which could happen if there were no merge step), so I think this is the right thing.

I actually disagree, if I want to merge a branch, I want to get tested that branch specifically, and not have a silent merge committed.

Incidentally it turned out that restarting the tests "resolved" the issue by pure luck: conda installed an older pytest version now, probably because the dependencies of one of the packages we installed changed such that the latest pytest does not get installed anymore.

hoechenberger:
But will invesitage this further and report back :)

hoechenberger:
So Travis is def. not performing any kind of merge. This is all the `git` magic at work:
```
$ git clone --depth=50 https://github.com/psychopy/psychopy.git psychopy/psychopy
Cloning into 'psychopy/psychopy'...
remote: Enumerating objects: 4528, done.
remote: Counting objects: 100% (4528/4528), done.
remote: Compressing objects: 100% (2597/2597), done.
remote: Total 4528 (delta 2223), reused 3398 (delta 1871), pack-reused 0
Receiving objects: 100% (4528/4528), 14.01 MiB | 28.92 MiB/s, done.
Resolving deltas: 100% (2223/2223), done.
$ cd psychopy/psychopy
0.66s$ git fetch origin +refs/pull/2190/merge:
remote: Enumerating objects: 51, done.
remote: Counting objects: 100% (43/43), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 11 (delta 9), reused 6 (delta 5), pack-reused 0
Unpacking objects: 100% (11/11), done.
From https://github.com/psychopy/psychopy
 * branch            refs/pull/2190/merge -> FETCH_HEAD
$ git checkout -qf FETCH_HEAD
```

PR #2198, opened 2019-01-11
packaging and new joystick component

peircej:


dvbridges:
This should resolve the merge conflicts resulting from the updates to versionchooser.py, and 'Use version' param in settings.__init__.py

PR #2208, opened 2019-01-15
TESTS: Enable `core.rush()` functionality

hoechenberger:


peircej:
That's cool if this is a way we can insert extra jobs that need doing "on install" while user is sudo. Cool!

PR #2213, opened 2019-01-16
Adding start/stop times to data file, optionally synced with screen refresh

peircej:


peircej:
This solves the basic issue of adding timestamps of start/stop to data files (rather than log files, which has been done for years). We needed the option to get the timestamp of the screen refresh for visual-based components (which occurs _after_ the call to log it!). 
Steps were roughly:

- add `Window.timeOnFlip(object, attributeName)` so that the timestamp can be scheduled for the future
- add code to the BaseComponent to call this function
- give Components the attributes tStart, tStartRefresh, tStop, tStoprefresh all of which are set to None at the start of the Routine
- at the end of the Routine, for all Components, add these new timestamp values to the data storage according to Component settings (saveStartStop=True/False, syncScreenRefresh=True/False)

These steps have been applied to nearly all stimuli (expect code compenents?) and for visual stimuli the `syncScreenRefresh` is fixed to be read only.

The new Component Params (saveStartStop and syncScreenRefresh) show up in the new Data tab of the Component dialog, but maybe this should be called "Timing"? And/or maybe other settings should move there?

peircej:
I think this is working nicely and is generating precise times (for visual stimuli).
It has (already) made it more apparent when there are issues around stimulus duration not matching what was requested when duration is set in seconds (nFrames is exact as expected).

The issue is with the old question of whether to present one more frame or not. In the past we waited simply until t>desired but that would often lead to an overshoot. We inserted code to see if most of (75%) of a frame interval period was left, but on my macbook air I'm then seeing some undershoots. We need to test this more widely on several OSs to see if we can find a common setting (0.25*screen interval? or just a 5ms fixed interval?)

Issue #2214, opened 2019-01-16
MovieStim demo doesn't work


hsogo:
Hi,

I tried StandalonePsychoPy3_PY2-3.0.0-win32 on my Windows10 PC and got following error when running demos/coder/stimuli/MovieStim.py.
```
From cffi callback <function callback_ptr at 0x0AD7C4F0>:
Traceback (most recent call last):
  File "C:\Program Files (x86)\PsychoPy3_PY2\lib\site-packages\sounddevice.py", line 740, in callback_ptr
    return _wrap_callback(callback, data, frames, time, status)
  File "C:\Program Files (x86)\PsychoPy3_PY2\lib\site-packages\sounddevice.py", line 2516, in _wrap_callback
    callback(*args)
  File "C:\Program Files (x86)\PsychoPy3_PY2\lib\site-packages\psychopy\sound\backend_sounddevice.py", line 192, in callback
    for thisSound in self.sounds.copy():
AttributeError: 'list' object has no attribute 'copy'

```

As error message says, List object doesn't have `copy()` method in Pyhton2.  `list()` works both in Python2 and 3 ([https://stackoverflow.com/questions/2612802/how-to-clone-or-copy-a-list](url)).

```
        for thisSound in list(self.sounds):
            dat = thisSound._nextBlock()  # fetch the next block of data
```

After applying this fix, the movie was played; however, I got following error after the movie was finished.

```
Exception AttributeError: "VideoFileClip instance has no attribute 'close'" in <bound method MovieStim3.__del__ of <psychopy.visual.movie3.MovieStim3 object at 0x10C9F750>> ignored
```

In psychopy/visual/movie3.py, `__del__()` calls `_unload()` and `self._mov.close()` is called in it.  In confirmed that the content of `self._mov` was moviepy's `VideoFileClip` object.  According to [document of VideoFileClip](https://zulko.github.io/moviepy/ref/VideoClip/VideoClip.html#videofileclip), this object doesn't have `close()` method.

```
    def _unload(self):
        # remove textures from graphics card to prevent crash
        self.clearTextures()
        if self._mov is not None:
            self._mov.close()
        self._mov = None
        self._numpyFrame = None
        if self._audioStream is not None:
            self._audioStream.stop()
        self._audioStream = None
        self.status = FINISHED
    
    # (snip)

    def __del__(self):
        self._unload()
```

I'm not sure what is the aim of `self._mov.close()` here.  Is it possible that the content of `self._mov` is an object which has `close()` method?

PR #2220, opened 2019-01-18
BF: sounddevice backend raises exception in Python2

hsogo:
sounddevice backend raises exception in Python2 because `list` object in Python2 doesn't have `copy() `method.
Please see #2214 for detail.


peircej:
Looks good. Thanks for tracking it down. Yes, the key is for a copy to be produced from the original list (I would have gone with copy.copy() but your solution is neater. :-)

peircej:
I'm going to release 3.0.2 today and this will be in that release, so super fast to publish the fix this time! :-)

PR #2226, opened 2019-01-24
PKG: Update some package dependencies

hoechenberger:


peircej:
Why isn't pyparallel needed on windows?

hoechenberger:
Hi Jon,

we don't use `pyparallel` on Windows; [the parallel port interface uses `inpout` or `dlportio` on this platform.](https://github.com/psychopy/psychopy/blob/master/psychopy/parallel/__init__.py#L42)

Good remark on `wxpython`: of course, I only checked `conda` availability and forgot about PyPI. BUT I _just_ checked and there are [binaries available](https://pypi.org/project/wxPython/#files) at least for Phoenix!

peircej:
wxphoenix does have wheels for linux but not very unified:

2 flavours of gtk, and then another folder for 11 linux distribution/versions and within that they have the 3 or 4 versions of Python. So about 70 linux wheels per version of wx!

e.g.  https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-18.04/wxPython-4.0.4-cp36-cp36m-linux_x86_64.whl

PR #2234, opened 2019-01-28
BF: Allows users to control loops.

dvbridges:
This fix adds a conditional that ends loops if routine is in a loop.
Users can now use ```{loopName}.finished = true;``` to end loops as with Python scripts.


dvbridges:
Ends loop during following current routine, which is incorrect, and it would skip any remaining routines in the loop.

Issue #2238, opened 2019-01-29
Travis currently not building


hoechenberger:
I believe the problem we're having with Travis right now is the one mentioned here: https://github.com/pypa/pip/issues/6165

It's related to a recent `pip` update to 19.0:

> pip 18.1 was a lot kinder, but would generate invalid RECORD files and fail to properly handle uninstalls.

peircej:
We just noticed that one too. _Strict_ is OK, but _working_ is better! 😉 
Should we disallow 19.0? I bet they'll fix this in 19.1 because I imagine a lot of travis python testing just broke!

hoechenberger:
@peircej ;)

I just proposed a PR to (temporarily) pin pip to <19.0. I suggest we follow the discussions about this issue on their side. I wonder why we're running into this problem anyway, are we doing something wrong?

hoechenberger:
I had a look at the `RECORD` file created by `pip wheel`. There is a number of invalid entries (lines 998ff):

```
psychopy/tests/data/winScalePos_ori0_scale-1,-1_pos-0.4,0.png,sha256=GzFxAFXYZKYHbSl6-csyIL0BytopsEnlIvw9hgOAj8A,1530
psychopy/tests/data/winScalePos_ori0_scale-1,-1_pos0,0.png,sha256=2QU6uKnT4JniZWVXzH-JvM72xx5fdGNqKlmrHfDbCag,1538
psychopy/tests/data/winScalePos_ori0_scale-1,1_pos-0.4,0.png,sha256=Drh-P9SI8A_78q6-6sZpDge01hFRBHmm81O_1XO_AWs,1549
psychopy/tests/data/winScalePos_ori0_scale-1,1_pos0,0.png,sha256=JuJgP-EISHEql5-KXY4_3-6-fde5esfnlYZ8AWS0LO8,1549
psychopy/tests/data/winScalePos_ori0_scale-2,-2_pos-0.4,0.png,sha256=rxPyVKxvQWV44DpBuDVBbkoIcNwowEzgVEIu8GzhFzM,2294
psychopy/tests/data/winScalePos_ori0_scale-2,-2_pos0,0.png,sha256=5lFV8g-MdUD5aq2VfA3hTMWR_fNv48KQsf9SznDCp3Y,2328
[...]

PR #2239, opened 2019-01-29
TESTS: Pin pip to version <19.0 and ensure packages are not updated when installing a new conda package

hoechenberger:
Addresses GH-2238 and ensures `conda` will not update arbitrary packages when we try to install another one.

### DIFF
--- a/conda/environment-2.7.yml
+++ b/conda/environment-2.7.yml
@@ -32,7 +32,7 @@ dependencies:
 - openpyxl
 - pandas
 - pillow
-- pip
+- pip<19.0
 - pixman
 - prompt_toolkit
 - psutil
--- a/conda/environment-3.6.yml
+++ b/conda/environment-3.6.yml
@@ -32,7 +32,7 @@ dependencies:
 - openpyxl
 - pandas
 - pillow
-- pip
+- pip<19.0
 - psutil
 - pyopengl
 - pyosf
--- a/conda/environment-3.7.yml
+++ b/conda/environment-3.7.yml
@@ -32,7 +32,7 @@ dependencies:
 - openpyxl
 - pandas
 - pillow
-- pip
+- pip<19.0
 - psutil
 - pyopengl
 - pyosf
--- a/travis/20_install_miniconda_python.sh
+++ b/travis/20_install_miniconda_python.sh
@@ -16,5 +16,5 @@ ls -la ./conda/environment-$PYTHON_VERSION.yml
 conda env create -n psychopy-conda -f ./conda/environment-$PYTHON_VERSION.yml
 conda env list
 source activate psychopy-conda
-if [ -n "$WXPYTHON" ]; then conda install -c conda-forge wxpython=$WXPYTHON; fi
-if [ -n "$OPENPYXL" ]; then conda install -c conda-forge openpyxl=$OPENPYXL; fi
+if [ -n "$WXPYTHON" ]; then conda install --freeze-installed -c conda-forge wxpython=$WXPYTHON; fi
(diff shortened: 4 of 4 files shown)


hoechenberger:
There was a change to the `qt` `conda-forge` package 3 days ago: https://github.com/conda-forge/qt-feedstock/pull/90

peircej:
Conda and pip seem to break our tests much more often than our own code!

hoechenberger:
Hah. These filenames actually contain commas. We should rename them.

hoechenberger:
Related issue: https://github.com/pypa/wheel/issues/280

hoechenberger:
Tests at least pass locally for me on macOS.

hoechenberger:
Travis installs `gstreamer` via `apt` -- probably as a dependency of something, while the change that was made to the `qt` package in conda-forge/qt-feedstock#90 is simply an adjustment of deps -- towards `gstreamer`. I guess there's a collision of some kind now on our testing system.

hoechenberger:
Potentially related: https://github.com/conda-forge/gstreamer-feedstock/issues/23

We could try to pin to the previous `qt` version for now, or just wait a little for a new `gstreamer` release and see if that fixes things.

peircej:
Not sure what gstreamer is there for unless it's needed by something else like QT. I looked at it briefly as an alternative to opencv for MovieStim (I believe OpenSesame uses GStreamer) but decided against it because it was painful to install. So maybe the gstreamer install is a hangover from that

hoechenberger:
It seems to get installed as one of the many dependencies of `libwebkitgtk-1.0-0`.

hoechenberger:
I looked through the Travis log again. I now believe the issue is _not_ a conflict between system and conda libs (so the commit I just made is not going to fix it); rather, when installing `openpyxl` and / or `wxpython` "post-hoc" (as we do in the Python 2 environments), suddenly qt and gstreamer versions from the `defaults` channel get installed, replacing the `conda-forge` packages that were installed previously (via the environments files).

This explains why we see the problem with our Python 3 tests only – they don't do this kind of install-packages-after-initial-setup thing.

hoechenberger:
Followup:

- `wxpython` 3 is not available from `conda-forge`, therefore installing this version will fetch it from `defaults`, including `qt` and `gstreamer` deps from this channel
- installation of `openpyxl`, for some reason, retrieves a bunch of additional packages from `defaults`, including `qt` and `gstreamer`

hoechenberger:
Potentially relevant: https://github.com/conda/conda/issues/7261

hoechenberger:
@peircej I rebased to clean the commit history. This should be good to merge once Travis is done testing.

peircej:
Still amazed that any of these things are installing gstreamer! Potentially we could replace openpyxl with either xlsxwriter/xlsreader combo  (which I think pandas uses by default) or from xlwings (which also looks cool for the fact we can write python plugins from excel!)

hoechenberger:
@peircej `qt` depends on `gstreamer` (on `conda-forge`, that is); neither `openpyxl` nor `wxpython` directly depend on the former two. Not really sure why `conda` tried to update some of the (implicit?) deps when installing `openpyxl` and `wxpython`. But this is now resolved thanks to using the `--freeze-installed` option we pass to `conda`. Before this parameter was set, we would apparently sometimes end up with a mixture of `conda-forge` and `defaults` packages, which is discouraged and probably the cause of the `qt`-related issue we saw here. To prevent this from happening in the future, I have prepared #2240, which ensures all deps are getting installed from `conda-forge` only.

PR #2240, opened 2019-01-30
TESTS: Updates to conda environment setup

hoechenberger:
- slightly change how `conda` environment gets activated (to match recommended procedure for newer `conda` versions)
- ensure all dependencies listed in the environment files are installed from `conda-forge` only; this will avoid future mixups of `conda-forge` and `defaults`, which can lead to unexpected behavior as seen in #2239
- provide a list of all `conda` packages installed  in the testing environment to simplify future debugging

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,12 +36,13 @@ before_install:
 
   # Install & activate conda environment, if requested; otherwise,
   # install system Python dependencies.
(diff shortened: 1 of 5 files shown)


hoechenberger:
Errm. For whatever reason, Travis didn't run its tests on this branch… Need to figure out why.

PR #2252, opened 2019-02-04
BF: Convert NoneTypes to undefined for HTML export, and Pavlovia gui fix

dvbridges:
--For textstim, this changes any NoneTypes to 'undefined', except actual text,
where NoneTypes become empty string (as with Python).
--The Pavlovia sync project function was throwing a TypeError if the user
chose not to recreate project on sync, if the exp was pointing to a remote
that doesn't exist. Error thrown because the line was comparing Int and None
if the user aborted.

### DIFF
--- a/psychopy/app/pavlovia_ui/project.py
+++ b/psychopy/app/pavlovia_ui/project.py
@@ -434,7 +434,7 @@ def syncProject(parent, project=None, closeFrameWhenDone=False):
         # project didn't exist at Pavlovia (deleted?)
         recreatorDlg = ProjectRecreator(parent=parent, project=project)
         ok = recreatorDlg.ShowModal()
-        if ok is not None:
+        if ok > 0:
             project = recreatorDlg.project
         else:
             logging.error("Failed to recreate project to sync with")
@@ -578,5 +578,7 @@ class ProjectRecreator(wx.Dialog):
             elif choice == 2:
                 raise NotImplementedError("Deleting the local git repo is not "
                                           "yet implemented")
+            elif choice == 3:
+                return -1  # user aborted
         else:
             return -1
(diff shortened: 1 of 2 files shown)


psychopy/app/pavlovia_ui/project.py
```diff
@@ -434,7 +434,7 @@ def syncProject(parent, project=None, closeFrameWhenDone=False):
         # project didn't exist at Pavlovia (deleted?)
         recreatorDlg = ProjectRecreator(parent=parent, project=project)
         ok = recreatorDlg.ShowModal()
-        if ok > 0:
+        if ok is not None:
```
peircej: Looking at the code for `ProjectRecreator.ShowModal` we shouldn't ever be seeing `None`. We should be returning 1 for "OK" and -1 for "Cancel". I think this change to the code will cause cancel not to cancel anymore

peircej: Although if there is a `None` case coming out somehow, we should then use `if ok is not None and ok>0:`

dvbridges: Ah, yes it returns None because when OK is clicked for choice 3 ("Do nothing and abort the sync"), there is no conditional for choice 3, so it returns None. Although, there is no "cancel" button either. I will push up the change to include your fix, but also need to add a cancel button then?

peircej: Or let's return -1 from that option (as if it is the cancel button?). We could add the cancel button too I guess

dvbridges: Yes, perhaps a cancel button is faster - user can cancel without having to read options and select radio - however just added the return -1. Will look at adding a cancel now.

PR #2258, opened 2019-02-06
BF/RF: Fixes sync problem occuring when user was not actively logged in.

dvbridges:
If a project was opened, but the user had not logged in, and tried to sync,
the user was asked to create a project, even though the project existed.
The failed sync then resulted in the user being logged in and the sync
working on the next attempt. This fix attempts to log the user in on a sync
if the user is known, or in the case of multiple users where the namespace
is not indicative of the user account, uses the project name to determine
membership to known user and which known user should be logged in. Also
increases number of items listed to 100 when retrieving user projects from
Pavlovia. Fixes and refactors warning messages.

### DIFF
--- a/psychopy/projects/pavlovia.py
+++ b/psychopy/projects/pavlovia.py
@@ -384,7 +384,7 @@ class PavloviaSession:
         projs = []
         projIDs = []
         for proj in own + group:
-            if proj.id and proj.id not in projIDs:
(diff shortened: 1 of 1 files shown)


dvbridges:
See [advanced merge tools](https://git-scm.com/book/en/v2/Git-Tools-Advanced-Merging).

PR #2259, opened 2019-02-07
Many changes

peircej:


setup.cfg
```diff
@@ -54,6 +54,8 @@ install_requires =
     tables
     pyzmq
     moviepy
+    imageio
+    imageio-ffmpeg; python_version > "2.7"
```
hoechenberger: This might require adjustment for `conda` installs, at least on Travis: we currently install `ffmpeg` there, which retrieves compatible binaries from `conda-forge`.

hoechenberger: Also, shouldn't we depend on `imageio-binaries` instead of `imageio` to avoid having to download the binaries during startup?

peircej: See https://github.com/imageio/imageio/issues/318

You can't pip install `imageio-binaries` that's just a git repo storing the dlls. This new imageio-ffmpeg was a way of taking those dlls and making wheels.

I also tried simply packaging the raw dll file into the Standalone but that gave a permissions error that I couldn't easily fix. So this option of the wheel installing ffmpeg seems great if we can get it to work, but annoying that it doesn't work with conda

hoechenberger: We simply need to package `imageio-ffmpeg` for `conda-forge`. I can do that tomorrow evening. – As a workaroung, we could also install it via `pip` for the time being…

x-ref #2260

setup.cfg
```diff
@@ -55,8 +55,7 @@ install_requires =
     tables
     pyzmq
     moviepy
-    imageio
-    imageio-ffmpeg; python_version > "2.7"
+    imageio <= 2.4.1
```
hoechenberger: This new line should be removed from `master`.

peircej: Yup. Done in my master

PR #2269, opened 2019-02-10
BF: "Save onset/offset times" doesn't work

hsogo:
Save onset/offset times are output even though "Save onset/offset times" in the Properties dialog are unchecked.

### DIFF
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -144,14 +144,14 @@ class BaseComponent(object):
         """Write the code that will be called at the end of
         a routine (e.g. to save data)
         """
-        if 'saveStartStop' in self.params and self.params['saveStartStop']:
+        if 'saveStartStop' in self.params and self.params['saveStartStop'].val:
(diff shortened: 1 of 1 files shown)


peircej:
Thanks. This is fixed more generally by 301be457e0993f069dc706b5990dc4d3389b356b
meaning that this fix isn't technically needed but doesn't hurt (belt as well as braces)

peircej:
By the way, I only inserted that on seeing the issue you found here, so thanks! :-)

hsogo:
Ah, that's nice!
Actually, I often forget to add `.val` when coding components. :sweat_smile:

peircej:
Yes, me too! It happened so many times and then I discovered about `__eq__` and `__ne__` methods a while ago, so we could do comparisons like `param==somethingElse` and it did the right thing. Your post made me realise that we need the boolean option too. And now I'm wondering whether we should implement some of the others as well _before_ we get bitten again!

https://docs.python.org/2/reference/datamodel.html#basic-customization

PR #2276, opened 2019-02-11
BF: Add newline arg when opening py script in Coder.

dvbridges:
Some users had issues with newline args when working on code x-platform.
This fix allows specific newline convention as per prefs to be used when
opening files. Note, None is used for same as in file" because None enables
 universal newlines, i.e. interprets unix, Windows and old Mac EOLs.

Would perhaps be good to update the Dos pref to Win.


dvbridges:
Closing this pull request for now, because we need to deal with cases of mixed EOL.

PR #2278, opened 2019-02-12
BF: converts EOL on opening and compiling scripts to Coder.

dvbridges:
Should fix issue of mixed EOLs causing lines of code to disappear, see [Discourse](https://discourse.psychopy.org/t/bug-disappearing-line-breaks-in-code-cells-in-builder/6607/9).
Also, adds Legacy Mac line endings as a preference.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -53,6 +53,7 @@ from psychopy.app import pavlovia_ui
 from psychopy.projects import pavlovia
 
 from psychopy.scripts import psyexpCompile
+from psychopy.app.coder import BaseCodeEditor
 
 canvasColor = [200, 200, 200]  # in prefs? ;-)
 routineTimeColor = wx.Colour(50, 100, 200, 200)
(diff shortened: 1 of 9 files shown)


peircej:
Just want to check a couple of things before pulling this in.

As we have the current preference set to "keep" I'm guessing that we currently wouldn't change the EOLs and so the problem wouldn't be fixed for Ben until he changes his preference to a particular line ending?

I think the original post was to do with code being pasted into the script (and/or into a code component?) so we might need to force a check just after any paste operation?

dvbridges:
Yes, this would not take effect unless you changed your preferences, so not going to cover the pasting problem unless the file is opened or compiled.  I like the idea of checking the EOLs on paste, will have a look now.

dvbridges:
Just looking at whether to convert the getEOL method to a classmethod, so I can call it without constructing a BaseCodeEditor object

peircej:
This pull request **introduces 1 alert** and **fixes 2** when merging 2401cb9aaf6d4b291ac09c266f91ac8cdbf59e31 into 4a576ebb2684277677aebdb7468fa1f1e389f499 - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-961589f2081e7146cc73dcec3a0073719c22ab29)

**new alerts:**

* 1 for Unused import

**fixed alerts:**

* 2 for Except block handles &#39;BaseException&#39;

---

PR #2281, opened 2019-02-13
Packaging fixes for 3.0.4

peircej:


dvbridges:
@peircej  would you be able to hold off on the build until I make a new pull req for slider?

peircej:
I've actually made the build already. Make the PR anyway, and we can decide whether it's worth rebuilding.

dvbridges:
Ok, the problem is:

```
if (slider.getRating() !== 'undefined' && slider.status === PsychoJS.Status.STARTED) {
    continueRoutine = false; }
```

Is always true, because ```slider.getRating() !== 'undefined'``` is always true, because the getRating() does not return a string of 'undefined' when there is no rating, but an actual undefined property.

PR #2282, opened 2019-02-14
BF: slider.getRating() returns undefined, not 'undefined'

dvbridges:
The fix changed the conditional to end the routine, as currently it
always ends routine without a response because 'undefined' !== undefined.
Note, you cannot see the actual marker if you ask the Slider to end the
routine. The routine will end without a marker being drawn.

### DIFF
--- a/psychopy/experiment/components/slider/__init__.py
+++ b/psychopy/experiment/components/slider/__init__.py
@@ -263,7 +263,7 @@ class SliderComponent(BaseVisualComponent):
         forceEnd = self.params['forceEndRoutine'].val
         if forceEnd:
             code = ("\n// Check %(name)s for response to end routine\n"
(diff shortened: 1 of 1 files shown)


dvbridges:
This also highlights the issue of inconsistent use of 'undefined' and undefined that would be good to address e.g., 

```
function endLoopIteration(thisTrial) {
  // ------Prepare for next entry------
  return function () {
    if (typeof thisTrial === 'undefined' || !('isTrials' in thisTrial) || thisTrial.isTrials) {
      psychoJS.experiment.nextEntry();
    }
  return Scheduler.Event.NEXT;
  };
}

```

vs 
```
 text = new visual.TextStim({
    win: psychoJS.window,
    name: 'text',
    text: 'Any text\n\nincluding line breaks',
    font: 'Arial',
    pos: [0, 0], height: 0.1,  wrapWidth: undefined, ori: 0,
    color: new util.Color('white'),  opacity: 1,
    depth: -1.0 
  });
```

PR #2288, opened 2019-02-18
BF: Reverts all commits from #2278 and #2279

dvbridges:
The original issue from Discourse appeared to be a problem with interpretting
mixed EOLs in Coder and code components. The original fix
attempted to convert EOLs to those defined in the preferences to provide
uniform EOLS. However, this did not address the problem described in
the original post.

PR #2289, opened 2019-02-19
BF: Fixes how EOL are converted in XML Builder file

dvbridges:
Previous problems with EOLS were found when legacy Mac EOLs were used in
code components, because PsychoPy could not convert those EOLs into the
correct XML format. This fix now also checks for legacy Mac EOLS in code
components only, but in doing so, covers three EOL styles (\\r, \\n and \\r\\n for legacy Mac, Unix
and Windows). The previous fix for controlling and converting EOLs using
preferences may still be useful, because it creates uniform EOLs in the
PsychoPy when it is compiled from Builder.

The prefs for newline convention were causing a global reconfig of EOLS
in components that use CodeBox class e.g., text and RatingScale.
This fix also keeps the EOLs as standard UNIX in components that
use CodeBox, but are not code components. This way, there are no
differences presenting text locally or online that may be caused by
newline prefs.


dvbridges:
Closing in place of #2288

PR #2292, opened 2019-02-20
ENH: update translation

hsogo:
Add \_translate() to untranslated dialog strings.


peircej:
Oo, I was about to release 3.0.5, but I guess I should rebuild with these included! Great!

PR #2293, opened 2019-02-20
BF: Code component was not using Paste method from codeEditorBase class

dvbridges:
This fix adds a keycheck when using paste shortcut and redirects
to the codeEditorBase paste method. It also enables the
context menus for both Coder and Code Component to be directed
to the Paste function in the codeEditorBase.

Edit: The context menu **DOES** work when switching the tabs - the tab just needs
to be active. The keyboard shortcut is independent
of the context menu fixes, and can be cherry picked if the menu is not
wanted.

### DIFF
--- a/psychopy/app/coder/codeEditorBase.py
+++ b/psychopy/app/coder/codeEditorBase.py
@@ -85,7 +85,7 @@ class BaseCodeEditor(wx.stc.StyledTextCtrl):
                               'Cut': {'id': None, 'item': '', 'method': self.onCut, 'enabled': self.CanCut},
                               'Copy': {'id': None, 'item': '', 'method': self.onCopy, 'enabled': self.CanCopy},
                               'Paste': {'id': None, 'item': '', 'method': self.onPaste, 'enabled': self.CanPaste},
-                              'Delete': {'id': None, 'item': '', 'method': self.onDelete,'enabled': self.CanCopy},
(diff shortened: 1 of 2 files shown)


psychopy/app/coder/codeEditorBase.py
```diff
@@ -72,6 +72,83 @@ def __init__(self, parent, ID, pos, size, style):
...
+
+    def OnContextMenu(self, event):
+        """Sets the context menu for components using base code editor"""
+        menuDict = {'Undo': {'id': None, '': '', 'method': self.onUndo, 'enabled': self.CanUndo()},
+                    'Redo': {'id': None, 'item': '', 'method': self.onRedo, 'enabled': self.CanRedo()},
+                    'Cut': {'id': None, 'item': '', 'method': self.onCut, 'enabled': self.CanCut()},
+                    'Copy': {'id': None, 'item': '', 'method': self.onCopy, 'enabled': self.CanCopy()},
+                    'Paste': {'id': None, 'item': '', 'method': self.onPaste, 'enabled': self.CanPaste()},
+                    'Delete': {'id': None, 'item': '', 'method': self.onDelete, 'enabled': True},
+                    'Select All': {'id': None, 'item': '', 'method': self.onSelectAll, 'enabled': True},}
```
dvbridges: I will change this, as a wx.NewId is created every time the menu is created.

PR #2296, opened 2019-02-21
Remove pygame requirement

cbrnr:
I noticed that the most recent version on PyPI still requires lots of packages, including the deprecated `pygame`. I've fixed that in this PR, but maybe it would be a good idea to do some more cleaning (especially since the [install docs now mention a minimal setup](https://www.psychopy.org/installation.html), which currently cannot be installed because all requirements get pulled in). Let me know if you want me to do that.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,6 @@ install_requires =
     pandas
     pillow
     pyglet
-    pygame
     pyopengl
     soundfile
     sounddevice


peircej:
I can't decide whether this is right to include as a dependency or not. The thing is that some people still use it and to run some historical experiments it is needed. But for most use cases it isn't We could say the same about several other libs (e.g. most people don't need mspack_numpy but when they run the study that needs it they're surprised).

Happy to take a steer here (@hoechenberger ?). In a sense, as long as the standalone application does include these libs it won't be the end of the world if the pip/conda installations don't (except that I might forget to include them in my installations!)

cbrnr:
I think your idea to include everything in the standalone release but only the hard dependencies in the PyPI release is great. I really think that if I pip install a package, it should not pull in unnecessary dependencies, but if people want to run historical experiments they can download specific historical standalone releases (which I think is common practice anyway).

According to the [docs](https://www.psychopy.org/installation.html), these are the hard dependencies: `numpy`, `scipy`, `matplotlib`, `pandas`, `pyopengl`, `pyglet`, `pillow`, `moviepy`, `lxml`, `openpyxl`, and `configobj`. From these, it looks like `pyopengl` is not needed anymore (since 1.74.00). Also, I didn't find any imports or usages of `pillow` in the source code. Both `openpyxl` and `configobj` are also not required, because imports are safe-guarded with try/except clauses (and PsychoPy has its own `configobj` implementation if needed) so people get a meaningful message if they really use this functionality. Furthermore, `lxml` is only used in one test but not anywhere in the source.

Therefore, I think these would be the final hard dependencies:
numpy, scipy, matplotlib, pandas, pyglet, moviepy

cbrnr:
There are probably other hard dependencies that are not mentioned in the docs, e.g. `sounddevice` or packages required for `iohub`. In the latter case, a good solution would be to wrap specific imports in try/except blocks so that only people that really use `iohub` need to install these dependencies.

cbrnr:
And `pip install wxPython` should also work out of the box now (no need for a manual install).

peircej:
Very few things are actually hard dependencies. It nearly always depends on your study/usage. e.g.: 
- if you don't use arabic text you don't need arabic_reshaper
- if you don't need online studies then you don't need esprima
- If you don't need a dialog box then wxpython and pyqt aren't needed
- If you don't play sounds then you don't need pyo, sounddevice or pygame. If you do then you only need one of them (but we don't know which will perform best)
- Similarly, we only really need one of xlread and openpyxl (and neither if you always save/load csv files) but it's much easier to provide both from the outset
- ...

Basically, _most_ of the entries in https://github.com/psychopy/psychopy/blob/master/setup.cfg are only needed for some specific use case, but most are needed for many experiments. For the user I think it's nicer to install all the things you might _reasonably_ need.

cbrnr:
I agree if you are talking about the user of the stand-alone package. In this case, I think it is good to include everything.

However, for users who `pip install psychopy`, it seems unusual and unnecessary to automatically install all packages that provide extra functionality for special cases/studies (but are not essential for `psychopy` to work). In that case, I think it is a good idea to keep the list of hard dependencies as short as possible and use nested imports (i.e. within a function or try/except blocks) for packages that provide extra functionality.

A good starting point would be to come up with a list of hard dependencies, and then make sure that the source code directly imports only these packages. For example, it might be reasonable to assume that presenting audio is an essential feature, so `sounddevice` should be included as a hard dependency. On the other hand, it might also be reasonable to decide that audio is an optional feature so `sounddevice` should be optional (and therefore not get installed automatically).

In any case, it seems like the PsychoPy source already contains hard dependencies (direct imports) and soft dependencies (nested), but this is not reflected in the PyPI package, which requires all hard and soft dependencies.

hoechenberger:
> However, for users who `pip install psychopy`, it seems unusual and unnecessary to automatically install all packages that provide extra functionality for special cases/studies (but are not essential for `psychopy` to work). In that case, I think it is a good idea to keep the list of hard dependencies as short as possible and use nested imports (i.e. within a function or try/except blocks) for packages that provide extra functionality.

I believe we discussed the introduction of `extras_require` settings in `setup.py` before and the consensus was that we didn't want it at that time. The idea was to provide users with a fully-featured PsychoPy install out-of-the-box when they `pip install` it.

I really don't mind either approach. I've learned to bite the bullet and every PsychpPy installation on my system gets its dedicated `conda` / `virtualenv` anyway, so I don't really care whether it pulls in 5 or 100 deps. (That's not quite true, actually. My actual approach is: I only install the bare number of deps I really need, and then `pip install --no-deps psychopy`. So now, considering this, having `extras_require` would make _my_ life easier)

cbrnr:
@hoechenberger I think having `extras_require` would make the life of everyone who installs via `pip` much much easier. Let me know if I should modify this PR accordingly.

peircej:
> I think having extras_require would make the life of everyone who installs via pip much much easier

That's definitely an overstatement. My sense is that a lot of (probably most) users don't care about the time/space of a large install but want everything to "just work". 

If we stick to the list that you're considering hard dependencies nearly all the demos won't run. I think most users would find that annoying. 

> ... Also, I didn't find any imports or usages of pillow in the source code. 

This is a nice example. Pillow is technically a soft dependency. I could write a script that never uses it (although the app certainly does and so does ImageStim). How many seconds does it take to install pillow using pip? Conversely, how many seconds does it take a user to debug the message "ImportError: No module named Image" and work out that this was because they need to back to the terminal and do `pip install pillow` (not `pip install Image`

As another for-instance, PyQt is not a hard dependency. Some people don't need ever to open a dialog box. Those people are rare and I would not support suggesting that PyQt is soft even though some people might never need it. When they find that a(nother) demo has failed to run, and spent time working out how to install it (`pip install pyqt5` not `pip install pyqt` nor `pip install PyQt5`...) they get annoyed that things don't "just work". 

In short, I'm willing to accept dropping something like pygame, but definitely not the more drastic drops that I think you're asking for. I think most people prefer "batteries included".

peircej:
I figured this was a lot of discussion for a 1-line change! :-) Merged

cbrnr:
Thanks @peircej, but I think we're not done with the discussion yes 😄 - at least I have more to say. Should I open a new issue (because no one reads closed PRs)?

peircej:
I suggest discussing it in the developers section of Discourse.

hoechenberger:
I just uninstalled `pygame` and cannot even start the PsychoPy app (Builder) after that. It seems it's trying to initialize the `SoundComponent` and then fails hard (I don't have any other audio lib installed). – Seems like a bug in the Builder / Component, but just saying that dropping `pygame` may have weird side-effects.

cbrnr:
Which platform? I tried this on macOS and everything works after uninstalling `pygame`.

hoechenberger:
It's not related to removing `pygame` _per se_, it's about PsychoPy being unable to load the Builder GUI when no audio lib is present, which should probably be considered a bug:

To reproduce with the minimal number of deps:
```
$ conda create --yes -n psychopy-no-audio -c conda-forge python=3 wxpython configobj \
                                                         future numpy json_tricks python-gitlab \
                                                         pillow astunparse esprima-python pyglet
$ conda activate psychopy-no-audio
$ pip install --no-deps git+https://github.com/psychopy/psychopy.git
$ rm -rf ~/.psychopy3
```

Then, running `psychopy` (to start the app) will fail:
```python
$ psychopy
2.8196 	WARNING 	sounddevice audio lib was requested but not loaded: sounddevice not working
2.8218 	WARNING 	pyo audio lib was requested but not loaded: ModuleNotFoundError("No module named 'pyo'")
2.8225 	WARNING 	pygame audio lib was requested but not loaded: ModuleNotFoundError("No module named 'pygame'")
2019-03-14 09:06:36.794 python[7394:514097] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to (null)
09:06:37 AM: Debug: Adding duplicate image handler for 'Windows bitmap file'
09:06:37 AM: Debug: Adding duplicate image handler for 'Windows bitmap file'
09:06:37 AM: Debug: Adding duplicate animation handler for '1' type
09:06:37 AM: Debug: Adding duplicate animation handler for '2' type
09:06:37 AM: Debug: Adding duplicate image handler for 'Windows bitmap file'
09:06:37 AM: Debug: Adding duplicate animation handler for '1' type
09:06:37 AM: Debug: Adding duplicate animation handler for '2' type
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/psychopyApp.py", line 104, in <module>
    main()
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/psychopyApp.py", line 98, in main
    start_app()
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/psychopyApp.py", line 28, in start_app
    app = PsychoPyApp(0, showSplash=showSplash)
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/_psychopyApp.py", line 180, in __init__
    self.onInit(testMode=testMode, **kwargs)
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/_psychopyApp.py", line 321, in onInit
    self.showBuilder(fileList=exps)
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/_psychopyApp.py", line 505, in showBuilder
    self.newBuilderFrame()
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/_psychopyApp.py", line 491, in newBuilderFrame
    fileName=fileName, app=self)
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/builder/builder.py", line 1069, in __init__
    self.componentButtons = ComponentsPanel(self)
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/app/builder/builder.py", line 693, in __init__
    self.app.prefs.builder['componentsFolders'])
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/experiment/components/__init__.py", line 58, in getAllComponents
    components = getComponents(fetchIcons=fetchIcons)  # get the built-ins
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/experiment/components/__init__.py", line 149, in getComponents
    module = import_module(explicit_rel_path, package=pkg)
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/experiment/components/sound/__init__.py", line 15, in <module>
    from psychopy.sound._base import knownNoteNames
  File "/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/lib/python3.7/site-packages/psychopy/sound/__init__.py", line 107, in <module>
    .format(prefs.hardware['audioLib']))
psychopy.exceptions.DependencyError: No sound libs could be loaded. Tried: ['sounddevice', 'pyo', 'pygame']
Check whether the necessary sound libs are installed
Error in atexit._run_exitfuncs:
wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed at /private/tmp/pip-req-build-q590exde/ext/wxWidgets/src/common/wincmn.cpp(478) in ~wxWindowBase(): any pushed event handlers must have been removed
/Users/hoechenberger/miniconda3/envs/psychopy-no-audio/bin/pythonw: line 3:  7394 Segmentation fault: 11  /Users/hoechenberger/miniconda3/envs/psychopy-no-audio/python.app/Contents/MacOS/python "$@"
```

cbrnr:
`sounddevice` is still a required dependency and therefore this error can't occur if you install via `pip` even after `pygame` was removed.

hoechenberger:
Yes, exactly. So this is a different issue: Builder should not crash when not audio lib was found. Not directly related to `pygame`.

peircej:
Yes, I think even in the absence of any installed sound lib we should avoid crashing! I suspect the issue is that sound component now provides the option of "which sound card would you like to use." That requires us to load pyo or sounddevice, so we try to load sound to get to them. But if there are no sound libs we should handle it more gracefully and leave the device choice as 'auto'

mirh:
I'm also not clear on why python-gitlab is required to even just start the GUI. 

p.s. wxPython (and also pyqt5 if you are on 32bit) still requires manual "system-specific" installation in linux

PR #2300, opened 2019-02-22
BF: Update control is contaminated with translated string

hsogo:
I noticed that Builder fails to save experiment if it includes Static component on my Japanese environment.
Error message is as follows:

```
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\builder.py", line 1755, in fileSaveAs
    self.fileSave(event=None, filename=newPath)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\builder.py", line 1714, in fileSave
    filename = self.exp.saveToXML(filename)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\experiment\_experiment.py", line 277, in saveToXML
    parent=componentNode, param=param, name=paramName)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\experiment\_experiment.py", line 333, in _setXMLparam
    thisChild.set('updates', "{}".format(param.updates))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u306e' in position 7: ordinal not in range(128)
```
Replacing `"{}"`  in line 333 of \_experiment.py with `u"{}"`, I could save the experiment without error.  Looking into created .pyexp file, I found that value of `updates` was not original (i.e. `"set during: trial_2.ISI"` but translated string.

```
      <MovieComponent name="movie_2">
        <Param name="movie" updates="trial_2のISIの間に更新" val="2.mp4" valType="str"/>
```
Update control in Property dialog is created at lines 230-242 of psychopy/app/builder/dialogs/\_\_init\_\_.py.

```
            # updates = display-only version of allowed updates
            updateLabels = [_localized[upd] for upd in param.allowedUpdates]
            # allowedUpdates = extend version of allowed updates that includes
            # "set during:static period"
            allowedUpdates = copy.copy(param.allowedUpdates)
            for routineName, routine in list(self.exp.routines.items()):
                for static in routine.getStatics():
                    msg = _translate(
                        "set during: %(routineName)s.%(staticName)s")
                    vals = {'routineName': routineName,
                            'staticName': static.params['name']}
                    updateLabels.append(msg % vals)
                    allowedUpdates.append(msg % vals)
```

Role of `updateLabels` is to hold translated labels, while that of `allowedUpdates` is to hold original values.  However, translated string is appended to `allowedUpdates`.  Fixing this, Builder can save experiment with Static component in Japanese environment.


hsogo:
Ouch! my fix prevent poedit from finding this message!
I'll commit fix ASAP.

peircej:
If I understand correctly this is translating the name of the Static Component, but I'm not sure that's wise. Shouldn't we use:
```
msg = _translate("set during:")+str(routineName)+str(staticName)
```
Routines and Components really shouldn't have non-ascii names and it could be confusing for the user to see it sometimes translated and sometimes not?

hsogo:
Exactly to say, "update" parameter must include `'during:'` in order to make Static component work properly.  Codes for Static Component are generated in lines 596-617 of psychopy/experiment/\_experiment.py.  Line 604 is critically important.:

```
        # for each component that uses a Static for updates, we need to set
        # that
        for thisRoutine in list(self.routines.values()):
            for thisComp in thisRoutine:
                for thisParamName in thisComp.params:
                    thisParam = thisComp.params[thisParamName]
                    if thisParamName == 'advancedParams':
                        continue  # advanced isn't a normal param
                    elif thisParam.updates and "during:" in thisParam.updates:  # <- Line 604
                        # remove the part that says 'during'
                        updates = thisParam.updates.split(': ')[1]
                        routine, static = updates.split('.')
                        if routine not in self.routines:
                            msg = ("%s was set to update during %s Static "
                                   "Component, but that component no longer "
                                   "exists")
                            logging.warning(msg % (thisParamName, static))
                        else:
                            self.routines[routine].getComponentFromName(  # <- this line must be called to generate codes for Static Component!
                                static).addComponentUpdate(
                                thisRoutine.params['name'],
                                thisComp.params['name'], thisParamName)
```

Because values in `allowedUpdates` in psychopy/app/builder/dialogs/\_\_init\_\_.py are assigned to `thisParam`, they must include  `'during:'`. So, they must be untranslated strings.  Values in `allowedUpdates` are used only for internal processing.  From users, only values in `updateLabels` are visible.  I think users wouldn't be confused.

In addition, if we allow using translated strings here, Experiments saved in a certain locale may be impossible to load in other locale.  For example, .psyexp file shown in my first post (see below) cannot be loaded in English locale.  I think we should avoid such a situation.

```
      <MovieComponent name="movie_2">
        <Param name="movie" updates="trial_2のISIの間に更新" val="2.mp4" valType="str"/>

```

hsogo:
I'm afraid I might have misunderstood...

`_translate("set during:")+str(routineName)+str(staticName)` might be OK, but it will make translation difficult because object of verb is placed before the verb in Japanese.  I hope this string is not changed if possible.

As for non-ASCII characters for Routines and Components, it wouldn't work with Python2 because they are used as variable names.  Python2 doesn't allow non-ASCII variable names.

Anyway, as described at lines 49-74 of psychopy/app/builder/dialogs/\_\_init\_\_.py, `fieldName`s value has to be en_US in current implementation.  The same applies to values in `allowedUpdates`.  We have to do refactoring if we want to use non-ASCII characters in `allowedUpdates`.

peircej:
Yes, exatly. Unicode should not be allowed as a variable name, so we shouldn't translate routineName or staticName. Also, given that these could be any arbitrary name set by the user I don't think we _could_ translate them.

So we can either translate just the preceding label ("set during") or not translate this parameter at all if that's too hard to do. But the code 
```msg = _translate("set during: %(routineName)s.%(staticName)s")```
seems not good

**Comment by hsogo** [comment]
I see.  I think that `_translate("set during:")` is better than untranslated.  I'll fix it.

**Comment by peircej** [comment]
Yes, agreed. `"{}.{}".format(routineName,staticName)` is better.

**Comment by hsogo** [comment]
I've fixed codes.

I looked into commit history and found that  1e67b15eb4042c298cf070eb25ff24d6f5fe48c8 is the origin of this issue.  Before this commit, the code was as follows.  Raw (en_US) string was appended to `allowedUpdates`, while translated string was appended to `updateLabels`. I'm sorry I hadn't noticed this issue for years!

```
            # allowedUpdates = extend version of allowed updates that includes
            # "set during:static period"
            allowedUpdates = copy.copy(param.allowedUpdates)
            for routineName, routine in self.exp.routines.items():
                for static in routine.getStatics():
                    updateLabels.append(_translate("set during: %(routineName)s.%(staticName)s") % {
                                        'routineName': routineName, 'staticName': static.params['name']})
                    allowedUpdates.append("set during: %(routineName)s.%(staticName)s" % {
                                          'routineName': routineName, 'staticName': static.params['name']})
```

peircej:
Cool. Glad it worked. Yes, I think static intervals aren't used much, so this probably didn't hit many people.

PR #2304, opened 2019-02-25
ENH: Added Channel field to Microphone properties

bakablah:
This is so that the experiment builder can auto generate 'chnl' attribute for AdvAudioCapture, as not all devices share the same default '0' for input device.

**Detailed explaination of fix in below thread**
https://discourse.psychopy.org/t/potential-improvements-to-microphone-component-of-experiment-builder/6814

### DIFF
--- a/psychopy/experiment/components/microphone/__init__.py
+++ b/psychopy/experiment/components/microphone/__init__.py
@@ -19,7 +19,7 @@ iconFile = path.join(thisFolder, 'microphone.png')
 tooltip = _translate('Microphone: basic sound capture (fixed onset & '
                      'duration), okay for spoken words')
 
-_localized = {'stereo': _translate('Stereo')}
+_localized = {'stereo': _translate('Stereo'),'channel': _translate('Channel')}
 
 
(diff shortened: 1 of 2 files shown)


peircej:
To fix the components template did you do it manually, or did you know to run it with:
```bash
cd psychopy/tests/test_app/test_builder/
python genComponentsTemplate.py --out
```

cheers,
Jon

bakablah:
Jon,
I did it manually, was I not meant to do that?

peircej:
Well, you could I guess, but that's why there's still an error (your manual method isn't giving quite the same results). Using the commands above should mean the build and test get the same answer

bakablah:
Hey Jon,
I ran the genComponentsTemplate.py, diff'ed the resultant componsTemplate.txt against the one in master branch, the changes are massive (at least 48 sections are different), not sure if it's because I'm running it in Windows?
In any case, I'm not going to check in the generated version, seems a bit risky to me.

Looks like CI is happy now...

peircej:
OK, yes, maybe to do with windows line endings or something
cheers

PR #2306, opened 2019-02-25
BF: Fixes file and exp name mismatch causing online studies to fail

### LINKED ISSUE #2305: Mismatching filenames and experiment names cause online studies to fail
When users change the experiment name (e.g., to ```myExp```) in experiment settings to one that is different from the original filename (e.g., ```myOriginalName.psyexp```), the HTML file calls a JavaScript file by the experiment name defined in the  experiment settings e.g., ```myExp.js```, however the actual JavaScript file has the same name as the psyexp file (```myOriginalName.js```).

For solutions:
1) We could instead populate HTML and JS outputs using the psyexp filename
2) Write the JS output scripts using filenames created from experiment settings exp. name
3) Make experiment name field immutable
4) Warn the participant of what is happening, but do nothing

dvbridges:
Fixes #2305

### DIFF
--- a/psychopy/experiment/components/settings/__init__.py
+++ b/psychopy/experiment/components/settings/__init__.py
@@ -495,11 +495,11 @@ class SettingsComponent(object):
     def writeInitCodeJS(self, buff, version, localDateTime, modular=True):
         # create resources folder
         self.prepareResourcesJS()
-
+        jsFilename = os.path.basename(os.path.splitext(self.exp.filename)[0])
         # html header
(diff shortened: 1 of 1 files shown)


dvbridges:
@peircej , just wondering if you plan on pulling this into the next release?

peircej:
Actually I thought we *had* released it. Did I miss it last time?

dvbridges:
Ah yes, 3.0.6 - I should have checked the remote repo for the latest version, not my local standalone. Apologies!

PR #2310, opened 2019-02-27
BF: add spacebar to the valid keys

superabe:
I found the spacebar is missing 
```
  File "stimulus_show.py", line 40, in <module>
    event.globalKeys.add(key='space', func=pause)
  File "/usr/local/lib/python3.6/site-packages/psychopy/event.py", line 1090, in add
    raise ValueError('Unknown key specified: %s' % key)
ValueError: Unknown key specified: space

```

And adding the spacebar to the valid keys can fix this.

### DIFF
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -990,7 +990,7 @@ class _GlobalEventKeys(MutableMapping):
 
     _valid_keys = set(string.ascii_lowercase + string.digits
                       + string.punctuation + ' \t')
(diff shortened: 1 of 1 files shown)


superabe:
Sure! Already removed.

hoechenberger:
Thanks!

peircej:
For future reference please use a commit message with one of BF: FF:... as listed in

Without them we can't readily filter what should should not be included in bug-fix releases.

thanks for the contribution

Issue #2317, opened 2019-02-28
iohub fails on large packet size on python3


ariekahn:
Ran into the following crash, though not consistently:

Python: 3.7.2

It looks like iohub still contains a few references to `xrange`, which crashes on python 3.

```bash
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
6.2647 	WARNING 	User requested fullscreen with size [1280  800], but screen is actually [1920, 1080]. Using actual size
=============================
Error trying to send data to experiment process:
max_pkt_sz: 32748
reply_data_sz: 39589
pkt_cnt: 2
packet index, byte index: -1-1
Traceback (most recent call last):
  File "/Users/bassettlab/.local/share/virtualenvs/task/bin/twoday_task_fmri", line 11, in <module>
    load_entry_point('fmri-network-learning', 'console_scripts', 'twoday_task_fmri')()
  File "/Users/bassettlab/task/experiment_twoday.py", line 673, in main
    run_experiment(args.subject_id, 'scan1', args.fullscreen, args.in_scanner)
  File "/Users/bassettlab/task/experiment.py", line 586, in run_experiment
    score = block.run()
  File "/Users/bassettlab/task/models.py", line 548, in run
    self.keyboard.getKeys()
  File "/Users/bassettlab/.local/share/virtualenvs/task/lib/python3.7/site-packages/psychopy/iohub/client/keyboard.py", line 285, in getKeys
    self._syncDeviceState()
  File "/Users/bassettlab/.local/share/virtualenvs/task/lib/python3.7/site-packages/psychopy/iohub/client/keyboard.py", line 190, in _syncDeviceState
    self._reporting = kb_state.get('reporting_events')
AttributeError: 'int' object has no attribute 'get'
Traceback (most recent call last):
  File "/Users/bassettlab/.local/share/virtualenvs/task/lib/python3.7/site-packages/psychopy/iohub/server.py", line 329, in sendResponse
    for p in xrange(pkt_cnt - 1):
NameError: name 'xrange' is not defined
=============================
ioHub Server Process Completed With Code:  0
```

hoechenberger:
That should actually be trivial to fix. I can look into this (but not test -- would you be willing to test a patched version of iohub?)

PR #2318, opened 2019-02-28
BF: xrange not available in Python 3

hoechenberger:
Closes GH-2317.


peircej:
Maybe we should do the systematic application of future to iohub. I think I didn't dare before because I didn't know what would go wrong!

PR #2319, opened 2019-03-01
NF: Allow routines to be disabled

hoechenberger:
This PR allows routines to be disabled via a context menu entry.

Note that because of the  way PsychoPy adds routines to the flow, there is currently no easy way to allow users to disable individual representations of the same routine. If a routine is
placed in the flow multiple times, disabling one of them will disable _all_ occurrences.

More background info on this:
When adding a routine to the flow, PsychoPy simply adds this routine to a list. When adding another representation of the same routine to the flow, effectively the very same object will get added again -- they both point to the same routine object in memory. Modifying one will automatically alter the other too (since, well, it is not "another", but "the same" routine). To fix this, we'd have to come up with a solution to store metadata for each representation of a routine in the flow individually.

screenrecording20190301at1

cc @m-macaskill

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2288,6 +2288,12 @@ class BuilderFrame(wx.Frame):
         """
         self.routinePanel.createNewRoutine()
 
+    def toggleRoutineDisabled(self):
+        routine = (self.routinePanel.GetPage(self.routinePanel.GetSelection())
+                   .routine)
+        routine.params['disabled'] = not routine.params['disabled']
+        return routine.params['disabled']
+
     def renameRoutine(self, name, event=None, returnName=True):
         """Defines ability to rename routine in the routine panel
         """
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2289,6 +2289,13 @@ class BuilderFrame(wx.Frame):
         self.routinePanel.createNewRoutine()
 
     def toggleRoutineDisabled(self):
+        """Toggle the `disabled` parameter of a routine.
+
+        Returns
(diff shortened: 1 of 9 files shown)


peircej:
This pull request **fixes 1 alert** when merging e511aad263388b4478ea266478d2f90839837e69 into f89b8d5e4cd80689875abf99065976a79e09afe3 - [view on LGTM.com](https://lgtm.com/projects/g/psychopy/psychopy/rev/pr-46d2fd91dd22588596a4ab6c9545bae54f5d38f2)

**fixed alerts:**

* 1 for Unused import

---

peircej:
For the Flow, instead of setting this as an attribute of a Routine, I'd suggest it's controlled by a new class (Exclusion Component?). We would have a region that we can drag across the flow to exclude any given section of it. That would be be 

- more flexible (choose to exclude 1 _or more_ object within an Exclusion section)
- easy to store without changing existing data format (as 
- really easy to code in the exported code (the beginning of the section would do `if False:` and then increment the indentation and the end of the section would simply decrement the indentation, much like current LoopInitiator/LoopTerminator objects)
- this would extend easily to allow conditions for exclusion (exclude if expInfo['group']=='A')

hoechenberger:
Shouldn't we revert this PR then, and wait for somebody to implement the `ExclusionComponent`? Otherwise we'll have to deal with backward-compatibility issues in the future…

peircej:
To be honest, I hadn't realised how far you'd got with it! Now I don't know. Maybe your solution is sufficient and we tweak the Flow so that it holds additional data about each Routine. Or do you think the exclusion thing is good. I'm easy bother ways to be honest

PR #2321, opened 2019-03-04
BF: Misc iohub fixes

isolver:
Fixes iohub.datastore errors for records that had a string column.

The demos demo/coder/iohub/launchHub.py, keyboard.py, and mouse.py should now run on Windows 7 / 10 using Python 2.7 or 3.6. 

Note: due to changes in the gevent pkg implementation,  the smallest iohub device poll interval on windows is apparently 1 msec. If you put anything small than 0.001 in an iohub device interval field you get a warning from gevent and it is set to 1 msec. Devices that use callbacks are not affected by this.


peircej:
Wonderful to see iohub getting some love! I'm sure many users are going to be delighted!

PR #2322, opened 2019-03-04
DOCS: Updating docstrings for trialhandler

jfkominsky:
see https://discourse.psychopy.org/t/what-happened-to-trials-next-and-trials-nextentry/6903/2

Extremely minor change, but one that was causing confusion.


peircej:
I think we should update the code so that the documentation is correct, rather than the other way. The `__next__` method is needed for Python3 but I'd like to keep `next()` as an alias (that isn't considered private). I think we just need to add a line after the __next__ is created to say `next = __next__` thereby creating an alias.

PR #2332, opened 2019-03-08
BF: Set new remote URL when recreating projects in existing local repo

dvbridges:
This fixes an error where remote repos were not updated when
recreating projects in an existing repo. This was causing an error where
the project was not found remotely because the URL was pointing to a
different accounts remote repo, which may or may be visible due to
privacy settings. This scenario lead to users having to
keep recreating their project locally on every sync.


dvbridges:
Needs a rethink, only works for Coder. Builder seems to work differently from Coder.

PR #2339, opened 2019-03-11
BF: Writes custom code for StaticComponent

### LINKED ISSUE #2337: Setting "Custom code" in StaticComponent doesn't seem to have any effect
The generated script doesn't contain any traces of the `Custom code` entered in the `StaticComponent`'s properties dialog.

`psychopy:master`

dvbridges:
Fixes #2337.

### DIFF
--- a/psychopy/experiment/components/static/__init__.py
+++ b/psychopy/experiment/components/static/__init__.py
@@ -163,4 +163,11 @@ class StaticComponent(BaseComponent):
                                       updateType=prms[fieldName].updates,
                                       params=prms)
             code = "# component updates done\n"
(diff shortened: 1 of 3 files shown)


hoechenberger:
(We should probably add tests for this too *cough* ;))

dvbridges:
I think I will use @hoechenberger 's idea for testing routine disabling and create Builder files for each component (or one containing all components), and create some basic tests for each component i.e., does .py contain an entries for component N. I can create the test for the static component and push to this branch, then add the rest in a separate branch. We can also do the same with JavaScript, and have at least a minimal test for the Py/JS outputs :)

PR #2346, opened 2019-03-12
TEST: Tests Python component code is compiled correctly against correct version

dvbridges:
This new test creates and compares PsychoPy component code compiled
with Builder with correct versions stored in the data folder.

- Removes need for previous tests specific to components.

- Includes custom code checks for Static Component, fixing #2337 & #2339

- Allows for FileNotFound error on first test, so that a correct file can
be created. However, correct version contained in this push.

- Adds improved error handling to provide information for failures e.g.,
which file contains the error.

- Adds a tolerance for the compareTextFiles function. This is needed
when comparing files with different dates, and version numbers etc.

- Fixes an UnboundLocalError in the compareTextFiles when delim is None

- Fixes small ascii encoding error in pump component giving a Py2 failure


hoechenberger:
Wow, this seems like a lot of work you’ve put into this!

I wonder if the file comparison won’t fail if the PsychoPy version increases? Because the version number is stored in the compiled .py file?

Also, if we ever touch, say, the common includes, won’t we have to re-generate all the experiment scripts?

dvbridges:
Thanks, although its mostly automated code, just had to add the correct code scripts otherwise they get recreated on every test and so would not be a true baseline. Was Jons idea to create correct vs new scripts. 

I added a tolerance for the text file comparison, which allows a % difference (can be any value), to allow for differences from dates, versions etc. I chose 5% tolerance for the tests. 

Yes if something changes like that, we have to recreate all the baseline correct files again, but this is easily done by deleting the existing correct files from the data folder and rerunning the test locally. This will recreate all the files automatically.

psychopy/scripts/psyexpCompile.py
```diff
@@ -182,7 +182,7 @@ def _makeTarget(thisExp, outfile, targetOutput):
             else:
                 # We already have the text
                 scriptText = scriptDict[scripts]
-            with io.open(eval(scripts), 'w', encoding='utf-8-sig') as f:
+            with io.open(eval(scripts), 'w', encoding='utf-8-sig', newline='\n') as f:
```
hoechenberger: Is this _really_ what we want? Shouldn't we take care of the line endings when _reading_ files?

dvbridges: Good point, I will change from using codecs to read file to io, which has the "newline" parameter.

dvbridges:
@hoechenberger @peircej , seems like it might be better to create a folder for correct scripts in the tests/data folder. If there are no objections (for reasons I may not know!), I will create a scripts folder for the correct scripts.

hoechenberger:
No objections from my side, good idea!

peircej:
Well done on this by the way Dave. A +1.27 in coverage over the whole project might not sound like a lot but it's actually very impressive!

dvbridges:
Cheers Jon, yes all that green coverage does look good :rocket:

PR #2350, opened 2019-03-13
TEST: Tests JS component code is compiled correctly against correct version

dvbridges:
- Adds correct scripts, compiled in Py3

- Fixes parsing error causes by ast.parse not receiving a string. Also
imports unicode from past.builtins for Py3, as unicode was undefined.

- Sets Python as targets for Cedrus and ioLabs, at least until they
output JS.

- Couple of typos in psyexpCompile

- Sets screen units for image and text components as undefined if using
experiment settings. Further work to be done with units and JS, and need
clarity with what happens when units are undefined.


dvbridges:
Made this branch from the last pull req with the new Python tests, so looks like there are more changes that there really is.

PR #2351, opened 2019-03-13
DOCS/RF: Documentation fixes for colorspacetools

mdcutone:
Just doing some documentation work on modules I've contributed to in the past. Added __all__statment to keep internal functions from gunking up the namespace.

Also, it might be helpful to consider using Colour Science for Python libraries (https://www.colour-science.org/) to handle future colorspace conversion routines.

### DIFF
--- a/docs/source/api/tools/colorspacetools.rst
+++ b/docs/source/api/tools/colorspacetools.rst
@@ -15,6 +15,10 @@
     lms2rgb
     rgb2lms
     dkl2rgb
+    cielab2rgb
+    cielch2rgb
+    srgbTF
+    rec709TF
     
 Function details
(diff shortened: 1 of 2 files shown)


peircej:
Using colour-science for CIE and LAB is probably reasonable (although I also have the philosophy that if it isn't broken don't fix it!) but some of the colour spaces we support they don't. DKL and LMS are spaces that are mostly used by vision scientists trying to selectively stimulate individual cones, and not much by engineers but DKL also makes it relatively easy to create isoluminant stimuli (set the elevation to 0 and you're done) which I don't know how to do in CIE or LAB.

mdcutone:
I wouldn't replace existing functions here and I think my CIELAB function is good enough for most people.

PR #2358, opened 2019-03-14
BF: Fix escape key behaviour for online experiments

dvbridges:
Escape to quit behaviour was hardcoded into the routine code for JS. This
fix removes hardcoded escaping using "escape" and this behaviour is now
only determined by experiment settings option.


dvbridges:
Will make independent from keyboard component in next pull

PR #2359, opened 2019-03-15
BF: Fixes how escape is used to quit the experiment online

dvbridges:
Previously, the JS code would always quit the experiment if escape was
pushed. This fix allows user to control whether or not escape quits the
experiment, using the experiment settings.

- Tidies up some spacing inconsistencies between conditionals in JS
outputs

- Only writes code if code exists - this prevents blank lines from being
written into the code


peircej:
For the python implementation escape behaviour doesn't depend on whether escape is in allowedKeys does it? I don't think this conditional needs the query about it being in keyList or keyListStr (because it isn't used as a response)

dvbridges:
> For the python implementation escape behaviour doesn't depend on whether escape is in allowedKeys does it? I don't think this conditional needs the query about it being in keyList or keyListStr (because it isn't used as a response)

No, it is not the same. Unless the user adds the escape key to the allowed keys, escape will not be found in "theseKeys" and so checking for escape with the keyboard is redundant. However, if the keylist is empty (thus allowing all keys) this fails. So, looks like it makes more sense to leave the escape code in the keyboard as well.

PR #2361, opened 2019-03-15
ENH: Don't alphabetically sort exp info dict by default in Builder

hoechenberger:
This should work on Python 3.6+, where dictionaries retain insertion order by default.

Example settings:
Screen Shot 2019-03-15 at 11 43 52

Old behavior – sorted alphabetically:
Screen Shot 2019-03-15 at 11 44 44

New behavior – sorted as in Experiment Settings:
Screen Shot 2019-03-15 at 11 44 11

### DIFF
--- a/psychopy/experiment/components/settings/__init__.py
+++ b/psychopy/experiment/components/settings/__init__.py
@@ -589,7 +589,7 @@ class SettingsComponent(object):
         buff.writeIndented("expInfo = %s\n" % repr(expInfoDict))
         if self.params['Show info dlg'].val:
             buff.writeIndentedLines(
-                "dlg = gui.DlgFromDict(dictionary=expInfo, title=expName)\n"
+                "dlg = gui.DlgFromDict(dictionary=expInfo, sortKeys=False, title=expName)\n"
                 "if dlg.OK == False:\n    core.quit()  # user pressed cancel\n")
         buff.writeIndentedLines(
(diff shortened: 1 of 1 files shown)


peircej:
Hmm, but it does maintain the original order? I expected this to be random order if not sorted.

Maybe it's because Py3 uses OrderedDict by default whereas Py2 dict didn't store original order in any way

hoechenberger:
> Hmm, but it does maintain the original order? 
I'm not sure I follow you here… By default, `dlgFromDict()` sorts alphabetically. So this typically doesn't represent the order a user has specified in the Experiment Settings.

> I expected this to be random order if not sorted. Maybe it's because Py3 uses OrderedDict by default whereas Py2 dict didn't store original order in any way

Python 3 retains dictionary insertion order since Python 3.6; it's not using the `OrderedDict`, but "an ordered dict" :)

In Python 2, no specific sorting order is guaranteed – CPython typically sorts alphabetically, but that's an implementation detail.

What we could do is pass `sortKeys=False` for Python 3 to retain sorting order, and continue with the default (`sortkeys=True`) for Python 2.

peircej:
Yes, I think that's best. I think I switched to sorted at one point so that there would be at least *some* rule to the order (before I knew of the existence of OrderedDict)!
 
Alternatively, we could use OrderedDict for the job, which would provide the "right" behaviour and work in both environments.

PR #2362, opened 2019-03-15
FF: fix multiple PsychHID keyboards on mac

peircej:
On windows we can use (and need!) `PsychHID('Keyboard', None)` to get
all keyboards.

On Mac that currently returns only the first keyboard, not a combo, so
on Mac (and linux?) we need to fetch all keyboard-looking devices again
and create a virtual unified keyboard ourselves.


isolver:
This fixes the MS X4 keyboard on macOS too. It still kills PsychHID on Windows; I have not tested it on Linux yet.

PR #2367, opened 2019-03-18
Another attempted fix to incompatible freetype library

peircej:
It seems there is a freetype and a py-freetype (and I think the former is installed by the latter) so maybe we need to fix them both to a particular version rather than just one of them?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,8 +59,7 @@ install_requires =
     gitpython
     astunparse
     esprima
-    freetype<2.10.0
-    freetype-py<2.10.0
+    freetype-py
     # Platform-specific dependencies.
     imageio < 2.5; python_version < "3"
     imageio >= 2.5; python_version >= "3"
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,7 @@ install_requires =
     gitpython
     astunparse
     esprima
-    freetype-py
+    freetype-py<2.10.0
     # Platform-specific dependencies.
     imageio < 2.5; python_version < "3"
     imageio >= 2.5; python_version >= "3"
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,6 +59,7 @@ install_requires =
     gitpython
     astunparse
     esprima
+    freetype<2.10.0
     freetype-py<2.10.0
     # Platform-specific dependencies.
     imageio < 2.5; python_version < "3"
--- a/conda/environment-2.7.yml
+++ b/conda/environment-2.7.yml
@@ -13,6 +13,7 @@ dependencies:
 - esprima-python
 - gitpython
 - ffmpeg
+- freetype<2.10.0
 - freetype-py<2.10.0
 - future
 - gevent
--- a/conda/environment-2.7.yml
+++ b/conda/environment-2.7.yml
@@ -14,7 +14,7 @@ dependencies:
 - gitpython
 - ffmpeg
 - freetype<2.10.0
-- freetype-py<2.10.0
+- freetype-py
 - future
 - gevent
 - greenlet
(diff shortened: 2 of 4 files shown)


hoechenberger:
My understanding is that only freetype needs to be fixed to <2.10, the Python bindings (py-freetype) have been fixed in the development branch to support freetype 2.10, so the next release of those should make the error go away. 

--

setup.cfg
```diff
@@ -59,7 +59,8 @@ install_requires =
     gitpython
     astunparse
     esprima
-    freetype-py
+    freetype<2.10.0
```
hoechenberger: I don’t think this should go here, this is not on PyPI or is it?

peircej: I think you're right. Looks like 
- on PyPI we have freetype-py
- on conda we have freetype and freetype-py (and the latter auto-installs the former)
- on conda freetype needs to be <2.10.0 but setting freetype-py to that has no effect (because it doesn't constrain the freetype version!)

No wonder we were confused @dvbridges ! Man, I hate conda!! ;-)

dvbridges: Ha, glad its fixed :)

hoechenberger:
I am currently in the process of "fixing" `freetype-py` upstream. No need to change / pin anything on our side once these changes are effective.

https://github.com/conda-forge/freetype-py-feedstock/pull/2

setup.cfg
hoechenberger: > No wonder we were confused @dvbridges ! Man, I hate conda!! ;-)

I can see where to confusion came from. But: Don't hate, learn to appreci-ate! ;) 

Seriously, what tripped you up (?) is actually an awesome property of `conda` (or at least `conda-forge` packages): You language install bindings for a library? Hell of course they depend on that library! So this one gets installed too. You need to think of `conda` more as an `apt` instead of a `pip`. The great thing is that stuff one once used to have to install system-wide can now go into a virtual (`conda`) environment, and you can have multiple different versions installed without them interacting in a strange manner.

`freetype-py` (the Python project, not the `conda-forge` package) had a bug that would only show with the latest `freetype` release. It's already been fixed and will go away with the next `freetype-py` release.

I have now pinned the `conda-forge` `freetype-py` package to depend on `freetype <2.10`, as a temporary workaround.

Once CI finishes on https://github.com/conda-forge/freetype-py-feedstock/pull/2, `freetype-py` on `conda-forge` will be updated to include this workaround, and we can revert any changes we did to `PsychoPy` in that matter.

peircej: Yes, no worries. Don't mind my ribbing. I do understand the idea behind it now. The reason I get stuck more with confusing conda bugs is just because I don't use it as often as pip.

hoechenberger: It's done.

https://anaconda.org/conda-forge/freetype-py/files

peircej: Oh wow, that was quick. I assumed this would take some time hence bothering to pin the version

hoechenberger: I see there's already a new release of `freetype-py` with the fix included:
https://github.com/rougier/freetype-py/releases

I will package this later tonight or tomorrow.

hoechenberger: > I will package this later tonight or tomorrow.

Done. Version `2.1.0.post1` Should be on `conda-forge` in half an hour or so.

hoechenberger: > Yes, no worries. Don't mind my ribbing. I do understand the idea behind it now. The reason I get stuck more with confusing conda bugs is just because I don't use it as often as pip.

I don't mind that at all, I can totally see what confused you and I wonder if there'd be ways to make such things more clear. Generally, I'm not too happy with Python packaging in general. It's such a pain. And `conda` adds yet another layer of complexity… :S

hoechenberger: > Oh wow, that was quick. I assumed this would take some time hence bothering to pin the version

I could've fixed this 3 days ago already, I just had forgotten I was the maintainer of the `freetype-py` `conda-forge` package!! Just realized today, accidentally ;) And then thought, hey, wow, so, let's fix this thing right away :D

PR #2371, opened 2019-03-19
ENH: Adds new keyboard code to Builder!

dvbridges:
- New keyboard code is written dependent on the existence of the
psychtoolbox module (PTB). If PTB does not exist, then keyboard
reverts to using event module code.

- The new keyboard module will make different calls depending
on whether PTB exists. Currently, only getKeys is implemented. If no PTB,
then a new response object is created from the event.getKeys output
containing the key name and the response time. This differs from previous
calls to event.getKeys, which returned a list of key names.

- BuilderKeyResponse container class is now redundant and replaced with
keyboard attributes. These are refreshed on every repeat of the
routine, because a new keyboard object is instantiated on each repeat.

- To smooth call signature to new getKeys methods, the event.getKeys
kwargs are added to the new getKeys method.

- clear method in new Keyboard renamed to clearEvents (like event module).
Also added code to clear buffer - @peircej to check. Also adds kwargs to
make compatible with calls to event.clearEvents method.

## Other changes:
- key_resp is the first name of the first keyboard, but because it is in
reserved namespace, all new keyboards are called "key_resp_2". Replaced
with 'keyboard', because overwriting keyboard will
potentially cause new keyboard module to fail.

### DIFF
--- a/psychopy/experiment/components/cedrusBox/__init__.py
+++ b/psychopy/experiment/components/cedrusBox/__init__.py
@@ -61,6 +61,12 @@ class cedrusButtonBoxComponent(KeyboardComponent):
 
         self.type = 'cedrusButtonBox'
         self.url = "http://www.psychopy.org/builder/components/cedrusButtonBox.html"
+
+        # Remove inherited import from keyboard
+        for idx, imports in enumerate(self.exp.requiredImports):
+            if imports.importName == 'keyboard':
+                del self.exp.requiredImports[idx]
+
         self.exp.requirePsychopyLibs(['hardware'])
 
         self.params['correctAns'].hint = _translate(
--- a/psychopy/experiment/components/keyboard/__init__.py
+++ b/psychopy/experiment/components/keyboard/__init__.py
@@ -16,6 +16,10 @@ from os import path
 
 from psychopy.experiment.components import BaseComponent, Param, _translate
 from psychopy.experiment import CodeGenerationException, valid_var_re
+from pkgutil import find_loader
+
+# Check for psychtoolbox
+havePTB = find_loader('psychtoolbox') is not None
 
 # the absolute path to the folder containing this path
 thisFolder = path.abspath(path.dirname(__file__))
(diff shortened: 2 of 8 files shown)


psychopy/experiment/components/keyboard/__init__.py
```diff
@@ -225,49 +232,46 @@ def writeFrameCode(self, buff):
...
+        # Check for response
+        buff.writeIndented("if len(theseKeys):\n")
+        buff.setIndentLevel(1, True)
+        dedentAtEnd += 1  # indent by 1
+        buff.writeIndented("theseKeys = theseKeys[0]  # at least one key was pressed\n")
 
         if self.exp.settings.params['Enable Escape'].val:
             code = ('\n# check for quit:\n'
-                    'if "escape" in theseKeys:\n'
+                    'if "escape" in theseKeys.name:\n'
```
peircej: Actually... I wrote it so that this could be a comparison `if "escape" == theseKeys:` (without you needing to include `.name`) although it looks like I need to do something similar for `includes` operations

dvbridges: Ah, yes that should actually be a equality operator. I had written it at the time in case the name attribute was also a list. Equality check will work for both types of object returned PTB and event modules.

dvbridges: Fixed in #2373

psychopy/experiment/components/cedrusBox/__init__.py
```diff
@@ -61,6 +61,12 @@ def __init__(self, exp, parentName, name='buttonBox',
 
         self.type = 'cedrusButtonBox'
         self.url = "http://www.psychopy.org/builder/components/cedrusButtonBox.html"
+
+        # Remove inherited import from keyboard
+        for idx, imports in enumerate(self.exp.requiredImports):
+            if imports.importName == 'keyboard':
+                del self.exp.requiredImports[idx]
```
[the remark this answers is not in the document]
dvbridges: Fixed in #2373

PR #2373, opened 2019-03-19
BF: Removes unnecessary imports from Cedrus/ioLabs safely

dvbridges:
This fixes a problem where cedrus and ioLabs imports would wipe keyboard
imports if they appeared in the same experiment. This fix adds new
functionality to the experiment class by giving ability to check more
abstractly whether a particular type of component exists in an experiment.
For example, whether a component of type keyboard exists. With this new
functionality, Cedrus and ioLab boxes will only delete keyboard imports
if there are no keyboard components in the experiment.

Other commit also adds tests for new getComponentFromType function, which perhaps could be in
better place.


peircej:
Let's chat about this one tomorrow. I think I need to understand better what's going on (notably, why we were removing the keyboard anyway)

dvbridges:
> Let's chat about this one tomorrow. I think I need to understand better what's going on (notably, why we were removing the keyboard anyway)

This problem is because Cedrus and ioLabs inherit the keyboard, and so the new required imports for keyboard get inherited as well. So, if you had a routine with an image and Cedrus box, the code would include the following unnecessary import:

```
from psychopy.hardware import keyboard
```

With the latest fix, we only remove this import if keyboard do not exist in the experiment.

PR #2374, opened 2019-03-19
NF: Adds glfw to winType menu in Gen. Prefs and outputs winType in code

dvbridges:


dvbridges:
@peircej , because of the different branches I am working from, please pull this one in before #2373 because the correct code are different for keyboard (will be wrong for this branch, and correct for #2373)> It is only wrong on this branch because the equality check with theseKeys was changed in the most recent commit in #2373.

PR #2378, opened 2019-03-27
Fix typo and comment correspondence

Reoge:
Not sure that PR is right way to do such fixes (: 

newXYs = globForm.xys.copy() # create real copy of current xys (it was the same object, i.e. newXYs  is globForm.xys was True)

newOris = globForm.oris.copy() # create real copy of current oris (it was the same object, i.e. newOris is globForm.oris was True)


peircej:
Hmm, technically you're right that we refer to a copy but don't explicitly take a copy. On the other hand we don't technically need a copy here, and creating one is presumably marginally slower than the work in-place version. I can't see anything going wrong here by *not* creating a copy, is there?

Maybe we can change the comment to be more accurate rather than changing the code?

peircej:
I think I'm going to close this one for now for fear that it's slower creating the copies

PR #2379, opened 2019-03-29
FF: New keyboard no longer creates syntax error if keyList empty

dvbridges:
To fix this syntax error, the keyList param is hardcoded, but the
keyList arg is written as either a list (if filled) or None.

Also refactored the evaluation of keylist as a a variable, as conditional
statements are more readable using correct line breaks and indentation.


dvbridges:
@peircej , the second BF commit (157850a)  can be treated as a separate bug fix, if you want to pull that into a new BF release before the new keyboard is released.

PR #2389, opened 2019-04-04
Fix to JS compiled code

peircej:


dvbridges:
The code will work without the braces, but see:

>JavaScript allows the omission of curly braces when a block contains only one statement. However, it is considered by many to be best practice to never omit curly braces around blocks, even when they are optional, because it can lead to bugs and reduces code clarity. So the following:

https://eslint.org/docs/2.0.0/rules/curly

PR #2393, opened 2019-04-10
Explain alignVert positioning

alexholcombe:
Explain that alignVert is unlikely to align the exact center of text on the position set for the text object. However, I'm not sure I got the jargon quite right.


peircej:
You're right. I wish I'd renamed the (pyglet) text property to be `verticalAnchor` rather than vertical alignment.
FWIW I'm hoping during the summer we'll get TextBox2 finalised, which will have genuine text alignment option as well. Then we have to make a decision on whether to switch the name to the "right" thing even though that will then differ between the classes. (I think we probably should do, for long-term sanity)

PR #2400, opened 2019-04-14
FF: Creates Psychopy version class for handling version formatting

dvbridges:
Version formatting varies for html exports. When no version is
requested in experiment settings, no version number is added to the JS
code meaning the latest PsychoJS version will be used for a task.
This fix handles different version formats by creating a version class
for all version string formatting. The alternative was to use a function
or if statement to convert the version string, but having a version class
keeps things tidy and object-oriented. It also adds clarity which version
is actually being requested in the version compiler e.g., current or
requested version.


dvbridges:
Closing, as agreed this is overkill.

PR #2402, opened 2019-04-15
Fix windows NSIS installer to work in CurrentUser mode

### LINKED ISSUE #2127: Window standalone installation should not require admin access
Hi!

I'm trying to install psychopy on a Windows machine where I don't have local admin access. If my memory serves me, in the past I've been able to copy a psychopy installation from Program Files into another folder and have it work fine in the new location.

Assuming that's really true, can you change the installer so it doesn't require admin access? Or put up a standalone .zip file with the contents of the installation?

For what it's worth: I do have access to write to C:, but not to \Program Files or anything.

I guess another workaround would be to install anaconda and then install psychopy through that, but that seems like a really complicated workaround.

My ultimate goal is to have researchers "bring psychopy with them" rather than relying on systemwide installed tools, so systemwide upgrades don't cause surprises for studies that collect data for 5+ years. Well, not quite as often ;)

dgfitch:
With these changes, the windows installer no longer requires admin. 

Fixes #2127 

- If run with admin privileges, allows user to choose "Current User" or "All Users"
- If run without admin, defaults to "Current User" and warns user about installing avbin

The main changes to the NSI file are:

- using `!include MultiUser.nsh` and its needed settings
- using LogicLib to do some conditional things on install/uninstall

I changed the .ps1 script to make it less path-dependent as well.

Likely needs testing on more bitness/platforms, but the installer is largely the same; just slightly different methods to set PATH and create shortcuts, etc.

### DIFF
--- a/buildCompleteInstaller.nsi
+++ b/buildCompleteInstaller.nsi
@@ -83,16 +83,26 @@ Function multiuser_pre_func
 
 FunctionEnd
 
-;if previous version installed then remove
 Function .onInit
   !insertmacro MULTIUSER_INIT
 
   ${If} $MultiUser.InstallMode == "CurrentUser"
     StrCpy $InstDir "$LOCALAPPDATA\${PRODUCT_NAME}"
+    StrCpy $PRODUCT_REGISTRY_ROOT "HKCU"
+    IfFileExists $SYSDIR\avbin.dll continue_init 0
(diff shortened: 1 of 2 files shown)


peircej:
Cool. Thanks for your work on this. I'll run some tests on it tomorrow and make sure it all still flies. Debugging the installer is pretty painful due to the slow debug cycle (build/uninstall/install) so the help is much appreciated! :-)

PR #2425, opened 2019-04-26
Enh added channel field to mircophone properties

wilfrid-li:


wilfrid-li:
Updated Channel field to the correct valType: int

Wrong valType caused ExpBuilder to generate wrong attribute for AdvAudioCapture

PR #2428, opened 2019-04-29
WIP - BF/EN: Fixed drawing on multi-window displays with heterogeneous configurations.

mdcutone:
Fixes the issue posted here: https://discourse.psychopy.org/t/scaling-problems-with-multiple-windows-using-different-resolutions/7476/5

This patch fixes issues around switching windows which cause things to be incorrectly drawn in some use cases (e.g. when using windows with differing resolutions). It also makes window switches a touch more efficient by reducing redundant calls when a context change is requested on a current window. 

Also has support for properly configuring windows on retina displays. However, I don't have one to test with.

### DIFF
--- a/psychopy/visual/backends/glfwbackend.py
+++ b/psychopy/visual/backends/glfwbackend.py
@@ -447,6 +447,9 @@ class GLFWBackend(BaseBackend):
                 GL.glBindTexture(GL.GL_TEXTURE_2D, 0)
                 GL.glEnable(GL.GL_STENCIL_TEST)
 
+                GL.glViewport(0, 0, win.size[0], win.size[1])
+                GL.glScissor(0, 0, win.size[0], win.size[1])
+
     def dispatchEvents(self):
         """Dispatch events to the event handler (typically called on each frame)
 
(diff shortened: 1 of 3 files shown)


mdcutone:
As I was working on this I realized some limitations associated with multiple windows. Some stimuli cannot be drawn between them without context sharing enabled. When creating stimuli objects *after* initializing a window, objects will be bound to the context of the last created window, regardless if you passed a differnt window to the ``win`` argument. For GLFW, you must explicitly specify which window to share objects using the ``share`` argument when opening a window. Pyglet seems to do this automatically, therefore there is some inconsistent behavior there. Certainly needs to be looked into more.

peircej:
Yes, Pyglet uses a thing called a "shadow window" which means, I think, that everything is attached to that implicitly. That's the thing that isn't supported on some machines (Microsoft OpenGL driver didn't support it) and then users have to turn it off for most other uses

mdcutone:
Might be a good idea to make the GLFW backend automatically share resources with some *root* window to avoid confusion. 

On another note, the changes seem to break the view projection support I added. However, regular psychopy stimuli seem to draw perfectly fine.

PR #2430, opened 2019-04-30
BF/EN: Fixed view projection support and misc. window class enhancments.

mdcutone:
This patch fixes view projection support which was broken in the last patch. For some reason, a transposed numpy matrix no longer changes the order of values in memory and the values passed to OpenGL were not in column-major order. 

Also cleaned up some documentation and extended ``setMouseType`` support to Pyglet.


mdcutone:
Closing until I complete my tests to make sure changes aren't breaking.

PR #2435, opened 2019-05-03
FF: Create resource list as JSON file containing paths to resources.

dvbridges:
This fix creates a resources.json file for containing all paths to
resource files. The JSON contents are passed to the psychoJS.start
function.


dvbridges:
This is not relevant for now, and is part addressed in https://github.com/psychopy/psychopy/pull/2654 where a resource list is created in the JS, rather than a JSON file.

PR #2447, opened 2019-05-09
BF: fixed mic for pyo backend when prefs did not give pyo as preferred

peircej:
Mic is currently ONLY supported under pyo backend so, if another backend
is set as the default, we should initialise the pyo engine anyway and give
a warning that the mic and the preferred output lib may clash

### DIFF
--- a/psychopy/microphone.py
+++ b/psychopy/microphone.py
@@ -41,7 +41,8 @@ else:
 import json
 import numpy as np
 from scipy.io import wavfile
-from psychopy import core, logging, sound, web, prefs
+from psychopy import core, logging, web, prefs
+from psychopy.sound import backend_pyo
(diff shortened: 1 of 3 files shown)


mirh:
```
  File "d:\psychopy\demos\coder\input\latencyFromTone.py", line 38, in <module>
    names, idx = sound.backend.get_input_devices()
AttributeError: module 'psychopy.sound.backend_sounddevice' has no attribute 'get_input_devices'
```
Not sure if this falls under the more or less expected clashing or not. 

Also, unless I add directsound somewhere in the audio driver list, I trigger
https://github.com/psychopy/psychopy/blob/d4c11643fd695d1f1780ace9ffa7e27e693909a4/psychopy/sound/backend_pyo.py#L231-L232

PR #2449, opened 2019-05-10
BF/RF: Window fixes 

mdcutone:
Fixes some issues where ``viewPos`` and ``viewScale`` are overwritten by the new matrix logic. Added a bunch of fixes for retina displays when switching contexts. I drew a range of stimuli and everything draws correctly now. 

I'm going to play it safe an avoid going all in with regards to the view/projection matrix system until the next major release. If anyone can test this patch out on a Mac, that would be very helpful to see if retina support is working correctly.

Also for some reason there are lots of duplicate commit entries. Seems like git is acting up a bit.

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -822,13 +822,6 @@ class Window(object):
                 GL.glColor3f(1.0, 1.0, 1.0)  # glColor multiplies with texture
                 GL.glColorMask(True, True, True, True)
 
-                # clear the projection and modelview matrix for FBO blit
-                GL.glMatrixMode(GL.GL_PROJECTION)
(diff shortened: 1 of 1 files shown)


peircej:
Duplicate commits can occur if you've done some strange merging (e.g. a merge with rebase after some commits were already pushed to github).

The changes look relatively small and I'm tempted to squash them into a single commit. Can you just confirm this is a simple BF commit (should be included in 3.1.1 rather than waiting for 3.2). i.e. you're reasonably confident it only bugs in 3.1.0 and doesn't break other code?

mdcutone:
Yeah this only deals with bugs. No new features here.

PR #2450, opened 2019-05-10
ENH: Adds power term to envelope grating

schofiaj:
Add fature to raise the modulator in an envelope grating to some power.
Adds this feature in both secondorder.py and the env_grating builder component.
Also adds code to user scripts to warn if using env_grating with an apparently uncalibrated monitor.


psychopy/experiment/components/envelopegrating/__init__.py
```diff
@@ -31,6 +31,7 @@
               'envori':_translate('Envelope orientation'),
               'envphase':_translate('Envelope phase'),
               'moddepth':_translate('Envelope modulation depth'),
+              'power':_translate('Power to which envelope is rasied'),
```
peircej: rasied = raised

schofiaj: fixed.

peircej: FWIW I wouldn't normally push for a recommit on a typo but this is in a piece of text that will be translated (e.g. into Japanese) and when the typo does get corrected it will show up as needing re-translation

schofiaj:
The comparision between the complies scripit and my upload is failing. Maybe because I build mine under a different version or something. Is there anyway to see the version that Travis is making?

peircej:
That test is too overzealous and doesn't ignore irrlevant changes. Don't worry - we'll take a look at this end

schofiaj:
I think it might be an extra blank line in the doc string at the top. Shall I edit and give it a go?

peircej:
Sure, if you fancy, but no big deal

PR #2452, opened 2019-05-10
TESTS: tests new noise and secondorder features

schofiaj:
Adds stimulus tests for recently added features in noise.py and secondorder.py


schofiaj:
needs a BF first

PR #2455, opened 2019-05-11
BF: Return keys if maxWait <= 0

### LINKED ISSUE #2454: iohub.client.keyboard.waitForKeys fails for maxWait=0
The documentation for `iohub.client.keyboard.waitForKeys` states that if `maxWait` is 0, the function should act like `getKeys()`.

 >  :param maxWait: Maximum seconds method waits for >=1 matching event.
 >                  If 0.0, method functions the same as getKeys().
 >                  If None, the methods blocks indefinitely.

In practice, if `maxWait` is 0, `getKeys()` never is called and the function returns an empty list.

The only two checks where `getKeys` can get called are
https://github.com/psychopy/psychopy/blob/4ffc99ec7f2724bf85034cd53f0a69698446c070/psychopy/iohub/client/keyboard.py#L430-L437
and
https://github.com/psychopy/psychopy/blob/4ffc99ec7f2724bf85034cd53f0a69698446c070/psychopy/iohub/client/keyboard.py#L439-L443

I'm thinking the documentation makes more sense here than the current implementation, specifically in the case where `maxWait` is based on a timer that can drop below 0.

Happy to submit a pull request if the documented behavior is intended.

ariekahn:
This fix ensures that `waitForKeys` still returns the current set of queued events if `maxWait` is a number but not greater than 0. Previous behavior was to return an empty list.

Fixes #2454

### DIFF
--- a/psychopy/iohub/client/keyboard.py
+++ b/psychopy/iohub/client/keyboard.py
@@ -394,7 +394,7 @@ class Keyboard(ioHubDeviceView):
         Returned events are sorted by time.
 
         :param maxWait: Maximum seconds method waits for >=1 matching event.
-                        If 0.0, method functions the same as getKeys().
+                        If <=0.0, method functions the same as getKeys().
                         If None, the methods blocks indefinitely.
         :param keys: Include events where .key in keys.
         :param chars: Include events where .char in chars.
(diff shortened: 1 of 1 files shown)


peircej:
This seems sensible enough to me. @isolver do you agree?

isolver:
Good catch. Thank you.

PR #2456, opened 2019-05-12
BF: FormComponent crashes if non-ASCII characters are used in Python2

hsogo:
Form component crashes if non-ASCII characters are used in form items in Python2.
Error message is as follows:

```
  File "C:\Users\Hiroyuki\Desktop\untitled.py", line 81, in <module>
    itemPadding=0.05)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\visual\form.py", line 99, in __init__
    self._doLayout()
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\visual\form.py", line 605, in _doLayout
    question, questionHeight, questionWidth = self._setQuestion(item)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\visual\form.py", line 269, in _setQuestion
    psychopy.logging.exp("Question text: {}".format(item['questionText']))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```
If non-ASCII characters are included in `item['questionText']`, `u` is necessary to output this string to logfile.

In addition, in lines 174-175 for visual/form.py, type of `item['options']` is not `str` but `unicode ` in Python2.

### DIFF
--- a/psychopy/visual/form.py
+++ b/psychopy/visual/form.py
@@ -15,6 +15,8 @@ from psychopy.visual.basevisual import (BaseVisualStim,
                                         ColorMixin)
 from random import shuffle
 
+from psychopy.constants import PY3
+
 __author__ = 'Jon Peirce, David Bridges, Anthony Haffey'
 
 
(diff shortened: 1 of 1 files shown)


psychopy/visual/form.py
```diff
@@ -171,8 +173,12 @@ def _checkHeaders(fields):
                 _checkHeaders(item.keys())
         # Convert options to list of strings
         for idx, item in enumerate(items):
-            if isinstance(item['options'], str):
-                items[idx]['options'] = item['options'].split(',')
+            if PY3:
+                if isinstance(item['options'], str):
+                    items[idx]['options'] = item['options'].split(',')
+            else: # Python2
+                if isinstance(item['options'], unicode):
```
peircej: I think this needs to be `basestring` instead of `unicode` for Py2

hsogo:
Ah, that's better!  I've fixed it.

PR #2458, opened 2019-05-13
BF: Fixes issue in filtered noise

schofiaj:
Fixes a rounding error which was slightly altering the frequency spectrum of Butterworth filtered noise and 'filtered' noise images and doing so in a way that varied between PY2 and PY3


schofiaj:
Failing on a test for Radial images which I've not touched and which I gather is tempremental. How do I re-run the test?

peircej:
Yes, I spotted that and started the rerun already. There are restart buttons to the right of each test:
image

peircej:
I'll bear in mind your comments about it being broken in the wild and I'll plan to release quickly (it just missed one release today but I restarted building that one so many times in the last few days I was sick of it and wanted it out!)
I'm building new releases pretty regularly right now though so it won't be long. A week or 2 I expect.

PR #2463, opened 2019-05-14
BF: Ensure events cleared and clock reset for KB when using visual Sync

dvbridges:
When KB was synced to visual, the KB clock was not reset and events
not cleared before first response was collected, only when
a response happened immediately after keyboard status was changed
to "STARTED". The clearEvents call now added to callOnFlip when visual
sync for KB is True. The new waitOnFlip var does not allow
getKeys to be called if the initial callOnFlip, used for resetting clock
 and clearing events, has not been completed.


dvbridges:
@peircej , with this fix I am not getting negative RTs using getKeys (or the other issue that was fixed), but I am able to achieve 1ms response times with a keyboard.

peircej:
For info, the fact that the keyboard is reporting 1ms response times just means that it's detecting events in the backlog from the keyboard

PR #2468, opened 2019-05-16
DOCS: Numpy style docs for the window class

mdcutone:
The documentation seems a bit out-of-date and broken in some places. I tried to convert much of the documentation of the Window class to Numpy style. I fixed loads of broken cross-refs and links in the process. I also added directives to hide private and deprecated attributes from being included in the documentation. I generated the HTML on my end and it looks great.

~However, I didn't convert the parameter list for the Window class to numpy style just yet. That's going to be pretty involved.~

I also made a bug-fix where the ``setMouseType`` function doesn't do anything. Should probably be converted to a property someday.

### DIFF
--- a/docs/source/api/visual/window.rst
+++ b/docs/source/api/visual/window.rst
@@ -8,5 +8,9 @@ A class representing a window for displaying one or more stimuli.
 .. autoclass:: psychopy.visual.Window
     :members: 
     :undoc-members: 
-    :exclude-members: go, whenIdle, onResize, fullScr, update, multiFlip, dispatchAllWindowEvents
+    :exclude-members: go, whenIdle, onResize, fullScr, update, multiFlip,
+        dispatchAllWindowEvents, setScale, setViewPos, setUnits, setRGB,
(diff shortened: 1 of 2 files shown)


peircej:
Cool, nice one. Yes, there's a ton of work to convert everything to numpy format docs and I'm not sure I see anyone bothering to tackle it large-scale! But good to see the docs getting a bit of love!

By the way, I didn't dare pull in your bug fixes for the 3.1.x series. There were enough changes to the files that I was worried I'd mess it up cherry-picking them across so just left them for the 3.2 series (which will only be a few weeks away anyway I think).

mdcutone:
I'll look into why that happened. I can probably make a clean branch and apply the changes again. There actually weren't many.

mdcutone:
Usually people with very little programming background are using PsychoPy, so docs are very important given the target audience. Particularly, the docs should include lots of examples to help people understand how things work in context. 

On a side note, I usually judge the level of professionalism around a project based on the quality of the documentation. It's a place where first impressions are made.

PR #2470, opened 2019-05-18
NF: Mathtools

mdcutone:
This commit adds a module for math functions. Particularly for performing transformations on vectors, quaternions, and vectors. This is useful for the upcoming 3D stimulus classes and for people who wish to render stimuli by computing rotation matrices. More math functions will be added over time to handle deprecated OpenGL commands.


peircej:
It feels like the sort of code that should be included in the test suite pretty to check that the results being returned are as expected.

mdcutone:
Good idea, I'll put that in too. 

I added support for deprecated OpenGL functions 'glScale', glRotate' etc. to follow modern OpenGL conventions.

mdcutone:
Added tests for quaternion operations and rotation matrices. I know these work since I've used these exact routines for rendering 3D objects in the past, but having tests never hurt :).

BTW, all these default to float32 precision since that's what is typically used for OpenGL.

peircej:
Great. Thanks. Yes, the idea of the test is to make sure it doesn't start failing in the future (e.g. with some change to numpy/python)

PR #2475, opened 2019-05-21
BF,ENH: update translation

hsogo:
I've added Japanese translation to new dialog strings.
In addition, I've fixed an error that message on the Loop properties dialog (shown below) was not translated because of wrong use of _translate().

By the way, this message should be shown in red as error/warning messages on Components' properties dialogs.  I added some lined to do this, but some messages are still shown in black.  This message is the return value of `app.builder.dialogs.DlgLoopProperties.getTrialsSummary()`, and I think that this method should additionally return whether the message is warning/error or not to set font color.  Can I make this change?


peircej:
Yes your suggestion of tracking whether the message includes a color suggestion makes sense. Feel free to make that work if possible :-)

hsogo:
I've modified 'getTrialsSummary()' to return whether the message is warning/error or not.
I noticed that error message is not shown on the Loop Properties dialog when duplicate name names are in conditionsFile.  Instead, following message is output to Coder's output pane (in this example, `stim` is duplicated name).

```
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\builder\dialogs\__init__.py", line 1569, in onBrowseTrialsFile
    returnFieldNames=True)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\data\utils.py", line 285, in importConditions
    trialList, fieldNames = pandasToDictList(trialsArr)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\data\utils.py", line 248, in pandasToDictList
    _assertValidVarNames(fieldNames, fileName)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\data\utils.py", line 223, in _assertValidVarNames
    (fileName, msg, os.linesep * 2, name))
ValueError: Conditions file C:\Users\Hiroyuki\Documents\cnd_ee.xlsx: Parameters (column headers) cannot contain punctuation or spaces

"stim.1"
```

There are two problems. Firstly, only ImportError is caught when reading conditionsFile. I think that ValueError should also be caught  (see [this commit](https://github.com/psychopy/psychopy/pull/2475/commits/7c5ad9cb8e78ff12f34228204bbf3163309d9894)).  Applying this fix, the error message appears as a dialog box.



The second problem is that this error message is puzzling.  As said above, this error occurs because `stim` is duplicated in the conditionsFile.  However, pandas automatically appends `.1` to 
duplicated name and then the error raised because `stim.1` is not valid name. Although `onBrowseTrialsFile()` contains codes that handle duplicated name, it doesn't work with this pandas behavior (I guess these codes were designed to work with openpyxl?).

peircej:
That's annoying! It's probably a new feature they've added. That's where there was an advantage of parsing the files ourselves, which is what we used to do :-/

In the `pandas.read_csv()` the docs say it raises a warning if it finds duplicate headers. On the other hand, for `read_excel()` it has a setting called `mangle_dupe_cols` which seems to be the source of the behaviour. If we set that to False it will no longer append `.1` (we definitely need it not to do that) but it will overwrite columns during reading, which also doesn't seem great. To handle this properly maybe we need to read it in twice and check if the two imports result in the same set of header names (suggesting that pandas had not performed any auto-correction for us). Yuck!

peircej:
For info, the failing tests are due to a deprecation in scipy, which we'll need to fix separately. Don't worry about it here.

hsogo:
I tried pandas 0.24.2, but `mangle_dupe_cols=False` didn't work because it is not supported yet!
```
(snip)
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\site-packages\pandas\io\parsers.py", line 910, in _get_options_with_defaults
    raise ValueError('Setting mangle_dupe_cols=False is '
ValueError: Setting mangle_dupe_cols=False is not supported yet
```
So, I think that it would be better to show a message like "Parameters (column headers) cannot contain dots or be duplicated." when erroneous column header matches  `'".+\.[0-9]+$"'`.  For example, how about generating error dialog in `getTrialsSummary()`? 

```
                    mo = re.search('".+\.[0-9]+"$', msg)
                    if 'cannot contain punctuation or spaces in msg' and mo:
                        # column name is something like "stim.1", which may be 
                        # in conditionsFile or generated by pandas when column
                        # names are duplicated.
                        m2 = 'Parameters (column headers) cannot contain dots' \
                             ' or be duplicated.'
                    else: 
                        # other exceptions
                        sep2 = os.linesep * 2
                        m2 = msg.replace(
                            'Conditions file ','').replace(': ', sep2)
                    _title = _translate(
                        'Configuration error in conditions file')
                    dlgErr = dialogs.MessageDialog(
                        parent=self.frame, message=m2,
                        type='Info', title=_title).ShowModal()
```

peircej:
Damn pandas! ;-) Sure, that solution is easy enough.

PR #2482, opened 2019-05-28
NF: New Brush component for freehand drawing 

dvbridges:
The Brush class creates a new ShapeStim on every mouse click, and updates ShapeStim vertices according to mouse position. JavaScript output does not use a Brush class, but uses functions and code to mimic the Python behaviour. As a result, updating on every repeat for line color and line width is not functional online.  The JS code will be replaced with a JS class for the Brush component in the coming months. Tests and docs included in this pull request.


dvbridges:
Py3 failing on :

- test_noiseAndRaisedCos 
- test_noiseFiltersAndRaisedCos

dvbridges:
This could be replaced with something like https://github.com/Ianmichael7/SimplePaint

PR #2490, opened 2019-06-11
TESTS: scipy.imresize is deprecated in scipy 1.2 so swith to using PIL

peircej:


psychopy/tools/mathtools.py
```diff
@@ -495,7 +495,7 @@ def translationMatrix(t):
     return T
 
 
-def concatenate(*args, out=None):
+def concatenate(*args, **kwargs):
```
peircej: @mdcutone for now I'm still running the test suite on Py2 as well as Py3 so a syntax error breaks tests. I had to switch to the slightly uglier format to cope with that

psychopy/visual/noise.py
```diff
@@ -35,7 +35,6 @@
 import numpy
 from numpy import exp, sin, cos
 from numpy.fft import fft2, ifft2, fftshift, ifftshift
-from scipy.misc import imresize, imrotate
```
peircej: @schofiaj the new versions of scipy have dropped support for these (but they were previously using PIL before apparently) so all the changes below are about handling that change for future

peircej: Although the code looks longer I think the compute time will be the same (becuase I think this is probably exactly what scipy.imresize was actually doing!)

schofiaj: Thanks Jon, I'd noticed that these image functions were depreciated in scipy and had it on my list to recode to avoid them. Hopefully your changes will deal with that approriately.

psychopy/tools/mathtools.py
mdcutone: @peircej 

I think I'm going to change that function to take a list as an argument for py2 compatibility. 

I actually re-wrote all those functions as super-efficient vectorized versions, however they will yeild the same results as the current version. So we can hold off releasing that until the next point release. The functions cover 95% of what needed for 2/3D graphics and animation (point walkers, etc.) and will make working with PsychXR data much easier.

PR #2496, opened 2019-06-18
BF: Find and save orphaned data at end of online task

dvbridges:
This fix solves a problem where routine data was not saved
if no loops came after that routine (because nextEntry was not called).

This current implementation will not work until PsychoJS
Experiment Handler attributes of `_trialsData` and `thisCurrentTrial`
are changed  to `entries` and `thisEntry`, respectively.

### DIFF
--- a/psychopy/experiment/components/settings/__init__.py
+++ b/psychopy/experiment/components/settings/__init__.py
@@ -813,6 +813,10 @@ class SettingsComponent(object):
                     "}\n")
         buff.writeIndentedLines(recordLoopIterationFunc)
         quitFunc = ("\nfunction quitPsychoJS(message, isCompleted) {\n"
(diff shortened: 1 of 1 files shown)


dvbridges:
@peircej , another situation has arisen where data not being saved. In a nested loop, with the outerloop `isTrials` is False, and the inner loop `isTrials` is True, a participant ending their inner loop on the first iteration, but repeats inner loop n times based on number of trials in the outerloop, will not have data saved. This is because breaking the inner loop on the first iteration does not call `nextExtry()`, and the outer loop where isTrials is False does not call `nextEntry()`. Calling at the end of the task will only write the last trial recorded. The solution is to call next routine in `endLoopIteration()` when `currentLoop` is true.

PR #2498, opened 2019-06-19
BF: Save data when calling loop.finished in nested loop

dvbridges:
In a nested loop, with the outerloop isTrials is False, and the
inner loop isTrials is True, a participant ending their inner loop
on the first iteration, but has repeating inner loop n times based on number
of trials in the outerloop, will not have data saved. This is because
breaking the inner loop on the first iteration does not call nextExtry(),
and the outer loop where isTrials is False does not call nextEntry().
Calling at the end of the task will only write the last trial recorded.
This fix calls nextEntry in endLoopIteration() when currentLoop
 is true and orphan data exists.


peircej:
Just wanted to make sure that BF is the correct prefix here. I don't want to include in 3.1.4 if it breaks existing studies

Issue #2504, opened 2019-06-26
Drop cffi dependency?


hoechenberger:
I realized we [depend](https://github.com/psychopy/psychopy/blob/59663caac65b12eba29ca14fe5d680ab64268074/setup.cfg#L42) on `cffi`, but we never import / use this package..? Is it safe to drop this dependency?

PR #2520, opened 2019-07-07
PKG: Remove superfluous cffi dependency

hoechenberger:
Closes GH-2504.


peircej:
Although we don't import it, several dependencies depend on it (e.g. sounddevice and I can't remember what others) but I see that sounddevice specifies it in its own setup.py so this should be safe

hoechenberger:
Yes, exactly, it’s an implicit dependency, and packages that need it should specify that dependency on their own :)

PR #2530, opened 2019-07-13
ENH: add `gammaErrorPolicy` option to control gamma table error handling

djmannion:
This is a request for comment on a potential strategy to deal with OS-level breakages in gamma table handling.

We have seen these recently in both Linux and Windows, and they are currently fatal (under the `pyglet` backend, at least). Since it is probably a minority of users that care about gamma settings, it seems like we ought to have a way that the code can still run if system gamma problems are encountered.

This isn't ready for merging or anything yet (it is untested, and I don't have an updated Windows 10 system to test it on at the moment - and I just noticed the `"` and `'` inconsistency that I will need to fix, and I think the message could be more informative), but it is one idea for how it might be handled; basically, by adding a `gammaErrorPolicy` parameter to the `Window` that then determines what happens if gamma problems are encountered.


peircej:
I agree with the concept and I've long had plans in my head to take such an approach more generally. @dvbridges is currently working on a system of "Issues" (or did we call it "alerts"?) in PsychoPy that allow us to track issues according to codes. Within the PsychoPy code we would "raise" one of these issue codes but then PsychoPy decides what to do with it.
The idea is that the issue code can be brief but carries info about a URL for further help. 
Additionally it allows us to set user-dependent actions for each of the codes. E.g. I really care about gamma so raise error on gamma fail, but just log an error on frame drop, and log a warning on window size mismatch

djmannion:
That sounds neat! Should I pursue this PR or is the direction you mention likely to be in place soon? I'm a bit worried that the Windows 10 update will be rolled out more broadly and many more experiments will stop working due to gamma problems outside our control.

peircej:
Yes, fair point, let's pursue it here so it's done faster. 

We can add an option to experiments called "Gamma corrected" that sets this policy to 'raise' and otherwise leaves it as 'warning'. Actually this allows us to solve another issue, which is that we could use the same flag to check whether a valid calibration file is found for that monitor (currently that just provides a warning if not and I know of one lab that lost lots of data-collection becuase of that problem).

I guess for a default we should set the "Gamma corrected" flag to be False and then vision scientists have to take a manual step to change it. Long-term I intend to create a new PsychoPy installer that asks questions about these sorts of things and sets defaults for an entire installation. But that's another day! :-)

cheers

peircej:
Was this roughly ready to go @djmannion or should I wait? When you said "pursue" it sounded like you had more to add.
We're planning a release (3.2) in next week or so

djmannion:
A few things that I need to tidy up and I need to do more testing - within the next few days.

djmannion:
I have been able to reproduce the problem on a Windows machine, and the code in this PR seems to be working as intended (i.e., it allows the error to be bypassed).

Could you please take a look and see if it seems OK. Perhaps we could post on discourse looking for other testers too.

peircej:
Great. I'll pull this in for now. I've got a linux machine with intel card currently failling for the same reason so I'll test there

PR #2540, opened 2019-07-19
DOC: enable sphinx.ext.viewcode

sappelhoff:
With [`sphinx.ext.viewcode`](https://www.sphinx-doc.org/en/master/usage/extensions/viewcode.html), users can click on a very subtle "source" link next to each function/method/... in the API docs and that link leads them ... to the source :-)

For example, I was just inspecting [`getActualFramRate` in the API docs](https://www.psychopy.org/api/visual/window.html#psychopy.visual.Window.getActualFrameRate) and wanted to learn a bit more about it ... `sphinx.ext.viewcode` would have made that easier than going to GitHub and clicking on lots of different folders and files, or using a search function in an editor etc.


peircej:
Nice. Thanks Stefan!

mdcutone:
I like this idea, going to use it in the future.

PR #2550, opened 2019-07-31
NF/DOCS: Added shader functions to gltools

mdcutone:
Added a bunch of useful functions for working with shaders. Includes functions for generating shader programs compatible with existing PsychoPy shader code and newer GLSL specs.

I tried re-writing the `compileShaders` function (haven't committed it) to be dual compatible, using the legacy shader functions when running on Intel drivers. Nvidia doesn't complain when compiling older spec drivers with the newer functions. Not sure about AMD at this point. The point of all this is to deal with shaders failing to work on GLFW when using vertex arrays, they need to use core profile functions to work.

### DIFF
--- a/docs/source/api/tools/gltools.rst
+++ b/docs/source/api/tools/gltools.rst
@@ -5,7 +5,26 @@
 .. currentmodule:: psychopy.tools.gltools
     
 .. autosummary:: 
-
+    
+    createProgram
+    compileShader
+    deleteObject
+    attachShader
+    detachShader
+    linkProgram
+    validateProgram
+    useProgram
+    createProgramObjectARB
+    compileShaderObjectARB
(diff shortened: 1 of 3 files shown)


peircej:
It looks like these functions would supercede `psychopy.visual.shaders.compileShader()`
Might I suggest that visual.shaders starts by importing your function(s) as well?

mdcutone:
I rewrote ``compileShader`` to use these functions. However, I'm holding off committing that until I can ensure nothing breaks. Stimuli render fine on my nvidia and intel machines, but I have yet to test on AMD. However, using the *ARB functions, the resulting program object is identical to the one ``compileShader`` was returning so it should work.

mdcutone:
Okay, I added the new ``compileShader`` function.

PR #2562, opened 2019-08-12
FF: Have `embedShaderSourceDefs` use proper indentation level.

mdcutone:
Minor fix, have embedded `#define` statements match the indentation level of the GLSL source code they are being inserted into. This make printing the modified GLSL source code appear neater. Also removed quotations from define values.

### DIFF
--- a/psychopy/tools/gltools.py
+++ b/psychopy/tools/gltools.py
@@ -76,7 +76,7 @@ def createProgramObjectARB():
 
     This creates an *Architecture Review Board* (ARB) program variant which is
     compatible with older GLSL versions and OpenGL coding practices (eg.
-    immediate mode) on some platforms. Use *ARB variants of shader helper
+    fixed function) on some platforms. Use *ARB variants of shader helper
     functions (eg. `compileShaderObjectARB` instead of `compileShader`) when
     working with these ARB program objects. This was included for legacy support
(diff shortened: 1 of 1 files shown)


mdcutone:
Added some examples to emphasize how useful `getAttribLocations` and `getUniformLocations` are. You can bind uniforms/attributes conditionally on whether they are supported by the shader. This eliminates the need to keep track of what code paths the shader supports. For instance, you can enable texturing in your `draw()` method only if the shader has the uniform defined:

```
        # Call `getUniformLocations` after compiling the shader and store it as an attribute somewhere
        self.shaderUniforms = getUniformLocations(self.shader)

        # If the shader has `uniform sampler2D diffuseTexture` defined, we enable textures and 
        # bind it to the appropriate texture unit.
        if 'diffuseTexture' in self.shaderUniforms.keys():
            # enable textures if the shader calls for it
            glEnable(GL_TEXTURE_2D)
            glActiveTexture(GL_TEXTURE0)
            glColor4f(1.0, 1.0, 1.0, 1.0)
            glColorMask(True, True, True, True)
            glBindTexture(GL_TEXTURE_2D, texId)

            # diffuse at texture unit `0`
            glUniform4i(self.shaderUniforms['diffuseTexture'], 0)
            # remember to disable textures when done!
        else:
            # no diffuse texture, just have the material track the current color
```

PR #2563, opened 2019-08-13
new demos: adaptive methods for delayed and risky choice tasks

drbenvincent:
Requires this https://pypi.org/project/darc-toolbox/ which should auto-install it's dependency https://pypi.org/project/badapted/


hoechenberger:
@peircej Shouldn’t the darc toolbox be added as a package requirement in setup.cfg as well? Otherwise the demos won‘t run...

peircej:
For now I've added it to the Standalone distributions (using the setupApp.py settings). We have an ongoing debate about what we should list as a PsychoPy `requirement`. Some want us to list only *strictly required* libs (which darc is not), some want all libs that PsychoPy ever needs for anything (which darc now is), and some want the libs that are required for *most studies* which is where I roughly fall, and I'd say darc isn't there yet.
So, for now, I'd suggest we leave it as is being distributed in Standalone installations and trusting self-installing users to be able to work out what to do if they need an additional install due to an error message

drbenvincent:
I don't have any knowledge of your behind the scenes build system etc, but this sounds reasonable to me.

drbenvincent:
I've tried the demos in the 3.2.0 testing version. Demos are included and unpack ok. But it looks like it's not finding the darc_toolbox code

 ```
Running: /Users/btvincent/Documents/PsychoPy3 Demos/adaptive_hyperbolic_discounting/delayed_choice_lastrun.py 
2019-08-15 15:14:29.944 python[3009:16394] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to (null)
  File "/Users/btvincent/Documents/PsychoPy3 Demos/adaptive_hyperbolic_discounting/delayed_choice_lastrun.py", line 121, in <module>
    from darc_toolbox.delayed import models
ModuleNotFoundError: No module named 'darc_toolbox.delayed'
```

**Comment by peircej** [comment]
Ah, it looks like your wheel is incomplete. It  doesn't have any of your folders inside, just the 4 files in the main folder

**Comment by drbenvincent** [comment]
Will have a look at building wheel for pip again and get back to you.

**Comment by peircej** [comment]
I've sent you a pull request for a fixed (hopefully) setup that will then include those other sub-packages

**Comment by drbenvincent** [comment]
Ok. After all that, new pip package is here https://pypi.org/project/darc-experiment-toolbox/ Had to do a name change to `darc-experiment-toolbox` because I'm new at this and am making mistakes. But as far as I can tell, this should solve the problem

**Comment by drbenvincent** [comment]
Just tried the demos with version `2nd TESTING release for PsychoPy 3.2.0` and get the following error. Have to say I have no clue about what this means. Looks like something PsychoPy related rather than in the demo code or associated packages, but I'm not certain.

```
 Running: /Users/btvincent/Documents/PsychoPy3 Demos/adaptive_risky_choice/risky_choice_lastrun.py before 1
after 2
  File "/Applications/PsychoPy3.app/Contents/Resources/lib/python3.6/psychopy/app/builder/builder.py", line 2247, in whileRunningFile
    wx.Yield()
wx._core.wxAssertionError: C++ assertion "Assert failure" failed at /Users/robind/projects/buildbots/macosx-vm6/dist-osx-py36/Phoenix/ext/wxWidgets/src/common/evtloopcmn.cpp(110) in Yield(): wxYield called recursively
  File "/Applications/PsychoPy3.app/Contents/Resources/lib/python3.6/psychopy/app/builder/builder.py", line 2233, in whileRunningFile
    self.onProcessEnded()
  File "/Applications/PsychoPy3.app/Contents/Resources/lib/python3.6/psychopy/app/builder/builder.py", line 2255, in onProcessEnded
    wx.Yield()
wx._core.wxAssertionError: C++ assertion "Assert failure" failed at /Users/robind/projects/buildbots/macosx-vm6/dist-osx-py36/Phoenix/ext/wxWidgets/src/common/evtloopcmn.cpp(110) in Yield(): wxYield called recursively
/Applications/PsychoPy3.app/Contents/MacOS/python: can't open file '/Users/btvincent/Documents/PsychoPy3\': [Errno 2] No such file or directory
```

peircej:
Hmm, OK, that error is caused by me changing the method for launching the experiments in a process. Clearly I've got mroe testing to do on that. I might need to get back to you if I can't recreate the error locally

drbenvincent:
Feel free. At a conference currently, but can try my best to fit some time in in the evenings.

Issue #2518, opened 2019-07-07
Support for pyglet 1.4


luketudge:
When installing psychopy's dependencies, pip now retrieves the latest version of pyglet (at least for me), pyglet 1.4. But some psychopy code is only compatible with pyglet<=1.3.

Opening a window with `visual.Window()` and the pyglet backend raises an error with traceback leading to `pyglet.window.get_platform()`. This problem is noted in #2399, with a downgrade to pyglet 1.3 as the suggested fix.

In pyglet 1.4, `pyglet.window.get_platform().get_default_display()` seems to have been refactored to `pyglet.canvas.get_display()` (see pyglet issue [here](https://bitbucket.org/pyglet/pyglet/issues/135/ubuntu-pygletwindow-return-nonetype-object)).

I am not familiar enough with the psychopy project to know which fix is preferable:
* hold back pyglet to <=1.3 in psychopy's dependencies
* switch to supporting only the latest pyglet
* offer compatibility for both pyglets with a version check

If the preference is for full compatibility, then the necessary fix is probably in [pygletbackend.py](https://github.com/psychopy/psychopy/blob/c6389180a0fb2b47423456a15bfa09bf69e11db7/psychopy/visual/backends/pygletbackend.py#L101), which would need to become something like:

```
if pyglet.version > '1.3':
    defDisp = pyglet.canvas.get_display()
else:
    defDisp = pyglet.window.get_platform().get_default_display()
```

luketudge:
I tested the change I suggested above, and it fixes the immediate problem, so that opening a window with the pyglet backend and pyglet 1.4 runs without error.

But the refactoring in pyglet 1.4 clearly goes deeper. Creating a `TextStim` for the window leads to a similar-looking problem in [text.py](https://github.com/psychopy/psychopy/blob/c6389180a0fb2b47423456a15bfa09bf69e11db7/psychopy/visual/text.py#L360), where `pyglet.font.Text()` no longer matches the new module structure in pyglet 1.4.

So I guess there are many such changes and providing compatibility for pyglet 1.4 would be a bigger task.

peircej:
Thanks for detailing a couple of the issues. Yes, I imagine there are a number of changes we need to make to enable this switch and, for now, fixing the pyglet version is probably needed. I'll try and look later this week at what else would be required

luketudge:
[...]

PR #2565, opened 2019-08-13
BF: Pygelt 1.3 and 1.4 compatibility

mdcutone:
Fix allows psychopy to use Pyglet version 1.3 and 1.4. Not sure if I caught all the issues but the demos seem to work fine.

### DIFF
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -52,6 +52,15 @@ if haveGLFW:
 else:
     useGLFW = False
 
+
+if havePyglet:
+    # get the default display
+    if pyglet.version < '1.4':
+        _default_display_ = pyglet.window.get_platform().get_default_display()
+    else:
+        _default_display_ = pyglet.canvas.Display().get_default_screen()
+
+
 import psychopy.core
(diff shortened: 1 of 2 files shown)


mdcutone:
Don't pull this in until I refine it a bit.

hoechenberger:
pyglet 1.4.2 also [added](https://github.com/pyglet/pyglet/blob/f5491971f21d32d1403c8749fe10c1c9ec9cf0bd/RELEASE_NOTES#L8) a
> `Window.get_pixel_ratio` function for calculating scaling on HiDPI displays.

wonder if that could make things easier and less error-prone for Macs?

peircej:
I wanted to pull this in ready for a 3.2.0dev1 release, to test a range of things in a full release

peircej:
@hoechenberger that function is helpfull in detecting that this is a system with silent upscaling, but doesn't ultimately fix the central issue of how we present these screens to users.

mdcutone:
I'm going to look into the suggestions posted in  #2518. He seems to have better figured out what was exactly refactored on the Pyglet end. I'm trying to only version check once upon loading the backend, instead of evaluating the ``if version <1.4`` every time.

PR #2583, opened 2019-08-28
Update translation

hsogo:
There are some points I'd like to ask:

**1. Use prefs** 
"Use prefs" in "Audio library" parameter of the Preference dialog starts with an uppercase letter, while "use prefs" in "Units" parameter starts with a lowercase letter.  Because _translate() treats these as different strings, "Use prefs" is not translated.  I've replaced "Use prefs" with "use prefs".  Is it OK?

**2. Hint of Audio latency mode** 
Hint of "Audio latency mode" is defined multiple times.  Hint strings comes from comment lines in  psychopy/preferences/*.spec files.  In baseNoArc.spec, it is defined as follows.

```
    # latency mode for PsychToolbox audio (3 is good for most applications. See
    audioLatencyMode = option(0, 1, 2, 3, 4, default=3)
```

On the other hand, other *.specs files, it is defined:

```
    # priority
    audioLatencyMode = option(0, 1, 2, 3, 4, default=3)
```

Because definition in baseNoArc.spec is overwritten by other *.spec files, we see "priority" as the hint of "Audio latency mode" in the Preference Dialog.  I think that the hint defined in baseNoArc.spec would be more informative (is "See" at the end of string a typo?).  Can I replace definitions in other *specs files?  

**3. translation of pull-down menu items**
I noticed that `allowedLabels` attribute is added to `psychopy.experiment.params.Param` to translated "Audio latency mode" pull-down menu items.  Although it's nice feature, I'm not sure which of  `allowedLabels` or `psychopy.app.builder.localizedStrings._localized` should I use in future.  Is there any policy?  In my opinion, common parameters like "time (s)", "frame N" and so on should be translated at once using `_localized`.


peircej:
## use prefs

Agreed. I'm pushing an updated version

## hint in base spec file

Yes, my mistake. I'd failed to run generateSpec.py which transfers these strings to the specific files.
I've done that now and pushing upstream

## `_localized` or `allowedLabels`

For this particular instance I think we do need `allowedLabels` as it allows us to add a verbal  message to the name. If we used localised to turn '0' into '0: latency not important' that would be applied to all other times that '0' was used as a setting. So I think the two things allow us different options. _localised allows us to convert a string in all cases (e.g. if we want all labels to have upper case we could do it here) and allowedLabels allows custom control over each separate parameter. I'm not actually very worried about the translation being a performance issue though - I expect` _translate()` is a pretty fast lookup table (although I haven't actually tested)

hsogo:
Thank you for reply.  I understand the aim of `allowedLabels`.

PR #2591, opened 2019-08-29
ENH: An alerts package for PsychoPy integrity checks

dvbridges:
AlertLoggers are objects created in components and used for creating alerts. AlertLoggers raise error codes, which are fed into AlertEntry class data types. AlertEntry data type attributes are filled using the AlertCatalogue class, which holds a set of code definitions obtained from easy to read yaml files. AlertEntry data types are written to the AlertLog, a controller class created as a root object in the alert class. AlertsLog has a write method for adding each alert to storage (list), and a flush method for sending information to the standard stream and to the MasterLog class, and for clearing the alerts from storage.  The MasterLog stores all alerts in alert log files, one file per Python process. Only 5 most recent log files are stored. MasterLog is saved in the experiment folder.


dvbridges:
Passes for Python 3, needs pathlib dependency to pass for python 2

peircej:
CLosing for now pending changes we've discussed

PR #2598, opened 2019-09-03
TESTS: For test_scripts tests, create new files in a temp directory

dvbridges:
This fix stops the creation of many Python and JS "new" scripts
when test_scripts tests are run.


dvbridges:
Already fixed in #2597

PR #2600, opened 2019-09-04
BF: Bind EVT_CLOSE to stdoutFrame closeFrame method

dvbridges:
Without this, the stdoutFrame would not show again after being closed.


peircej:
ah. Good point

PR #2607, opened 2019-09-09
DOCS Correct typo

m-macaskill:


m-macaskill:
I could swear that this commit only affected one line but it seems to have magically grown to two (weirdly deleting the ''' at the end of the doc string, which now causes errors). I'm just going to close this now and will try again.

PR #2610, opened 2019-09-10
BF: QuestPlusHandler now accepts "unknown" kwargs

hoechenberger:
This is necessary e.g. when `QuestPlusHandlers` are created via a `MultiStairHandler`, and all "MultiStair conditions" are passed down to `QuestPlusHandler`. These are now all swallowed through a `**kwarg` in the signature. To warn users of potential typos in their code, a warning will be emitted if one or more unknown kwargs are encountered.


hoechenberger:
@peircej Thanks!!!

PR #2611, opened 2019-09-11
NF: Additional eye transformation options for stereoscopy

mdcutone:
Added some additional options to set the interocular offset and convergence distance for stereoscopic rendering. These features will be better utilized by the incoming stereomodes support, but can be used for multi-window or quad-buffered stereo already supported.


mdcutone:
Might need to squash those commits. Not sure why they are coming up here ...

peircej:
The mysteries of git ;-)

mdcutone:
Withdrawing this for now, going to add these in when the first round of stereo modes drops.

PR #2616, opened 2019-09-13
NF: Viewport rendering, OpenGL state flags, new shapes for GL tools.

mdcutone:
Adds the ability to create 3D boxes to `gltools`. Also splits `createPlane` into `createPlane` and `createMeshGrid`. `createMeshGrid` is more specialized for creating distortion meshes for display warping, they shouldn't really be used interchangeably under most uses.

### DIFF
--- a/docs/source/api/tools/gltools.rst
+++ b/docs/source/api/tools/gltools.rst
@@ -68,6 +68,8 @@ Overview
     loadMtlFile
     createUVSphere
     createPlane
+    createMeshGrid
+    createBox
     getIntegerv
     getFloatv
     getString
@@ -133,6 +135,8 @@ Details
 .. autofunction:: loadMtlFile
 .. autofunction:: createUVSphere
 .. autofunction:: createPlane
+.. autofunction:: createMeshGrid
+.. autofunction:: createBox
 .. autofunction:: getIntegerv
 .. autofunction:: getFloatv
 .. autofunction:: getString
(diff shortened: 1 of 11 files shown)


mdcutone:
Added some attributes to the window class to permit seamless 2d and 3d drawing. 2D stimuli just need to set the `draw3d` flag to False before drawing, and 3D to True. Stimuli classes can check if the window is already in the appropriate mode before drawing to prevent redundant GL calls. `draw3d` needs to be disabled prior to flipping. 

For now, only the 3D classes will check for this state and switch back to 2d when done. So, for now 2d classes will remain unchanged.

mdcutone:
Went all in with this PR, added additional OpenGL state flags to the window class. Furthermore, I went through with modifying how the `size` property works. It now reports the dimensions of the drawable area/viewport instead of the client area of the window. This is because the window size can differ from the framebuffer size (eg. retina displays), or the drawable area may be restricted to a region of the window (eg. for multiview rendering). Now when accessing the window `size` attribute, you get the framebuffer dimensions by default which is provided by the backend (deals with retina displays). Adjusting the viewport attribute causes stimulus classes to think they are drawing to a different sized window, so they are drawn correctly within the sub region.

I tested this code out with loads of stimuli and everything seems to work fine. This was a huge obstacle for stereo support that was solved by these changes. We now have seamless integration of 2d/3d drawing and the ability to compose multi-view displays.

I would greatly appreciate if someone can test this out on their retina display to double check if everything works on that end.

mdcutone:
All visual tests pass on my end on Windows. I assume they will to on Linux, but maybe not MacOS.

PR #2624, opened 2019-09-24
RF: In QuestPlusHandler, Initialize unused inherited attribs (to avoid confusion), and exclude them from being written to the MultiStairHandler log

hoechenberger:
`.currentDirection` and `.stepSizeCurrent` attribs don't make sense for QUEST+ and only lead to confusion when their default values appear in logs. The `.stepType` attrib is never used (we use `.stimScale` instead). This PR ensures that we don't write any of these attribs to the `MultiStairHandler` data files any longer: the columns `direction`, `stepSize`, and `stepType` will not be included anymore.

### DIFF
--- a/psychopy/data/staircase.py
+++ b/psychopy/data/staircase.py
@@ -1484,7 +1484,13 @@ class QuestPlusHandler(StairHandler):
             warnings.warn(msg, RuntimeWarning)
 
         super().__init__(startVal=startIntensity, nTrials=nTrials,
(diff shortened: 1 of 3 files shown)


hoechenberger:
No idea what's up with Travis there…

peircej:
I'm also puzzled why StaticTest is failing with on changes to it :-/

hoechenberger:
@peircej Travis is green now.

hoechenberger:
Thanks, Jon!

PR #2629, opened 2019-09-27
RF/BF/DOCS: gltools and mathtools

mdcutone:
A few minor changes to some functions and documentation.

* Added tessellation modes to `createMeshGrid`. This reduces artifacts when generating meshes for distortion.
* Added `lensCorrection` which generates barrel/pincushion distortion meshes and texture coordinates to correct for optical distortion due to lenses in the optical path of the eye and display (HMDs).
* Fixed an edge condition for converting matrices to quaternions, also added some examples.
* Added `calculateSurfaceNormals` to re-compute surface normals of meshes. ~Really slow right now but it works.~ Much faster calculation being used now.
* Added `createMeshGridFromArrays` which allows you to use 2D arrays to define a mesh, allowing for 3D gratings, displacements maps, terrain, etc.
* Small fixes to documentation to clarify things.
* Added `alignTo` to compute quaternions which rotate vectors to align with other vectors.
* Added support to the window class for setting up and using scene lights.


peircej:
Is this OK to be pulled in to master branch now?

mdcutone:
Fixed conflicts in merge.

PR #2632, opened 2019-09-30
BF: Check for valid Slider units on compile script.

dvbridges:


peircej:
This one bit me today as well! I think it was because we previously had code that said "not supported" when the real solution was as you've put it here.

PR #2637, opened 2019-10-04
EN: Faster sampling method for uniform circular dot fields.

mdcutone:
Sampling for circular dot fields is not deterministic, as points are sampled randomly until a condition is met. This patch changes the sampling algorithm so that sampling completes in a fixed amount of time. This might require updating the tests related to `DotStim`.

### DIFF
--- a/psychopy/visual/dot.py
+++ b/psychopy/visual/dot.py
@@ -386,25 +386,40 @@ class DotStim(BaseVisualStim, ColorMixin, ContainerMixin):
         GL.glPopMatrix()
 
     def _newDotsXY(self, nDots):
-        """Returns a uniform spread of dots, according to the
-        fieldShape and fieldSize
+        """Returns a uniform spread of dots, according to the `fieldShape` and
+        `fieldSize`.
 
-        usage::
+        Parameters
+        ----------
+        nDots : int
+            Number of dots to sample.
+
+        Returns
+        -------
+        ndarray
+            Nx2 array of X and Y positions of dots.
+
+        Examples
+        --------
+        Create a new array of dot positions::
 
             dots = self._newDots(nDots)
 
         """
-        # make more dots than we need and only use those within the circle
         if self.fieldShape == 'circle':
-            while True:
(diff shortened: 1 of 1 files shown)


psychopy/visual/dot.py
```diff
@@ -386,25 +386,40 @@ def draw(self, win=None):
...
-        # make more dots than we need and only use those within the circle
         if self.fieldShape == 'circle':
-            while True:
-                # repeat until we have enough; fetch twice as many as needed
-                new = numpy.random.uniform(-1, 1, [nDots * 2, 2])
-                inCircle = (numpy.hypot(new[:, 0], new[:, 1]) < 1)
-                if sum(inCircle) >= nDots:
-                    return new[inCircle, :][:nDots, :] * self.fieldSize * 0.5
+            length = numpy.sqrt(numpy.random.uniform(0, 1, (nDots,)))
+            angle = numpy.random.uniform(0., 2. * numpy.pi, (nDots,))
```
m-macaskill: As a joint distribution of uniformly-distributed distances and angles, I would have thought that this would produce a non-uniform distribution in cartesian coordinates (i.e. that there would be a higher density of dots in the centre compared to the periphery of the circle). Is the sqrt() the magic sauce that prevents that?

mdcutone: Yes, the square root makes density uniform across the disk. See, http://mathworld.wolfram.com/DiskPointPicking.html

bdsinger:
To my eye this seems to have fixed the issue reported on discourse with 400 dots but the researcher is doing some tests now to be sure.

psychopy/visual/dot.py
m-macaskill: Nice stuff (and great to learn something new too).

mdcutone:
Actually, I did a quick hack on the source for `DotStim` and now I can render 10K dots no problem ...

peircej:
Oh, great!

mdcutone:
Some fixes in #2647 bring the number of dots up substantially.

PR #2646, opened 2019-10-10
ENH: The alerts package

dvbridges:
- On import, the alerts catalogue id loaded into the catalogue variable as a dict
- ErrorHandler created as an app attribute, like stdoutFrame
- On compile script, the ErrorHandler and stdOutFrame are assigned to the stderr and stdout stream, respectively.
- On script compile, psyexpCompile calls integrity check method in base component and code component
- IntegrityCheck method calls alerttool tests
- When alert is found, the alerttool tests call _alerts.alert function, with code, component, strformat, and any tracebacks
- The alert function creates an AlertEntry object for each alert
- The alert function sends AlertEntry object to `receiveAlert` method in ErrorHandler, which should be on the standard error stream.
- ErrorHandler stores each alert object in its `alerts` attribute
- When flushed, the ErrorHandler, sends the `alerts` list to the `alertLog` container in the `_alerts` module. 
- When users open ProjectInfo, the `alertLog` is used to populate the new alert panel


dvbridges:
@peircej , apart from the travis error, this pull request passes.

PR #2647, opened 2019-10-10
EN/RF: Faster DotStim

mdcutone:
Went through the code and optimized some areas of `DotStim` to handle substantially more dots than before. Looks like most of the slowdown was caused by the Python `sum` function being used on `ndarrays` instead of the `Numpy` one which is several orders of magnitude faster. I also cleaned up some docstrings and comments. The documentation needs some love, really inconsistent style in there.

### DIFF
--- a/psychopy/visual/dot.py
+++ b/psychopy/visual/dot.py
@@ -508,15 +508,16 @@ class DotStim(BaseVisualStim, ColorMixin, ContainerMixin):
             outofbounds = np.hypot(normXY[:, 0], normXY[:, 1]) > 1.
 
         # update any dead dots
-        if np.sum(dead):
-            self._verticesBase[dead, :] = self._newDotsXY(dead.sum())
+        nDead = dead.sum()
+        if nDead:
+            self._verticesBase[dead, :] = self._newDotsXY(nDead)
 
         # Reposition any dots that have gone out of bounds. Net effect is to
         # place dot one step inside the boundary on the other side of the
         # aperture.
-        if np.any(outofbounds):
-            self._verticesBase[outofbounds, :] = self._newDotsXY(
-                outofbounds.sum())
+        nOutOfBounds = outofbounds.sum()
+        if nOutOfBounds:
+            self._verticesBase[outofbounds, :] = self._newDotsXY(nOutOfBounds)
 
         # update the pixel XY coordinates in pixels (using _BaseVisual class)
         self._updateVertices()
--- a/psychopy/visual/dot.py
+++ b/psychopy/visual/dot.py
@@ -44,7 +44,11 @@ from psychopy.visual.basevisual import (BaseVisualStim, ColorMixin,
                                         ContainerMixin)
 
 import numpy as np
-from numpy import pi
+
+# some constants
+_piOver2 = np.pi / 2.
+_piOver180 = np.pi / 180.
+_2pi = 2 * np.pi
 
 
 class DotStim(BaseVisualStim, ColorMixin, ContainerMixin):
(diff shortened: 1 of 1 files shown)


mdcutone:
Wow, looks like I can hit 100K dots on my machine without dropping frames. However, I think nvidia drivers handle client arrays nearly fast as VBOs. The ultimate test will be other drivers.

peircej:
I'm not surprised you can hit 10k dots with a good card. I'm a bit surprised that the changes you've made here have a substantial improvement (but happy that they do)! In fact from the test I just ran (below) it looks like going from `sig*np.pi*2` to `sig*np.pi*2.` is slightly *worse*!

Going further I think we should be preventing the repetitive calculation of `pi*2` and `pi/180`. On my tests,  improvement, but going to `sig*twoPi` cuts the calculation time in half (not very surprisingly).

``` python
import timeit

nReps = 1000

t1 = timeit.timeit('a*np.pi*2.', setup='import numpy as np; a=np.ones(1000, dtype=float)', number = nReps)/nReps
print(f"{t1*1000:.5f}ms")

t1 = timeit.timeit('a*np.pi*2', setup='import numpy as np; a=np.ones(1000, dtype=float)', number = nReps)/nReps
print(f"{t1*1000:.5f}ms with float casting")

t1 = timeit.timeit('a*twoPi', setup='import numpy as np; a=np.ones(1000, dtype=float); twoPi=np.pi*2.', number = nReps)/nReps
print(f"{t1*1000:.5f}ms with pre-calc")
```
gives
```
0.00536ms
0.00493ms with float casting
0.00205ms with pre-calc
```
so I think we should probably be creating some vars at top of module:
```
_piBy2 = np.pi/2.
_piBy180 = np.pi/180.
_2pi = 2*np.pi
```

I think the *ultimate* speed boost will be to do all this in shaders, where we can happily loop over dots with custom rules, but that's obviously a bigger job and the current solution will solve most issues.

mdcutone:
> I'm not surprised you can hit 10k dots with a good card. I'm a bit surprised that the changes you've made here have a substantial improvement (but happy that they do)! In fact from the test I just ran (below) it looks like going from sig*np.pi*2 to sig*np.pi*2. is slightly worse!

I think in C, maybe Python is faster converting `int` to numpy scalars than `floats`. 

> Going further I think we should be preventing the repetitive calculation of pi*2 and pi/180. On my tests, improvement, but going to sig*twoPi cuts the calculation time in half (not very surprisingly).

That's a good idea. One of the advantages to PsychoPy is it's real-time capabilities. Making optimizations only helps here. These optimizations are in response to user feedback, which is best indicator that optimizations should be made. 

> I think the ultimate speed boost will be to do all this in shaders, where we can happily loop over dots with custom rules, but that's obviously a bigger job and the current solution will solve most issues.

Can be done actually, we can also look into using other vertex attributes to set the colour of points individually very quickly and using GPU buffers. We can also move unit scaling into the shader instead of computing them directly on the arrays, taking up CPU resources. However, using client side arrays seems to be fast enough here, at least on my machine and drivers.

mdcutone:
I wrote a version of DotStim which can use fragment shaders (super easy with gltools). Problem however is that I need multisampling enabled to get smooth dots. I need to find a way to alpha blend pixels near the edges of the dots to make them smooth.

mdcutone:
I wrote a version of DotStim which can use fragment shaders (super easy with gltools). Problem however is that I need multisampling enabled to get smooth dots. I need to find a way to alpha blend pixels near the edges of the dots to make them smooth.

mdcutone:
I found a way to anti-alias points in shaders (using distance based AA), but it looks nasty with small points for some reason due to some artifacts that arise. Like it works, but points don't look the same as those in fixed-function mode with `GL_POINT_SMOOTH` enabled. I really don't want to push that out since those artifacts can introduce visual noise that can mess up experiments. We could use point sprites to display pre-rendered point graphics stored in textures to get around this issue. Like we can render a point within a numpy array with the correct size and fancy smoothing once at initialization and instance it all over the place during rendering.

peircej:
As you point out, the existing updates allow us to create huge numbers of dots on decent hardware. Let's leave it at that. Thanks for your efforts in improving this

mdcutone:
`glPoint*` is still used all the way up to OpenGL version 4.5 so I doubt we need to worry about shaders any time soon.

PR #2648, opened 2019-10-11
ENH: update tips_ja_JP

hsogo:
I had forgotten to update translation of TIPS!

By the way, "users list" is mentioned in the following TIP.  I think that users list was replaced by PsychoPy Discourse.  So should this be "post it to PsychoPy Discourse"?

- If you have a tip that you'd like to added here, email it to the users list.

And, should Peirce et. al. (2019) be added to the following TIP?

- PsychoPy is free. Please cite the relevant papers (Peirce 2007; 2009) if you use PsychoPy in published work


hsogo:
Thank you for your concern. There was little damage where I live.

PR #2656, opened 2019-10-18
Update staircase.py

LegrandNico:
Remove invalid `stepType` parameter description in doc (duplicate entry)


peircej:
Great. Thanks for contributing the fix. 

Ideally, for PsychoPy commits, could you add a prefix to you commit title to highlight what sort of contribution it is. In this case a better commit title would be `DOCS: removed duplicate argument in in staircase documentation`. It just helps me later track things that need to go into the changelog or pushed into a release. 

Thanks again. All contributions are useful! :-)

LegrandNico:
Ok, indeed, I should have checked the guidelines :)

peircej:
Is all good. :-) Your change is now live on the site.

Nice seahorse logo by the way

Issue #2635, opened 2019-09-30
MultiStairHandler should support method='fullRandom'


hoechenberger:
Currently, `MultiStairHandler` only supports `method='sequential'` and `method='random'`. The latter will randomize the order of staircases on each trial separately, which ensures that no staircase can ever be called on more than 2 consecutive trials. However, certain experimental designs demand a _full_ randomization across _all_ trials. This is not possible with the current implementation of `MultiStairHandler`. `MultiStairHandler` should support this behavior, which was dubbed `fullRandom` in the `TrialHandler`.

peircej:
Hmm, for TrialHandler this is done by concatenating the trials and then shuffling. The method can't be *quite* the same for staircases because they have an indetrminate number of trials (in advance) so we can't do the advance concatenation, but we could keep track of still-running staircases and do a resample-with-replacement on each trial? I think that would be the same effective strategy, right? Actually for TrialHandler2 there's the same issue and probably the same solution.

peircej:
(because in TrialHandler2 we might have an issue that a trial wants to be rerun and therefore appears again)

PR #2659, opened 2019-10-20
NF: Add fullRandom method to MultiStairHandler

hoechenberger:
`MultiStairHandler` now supports a `fullRandom` randomization method. We do this by sampling with replacement from the list of all running staircases on each "pass".

To make this testable, I also added a `randomSeed` kwarg to `MultiStairHandler`, which is used to seed a `np.random.RandomState` RNG. All randomization operations are then being done through that object.

Closes GH-2635.

### DIFF
--- a/psychopy/data/staircase.py
+++ b/psychopy/data/staircase.py
@@ -1646,8 +1646,8 @@ class QuestPlusHandler(StairHandler):
 class MultiStairHandler(_BaseTrialHandler):
 
     def __init__(self, stairType='simple', method='random',
-                 conditions=None, nTrials=50, originPath=None,
-                 name='', autoLog=True):
+                 conditions=None, nTrials=50, randomSeed=None,
+                 originPath=None, name='', autoLog=True):
         """A Handler to allow easy interleaved staircase procedures
         (simple or QUEST).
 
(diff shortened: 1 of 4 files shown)


hoechenberger:
So the tests fail because my changes actually exposed another bug in the previous implementation.

The `method` kwarg was never checked for valid values. In cases where the user wishes to invoke multiple QUEST staircases and to pass down e.g. `method='quantile'` to the `QuestHandler`, `MultiStairHandler` now raises because `method='quantile'` is not valid _for `MultiStairHandler`. So what we actually need now is an additional `method` kwarg _that gets passed down to the invoked staircase handlers and is separate from the kwarg controlling `MultiStairHandler` behavior._

Previously, `MultiStairHandler(stairType='quest', method='quantile', ...)` would invoke the `QuestHandler`s correctly, _but_ `MultiStairHandler` would always present them in sequential order.

hoechenberger:
Ideally, all kwargs that need to be passed down to the invoked staircase handlers would be passed as a separate dict, e.g., `MultiStairHandler(stairType='quest', method='fullRandom', stairParams=dict(method='quantile'))`

hoechenberger:
> Ideally, all kwargs that need to be passed down to the invoked staircase handlers would be passed as a separate dict, e.g.,
> 
> ```python
> MultiStairHandler(stairType='quest', method='fullRandom',
>                   stairParams=dict(method='quantile'))
> ```

Never mind, we already do that via `conditions`. Turns it there apparently was a bug in one of the old `QuestHandler` tests, which is now hopefully resolved.

PR #2663, opened 2019-10-28
NF/DOCS: 3D stimuli classes 

mdcutone:
A bunch of 3D stimuli classes, light class and a Phong shading material class. These classes allow you to draw 3D stimuli as easily as any of PsychoPy's 2D stimuli. 

The shader included is a basic Phong lighting model shader, which produces something that looks similar to OpenGL's legacy lighting code, however it's computed per-pixel if `useShaders` is enabled. Most stimuli can use diffuse textures loaded from file if needed. 

The coolest class is `ObjMeshStim`, which can load models from *.OBJ files and also load their materials. This class allows you to create stimuli with meshes imported from 3D editing software like Blender. Diffuse textures and materials will also be loaded automatically. Performance is very good, using vertex arrays and interleaved vertex attribute buffers. This makes these stimuli classes suitable for high performance applications like VR.

Classes `SphereStim`, `BoxStim`, and `PlaneStim` are procedurally generated within PsychoPy, you don't need to load anything from file to use them.

Materials use the built-in OpenGL lighting model if shaders are disabled, if not, you can use per-pixel lighting with the Phong shading model. Note that lighting is not real-world accurate, but is a decent approximation. Lighting and material colors us PsychoPy's color conventions, so you can use any colorspace PsychoPy supports.

3D stimuli are transformed using the `RigidBodyPose` class which shares a similar interface to PsychXR's `LibOVRPose` class. This allows you to use both to define the position of an object in the scene.


mdcutone:
Let me fix these LGTM issues before pulling.

mdcutone:
Most of those issues are not mine.

PR #2666, opened 2019-10-31
RF: Removed requirement for manually setting the depth mask.

mdcutone:
Setting the depth mask is handled automatically now. Also the draw3d attribute wont set any OpenGL flags that are already in the correct setting.

### DIFF
--- a/docs/source/api/tools/mathtools.rst
+++ b/docs/source/api/tools/mathtools.rst
@@ -91,6 +91,7 @@ Overview
     slerp
     quatToAxisAngle
     quatFromAxisAngle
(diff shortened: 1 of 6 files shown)


peircej:
I wonder why LGTM keeps saying this introduces a new alert when it clearly doesn't! :-/

peircej:
Ah, the commit page didn't show it (looks like travis needs to be set up differently to fit under "checks") but you have introduced a Python2 incompatibility by the looks:
in window.py this line isn't valid Py2 syntax:
```python
shaderFlags.append((i, *j))
```

PR #2674, opened 2019-11-07
NF: Some changes to RigidBodyPose and SceneSkybox

mdcutone:
* Added `posOri` attribute which sets the position and orientation at the same time. This is for compatibility with PsychXR objects and libraries like PyBullet.
* Skybox textures can now be specified as a list of image paths or as a single cubemap object. You can swap out cubemaps during runtime.


mdcutone:
Closing since the other PR has these features.

Issue #1143, opened 2016-03-23
"continueRoutine = True" in Builder generated code should occur earlier


m-macaskill:
To have conditional execution of a routine (e.g. a rest break every 10th trial), a code snippet like this is often suggested:

    if not loopName.thisN in [9, 19, 29]:
        continueRoutine = False

It would be logical if such code was placed in the "Begin routine" tab of a code component. But this is executed immediately prior to a routine:

    #-------Start Routine "trial"-------
    continueRoutine = True

So a change affecting `continueRoutine` in the "begin routine" tab is immediately counteracted by the template code. 

Should the above line in the compiled code be shifted to *before* where  any "begin routine" code is placed? Otherwise any attempt to end a routine has to occur in the "Every frame" tab. 

Needing to always put `continueRoutine` code in the "Every frame" tab has two disadvantages:
- Probably one frame of a routine will be presented before it can then be terminated. This could cause a visible flicker in stimuli, or some other undesired action.
- It encourages inappropriate mixing of logic, and needlessly performing a single check multiple times. That is, some `continueRoutine` checks should be evaluated only once, prior to a routine (e.g. checking a trial number), while other `continueRoutine` checks should be monitored continuously, on every frame. We should allow hose checks to be placed where they should be, logically and performance-wise.

peircej:
Good idea Mike. I went ahead and did it

m-macaskill:
Cheers. Wasn't sure if it would cause other issues. Great that it was so simple.

PR #2680, opened 2019-11-13
BF: Fix regression in continueRoutine behaviour

m-macaskill:
This addresses a regression in version 3.2.0 in which setting `continueRoutine = False` no longer has an effect in Builder code components (because the Builder-generated code overrides the custom code by always resetting `continueRoutine = True` immediately before the routine begins). All this pull request does its shift `continueRoutine = True` to be first rather than last in the order of execution prior to a routine, so that custom code can override it.

For recent bug reports, see:
https://discourse.psychopy.org/t/continueroutine-false-in-begin-routine-tab-has-no-effect/9801

and for previous development history, see:
https://github.com/psychopy/psychopy/issues/1143

### DIFF
--- a/psychopy/experiment/routine.py
+++ b/psychopy/experiment/routine.py
@@ -117,6 +117,9 @@ class Routine(list):
         # create the frame loop for this routine
         code = ('\n# ------Prepare to start Routine "%s"-------\n')
         buff.writeIndentedLines(code % (self.name))
+        code = 'continueRoutine = True\n'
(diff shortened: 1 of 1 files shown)


m-macaskill:
Unclear to me why this is failing on just one of the five Travis Build jobs, and on an unrelated test (of `test_StaticPeriod()`).

peircej:
Thanks Mike. I thnk the Travis error was a random dropout

PR #2688, opened 2019-11-21
BF: Updates to Python keyboard code.

dvbridges:
This fix adapts the keyboard code for the new keyboard component, allowing multiple keypresses to be stored so that first or last keypresses can be accessed from a list.

Waiting on #2686 before amending JS code.


dvbridges:
@peircej what could cause issues is that theseKeys is no longer a keypress object, but a list of keypress objects. Previously was `theseKeys = theseKeys[0]`, but this is gone now.

peircej:
Yes, I spotted that one, but I think the fact is that theseKeys was only ever a temporary variable (designed just to store the keys for a few lines of code) so I think it's fair to lose that. I suspect it was rarely used. For instance our demos all tend to use resp.keys instead.

PR #2690, opened 2019-11-23
TESTS, ENH: Enable Python 3.8

hoechenberger:
- Make `hardware/crs/bits.py` and `createInitFilt.py` work on Python 3.8
- Add `pytest` marker for PyGame visual tests
- Skip PyGame tests on Python 3.8 for now due to unavailability of Linux wheels

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,12 @@ before_script:
 #    fi;
 
 script:
-  - pytest --cov-config .travis_coveragerc --cov=psychopy -v -s -m "not needs_sound" psychopy
+  - |
+    if [ "${PYTHON_VERSION}" == "3.8" ]; then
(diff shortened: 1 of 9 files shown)


hoechenberger:
Python 3.8 currently failing bc `platform.dist()` has been removed. Alternative would be to use the `distro` package. https://github.com/nir0s/distro#background

hoechenberger:
@peircej I believe tests are all passing instead of those involving `pygame` -- there's currently no Python 3.8 wheel available. Can we temporarily disable `pygame` tests? How?

hoechenberger:
The pygame folks are apparently planning to release wheels of pygame 1.9.6 for Python 3.8: https://github.com/pygame/pygame/issues/1429

hoechenberger:
FYI the Python 3.8 `conda` environment creation doesn't fail although `pygame` fails to install via `pip`, this behavior has been reported here: https://github.com/conda/conda/issues/8537

hoechenberger:
@peircej Should be good to merge now.

PR #2693, opened 2019-11-25
Text pyglet1.4

peircej:


hoechenberger:
You're currently running these tests on `pyglet` 1.3.3, idk if. that's intentional :)

peircej:
Not exactly but a good thing to do I suppose!

peircej:
The changes I made are affecting both (I've moved to the new pyglet Label class even for 1.3) so we should test it there at least once! I think the issue is mostly just that the text rendering is slightly different in its aliasing - the screenshots that are non-matching look the same to the naked eye

hoechenberger:
Tests should now run on both pyglet 1.3 and 1.4.

hoechenberger:
🥳👌

PR #2695, opened 2019-11-27
ENH: Adds auto Py to JS translation for code components

dvbridges:
Translates code on change of code type, translate, and when code comp
closed using "OK" button (when translate set to auto).

Also:
- Metapensiero only supports Py 3.5+
- Py2 auto-translate provides user with msg - metapensiero required
- Adds test for transpiler

### DIFF
--- a/psychopy/app/builder/dialogs/dlgsCode.py
+++ b/psychopy/app/builder/dialogs/dlgsCode.py
@@ -101,15 +101,6 @@ class DlgCodeComponentProperties(wx.Dialog):
                 self.codeTypeMenu.Bind(wx.EVT_CHOICE, self.OnCodeChoice)
                 self.codeTypeName = wx.StaticText(self, wx.ID_ANY,
                                                   _translate(param.label))
-            elif paramName == 'Translator':
-                _translators = self.params['Translator'].allowedVals
-                self.translatorMenu = wx.Choice(self, choices=_translators)
-                self.translatorMenu.SetSelection(
-                    _translators.index(self.params['Translator'])
-                )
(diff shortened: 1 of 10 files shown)


dvbridges:
Failing because metapensiero required

hoechenberger:
It seems indentation is 8 spaces in `py2js_transpiler.py`, while we‘re using 4 spaces in other places. Don‘t know if that‘s intentional.

dvbridges:
> It seems indentation is done via tabs in `py2js_transpiler.py`, while we‘re using 4 spaces in other places. Don‘t know if that‘s intentional.

Ah, thanks for pointing this out. I thought I had changed all tabs to whitespace when copying over the code. Will convert and make another pull request.

dvbridges:
@hoechenberger , I am trying to add javascripthon as a dependency, so these tests can pass. would you mind taking a look to see what needs adding?

hoechenberger:
> @hoechenberger , I am trying to add javascripthon as a dependency, so these tests can pass. would you mind taking a look to see what needs adding?

There is no `javascripthon` on `conda-forge` yet -- I can create one later this week. Until then, you may simply get it installed via `pip` by adding an entry to the `pip` section of the `travis/environment-*.yml` files.

peircej:
🍾

hoechenberger:
It seems that `metapensiero.pj` only officially supports Python 3.5 and 3.6, and the latest update to the [git repository](https://github.com/metapensiero/metapensiero.pj) was in May 2018. Not sure if the project is still alive…

PR #2697, opened 2019-11-28
BF: set nReps as undefined if passed None

dvbridges:
The code will now compile correctly in this instance, but the code will
not run correctly. This new behaviour matches the PsychoPy trialHandler.


peircej:
So the post that says it doesn't work is here:
https://discourse.psychopy.org/t/nreps-is-not-working-online/9810

Any idea what could have gone wrong?

PR #2698, opened 2019-11-28
BF: Block SaveAs to Desktop folder

dvbridges:
WIP: Still needs:

- Block SaveAs on Desktop for Coder
- Block create project on Desktop

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -1790,17 +1790,6 @@ class BuilderFrame(wx.Frame):
 
         if dlg.ShowModal() == wx.ID_OK:
             newPath = dlg.GetPath()
-            parentPath = os.path.split(newPath)[0]
-            parentDir = os.path.basename(os.path.normpath(parentPath)).lower()
-
-            # Block SaveAs to Desktop folder
-            if parentDir == "desktop":
-                wx.MessageBox("You cannot save projects to the Desktop. Please find another folder.",
-                              "File Save Error",
-                              wx.ICON_QUESTION | wx.OK)
-                self.fileSaveAs(filename)
-                return returnVal
-
             # update exp name
             # if user has not manually renamed experiment
             if usingDefaultName:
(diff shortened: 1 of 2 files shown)


hoechenberger:
Wow, why is this being implemented? Will there be a configuration setting to change this behavior?

hoechenberger:
Also, currently this would disallow saving to _any_ folder named `Desktop`, no? It doesn't seem to specifically check if it's _the_ Desktop Folder?

peircej:
The *need* is only really for online studies. When a user pushes their experiment/script to pavlovia this creates a repository of the folder in which the file sits. If that folder is the Desktop then they unwittingly push all their desktop and contained files to pavlovia in a git repo. Actually, the same is true for My Documents of course. The issue gets even worse for roaming profiles where the desktop doesn't even exist on the local machine and currently crashes. The first part is solved simply by saving a folder on the desktop and putting your experiment in there (the 3nd part needs further digging for a solution).

Now, we could wait until they try to sync before telling them "don't do that" but to undo their foolish deed the user would then need to move the files to the folder, remember to close them in the editor, and reopen from the new location. That's more hassle, and more prone to the user making mistakes, than simply requiring them to make a folder into which they save their file.

dvbridges:
> Also, currently this would disallow saving to _any_ folder named `Desktop`, no? It doesn't seem to specifically check if it's _the_ Desktop Folder? You should probably use sth like, `os.path.expanduser('~/Desktop')`.

Cheers for the feedback Richard, I'll take a look at using `expanduser` and make the changes.

hoechenberger:
> When a user pushes their experiment/script to pavlovia this creates a repository of the folder in which the file sits.

Isn't this actually the issue that should be tackled, then? I can very easily imagine privacy / data protection issues emerging if an entire experiment folder is auto-sync'ed to the cloud. Experimenters may store participants' contact info and other personally identifiable information alongside the data and code, at least as long as data collecting is ongoing. Once this stuff has ended up in a `git` repo, even deleting the file won't make all its traces disappear… jus my 2 ct!

peircej:
Detecting that some has included a file that would break GDPR (deliberately or otherwise) is very hard. Preventing them from syncing Desktop or MyDocuments is easy (and is a bad idea for many reasons, only one of which is accidentally sending unwanted files to pavlovia). Until we have a good way to detect that users are about to send files that are not associated with the current project I remain in favour of blocking sync from the desktop.
For me the debate is just whether that happens at save-time or at sync-time

PR #2728, opened 2019-12-17
Remove sounddevice in pyo backend

cbrnr:
The `pyo` backend currently uses `sounddevice.query_devices()` to get a list of audio devices. This means that `sounddevice` needs to be installed just because of this function even if `pyo` is used for everything related to sound input/output.

Actually, `pyo` includes a query function, which unfortunately doesn't return a value but prints to stdout. I've written a small function to convert this to what PsychoPy needs. This means that the `pyo` backend no longer needs the `sounddevice` module.

### DIFF
--- a/psychopy/sound/backend_pyo.py
+++ b/psychopy/sound/backend_pyo.py
@@ -12,21 +12,9 @@ from psychopy import prefs, exceptions
 from sys import platform
 from psychopy import core, logging
 from psychopy.constants import (STARTED, PLAYING, PAUSED, FINISHED, STOPPED,
-                                NOT_STARTED, FOREVER)
+                                NOT_STARTED, FOREVER, PY3)
 import os
 import io
-try:
-    from contextlib import redirect_stdout
-except ImportError:
-    from sys import stdout
-    import contextlib
-    @contextlib.contextmanager
-    def redirect_stdout(target):
-        original = stdout
-        stdout = target
-        yield
-        stdout = original
-
 travisCI = bool(str(os.environ.get('TRAVIS')).lower() == 'true')
 try:
     import pyo
@@ -40,6 +28,20 @@ from ._base import _SoundBase
 import atexit
 import threading
 from numpy import float64
+
+if PY3:
+    from contextlib import redirect_stdout
+else:
+    from sys import stdout
+    import contextlib
+    @contextlib.contextmanager
+    def redirect_stdout(target):
+        original = stdout
+        stdout = target
+        yield
+        stdout = original
+
+
 pyoSndServer = None
 audioDriver = None
 
--- a/psychopy/sound/backend_pyo.py
+++ b/psychopy/sound/backend_pyo.py
@@ -14,6 +14,8 @@ from psychopy import core, logging
 from psychopy.constants import (STARTED, PLAYING, PAUSED, FINISHED, STOPPED,
                                 NOT_STARTED, FOREVER)
 import os
+import io
+from contextlib import redirect_stdout
 travisCI = bool(str(os.environ.get('TRAVIS')).lower() == 'true')
 try:
     import pyo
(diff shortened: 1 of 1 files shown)


cbrnr:
Obviously this works only for Python 3, but given that [Python 2 will be unmaintained in two weeks](https://pythonclock.org/) I don't think it is still necessary to support it.

peircej:
I know that Python is about to be unmaintained but many scientists will still have it on their machines and continue to use it for a while. _We_ are still supporting it until it becomes hard to support

cbrnr:
Workaround for Python 2. Also, I noticed that the three Travis jobs for Python 2.7 warn that they now use the new default Python 3.6. Despite of that, somehow the script still installs and uses 2.7, but maybe you might want to update the job config.

hoechenberger:
> Also, I noticed that the three Travis jobs for Python 2.7 warn that they now use the new default Python 3.6. Despite of that, somehow the script still installs and uses 2.7, but maybe you might want to update the job config.

Our current test setup doesn't care about the pre-installed Python, as we [fetch and install Miniconda](https://github.com/psychopy/psychopy/blob/830c3133b70a1e22999ee5fb3efb5722d66ed175/travis/20_install_miniconda_python.sh) anyway :)

psychopy/sound/backend_pyo.py
```diff
@@ -15,7 +15,18 @@
                                 NOT_STARTED, FOREVER)
 import os
 import io
-from contextlib import redirect_stdout
+try:
+    from contextlib import redirect_stdout
```
hoechenberger: Could you change the `try...except` to an `if...else`, checking for the Python version via `psychopy.constants.PY3`? That way it's more obvious why there's two different ways to handle the issue

cbrnr: Done.

cbrnr:
I've also fixed PEP8 warnings. I'll try to fix the ones I've introduced with the generator...

hoechenberger:
Nice!

peircej:
Looks good to me (waiting for travis to pass). Happy Richard?

cbrnr:
Let's wait for Travis and give me a bit more time to clean up the generator function - any ideas how to avoid the warnings on lines 38 (local variable 'stdout' defined in enclosing scope on line 34 referenced before assignment) and 41 (local variable 'stdout'  is assigned to but never used)? Otherwise, we could just silence these with `# noqa` since it's doing the right thing.

hoechenberger:
Can only look at this in detail tonight, sorry!

cbrnr:
> Can only look at this in detail tonight, sorry!

NP, let's try to get rid of the warnings. LGTM also warns because it doesn't understand how generators work 😛.

hoechenberger:
@cbrnr Thanks, looking good to me :)

hoechenberger:
btw I didn't know pyo was still a thing, so I never looked into packaging it for conda-forge. Do you think it would be worth the effort? Also, as long as there's no conda-forge package, should it be added to the `pip` section of the conda environment file?

cbrnr:
I don't have a lot of experience with the various sound backends, but I'd say it's probably not worth packaging. There's sounddevice and PTB which should be preferred over pyo, and once PsychoPy supports microphone input via sounddevice (for cases where PTB doesn't work) there shouldn't be a need to use pyo anymore.

peircej:
pyo is definitely still a thing and it has the best timing other than psychtoolbox (it is actually better performance than souddevice but sunddevice was easier to install).
pyo now has wheels, including a manylinux wheel so it's pretty easy to pip install these days

hoechenberger:
Thanks for the explanation, @peircej! Guess I'll add `pyo` to the `pip`-installable dependencies then! (in another PR)

PR #2730, opened 2019-12-17
Add SoundFile to conda env

cbrnr:
Playing sounds with `psychtoolbox` requires `SoundFile`, which I've added to the conda env so that this gets installed automatically.

### DIFF
--- a/conda/psychopy-env.yml
+++ b/conda/psychopy-env.yml
@@ -9,3 +9,4 @@ dependencies:
   - psychtoolbox
   - pygame
   - pyparallel
+  - SoundFile


hoechenberger:
Thanks @cbrnr! Which OS are you using? I only see wheels for macOS and Windows on PyPI. Will it compile smoothly on Linux too?

hoechenberger:
Documentation says:

> On Linux, you need to install libsndfile using your distribution’s package manager, for example sudo apt-get install libsndfile1.

Do you think you could add this to our install instructions too?

hoechenberger:
`soundfile` has been packaged for `conda-forge` on macOS and Linux (as `pysoundfile`), so we should probably use that on these platforms.

cbrnr:
I'm using all three platforms. I think `pip install SoundFile` worked everywhere, but I'll check tomorrow. If this is already in `conda-forge` it'd be great if you just added it to the `psychopy` package - is this possible even though there's no version for Windows?

hoechenberger:
> If this is already in `conda-forge` it'd be great if you just added it to the `psychopy` package - is this possible even though there's no version for Windows?

Oh this reminds me. It _is_ a dependency on non-Windows platforms already! So it _should work already_ on those platforms, out of the box...

cbrnr:
Alright, I can confirm that `pysoundfile` is already installed on macOS and Linux. On Windows, `pip install SoundFile` is necessary though. Assuming that we don't want yet another tool (https://conda-devenv.readthedocs.io/en/latest/), the easiest solution is to create three env files for Linux, macOS, and Windows.

cbrnr:
Or we could just update the docs with the required commands and skip the env files completely (because we only need 2 lines anyway).

macOS and Linux:
```
conda create -n psychopy -c conda-forge python=3.6 psychopy
conda activate
python -m pip install psychtoolbox pyparallel
```

Windows:
```
conda create -n psychopy -c conda-forge python=3.6 psychopy
conda activate
python -m pip install psychtoolbox pyparallel SoundFile
```

Once installed, PsychoPy can be started as follows (all platforms):
```
conda activate psychopy
psychopy
```

hoechenberger:
@cbrnr I just checked and it seems we can use environment markers in the `pip` dependencies section too. So we could stick to a _single_ environment file, and simply add the following dep that will only be installed on Windows:

```
...
pip:
  - soundfile; platform_system == "Windows"
```

And the installation instructions could remain unchanged!

hoechenberger:
(btw `pyparallel` is not required on Windows as parallel port communication is achieved through different means on that platform)

hoechenberger:
@peircej this was actually not ready to be merged ^^

hoechenberger:
@cbrnr want to make a new PR for the environment marker I proposed above?

peircej:
Oh, sorry! 🤦‍♂ I thought that was a "wouldn't it be nice if" 🙂

cbrnr:
No problem, I'll make a new PR!

PR #2732, opened 2019-12-18
Use platform markers in conda env

cbrnr:
Follow up for #2730.

It is not necessary to specify `pip` because this package is installed by default in any conda env. Furthermore, what do you think about removing `pygame`? This is the oldest backend and PsychoPy ships with better options.

### DIFF
--- a/conda/psychopy-env.yml
+++ b/conda/psychopy-env.yml
@@ -4,6 +4,7 @@ channels:
 dependencies:
 - python=3.6
 - psychopy
+- pip
 - pip:
   - psychtoolbox
   - pygame
--- a/conda/psychopy-env.yml
+++ b/conda/psychopy-env.yml
@@ -4,9 +4,8 @@ channels:
 dependencies:
 - python=3.6
 - psychopy
-- pip
 - pip:
   - psychtoolbox
   - pygame
-  - pyparallel
-  - SoundFile
+  - pyparallel; platform_system == "Darwin" or platform_system == "Linux"
+  - SoundFile; platform_system == "Windows"
--- a/conda/psychopy-env.yml
+++ b/conda/psychopy-env.yml
@@ -8,5 +8,5 @@ dependencies:
 - pip:
   - psychtoolbox
   - pygame
-  - pyparallel; platform_system == "Darwin" or platform_system == "Linux"
+  - pyparallel; platform_system != "Windows"
   - SoundFile; platform_system == "Windows"


cbrnr:
I'm currently trying to come up with a canonical installation method for all three platforms, and so far it seems like the Anaconda/Miniconda way only works on Windows and macOS (using the updated conda env file). On Linux, I found it much easier to use packages provided by the system package manager as well as a plain old Python venv (no wxPython hacking necessary). I've successfully tested this on Arch Linux and will try on Ubuntu right now. We might want to update the installation instructions with these findings.

peircej:
Indeed, on ubuntu/debian it was great back in the day when psychopy itself was available to apt-get install. I got the files from @yarikoptic to (potentially) take that on myself but never got around to it and pip/conda seemed like they might work instead (wxpython being the one major hiccup). But if you had the appetite to battle with it I could pass them on

conda/psychopy-env.yml
```diff
@@ -4,9 +4,8 @@ channels:
 dependencies:
 - python=3.6
 - psychopy
-- pip
```
hoechenberger: Please leave this dependency in as new conda versions will emit a warning otherwise. (pip-installable deps listed but pip itself not included in the list of deps)

cbrnr:
Yes, dedicated and up-to-date packages would be great, but I'm not a Debian/Ubuntu user so I won't be able to package for this distribution. However, since I'm using Arch Linux, I might play around to see how difficult it is to create an AUR package ([I already package a few tools](https://aur.archlinux.org/packages/?SeB=m&K=cbrnr)).

In the meantime, it looks like it is easy enough to just `pip install psychopy` if some required system packages are installed. I'll come back to you once I have this information for Ubuntu.

conda/psychopy-env.yml
cbrnr: Done. What about `pygame` - leave it in or remove?

hoechenberger: I would leave it in unless you know of any drawbacks this would cause?

hoechenberger:
> On Linux, I found it much easier to use packages provided by the system package manager as well as a plain old Python venv (no wxPython hacking necessary).

What do you mean by that? The required system-wide installation of `libwegkitgtk`, since this is [not yet packaged](https://github.com/conda-forge/staged-recipes/issues/6157) for `conda-forge`?

conda/psychopy-env.yml
```diff
@@ -8,5 +8,5 @@ dependencies:
 - pip:
   - psychtoolbox
   - pygame
-  - pyparallel
-  - SoundFile
+  - pyparallel; platform_system == "Darwin" or platform_system == "Linux"
```
hoechenberger: matter of taste so I'll leave this up to you, but I believe `!= Windows` would work too and would be more concise.

cbrnr:
> What do you mean by that? The required system-wide installation of libwegkitgtk, since this is not yet packaged for conda-forge?

Exactly. I just tested this on both Arch and Linux, and it works if you just install the packages provided by `pacman` or `apt`.

conda/psychopy-env.yml
cbrnr: A potential drawback could be that at some point the package on PyPI fails to build on a specific platform. But right now it works, so this can be postponed until such a problem occurs.

conda/psychopy-env.yml
cbrnr: Yes, this is better. Changed.

cbrnr:
Alright, I'd say this is ready for merge from my side if Travis comes back green.

hoechenberger:
> Exactly. I just tested this on both Arch and Linux, and it works if you just install the packages provided by `pacman` or `apt`.

Yes it's annoying. It's just a workaround as long as the package hasn't been built for conda-forge proper. In this case, one can link against CentOS packages in the so-called [Core Dependency Tree (CDT)](https://conda-forge.org/docs/maintainer/knowledge_base.html#core-dependency-tree-packages-cdt) such that the binaries will run on other platforms too.

cbrnr:
OK. Though it's always good to know that there is a working alternative, in this case system Python with system packages, that works out of the box. Anaconda is great, but sometimes I want to avoid it.

hoechenberger:
> OK. Though it's always good to know that there is a working alternative, in this case system Python with system packages, that works out of the box. Anaconda is great, but sometimes I want to avoid it.

Why not provide an additional `requirements.txt` for those who prefer `pip` installs?

hoechenberger:
@peircej This is good to merge once CIs go green

cbrnr:
> Why not provide an additional requirements.txt for those who prefer pip installs?

For people who prefer to have everything in a `venv` this might be a good idea. However, because packages provided by the package manager usually work better than pip packages, this won't be a "manylinux" solution. Many pip packages require that an underlying library is installed, and it is still up to the user to install these. Installation instructions will across Linux distributions, but it would probably be OK to provide instructions for the most popular ones (Ubuntu) and the one I use (Arch) 😄.

hoechenberger:
Agreed. This is what we were hoping conda would provide, but do far it only manages to do so to some extent.  Btw great to hear that Arch still seems to be alive and popular, used to be my favorite Linux distro many years ago (before I switched to macOS)

hoechenberger:
BTW there is one major drawback with system packages: they install things globally. You cannot have multiple concurrent installations (e.g., different versions) of the same package. But this is crucial for my research work

peircej:
Thanks both! 😄

cbrnr:
@hoechenberger yes, conda is a great solution, let's hope that installing PsychoPy will be as easy as on Windows and macOS. I agree that system-wide packages do not cover all use cases, but it's good to have options. Reproducibility is extremely important and conda envs can help, but they don't containerize everything that might be important (there are cases where results differ due to different implementations and/or settings such as max threads in numerical libraries). Let's not start talking about Docker though 😄. And yes, Arch is still the best and very much alive (although I also mainly use macOS nowadays and even Windows sometimes, since it has become pretty usable).

peircej:
I've never used Arch. On Linux I'm mostly ubuntu (but with cinammon because it seems to fall over less often). Beware the mac for screen timing since 10.13. 😞

PR #2735, opened 2019-12-19
Replace platform.dist with distro.linux_distribution

### LINKED ISSUE #2733: platform.dist removed in Python 3.8
The attribute `platform.dist` has been deprecated since Python 3.5 and has been removed in Python 3.8 (https://docs.python.org/3.5/library/platform.html#platform.dist). This is used in [`sendUsageStats`](https://github.com/psychopy/psychopy/blob/master/psychopy/app/connections/sendusage.py#L42), [`updates.py`](https://github.com/psychopy/psychopy/blob/master/psychopy/app/connections/updates.py#L623), and [`testSoundLatency.py`](https://github.com/psychopy/psychopy/blob/master/psychopy/demos/coder/hardware/testSoundLatency.py#L99).

I don't know if this functionality can be replaced by something from the standard lib (if some loss of information is acceptable), otherwise this means yet another dependency such as https://pypi.org/project/distro/.

cbrnr:
Fixes #2733.

I've only replaced the `platform.dist` calls, but I noticed problems that should be fixed in another PR:

- https://github.com/psychopy/psychopy/blob/master/psychopy/demos/coder/hardware/testSoundLatency.py#L99 concatenates a `str` and a `tuple`
- https://github.com/psychopy/psychopy/blob/master/psychopy/app/connections/updates.py contains undefined names `unicode`, `time`, and `platform`

### DIFF
--- a/createInitFile.py
+++ b/createInitFile.py
@@ -6,8 +6,11 @@
 
 from __future__ import absolute_import, print_function
 from setuptools.config import read_configuration
-import os, copy, platform, subprocess
+import os
+import platform
+import subprocess
 from psychopy.constants import PY3
+
 
 thisLoc = os.path.split(__file__)[0]
 # import versioneer
@@ -30,7 +33,7 @@ def createInitFile(dist=None, version=None, sha=None):
     """
     # get default values if None
     if version is None:
-        with open(os.path.join(thisLoc,'version')) as f:
+        with open(os.path.join(thisLoc, 'version')) as f:
             version = f.read().strip()
     if sha is None:
         sha = _getGitShaString(dist)
@@ -48,12 +51,13 @@ def createInitFile(dist=None, version=None, sha=None):
                 'platform': platformStr}
 
     # write it
-    with open(os.path.join(thisLoc, 'psychopy','__init__.py'), 'w') as f:
+    with open(os.path.join(thisLoc, 'psychopy', '__init__.py'), 'w') as f:
         outStr = template.format(**infoDict)
         f.write(outStr)
     print('wrote init for ', version, sha)
     # and return it
     return outStr
+
 
 template = """#!/usr/bin/env python
 # -*- coding: utf-8 -*-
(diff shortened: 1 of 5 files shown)


hoechenberger:
Thanks @cbrnr, good idea to make this a platform-dependent import! We should probably do the same with `createInitFile.py` and the respective `setup_requires` entry in `setup.cfg`.

I noticed you call `distro.linux_distribution()` without further arguments; but back when I switched `createInitFile.py` to `distro`, I opted to [invoke it via `distro.linux_distribution(full_distribution_name=False)`](https://github.com/hoechenberger/psychopy/blob/ecef00b0dbe0a06192ce2154a03204fe23d0b3da/createInitFile.py#L144), because I felt only with this argument it would reproduce the original behavior of `platform.dist`. Was I mistaken?

cbrnr:
> Thanks @cbrnr, good idea to make this a platform-dependent import! We should probably do the same with `createInitFile.py` and the respective `setup_requires` entry in `setup.cfg`.

I can do that in this PR.

> I noticed you call `distro.linux_distribution()` without further arguments; but back when I switched `createInitFile.py` to `distro`, I opted to invoke it via `distro.linux_distribution(full_distribution_name=False)`, because I felt only with this argument it would reproduce the original behavior of `platform.dist`. Was I mistaken?

The argument `full_distribution_name` returns the full name if `True` and the short name if `False` (from [this list](https://distro.readthedocs.io/en/latest/#distro.id)). The deprecated [platform.linux_distribution](https://docs.python.org/3.5/library/platform.html#platform.linux_distribution) also had an argument `full_distribution_name` which was `1` by default, so it would return the full name. Therefore, I think the closest match is to call `distro.linux_distribution()` with the default `full_distribution_name=False`.

cbrnr:
Any idea why the platform marker doesn't work for `setup_requires`?

hoechenberger:
> Any idea why the platform marker doesn't work for `setup_requires`?

No clue, I googled a bit but couldn't find anything… this is weird!

hoechenberger:
Seems like it's trying to install a _package_ `platform_system`.

hoechenberger:
Seems to be working now!

cbrnr:
Alright, what a difference a newline makes.

peircej:
Indeed! Actually |I might have thought of that. I remember once before battling for ages with a cfg file - I thought it didn't support the version syntax I was using but in the end it turned out to be unhappy about a missing new line. Very picky parser!

Happy Christmas guys and thanks for the help! 

By the way @cbrnr you know we're going to be hiring soon? Any chance you're interested in working on psychopy professionally, full-time?

Issue #2741, opened 2020-01-05
Check number of available screens while initialising gui.Dlg


colinquirk:
Using Windows 7 and psychopy3-3.2.4 and window type=pyglet, I'm finding that dialog boxes are not correctly centering when gui.dlg.screen = 1 and there is only 1 available monitor. For a small dialog box, this resulted in the dialog being completely off the screen and unable to be interacted with. This issue does not appear on os x for me, dialog boxes are still centered. Furthermore, on both os x and windows, the visual.Window seems to be created and displayed normally.

Changing screen to 0 fixes the issue, but the code then is no longer correct when used on a different, 2 monitor set up. This is a small issue obviously, and I'm not even sure that the behavior I was relying on was ever purposefully designed, but it would be nice if it worked. I confirmed it worked on version 1.90.3. The following repex is enough to see the issue for me:

```
psychopy.gui.Dlg(screen=0).show()
psychopy.gui.Dlg(screen=1).show()
```

peircej:
We weren't doing anything within PsychoPy to assert this so it was presumably an artefact of the underlying WX/Qt implementations, which may have changed. But I agree that it would be good to do this (in both widget engines).

It should be easy enough to do:
- during `Dlg.__init__` we should detect the number of screens. For WX that can be done with this https://wxpython.org/Phoenix/docs/html/wx.Display.html and I'm sure there's somethign similar for PyQt
- set `screen = min(screen, nScreens-1)`

peircej:
The equivalent call in PyQt is uses the `Application.screens()` method:
https://doc.qt.io/qt-5/qguiapplication.html#screens

colinquirk:
> We weren't doing anything within PsychoPy to assert this so it was presumably an artifact of the underlying WX/Qt implementations, which may have changed.

I figured as much, I will try to take a crack at fixing this!

peircej:
Just touching base on this - it was done, right? At least in Qt. I thnk we don't care about wx anyway because it currently has a threading clash with pyglet that makes it useless!

colinquirk:
Then yup, I think you are free to close!

peircej:
Brilliant! Thanks for your help!

PR #2742, opened 2020-01-06
BF: Use primary display for Dlg if screen is unavailable

colinquirk:
Addresses #2741, but only in PyQt.

This change makes Dlg use the primary display if the screen that was asked for does not exist. As I use standalone and therefore PyQt, this fixed the issue for me.

### DIFF
--- a/psychopy/gui/qtgui.py
+++ b/psychopy/gui/qtgui.py
@@ -125,7 +125,9 @@ class Dlg(QtWidgets.QDialog):
             raise RuntimeWarning("Dlg does not currently support the "
                                  "style kwarg.")
         self.size = size
-        self.screen = screen
+
+        nScreens = len(qtapp.screens())
+        self.screen = -1 if screen >= nScreens else screen
         # self.labelButtonOK = labelButtonOK
         # self.labelButtonCancel = labelButtonCancel
 
--- a/travis/environment-2.7.yml
+++ b/travis/environment-2.7.yml
@@ -32,6 +32,7 @@ dependencies:
 - pip!=19.0,!=19.0.1,!=19.0.2
 - psutil
 - pyglet>=1.3,<1.4
+- glfw<3.3.1
 - pyglfw
 - pyopengl
 - pyosf
--- a/travis/environment-3.6.yml
+++ b/travis/environment-3.6.yml
@@ -32,6 +32,7 @@ dependencies:
 - pip!=19.0,!=19.0.1,!=19.0.2
 - psutil
 - pyglet>=1.3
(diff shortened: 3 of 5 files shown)


colinquirk:
When playing around with wxpython, I was getting complaints about using v4 instead of v3 ([related?](https://discourse.psychopy.org/t/psychopy3-wx-clashes-with-pyglet/4946)). Regardless, the Dlg defined in wxgui.py does not take a screen parameter, so it's a bit different. So my questions are:

1. Is there some easy way to force psychopy to use wxpython?
2. Given that I followed the instructions [here](https://www.psychopy.org/developers/repository.html) with the addition of a virtual environment, is there some known workaround for dealing with the error described above (it seems like wxpython==3.x is not pip installable)? FWIW, I'm working on macOS.
3. Is there a reason that the wxgui Dlg doesn't take a screen parameter and would you accept me adding one? If so, what should it be labeled? FF?

peircej:
Well done.

1. You can import a specific module like this: `from psychopy.gui import wxgui as gui`
2. Yes, wxPython 3 is indeed not pip-installable. They never built wheels for that. And I had forgotten the issue that wx4 was breaking pyglet windows (I actually can't now remember what it did to the pyglet environment to break it) meaning it shouldn't really be used for dialogs. Hmm, I guess that raises a questionmark about whether this is worth doing in wx
3. No, I suspect it's just that someone added it that wasn't using WX, so only added it to Qt. great if you felt like making the interface more consistent. Go ahead. I think you could call it ENH, in that you're extending the existing functionality. Arguably it's a BF, but in this case it doesn't actually matter because I'm not planning on releasing further bug-fix releases for 3.2.x and that's mostly what I use those tags for

peircej:
@hoechenberger any idea why these tests are failing? Seems to be something with glfw, but the code changes in this PR are unrelated and minor 😦

hoechenberger:
> @hoechenberger any idea why these tests are failing? Seems to be something with glfw, but the code changes in this PR are unrelated and minor 😦

Not sure, but `glfw` was updated about 5d ago, maybe they introduced some kind of change?

https://anaconda.org/conda-forge/glfw

hoechenberger:
@peircej we could try to work around this by explicitly adding `glfw` to the environments and pinning it to `<3.3.1`, i.e., `glfw <3.3.1`

hoechenberger:
@peircej Seems to be working now thanks to 5be0198

@mdcutone Do you have time to look into `glfw` 3.3.1 support?

mdcutone:
I'll take a look at it later today and see what I can do.

PR #2743, opened 2020-01-07
Versioneer2020

peircej:
OK @hoechenberger I'm reworking the use of versioneer at the same time as switching  to 
  <year>.<release>.<minor> 
format

My strategy is to use versioneer to create the version number during setup.py but then to bake this in to the distributed whl/zip/dmg files rather than trying to allow its use in the distributed versions too.

One issue I see with versioneer though - assuming that most developers don't pull the tags down when they do `pull upstream master` won't that mean they get a meaningless version number reported? That will break various other things (like data files and experiments knowing what version they were built with).

So, I feel like the default behaviour within the repo is *not* to use versioneer and only use this during setup.py to build the distributions. Or am I missing something?


hoechenberger:
@peircej I'm actually not sure if moving to `versioneer` in 2020 is such a great idea, the project seems to be pretty unmaintained :\ I guess we should be looking into using [`setuptools-scm`](https://pypi.org/project/setuptools-scm/) instead these days!

peircej:
Now he tells me! ;-)

hoechenberger:
@peircej I'm actually not sure if moving to `versioneer` in 2020 is such a great idea, the project seems to be pretty unmaintained :\ I guess 

> Now he tells me! ;-)

😬

Sorry!

peircej:
No worries. It didn't take me long. OK, so I'm going to revert fully to using the version file and remove all mention of versioneer!
`use_scm_version` does look more customisable but I think it suffers the same issue that developers won't typically have the repo tags downloaded so they won't have any version number (or they'll forget to `pull upstream master -t` and have an incorrect version number)

PR #2747, opened 2020-01-08
NF: API plugin loader (WIP)

mdcutone:
This is sort of a beginnings of a plugin system for PsychoPy which allows people to extend the PsychoPy API with their own packages. The concept is simple, just a function which can import packages and load their attributes into the namespace of an existing core PsychoPy module. This is a work in progress.

As an example of why you would want this, say you have a class you want people to use, but it's not ready for inclusion into core PsychoPy because it's under development, has an incompatible license, or whatever. You can package your class and upload it on PIP, where people can download it. If your package follows certain conventions, PsychoPy running on the same Python installation the package was installed to will be able to find and load it automatically. This can go further, like allowing you to distribute a custom backend or hardware interface.

Since plugins are developed by third-parties, the specification for plugins needs to be decided upon and stuck to since changes can break the software ecosystem.

**Tasks related to the specification for plugins that need dev team feedback and approval:**

**Package metadata**

- [ ] Identification of plugins installed on the system (metadata, etc.).
- [ ] Handling of platform specific plugins (eg. Windows only).
- [ ] Versioning and compatibility.
- [ ] Author and citation data.

**Packaging and usage**

- [ ] Packaging requirements and naming conventions.
- [x] Loading and assignment of objects to namespaces.
- [x] Handling of namespace conflicts.
- [ ] Determine how to extend Builder functionality with plugins.
- [x] Plugins for patching and hotfixes (optional).
- [ ] Add documentation for plugin developers.

**Software related**

- [x] Write plugin loader function, decide where to put it.
- [ ] Tests for plugin system.

### DIFF
--- a/docs/source/api/plugins.rst
+++ b/docs/source/api/plugins.rst
@@ -1,0 +1,5 @@
+:mod:`psychopy.plugins` - plugin utilities for extending PsychoPy
+=================================================================
+
+.. automodule:: psychopy.plugins
+    :members: loadPlugins
--- a/docs/source/developers/pluginDevGuide.rst
+++ b/docs/source/developers/pluginDevGuide.rst
@@ -71,6 +71,24 @@ operation of previously loaded plugins.
 Plugins can contain executable code which could run when loaded. For instance,
 a routine to initialize something so the user doesn't have to explicitly.
 
+Acceptable Use Policy for Plugins
+---------------------------------
+
+If one chooses to make available their plugins, regardless of the method of
+distribution, they must abide by the following policies for acceptable use.
+
+1) Plugin developers must be as transparent as possible regarding the intent of
+   their software.
+2) Plugins must not perform actions which may compromise the security or
+   privacy of the user.
+3) The software must not perform unsolicited: file operations (access,
+   modification, or deletion of files on the users computer or network), changes
+   to the user's computer hardware or software configuration, or transmission of
+   data over a network (eg. usage data).
+4) Plugins which modify PsychoPy's existing code should be focused on a specific
+   feature. Avoid creating a plugin which modifies multiple, unrelated aspects
+   of the software.
+
 Plugin packages
 ---------------
 
@@ -138,7 +156,7 @@ plugin loader. You can include ``__all__`` in your plugin module if you wish to
 allow it to be imported conventionally, or to expose non-PsychoPy related
 objects.
 
-Optional ``__load()`` and ``__shutdown()`` functions
+Optional ``__load()`` and ``__shutdown__()`` functions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Some plugins may need to execute code when loaded or to clean up when PsychoPy
--- a/docs/source/developers/pluginDevGuide.rst
+++ b/docs/source/developers/pluginDevGuide.rst
@@ -1,7 +1,7 @@
-.. _pluginGuide:
+.. _pluginDevGuide:
 
-Using Plugins to Extend PsychoPy
-================================
+Extending PsychoPy with Plugins
+===============================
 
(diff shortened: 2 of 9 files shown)


mdcutone:
Proof of concept for the current plugin system. This is how the above function works as of now with a package. For this example, it's just some class for rendering stimuli called `mystim` that I want to show up in `psychopy.visual` as `MyStim`.

You can create a package that has the following structure::

```
psychopy-mystim
|_ setup.py
|_ psychopy_visual_mystim
   |_ __init.py__
   |_ stuff.py

```

Your setup file should look like this:

```
from setuptools import setup

setup(
    name='psychopy-mystim',
    version='1.0.0',
    description='A PsychoPy plugin for are really cool stimulus.',
    long_description='',
    author='mdc',
    author_email='blah@blah.com',
    license='MIT',
    packages=["psychopy_visual_mystim"]
)
```

The `psychopy_visual_mystim/__init__.py` file should contain the following::

```
__all__ = ['MyStim', 'myFunc']
from psychopy_visual_psychxr.stuff import MyStim, myFunc
```

The last file is just a module containing `MyStim` and `myFunc`. It could look like this:

```
__all__ = ['MyStim', 'myFunc']

class MyStim(object):
    def __init__(self):
        pass

def myFunc():
    return 2 + 2
```

Calling the setup script with the ```python setup.py install``` command to install it alongside PsychoPy. You can access the plugin using the following statements:

```
import psychopy.visual as visual

import psychopy._plugins as _plugins
_plugins.loadPlugins('psychopy.visual')  # load the plugins

result = visual.myFunc()  # now myFunc is in visual
```

mdcutone:
Made a change so you can pass the module object itself to `loadPlugins`.

```
import psychopy.visual as visual
import psychopy.plugins as plugins
plugins.loadPlugins(visual)
```

mdcutone:
Added a `paths` argument so you can load plugin packages as zip files that reside outside of what's listed in `sys.paths`. This means that you can download plugin packages as ZIP files as store them in a local repository.

mdcutone:
Some ideas:

Have an `__extends__` directive in the top-level package of the plugin so the plugin defines where to put the objects. The value of `__extends__` can be a string, which is the fully qualified path to the module in PsychoPy to extend:

```python
__extends__ = 'psychopy.visual'
__all__ = ['MyStimClass']
```

Another option I'm leaning towards is to allow `__extends__` to be a dictionary so a plugin can install itself into multiple modules. Where the keys are the modules and the values are objects to import into the namespace:

```python
__extends__ = {'psychopy.visual': ['MyStimClass'], 'psychopy.core': ['myTimingFunction', 'anotherTimer']}
```

With a dictionary, we don't need to declare an `__all__` statement. This makes sense since plugins should never be imported explicitly with a `from ... import *` statement.

If `__extends__` is not defined in the file, then the plugin importer should probably throw an error. This should make it easier to delineate plugin modules from other packages. If `__extends__` is empty/None, then the plugin will still be loaded, but will not modify any namespaces. Since the plugin is still imported, the handle to the module will be present in the dictionary returned by `loadPlugins`.

mdcutone:
I tried out the slightly more complex version of ``__extends__`` and it seems to be the best route. It's pretty flexible and allows you to modify classes and modules without needing to do any monkey patching from within the module. I also added a mechanism to handle plugin conflicts. The system warns if two plugin attempt to export the same objects. This ensures that there is no undefined behavior arising from such a situation. There is also a bit of protection to prevent the plugin system from modifying anything other than PsychoPy. 

Through a test plugin, I was able to modify the behavior of several methods of the Window class without sub-classing it. Another benefit I've noticed is that I no longer need to work within PsychoPy's source code to make substantial changes.

Anyways, I split the documentation up so there is a plugin page for regular users and one for developers. I'm leaving out chunks of the developer page until we finalize the plugin implementation. I think we should also put a repo up somewhere with a sample plugin project.

PR #2756, opened 2020-01-10
Refactoring alerts

peircej:


peircej:
Looks like 2ec045f is the problem. All other commits are now pulled in

PR #2760, opened 2020-01-12
RF: alerts definition should be allowed to store all info

peircej:
Previously it was set so that the alert in the script/app didn't
have access to the full details (e.g. long descr) but that's unecessary
and for generating the web pages it's easier if it *does* have access to
all parts of the alert info


hoechenberger:
Hey @peircej I just realized there's still some British spelling in the paths, namely `psychopy/alerts/alertsCatalogue/` I guess you may want to change that to American at some point. Cheers!

PR #2767, opened 2020-01-19
Added instructions for installing using brew

yuchaoran2011:


peircej:
Thanks. Didn't know that was possible! Was it you that added the homebrew cask? Is this something that needs manually updating with new releases?
thanks, Jon

yuchaoran2011:
@peircej Thanks for merging. I didn't add psychopy to the brew cask. I happened to find it [here](https://formulae.brew.sh/cask/psychopy) and saw that it's the latest version in the cask. Not sure who has been maintaining it.

PR #2772, opened 2020-01-24
NF: Plugin framework for extending PsychoPy.

mdcutone:
Resubmitting to squash loads of pointless commits.

Initial work on a plugin framework for PsychoPy based around conventions established by `setuptools` (https://packaging.python.org/guides/creating-and-discovering-plugins/). Packages installed on a system can define entry points to objects to either extend or alter the behavior of PsychoPy in some way at runtime without modifying the local installation files. None of these changes persist across sessions unless the plugin is reloaded. For instance, you can create a plugin which adds a new stimulus class to `psychopy.visual` when loaded, or modify how the `flip()` method works. Existing projects can define entry points in their `setup.py` files, allowing one to add PsychoPy functionality without contributing code directly to the project, and allowing their libraries to be used standalone.

There are a few functions in the new `psychopy.plugins` module for doing this:

* `loadPlugin` - Loads objects from a target package if the package defines entry points. You can also specify optional arguments to this function which are passed to the module when registered.
* `listPlugins` - Discovers all packages installed on the system that advertise entry points for PsychoPy, or that have been already loaded. 
* `computeChecksum` - Plugin authors and users can compute checksums to validate the integrity of package archives they receive.

Here are some examples of loading plugins with this system:

```python
import psychopy.plugins as plugins
# Load a plugin by specifying its package/project name::

plugins.loadPlugin('psychopy-hardware-box')

# You can give arguments to this function which are passed on to the plugin's
# `register()` function which is called prior to resolving entry points:

plugins.loadPlugin('psychopy-hardware-box', switchOn=True, baudrate=9600)

#Load all plugins installed on the system into the current session::

for plugin in plugins.listPlugins():
    plugins.loadPlugin(plugin)

```


mdcutone:
Don't know why the build is failing here. I might just make a new PR with only the changes made.

peircej:
Not your fault. It's change in numpy (more strict about float/int in linspace) and travis has obviously updated some of its numpy versions. Is fixed in 30016b8521f10300ea47a17334cf028f6bb0f5be

PR #2773, opened 2020-01-28
Better runOnce code

peircej:
@hoechenberger might want to check that this doesn't break code for pump init from Builder


peircej:
Sorry @hoechenberger the key commits for you are 2911a1c  (and potentially the tests changed at f3a7392 ). I hadn't realised that the docs changes hadn't been pulled in already :-/

peircej:
I checked the import code for qpump still seems to work as before and Cedrus button boxes don't now error if they get used in multiple routines! 🎉

PR #2775, opened 2020-01-29
ENH: Optional enhancement for the experiment runner

dvbridges:
This enhancement uses AuiManager to set the RunnerFrame, allowing
resizable panels for the ListCtrL & buttons, and the stdout rich
text ctrl. The panels and frame have a minimum size set in order
to keep buttons visible. There is also a status bar.


dvbridges:
Needs more work, panels do not resize when hidden and reopenes

PR #2781, opened 2020-02-02
FF: Provide message when user tries to run without selecting task.

dvbridges:
If no task is selected, this fix catches the button event before
other functions can be called, and provides a message to the user
to select a task before attempting to run. This will become an alert.


peircej:
I'm going to close this one. I think the better approach is to disable the button when it doesnt make sense rather than allowing the click and then complaining.
I've made that and a few more tweaks here:
https://github.com/psychopy/psychopy/pull/2782

PR #2783, opened 2020-02-04
DOCS: add version control page, linked to from the FAQ section

AnonZebra:
This leads to, on the FAQ section main page, links to the different subheadings in the document also, see attached screenshot. I wasn't able to disable that unfortunately - please have a go at doing so if you want to.
FAQ section links

For background on the suggested new page, see https://discourse.psychopy.org/t/best-practices-for-using-multiple-psychopy-versions-on-one-computer/10749/10 .

### DIFF
--- a/docs/source/faqs/versionControl.rst
+++ b/docs/source/faqs/versionControl.rst
@@ -1,0 +1,29 @@
+How do I run experiments written with older versions of PsychoPy?
+------------------------------------------------------------
+
+If you perform experiments on computers shared by different research groups (e. g. at a shared experimental facility), it's possible that their experiments and yours are written using different versions of PsychoPy. Or maybe you yourself have some older and some newer experiments. In such a situation, it's important to ensure that the right version of PsychoPy is used for the right experiment.
+
+In `PsychoPy standalone <https://www.psychopy.org/download.html>`_, there is an easy-to-use system for controlling what version of PsychoPy is used. 
+
+Version control using Builder View
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


AnonZebra:
> Thanks for adding this very useful new section to the docs. It builds correctly and appears in the "FAQ" table of contents.
> 
> My only query is about this: "Add the function call psychopy.useVersion('<version_no>') (e. g. psychopy.useVersion('1.90.2')) directly below your import statements."
> 
> I'm not really sure about this myself, but from looking at the code that Builder generates, should the sequence actually be that you `import psychopy`, then call `psychopy.useVersion()`, and then do all other imports (including any from `psychopy` itself, so that the correct version is used)?
> 
> If that is the case, do you want to make that slight change, and then we can get this incorporated into the docs?
> 
> Lastly, I note that the Travis build failures here seem to be unrelated to this PR itself.

Good point, I've edited the .rst doc as per your comment after double checking the sources I had used. Please see the latest commit+commit comment. If I understand things correctly, the pull request doesn't need to be renewed or anything, the edit should just be included automatically. If I do need to do something more, please send me a message.

m-macaskill:
Looks good to me, great to have an explicit example of the import order. Yes, there is no need to renew anything for the PR to be accepted as-is. 

Thanks for contributing - this will help clarify things for many other users.

peircej:
Sorry for the delay on this. I was thinking a little bit about some restructuring (because the FAQ page isn't well-populated and because I wanted to check where else useVersion was documented). I.e. I was trying to work out if this was the right place for it to go, but let's just go with it! :-)

thanks for your addition!

PR #2799, opened 2020-02-11
FF: Stripping text field was causing error when field was None.

dvbridges:
This caused failure to compile script.

### DIFF
--- a/psychopy/alerts/alerttools.py
+++ b/psychopy/alerts/alerttools.py
@@ -1,10 +1,12 @@
 import ast
+from os import environ
 from esprima import parseScript
 from numpy import array
 
 from psychopy.tools import monitorunittools
 from psychopy.alerts._alerts import alert
 
+_TravisTesting = environ.get('TRAVIS') == 'true'
 
 class TestWin(object):
     """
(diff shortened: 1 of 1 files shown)


dvbridges:
@peircej , how about setting the win size manually, if Travis is running. That way, eval does not get called.

dvbridges:
@peircej , this has now passed the Travis testing, so good to pull in.

peircej:
Ideally I'd like to work out why it was failing (so we know if that can happen in any real-world setting too). When it failed the error was that eval (or literal_eval) needs a string. My guess was that it was sometime given None, but maybe sometimes it's given something else. Like, maybe a numpy array or a list instead of string (ie already converted so do nothing)?

dvbridges:
> When it failed the error was that eval (or literal_eval) needs a string. My guess was that it was sometime given None, but maybe sometimes it's given something else. Like, maybe a numpy array or a list instead of string (ie already converted so do nothing)?

Good point, I will have read about the failure and maybe some pull requests to test different scenarios.

dvbridges:
I think it is failing because it is passed a tuple, rather than a string. So, we should test whether it is a string before trying to eval it. Ill make another fix

Issue #2816, opened 2020-02-27
Working directory issues with 2020.1.1 (most likely related to the new run window)


jfkominsky:
See https://discourse.psychopy.org/t/working-directory-change-with-2020-1-and-later/11143 and https://discourse.psychopy.org/t/both-the-stim3d-py-and-trialhandler-py-programs-with-psychopy-2020-1-1-do-not-run-properly/11201

At least on Mac (have not tested on Windows), since upgrading to 2020.1.1, if you run a script, the working directory is set to /Applications/PsychoPy3.app/Contents/Resources, rather than the directory that contains the script. This completely breaks any script that uses relative paths for relevant files expecting the working directory to be the script's directory. This includes some of the demos (see second link above) and my PyHab project.

In the short run you can just add the following to the top of every file where it matters:

```
import os
os.chdir(os.path.dirname(os.path.realpath(__file__)))
```

but this is basically a kludge and bad for backwards-compatibility. Does it need to be this way or can we get the new running window to use the script's directory as the working directory by default?

peircej:
I'm sure there's no need for it to be this way. Looks like the code copied over to Runner class 

https://github.com/psychopy/psychopy/blob/3e8adf5debf381169a8598e799dbd0b71243bfa7/psychopy/app/runner/scriptProcess.py#L91

came from Builder 

https://github.com/psychopy/psychopy/blob/7d6b1a878f124999f986828e612bc6b36b68963c/psychopy/app/builder/builder.py#L2198

Now, Builder doesn't strictly need to update the cwd because your solution (set cwd to __file__) is done automatically in the Builder-output scripts, so it was probably left out but for hand-written scripts I definitely think it's amore expected thing to have the cwd automatically set to the script folder, so I think it would be cool to update scriptProcess as above to point the process cwd to the necessary location. Do you want to submit a PR on that or shall I?

jfkominsky:
I'm traveling, so probably faster if you do it, but if I get a moment and
you haven't already taken care of it I'll take a crack at it.

jfkominsky:
#2819 Submitted. NOTE: I don't have access to a Windows machine for a couple days to test it on Windows, and I highly recommend testing it. I couldn't just create a path object because the cwd argument of popen does not accept path objects in windows prior to Python 3.7, according to docs. It has to be a string. So, I had to manually get the last / or \ from the filename, in the conditionals that are already in the RunFile function for dealing with Windows-specific stuff.

[...]

PR #2819, opened 2020-02-29
BF: #2816

jfkominsky:
Attempted fix for #2816 

The new runner changed how the cwd was set so that it no longer used the file's directory as the working directory, and instead defaulted to the app folder. This fix uses the fileName string and uses rfind to locate the last instance of a folder delimiter ('\\' for windows, '/' for everything else) to set the working directory to the path of the current file. Tested on Mac, not yet tested in Windows.


jfkominsky:
Updated fix uses PathLib, and has been tested on Windows. Good to go.

peircej:
Thanks Jonathan. Oh, for commit messages try to prefx what type of change this is. BF indicates to me something that should be ported to bug-fix releases (we want this in 2020.1.3) whereas enhancements (ENH) and (new)FeatureFixes (FF) should be held back until 2020.2.0 
cheers!

PR #2827, opened 2020-03-06
NF/EN: Error dialog for unhandled exceptions in the GUI thread

mdcutone:
Adds an error dialog box which comes up when an unhandled exception occurs in the GUI application thread, preventing them from going unnoticed until the user checks the output box. It's also a bit more polished than simply dumping errors to the output.

errormsg

Displays a message informing the user that an internal error occurred and presents the traceback. The dialog has buttons to copy or save the traceback making it easier for the user to get the necessary information for the developers to fix/diagnose the problem. The message also keeps errors from the GUI out of the output window, so subprocess errors are presented separately. At this point, GUI still outputs errors to stdout along with being caught by the dialog (that could be changed now). The user can press OK and continue using the application if the error is not fatal. Errors raised by subprocesses will not cause this box to come up.

We could potentially make this generate full error reports (with OS and version info) and send them via email or something.

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -191,6 +191,9 @@ class PsychoPyApp(wx.App):
             sys.stderr = sys.stdout = lastLoadErrs = self._lastRunLog
             logging.console.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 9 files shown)


mdcutone:
Added preference to disable the error dialog if needed.

PR #2836, opened 2020-03-11
FF/RF: Coder IDE

mdcutone:
Made a few changes to GUI

- Use the pure-Python AUI implementation. Was using the AUI manager exported by the wxWidgets C++ lib which is buggy and incomplete. wxPython devs seem to recommend using the Python implementation for now. This was also changed in Builder.
- AUI theme tweaked a bit, captions look nicer now.
- Consistent title frame labels.
- Fixed error when trying to run a file when no file is present.
- Added captions and frame controls to panes.
- Removed file management from `Files` panel for now, `Files` tracks the location of the current editor file which is more in line with how Runner set the CWD.
- Removed menu options for changing the current working directory since Runner does that correctly now.
- Show home dir in `Files` pane if nothing is opened.

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -554,7 +554,7 @@ class PsychoPyApp(wx.App):
         # have to reimport because it is only local to __init__ so far
         from . import coder
         if self.coder is None:
-            title = "PsychoPy3 Coder (IDE) (v%s)"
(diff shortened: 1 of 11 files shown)


mdcutone:
Cant figure out why the builds are failing.

mdcutone:
Closing this since most of the changes made here have been applied in other PRs.

PR #2846, opened 2020-04-02
DOC: use default mathjax_path

### LINKED ISSUE #2845: Math formulas not rendering in documentation
Math formulas are not rendered in documentation on website, for example on https://www.psychopy.org/general/gamma.html. See image:

gamma

Math content is not rendered because the MathJax.js is not being loaded from http://mathjax.connectmv.com. It is blocked by Chrome, Firefox and Safari browsers with error message "Mixed Content: The page at 'https://www.psychopy.org/general/gamma.html' was loaded over HTTPS, but requested an insecure script 'http://mathjax.connectmv.com/MathJax.js'. This request has been blocked; the content must be served over HTTPS." in the browser developer console.

pn2200:
Use the default value for mathjax_path, so sphinx.ext.mathjax loads
MathJax from the cdnjs content delivery network. See documentation of [mathjax_path](http://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-mathjax_path). Should fix #2845 .

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -208,5 +208,5 @@ latex_use_parts = False
 # If false, no module index is generated.
 #latex_use_modindex = True
 
-# path to mathjax on a public server (or could put on our own in static path?)
-mathjax_path = 'http://mathjax.connectmv.com/MathJax.js'
+# path to mathjax. Use default to load from cdnjs content delivery network.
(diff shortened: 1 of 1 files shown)


peircej:
Updated docs rebuilt and pushed to psychopy.org. Thanks for that!

PR #2852, opened 2020-04-14
Emotiv support

peircej:


mdcutone:
Has anyone figured out why the builds are failing? I checked the logs but cant make sense of them to determine whether it's my code or something else causing the fault.

peircej:
I thin that error is to do with not having a .type attribute. Trying adding that now

peircej:
Ah @mdcutone The builds of yorur PRs appear to be failing on LGTM for some reason I don't understand unrelated to this

peircej:
OK, this appears to be passing now.
We've also updated the icons. The new version will have a new look (including dark mode) so we wanted the icons to be sharp. Hope these are OK

image
image

Issue #2849, opened 2020-04-13
BUG: Builder translates randint to inclusive boundaries


kdoelling1919:
I used builder to create an online experiment. In one custom code component I generate a random integer from 1 to 10. For this I put in the following code:
```python
num = randint(1, 11)
```
Python doesn't include the upper limit so it returns only numbers 1 through 10. 

Builder automatically translates this to the following:
```javascript
num = (Math.floor((Math.random() * ((11 - 1) + 1))) + 1);
```
This will occasionally return 11 and created an error for me when it did. I think removing the + 1 inside the inner parentheses will fix this issue.

My suggestion would be to adjust translation such that
 **python:** `num = randint(a, b)` --> **javascript:** `num = (Math.floor(Math.random() * (b - a)) + a);`

kdoelling1919:
This is somewhat related to #2847 but the problems brought up there are different so I decided to make a new issue.

peircej:
Good catch Keith, I think that makes sense

kdoelling1919:
Great! I made a pull request to fix the issue in #2853

kdoelling1919:
Closing as the merged pull request should fix this.

PR #2853, opened 2020-04-14
BF: Fix randint translation to remove upper limit inclusivity.

kdoelling1919:
This is a small edit to javascript transpiler to fix the bug specified in #2849  in which python code is upper limit exclusive but the JS code is upper limit inclusive.

### DIFF
--- a/psychopy/experiment/py2js_transpiler.py
+++ b/psychopy/experiment/py2js_transpiler.py
@@ -199,7 +199,7 @@ class pythonTransformer(ast.NodeTransformer):
                     keywords=[]
                 )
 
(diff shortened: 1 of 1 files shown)


peircej:
I think the better solution is going to be to add randint, and possibly a few similar functions and then not need to do any autotranslation. But well done for working this out in teh meantime! :-)

peircej:
For info, a randint function is now part of psychojs and used by psychopy, rather than requiring translation

See the psychopy change here: GH-3010
and the PsychoJS change here: https://github.com/psychopy/psychojs/pull/122

PR #2870, opened 2020-05-26
NF: More view and math tools

mdcutone:
Some additional view and math tools.

Changes to `viewtools`
-----------------------
* Added `visualAngle` which computes visual angle using vector inputs.
* Added `visible` and `visibleBBox` which are fast functions for testing whether points or bounding volumes are visible in the current projection.
* Added `computeFrustumFOV` which gives a frustum much like `gluPerspective` does.
* Added `projectFrustum` and `projectFrustumToPlane` which is used to determine the dimensions of the projection of a frustum on a screen or surface. Used for determining the drawing area when viewing through an aperture, but more generally, the size of a plane the frustum projects to at some distance.

Changes to `mathtools`
------------------------
* Added `normalMatrix` to compute the normal matrix used for lighting calculations in fragment shaders.
* Added `fitBBox` and `computeBBoxCorners` for fitting axis-aligned bounding boxes to a group of points or vertices. Used with `visible` and `visibleBBox` as well as the various raypicking routines in mathtools.
* Added `multMatrix` for fast chain multiplication of matrices.
* `rotationMatrix` and `quatFromAxisAngle` now accepts string names as a shorthand for rotations about world axes (eg. `-x`, `+z`, etc.) 
* Added `scale` function for uniform and anisotropic scaling of points.
* Added `accumQuat` for accumulating quaternion rotations. Used for forward-kinematics and armatures.
* Added `zeroFix` to truncate numbers in arrays close to zero to zero.
* Optimization of `invertMatrix` to invert affine/orthogonal matrices faster. Also handles inverting other matrices with different shapes. Useful for color space transforms which typically have 3x3 matrices.
* BF: `isAffine` strictly works with only 4x4 matrices, raised no error before. 
* Added 'fixTangentHandedness` to enforce handedness of tangent vectors for 3D models with messed up UV coordinates.
* Added `articulate` for forward-kinematics.
* Added `unproject` for getting object coordinates from screen and depth buffer coords.


peircej:
That's very odd error on Travis:
```
>    @pytest.mark.mathtools
>    def test_invertMatrix():
>       ...
>           inv = invertMatrix(m, homogeneous=True)

E           TypeError: invertMatrix() got an unexpected keyword argument 'homogeneous'
psychopy/tests/test_tools/test_mathtools.py:345: TypeError
```
but `homogeneous` is clearly there in the def

mdcutone:
Okay I'll fix it.

PR #2873, opened 2020-05-29
ENH: making fromFile function compatible with python 2

mservantufc:
filetools.py, modify the fromFile function line 55 as follows:

```
			try:
				contents = pickle.load(f)
			except:
				contents = pickle.load(f,encoding='latin1')# for compatibility with python 2
```


peircej:
hi @mservantufc this isn't quite right. It's trying to pull into branch 2020.1 and with lots of changes. Possibly it just needs adjusting the pull request to point to psychopy/master

mservantufc:
@peircej  It's the first time I use this pull request feature and I have no idea how to proceed. I couldn't even upload the adjusted filetools.py as this thing doesn't accept .py file. Could please take care of it ?

peircej:
OK, yes, that's no problem. I just like the people to get credit when they find the fix but I can handle it

PR #2877, opened 2020-06-02
BF: Fix import time race condition in monitors/calibTools.py

pdamodara:
Currently the module tries to create the monitors directory if it's missing.
The issue with this is that there is a race condition in how it is implemented:
- The directory could be created between the `os.path.exists` and `os.makedirs` calls.

This change modifies this to use a ask-for-forgiveness approach:
- Try to create the directories, but ignore any "file/directory already exists" errors.

`errno` should work cross-platform, so this will not break multi-platform support.

### DIFF
--- a/psychopy/monitors/calibTools.py
+++ b/psychopy/monitors/calibTools.py
@@ -25,6 +25,7 @@ try:
     haveSerial = True
 except Exception:
     haveSerial = False
+import errno
 import os
 import time
 import glob
@@ -63,8 +64,11 @@ else:
 if isinstance(monitorFolder, bytes):
     monitorFolder = monitorFolder.decode(sys.getfilesystemencoding())
 
-if not os.path.isdir(monitorFolder):
+try:
     os.makedirs(monitorFolder)
+except OSError as err:
+    if err.errno != errno.EEXIST:
+        raise
 
 
 class Monitor(object):


peircej:
> The directory could be created between the os.path.exists and os.makedirs calls.

This seems *relatively* unlikely but sure.

For the Exception, Python 3 (our preferred flavor) raises this as `FileExistsError` not `OSError` so we might need something like this:
``` python
try:
    os.makedirs(monitorFolder)
except FileExistsError as err:  # for Python3
    pass
except OSError as err:  # for Python2 compat
    import errno  # only import if needed. This is a one-off
    if err.errno != errno.EEXIST:
        raise err
```

pdamodara:
@peircej Thank you for the reviews!
This occurs quite often in our build system, actually - It is multi-processed on the same machine, which is probably making us run in to this more.

This works in python 3.6.3 on a Windows machine:
```
>>> import errno
>>> import os
>>> os.makedirs("test/directory")
>>> try:
...     os.makedirs("test/directory")
... except OSError as err:
...     print(err.errno == errno.EEXIST)
...
True
```

I took this approach because it was more concise, and py2/py3 compatible.
However, perfectly understandable if you'd prefer that `FileExistsError` is handled explicitly! Let me know if this is the case, I can update the PR.

peircej:
OK, cool. I'll pull in as is then

PR #2878, opened 2020-06-02
BF: Fix PermissionError on windows in Preferences.getPaths

pdamodara:
Currently the user prefs directory is tested for permissions, however they way this is done causes a `PermissionError` on windows:
```
>>> import os
>>> open("test", "r")
<_io.TextIOWrapper name='test' mode='r' encoding='cp1252'>
>>> os.remove("test")
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'test'
>>>
```

This change fixes this by opening the file within a context, so the file handle is closed appropriately.
```
>>> import os
>>> with open("test2", "r") as fh:
...     fh.read()
...
''
>>> os.remove("test2")
>>>
```

Debugging this within a build system that uses prebuilt wheels is cumbersome because the code hides the underlying error.
This change updates the code to append the underlying error message to the exit message.

### DIFF
--- a/psychopy/preferences/preferences.py
+++ b/psychopy/preferences/preferences.py
@@ -138,11 +138,12 @@ class Preferences(object):
                 tmp = os.path.join(self.paths['userPrefsDir'], '.tmp')
                 with open(tmp, 'w') as fileh:
                     fileh.write('')
-                open(tmp).read()
+                with open(tmp) as fileh:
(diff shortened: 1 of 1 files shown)


peircej:
That's an interesting one! I would have thought that would have caused issues for more people, which doesn't seem to be happening. But I'm happy with the change

PR #2883, opened 2020-06-12
Reskin

TEParsons:
Big update! Total aesthetic redesign. Shouldn't change how the software works, just how it looks.


peircej:
I think you're missing a file. When I try to launch I get 
"missing file coder//themes//psychopy.json"

The reason the tests don't then run is probably because @mdcutone's error dialog traps the exception and doesn't go away, so Travis eventually crashes out after no activity, but with no info about the error that was occuring

PR #2884, opened 2020-06-12
Todd's reskin branch with file location fix

peircej:


peircej:
Next issue is that codeFont has been removed from prefs (presumably because it wasn't needed due to the new code styling, but not all uses of that pref have been removed:
https://github.com/psychopy/psychopy/search?q=codeFont&unscoped_q=codeFont

Issue #2879, opened 2020-06-04
Error creating a psychopy.gui.Dlg() in PsychoPy3_PY2


pvdemael:
Hi,

I have an issue with the gui module of PsychoPy_PY2.

**Setup**
Concurrent installation of 
StandalonePsychoPy3_PY2-2020.1.3-win32
StandalonePsychoPy3-2020.1.3-win64
Windows 10 Home

**Code**
```
from psychopy import gui
myDlg = gui.Dlg(title="Test Dialog")
data = myDlg.show() 
```

**Issue**
Works in Psycho3_PY3

Not working in Psycho3_PY2, throwing error below
```
myDlg = gui.Dlg(title="Test Dialog")
File "C:\Program Files (x86)\PsychoPy3_PY2\lib\site-packages\psychopy\gui\qtgui.py", line 129, in __init__
nScreens = len(qtapp.screens())
AttributeError: 'QApplication' object has no attribute 'screens'
```

If I run the code in the Python shell (2.7.15) of the Coder, it works perfectly.

peircej:
Most likely this is a problem with the version of Qt that has been distributed in that version.

Makataui:
Apologies, just lurking through and saw this one - we've run into this for legacy setups before when trying to move on with some older code - in this distribution, the included Qt is version 4.xx or something like that where it used to be QDesktopWidget().screenCount() instead iirc. I've fixed this locally (hackily) by including a simple check in the past, just looking at qtgui.py I put something together - not sure if it's the right way to go about it but will send a PR in as well:

```Python
        if haveQt == 'PyQt5':
            nScreens = len(qtapp.screens())
        else: 
            nScreens = QtWidgets.QDesktopWidget().screenCount()
```

PR #2887, opened 2020-06-14
BF: Fix for PY2 counting screens for Qt4 distribution #2879

Makataui:
PsychoPy3_PY2 uses Qt4 which can run into errors when trying to count the number of screens (as seen in #2879) as code assumes Qt5 - this checks if using Qt5 or another version, and if not Qt5, applies the older QtWidgets.QtDesktopWidget().screenCount() method to return int for number of screens.


peircej:
Fixes #2879

Issue #2889, opened 2020-06-15
Cannot create data files in custom location


pn2200:
When using Builder in PsychoPy v2020.1.2, any output files will be written in same directory as the .psyexp file, followed by the user-specified path and file name from the 'Data filename' field in the experiment settings. However, the documentation (https://www.psychopy.org/builder/settings.html#data-settings), in the third example, documents that a file can be stored anywhere on the file system.

Following that example, where my experiment is in `/Users/user/src/`, then the data files are created in `/Users/user/src/~/dropbox/data/DEV/999-2020_Jun_15_1047`, which is not what I want to happen.

Either the documentation should be fixed to document actual behavior, or the way the files are written should be changed to match the current documented behavior.

pn2200:
Inspired by this discussion: https://discourse.psychopy.org/t/the-location-of-output-data-file/1988/5, this issue can be worked around by adding a CodeComponent and setting the `thisExp.dataFileName` and `filename` to the custom path, so the data files are written to the custom path. In the experiment settings, setting the Data filename field to an empty string `''` will prevent the creation of the `.log` file, and the `.log` file can be created in a custom path by setting `logging.LogFile(filename + '.log', level=logging.EXP)` in the CodeComponent.

m-macaskill:
The default location is indeed relative to the directory containing the `.psyxep` file. If you want to escape that folder to access other parts of the system, you could do that by constructing paths that include `../`. Have you tried that?

You're right, it looks like that example needs to be amended accordingly. Pull requests would be gratefully accepted to amend this file:
https://github.com/psychopy/psychopy/blob/4a86060a22abe78024c2a9c9f1cea319bb6a5190/docs/source/builder/settings.rst

pn2200:
Your suggestion is possible, but it is really difficult when I don't have control over where the `.psyexp` file is installed, but I do know the absolute path for the output.

Inspired by your suggestion I found that, without a CodeComponent, specifying the data filename in the experiment settings using `os.path.relpath` as in `$os.path.relpath(os.path.join(os.path.expanduser('~'), 'dropbox', 'data', 'DEV')))` works but that is unwieldy.

peircej:
Yes, the use of `~` is quite mac/linux-centric there! :-) And although absolute paths did once work it looks like they haven't done for a while and nobody noticed. I guess nearly everyone keeps their data next to their study. Paul

[...]

PR #2890, opened 2020-06-16
DOC: Fix data filename documentation example

pn2200:
For the Builder's data filename documentation, fix the example that
shows how to create data files anywhere on the filesystem.


pn2200:
Should fix #2889 .

Issue #2900, opened 2020-06-22
Files with unhandled lexer are not styled


TEParsons:
Specified base style isn't applying to files of a filetype which has a lexer but that lexer isn't specified by PsychoPy

Issue #2897, opened 2020-06-20
Code Component should appy JS lexer to the JavaScript code panels


peircej:
Currently the Python lexer is applied throughout

TEParsons:
Fixed in commit 7e6b0d917a7a60ec27285994d78a5e26a3a0d78b

PR #2901, opened 2020-06-22
Implementing feedback from Reskin

TEParsons:
Fixed:
#2900
#2897

Also updating theme format to include spec for caret, selection & app (darkmode/lightmode). App still requires a restart to apply app theme.


mdcutone:
I like a light editor theme with a dark app theme. I think it looks pretty neat.

peircej:
> I like a light editor theme with a dark app theme. I think it looks pretty neat.

@TEParsons and I talked about this a fair bit and I think this is relatively unusual. I personally find it annoying if I have to check in multiple places to switch the colour scheme (once for the app and once for the code). The solution we opted for was that the app dark/light mode is a part of the theme but that can be customised. So you can take one of Todd's dark themes, move it to your personal themes folder and change the app setting within that theme to light mode leaving the rest of the colours dark.

PR #2906, opened 2020-06-24
RF/NF: Coder touchups

mdcutone:
Fixes some issues around coder:

- Runner list should not allow multiple items to be selected, causes a crash if so.
- Collapsing 'Alerts' pane in Runner causes header text for other panes not the be redrawn.

More: 

- Removes features in the file browser and source tree that no longer work (eg. autoscroll for the source tree) or are not needed (file sizes). 
- Adds pref to disable autocomplete, makes editing faster on some systems. If disabled, you need to manually invoke autocomplete when prompted by the status bar text that appears using a key combo.
- Removes the "show file browser" menu item, that doesn't need to exist since it's integrated as a source assistant tab.
- Remove "close" buttons on source assistant tabs. Needed because there is no way to bring them back when closed without restarting.
- Bunch of cleanups that reduce code quality and would set off LGTM.
- Made code printing nicer. Output prints code using the same font and size the editor uses. Probably nicer for students submitting their assignments. Code will fit on the page as long as people keep text within 80 columns if using 10 point font. Headers all have horizontal lines when there are multiple pages, shows the page number, count, file name and date/time. Useful information if you are like me and keep everything loose leaf in/on a desk somewhere. Might be useful to have a pref to set printing font size so they don't have to change their editor font size to adjust. Keeping as-is for now since this is still a big improvement over how it was.


peircej:
I've combined these commits with the theming updates in PR #2924

PR #2908, opened 2020-06-26
Textbox

TEParsons:
Added working textbox interface


psychopy/visual/basevisual.py
```diff
@@ -590,6 +590,9 @@ def contains(self, x, y=None, units=None):
         else:
             poly = self.verticesPix  # e.g., tessellated vertices
 
+        print(poly)
```
peircej, line 593: probably best to remove the print statements :-)

peircej:
closing in favour of #2931 (where tests should at least be passing again)

PR #2918, opened 2020-07-03
LiveThemes

TEParsons:
Various fixes and enhancements to support live switching themes


TEParsons:
@peircej if I add a print statement to the code, will it print in Travis? I'd like to see what it actually reads `spec` as

peircej:
Yes, but usually only if the test errors - you don't see all the lines of output from successful tests normally

peircej:
These commits and more are included in PR #2924

PR #2922, opened 2020-07-07
Setuptools plugins

peircej:
This is a master merge for conflicts to try and pull in GH-2792


mdcutone:
This one is going to be hard to rebase. Also looking into removing the startup plugins feature if we can figure out how to add components and GUI elements during runtime. I'm pretty sure that's something we can figure out but will require some refactoring.

peircej:
It's my fault for not pulling it in a while ago
I'll work through it. We'll get there

PR #2925, opened 2020-07-13
LiveThemes2

TEParsons:
Fixes to the live theme switching feature


TEParsons:
Fixes #2933 and #2934

PR #2926, opened 2020-07-13
BF: Removal of the gitFeedback function causing error with new themes

dvbridges:
More informative git feedback should be reported in the Runner.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -39,8 +39,8 @@ except ImportError:
 
 from psychopy.localization import _translate
 from ... import experiment
-from .. import dialogs, icons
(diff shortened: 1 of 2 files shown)


peircej:
oooh that's why there was a red and green version of the globe icon. I had forgotten that this was there! @TEParsons and I were talking about it and I was guessing that redglobe and greenglobe were relics of us making up our mind on what looked nicest. Now I remember that they were _functional_!
It's easy enough for us to add them back in (Todd has actually made them in all versions) rather then getting rid of the functionality if people like it.

TEParsons:
I've put the icons back in :) PR #2925

dvbridges:
Cool :) Ill close the pull request.

PR #2960, opened 2020-07-20
DOC: Fix several typos in Sound API documentation

pn2200:


peircej:
thanks @pn2200 
I'm planning to update the docs online later today so this should go live then

PR #2964, opened 2020-07-21
Additional Resources Field

TEParsons:
Not integrated with anything else yet (might be worth leaving this open until the end of the day, I just wanted to run it by Travis to check that the presence of a new param won't break it)


peircej:
When you add a new param you need to run 
```
python genComponentsTemplate --out
```
to generate a new template file

TEParsons:
Where is that? I can't find it in the repo & just calling it when cd'd onto root doesn't work

TEParsons:
This is done now, can be pulled in pending Travis :)

PR #3026, opened 2020-08-03
FF: Remove offset when searching text.

mdcutone:
Remove the offset applied when searching for text in a Coder editor window. Must have been needed before, but now it's not.

### DIFF
--- a/psychopy/app/coder/coder.py
+++ b/psychopy/app/coder/coder.py
@@ -1039,9 +1039,6 @@ class CodeEditor(BaseCodeEditor, CodeEditorFoldingMixin, ThemeMixin):
                 start = 0
                 loc = textstring.find(findstring, start)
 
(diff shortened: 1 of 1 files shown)


peircej:
Hmm, that's a very recent addition from @TEParsons 
I wonder what changed in such a short time :-/ Todd do you want to check if it's still needed on yours?

TEParsons:
I added it to account for an existing offset, I'll check today whether that offset is still there

TEParsons:
Just went back and checked, it definitely is still needed - without it all searches are off by 2 characters.

peircej:
So we need it there on Todd's machine and we need it not there on Matthew's. Hmmmm

peircej:
I'm closing this until we can work out why it differs on the two

PR #3048, opened 2020-08-06
Bugfixes

TEParsons:
Bug fixes for 2020.2.1


peircej:
For the gender thing, I think the only reason to include it at all is to show how to use a fixed choice item, but as per our previous discussion I think drop it altogether

PR #3052, opened 2020-08-06
BF: Save mentalrot.pysexp data as text, rather than binary

mrkrause:
Without this tiny change, the mental rotation demo (Builder) crashes before the results are plotted:

 File "MentalRot_lastrun.py", line 826, in <module>
    fd.write(data_string)
TypeError: a bytes-like object is required, not 'str'


mrkrause:
The whole method for saving/plotting the results could be improved, but it seemed sensible to make a tiny change that gets it working first.

peircej:
Yes, it's actually a bit unusual to be doing this in the experiment at all (rather than letting PsychoPy save the data for you!)

Issue #3065, opened 2020-08-13
OSX 10.15 Font error  


j-ales:
Version: 2020.2.2
OS: OSX 10.15.6 Catalina

In 10.15 Running under the developer mode from GitHub. Not the packaged download which currently has another error precluding me from running it. 

Components rendering code fonts spawn an error: 
```
Traceback (most recent call last):
  File “/Users/ales/git/psychopy/psychopy/app/builder/builder.py”, line 1136, in setExperimentSettings
    dlg = DlgExperimentProperties(frame=self, title=title,
  File “/Users/ales/git/psychopy/psychopy/app/builder/dialogs/__init__.py”, line 1835, in __init__
    _BaseParamsDlg.__init__(self, frame, ‘Experiment Settings’,
  File “/Users/ales/git/psychopy/psychopy/app/builder/dialogs/__init__.py”, line 523, in __init__
    ctrls = self.addCategoryOfParams(theseParams, parent=page)
  File “/Users/ales/git/psychopy/psychopy/app/builder/dialogs/__init__.py”, line 636, in addCategoryOfParams
    self.addParam(fieldName, parent, sizer, currRow,
  File “/Users/ales/git/psychopy/psychopy/app/builder/dialogs/__init__.py”, line 789, in addParam
    ctrls.valueCtrl.SetFont(self.app._codeFont)
wx._core.wxAssertionError: C++ assertion “”m_refData != __null”” failed at /Users/robind/projects/bb2/dist-osx-py38/build/ext/wxWidgets/src/osx/carbon/utilscocoa.mm(104) in OSXGetNSFont(): invalid font
```

I've traced this and for some reason I don't understand it seems that calling SetFaceName() with '.AppleSystemUIFont' fails.   
'.AppleSystemUIFont'  is what is given by: wx.SystemSettings.GetFont(wx.SYS_ANSI_FIXED_FONT) and if there is no call to SetFaceName it also works for rendering the components.  

The calls that bork  the _codeFont variable are  done in _pscyopyApp.py when the theme is set at line 921: 
```
codeFont = themes.ThemeMixin.codeColors['base']['font']
self._codeFont.SetFaceName(codeFont)
```
I can't figure out why this error happens.  You can create a font with that name using"
`wx.Font(wx.Fontinfo().FaceName(".AppleSystemUIFont")`   but calling to SetFaceName(".AppleSystemUIFont") generates an invalid font error.   

[...]

PR #3066, opened 2020-08-13
BF:  FixFont error caused when SetFaceName() is called for .AppleSystemUiFont

j-ales:
This pull request implements the fix for the problem described in issue #3065


peircej:
thanks for this Justin :smile:

PR #3111, opened 2020-09-03
Bug Fixes

TEParsons:


peircej:
Subsequent alterations:
- #3124 improves on the start/stop timing for the FormComponent by maing it a subclass of BaseVisualComponent and using what's built-in there
- #3120 improves on fc24277 by raising NotImplemented if the user tries to set TextBox2 with 'norm' units

PR #3122, opened 2020-09-04
BF: handle multiple mouse objects in one Routine in JS

peircej:
Previously we were defining the `xys` and `buttons` with local scope.

Instead remove that local definition (const/let) and the script compiler
will then automatically give it global scope (var at the top of the script)

To avoid likely clashes we've given more obscure names:
- `_buttonXYs`
- `_mouseButtons`

fixes psychopy/psychojs-88


peircej:
fixes https://github.com/psychopy/psychojs/issues/88

PR #3130, opened 2020-09-07
BFI demo spacing fix

TEParsons:


peircej:
The let/const issue has been solved elsewhere (0e0b26d4dc) and I've updated the BFI demo to be more usable already (providing several shortened versions so not requiring the user to complete all 44 items) e8e8630e642b3e924

Issue #3127, opened 2020-09-05
iohub not working psychopy 2020.2.3


oricon:
On both mac and windows versions of 2020.2.3, I can't get ioHubServer to launch in my experiments. The demos just crash without running, so it doesn't appear to just be my code.

m-macaskill:
I'm seeing the same (macOS 10.15.6, PsychoPy 2020.2.3, no apparent error messages). We need to use `ioHub` for a textbook update, so I'll pursue this with @peircej to see if we can get it up and running properly again.

peircej:
@isolver do you think you'd be able to take a look at this? I believe gevent has made some changes (and possibly more dependencies have) in recent versions and I think that's caused iohub to fail

isolver:
Atleast on Windows, I think the crash is a side effect of [this change made to iohub\server.py](https://github.com/psychopy/psychopy/commit/1c360424535bbf3b285f42c86f1ff649f6e76ffa#diff-484411b67b479623304cd173e8d2faab). Lines 228 and 231 of the updated file are missing the 'not' part of the eval that was added. If 228 - 232 are changed to:

```
            if not isinstance(dclass, unicode):
                dclass = unicode(dclass, 'utf-8')
            dmethod = request.pop(0)
            if not isinstance(dmethod, unicode):
                dmethod = unicode(dmethod, 'utf-8')
```

Then the 4 demo's in coder\iohub run OK. Have not tried macos; will not be as easy for me to do.

The issue with " no apparent error messages" is, at least on Windows, actually a separate issue related to #2950. To see iohub prints or errors you now need to run Coder using python, not pythonw, otherwise the iohub process crashes when it tries to print. Not sure what is cause this.

Did you want me to submit a patch for server.py?

isolver:
Sorry, referenced wrong issue in my last comment; was referring to #2766, not #2950. Sorry for any confusion.

PR #3136, opened 2020-09-12
BF: #3127

isolver:
Only tested on windows 10.


peircej:
ooof, was that all it was? Embarassing! Thanks Sol

PR #3138, opened 2020-09-16
F-String Params

TEParsons:
Bit by bit updating the component code writing functions to use f-strings rather than % formatting


peircej:
I'm a bit wary of changes "just" to make the code prettier. Don't get me wrong - I love f-strings but changing working code purely to implement them is risky in my view

PR #3140, opened 2020-09-16
BF: Reset marker color after a response

### LINKED ISSUE #3139: RatingScale marker turns grey if the scale times out and does not refresh on next trial
Using PsychoPy v2020.1.2, the RatingScale marker turns grey if the scale times out and remains grey on all subsequent trials. The marker color is never reset to the original color.

This issue is substantially the same as #868 which was fixed, and then reintroduced in #1665.

pn2200:
Should resolve #3139 .

### DIFF
--- a/psychopy/visual/ratingscale.py
+++ b/psychopy/visual/ratingscale.py
@@ -857,6 +857,7 @@ class RatingScale(MinimalStim):
         self.markerColor = markerColor
         self.markerYpos = self.offsetVert + self.markerOffsetVert
         # save initial state, restore on reset
+        self.markerColorOriginal = markerColor
 
     def _initTextElements(self, win, scale, textColor,
                           textFont, textSize, showValue, tickMarks):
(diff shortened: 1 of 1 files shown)


peircej:
Thanks Paul. Couple of comments:
- Are both the following lines needed? Most attributes in PsychoPy include setter code so you don't need the second step (although Rating Scale is old and maybe never got retrofitted with that)
```
        self.markerColor = self.markerColorOriginal
        self._setMarkerColor(self.markerColor)
```
- I'd generally recommend moving to Slider instead of RatingScale. Rating is big and complex. Slider handles the same features and more (including styling optoins like radio buttons, vertical orientation...) and actually renders much faster too. It's also the only option that ports to JavaScript. That doesn't mean we shouldn't fix bugs in Rating but I'd say it's essentially deprecated for the long-term

pn2200:
- I think both lines are needed. `self.markerColor` is only an attribute on the RatingScale. Setting it doesn't modify the actual marker object. The second line tries to modify the marker object by setting the `color` or `fillColor` attributes. I think it tries to set the `color` or `fillColor` attributes to allow for custom marker objects that may only implement one or  the other.
- Fair point about RatingScale being essentially deprecated. I will keep that in mind when developing new tasks.

PR #3150, opened 2020-09-21
Use norm units in demo, fix typos

### LINKED ISSUE #3149: Sliders in slider demo not visible
Sliders in the second and third routines of the ratingsNew demo are not visible or are really small, making the demo not useful. See attached screenshots:
sliders_not_visible
sliders_really_small

pn2200:
Use norm units in experiment, and then use experiment settings in demo, so sliders are visible. Fix typos in ratingsNew demo also. Should resolve #3149.

### DIFF
--- a/psychopy/demos/builder/ratingsNew/ratingsNew.psyexp
+++ b/psychopy/demos/builder/ratingsNew/ratingsNew.psyexp
@@ -1,17 +1,21 @@
-<?xml version="1.0" ?>
-<PsychoPy2experiment encoding="utf-8" version="3.0.1">
+﻿<?xml version="1.0" ?>
+<PsychoPy2experiment encoding="utf-8" version="2020.2.4">
   <Settings>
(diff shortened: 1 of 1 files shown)


peircej:
Ah, yes, we should never have allowed a default unit for the application*, only for the experiment because it leads to inconsistent behaviour between machines/users

By the way, could you try and use our commit message prefixes that indicate whether this is a bugfix that needs porting to the existing release stream, versus a fix/feature for code that isn't yet released (should be held until next major version)

thanks for your contributions!

* the option dates back to the days when everything was pure code and I personally didn't want to specify 'deg' every time I created a window

pn2200:
Thanks for merging so quickly. I thought I was using the appropriate prefix; the two commit summaries are prefixed with `DOC:` because I thought demos corresponded to documentation improvements. Or do you mean that the pull request summary should also be prefixed?

peircej:
Oh, haha, yes, that's fine. I looked at the message of the pull request not the commit. Great job! 🥇

PR #3178, opened 2020-10-01
Runner Print Fix

TEParsons:
… Elementary, my dear Watson.


peircej:
So, we now move stdout there when the Runner is shown. What happens if the script is run while the Runner is _not_ being shown? Is that possible? 

I feel like maybe we should be tracking stdout ourselves (at all times) and then updating the runner output simply to show the contents of that buffer. Then the state of the Runner would be irrelevant to the print/stdout being received/stored

PR #3184, opened 2020-10-02
Add high DPI support for visual stimuli.

zeyus:
I haven't contributed to psychopy before, so if something more is needed, please let me know and I'll happily put in the extra work! :)
I found that on my laptop display, the stimulus images looked blurry and scaled. This PR adds an option to set high DPI support on windows for visual stimuli. This also fixes a problem where the reported size of the screen/display/windows differs from the actual size in pixels.

With high DPI support disabled:
`3.3480  WARNING         User requested fullscreen with size [3200 1800], but screen is actually [1829, 1029]. Using actual size`

```python
from psychopy import prefs
prefs.hardware['highDPI'] = False
# ...
visual.ImageStim(win, image)
````
image

With high DPI support enabled:
```python
from psychopy import prefs
prefs.hardware['highDPI'] = True
# ...
visual.ImageStim(win, image)
````

screenshot_f

### DIFF
--- a/psychopy/app/preferencesDlg.py
+++ b/psychopy/app/preferencesDlg.py
@@ -141,6 +141,7 @@ _localized = {
     'audioDevice': _translate("audio device"),
     'parallelPorts': _translate("parallel ports"),
     'qmixConfiguration': _translate("Qmix configuration"),
+    'highDPI': _translate('Try to support display high DPI'),
     # pref labels in Connections section
     'proxy': _translate('proxy'),
     'autoProxy': _translate('auto-proxy'),
--- a/psychopy/visual/__init__.py
+++ b/psychopy/visual/__init__.py
@@ -14,7 +14,17 @@ import sys
 if sys.platform == 'win32':
     from pyglet.libs import win32  # pyglet patch for ANACONDA install
     from ctypes import *
+    from psychopy import prefs
     win32.PUINT = POINTER(wintypes.UINT)
+    # get the preference for high DPI
+    if 'highDPI' in prefs.hardware.keys():  # check if we have the option
+        enableHighDPI = prefs.hardware['highDPI']
+        # check if we have OS support for it
+        if enableHighDPI:
+            try:
+                windll.shcore.SetProcessDpiAwareness(enableHighDPI)
+            except OSError:
+                pass
 
 from psychopy.visual import filters
 from psychopy.visual.backends import gamma

### CODE REVIEW

**Comment by mdcutone** [comment]
Thank you for addressing this issue. While this does make images clearer, the issue here is that treating high DPI this way causes a mismatch between window units PsychoPy uses and the framebuffer's coordinates. This is apparent in your demo where the image is smaller than it should be, we would expect the image to clearer and with the same dimensions. @peircej I think we should take what @zeyus has done here and use it to create a more generalized function that returns the DPI scaling factor of the display being used which works for both retina and Windows/Linux machines. 

Something like taking notes from wxPython:

```
scaleFactor = getDPIScalingFactor() 
```

Then using this to modify the reported window size stimuli use for drawing so window coordinates are properly mapped to framebuffer. This will also further simplify how high-DPI displays are treated in the future.

mdcutone:
@zeyus Now that I think about it, I think the GLFW backend might support this as-is, can you try running this again with `winType='glfw'` to see if you get the scaling correct? If that's the case, we can get your code integrated in and I can handle getting the other backends working.

zeyus:
> 
> 
> @zeyus Now that I think about it, I think the GLFW backend might support this as-is, can you try running this again with `winType='glfw'` to see if you get the scaling correct? If that's the case, we can get your code integrated in and I can handle getting the other backends working.

Yes that's true, it does work with GLFW, I've been using that as my backend for working with psychopy because I thought I could fix the issue just by trying different options, and I saw in the docs that GLFW is the preferred method so I left it at that. (the screenshots in my initial submission were using GLFW)

mdcutone:
Okay cool. I think we need to figure out how units are going to be handled here before committing these changes. I think we need to treat units like we do Retina displays (the original HiDPI display we support) to make things consistent across platforms. That requires a bit of work in many other parts of PsychoPy to get working.

peircej:
What to do with this one? With retina displays we certainly went with the approach that the user requesting something 256x256 pixels should actually get an image that was "looked the right size" which means it actually occupied 512x512 on a retina display. That's Apple's way of framing it, so that things looked similar across retina and non-retina laptops for instance. I have to say I have a slight preference not to sugar-coat the difference, and treat high-def display just like regular displays with higher resolution but. At this point, that would require behaviour change by PsychoPy (at least on macs) but might be worth it

peircej:
OK, you know, I'm going to pull this in for now pending further debate. It's subject to potential behaviour change in the future :-)

peircej:
It will be saved for next major release though (2021.1)

PR #3195, opened 2020-10-07
BF: Fixes exceptions in the shell being displayed in an error dialog.

### LINKED ISSUE #3173: Code "Shell" raises error box on exception
When using the interactive Python shell in Coder, the unhandled exception dialog is presented on error instead of stdout being diverted to the shell. We might have to hook an event on the shell to redirect stdout to the console on focus and back to the error dialog when focus is lost.

This reveals another longstanding issue, PyShell operates on the same interpreter thread as the GUI. This is useful for debugging, allowing us to inspect variables and run commands to mess around with the GUI at runtime. However, this is really bad idea for something user facing since commands being executed can crash the GUI (pretty much the same issue MATLAB has (had?) where typing in a while loop with no exit condition crashes it). Also the user can create objects and arrays that aren't garbage collected that can inflate the GUI's memory usage, requiring a restart to clear. Having the REPL talk to a Python interpreter in another process through pipes will allow users to restart a misbehaving interactive session without it affecting the GUI. I've written a REPL before for wx but it wasn't a great implementation. Now that I think about it, we'll probably need to write a GUI-less wxPython app that runs as a subprocess which wraps as system terminal, sending keypresses over to it using TCP or something.

mdcutone:
Fixes #3173


mdcutone:
Closing because my branches are are messed up.

PR #3215, opened 2020-10-14
BUGFIX: Setting volume with the PTB backend

suddhasourav:
While using the psychtoolbox backend, changing the sound volume (either by changing .volume or by using the setVolume() setter) had no effect on playing.

Setting the _track_'s volume, with a property, now seems to solve this. I have checked that it also works for basic sounds (e.g. 'A'), and setVolume() also works now without additional changes because of the property.

Original bug report: https://discourse.psychopy.org/t/cannot-set-volume-when-using-ptb/11055


peircej:
Thanks for digging into this one! :-) 

Ideally, could you use BF: as the prefix for bug fixes? When I search for what to include in a minor (bugfix release) I do that by searching the commit logs for BF and then yours wouldn't be found (but in this case I'm about to make a release so I'll probably remember!)

suddhasourav:
Thank you! I have changed the title now as well, sorry for forgetting the conventions - it's been a long time 😬

peircej:
No worries at all - it's actually the commit message that's the key, more than the pull request title. When I pull it all down to my local machine it's the commit messages I see. Thanks though!

PR #3222, opened 2020-10-16
BF: Fix to textbox colors

TEParsons:
Uses Color class as a one-off fix to be over-written when the class is fully implemented


TEParsons:
@peircej so there was already code in there to draw the caret even when no character vertices are available, it was just calculating default vertices wrong so the width became 0. The position isn't quite right still, but it's close enough for now as it'll all change when the Size/Position classes are in anyway.

Issue #3217, opened 2020-10-15
Issues with additional Resources box in Exp Settings


peircej:
The resource files need to be stored with a relative reference (relative to the experiment file) not an absolute reference, or they won't be found when the experiment is dragged to a new computer

The Additional Resources box should also expand with the dialog box

![image](https://user-images.githubusercontent.com/380582/96117142-89f69080-0ee1-11eb-9d02-e016beb44f00.png)

peircej:
The resizing is also now handled in 8888852

PR #3235, opened 2020-10-20
BF: Use relative paths for resource files

TEParsons:
Partially addresses #3217


TEParsons:
The other part of this issue is harder than originally anticipated - I was expecting to just add a `wx.EXPAND` flag, turns out it already has one so the problem is to do with the available space...

peircej:
> The other part of this issue is harder than originally anticipated - I was expecting to just add a wx.EXPAND flag, turns out it already has one so the problem is to do with the available space...

Check the proportion value (should be set to 1) as that works in combination with EXPAND

PR #3236, opened 2020-10-20
BF: More reasonable default values for Textbox fillColor and borderColor

TEParsons:


peircej:
When you set them to None I think they show up as "None" in the dialog box and then raise an error when being sent to JS, no? Did you check if the None option works in an experiment? I certainly see an issue like that with the padding value setting itself to "None" rather than "". Maybe we set all 3 to empty strings?

PR #3239, opened 2020-10-21
Template for form items

TEParsons:


peircej:
The test is reporting "Can't represent a Param of type table" meaning the Param class doesn't know how to represent this as code

psychopy/app/builder/dialogs/__init__.py
```diff
@@ -156,8 +156,10 @@ def __init__(self, dlg, label, param, parent, fieldName,
             self.valueCtrl = FileListCtrl(parent,
                                           choices=param.val,
                                           size=wx.Size(self.valueWidth, 100),
-                                          pathtype="rel"
-                                          )
+                                          pathtype="rel")
+        elif param.valType == 'table':
```
peircej, line 160: The valType is telling PsychoPy how to write the code in the script. Maybe you want this to be a 'fileList' type? Or a fixedList type? (I don't know what that one exists for)

PR #3263, opened 2020-11-03
BF: Import visual.line in shape

TEParsons:


peircej:
@TEParsons weren't we expecting the tests to fail on this final commit because of the import of psychopy.visual?

peircej:
Ah, not to worry anyway - it isn't needed because it's fixing an issue that isn't in 2020.2 (this warning about 127 pixels isn't there in 2020.2)

TEParsons:
Yeah but then the FINAL final commit was a revert

PR #3274, opened 2020-11-09
ENH: update translation

hsogo:
Add _translate() to new strings and update Japanese translation.


hsogo:
Hi @peircej ,
I can't start PsychoPy app after pulling the latest main branch.  Error message is as follows.  The error occurs during reading YAML files from psychopy/alerts/alertsCatalogue/

```
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\psychopyApp.py", line 104, in <module>
    main()
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\psychopyApp.py", line 100, in main
    start_app()
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\psychopyApp.py", line 22, in start_app
    from psychopy.app._psychopyApp import PsychoPyApp
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\_psychopyApp.py", line 13, in <module>
    from psychopy.app.colorpicker import PsychoColorPicker
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\colorpicker\__init__.py", line 15, in <module>
    from psychopy.app.themes import ThemeMixin
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\themes\__init__.py", line 1, in <module>
    from ._themes import ThemeMixin, IconCache
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\app\themes\_themes.py", line 16, in <module>
    from ...experiment import components
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\experiment\__init__.py", line 21, in <module>
    from .params import getCodeFromParamStr, Param
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\experiment\params.py", line 31, in <module>
    from ..alerts import alert
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\alerts\__init__.py", line 5, in <module>
    from ._alerts import alert, catalog
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\alerts\_alerts.py", line 182, in <module>
    catalog = AlertCatalog()
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\alerts\_alerts.py", line 32, in __init__
    self.alert = self.load()
  File "C:\Users\Hiroyuki\Dropbox\git-repository\psychopy\psychopy\alerts\_alerts.py", line 57, in load
    entry = yaml.load(ymlFile, Loader=yaml.SafeLoader)
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\site-packages\yaml\__init__.py", line 112, in load
    loader = Loader(stream)
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\site-packages\yaml\loader.py", line 34, in __init__
    Reader.__init__(self, stream)
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\site-packages\yaml\reader.py", line 85, in __init__
    self.determine_encoding()
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\site-packages\yaml\reader.py", line 124, in determine_encoding
    self.update_raw()
  File "C:\Users\Hiroyuki\Dropbox\WinPython-64bit-3.6.1.0Qt5\python-3.6.1.amd64\lib\site-packages\yaml\reader.py", line 178, in update_raw
    data = self.stream.read(size)
UnicodeDecodeError: 'cp932' codec can't decode byte 0x9d in position 805: illegal multibyte sequence
```
I examined YAML files in alertsCatalogue and confirmed that the problem is in the following line of **4305.yaml**.

> Re-enable the Component in the Builder view (component dialog boxes each have a “testing” tab: unselect the “Disable component” setting there). Otherwise just ignore this Alert if you intended it to be disabled.

The characters `“` (0xe2809c)  and `”` (0xe2809d) are unicode characters and therefore can't be read by cp932.  I don't see any reason to use these characters instead of ASCII double quotation mark  `"` (0x22) in this case.  However, if we will need to use Unicode characters in alert YAML files in the future, it would be better choice to use `codecs.open()` instead of `open` in line 55 of psychopy/alerts/\_alerts.py.  What do you think?

peircej:
Thanks @hsogo 
I totally agree - no need to use smart quotes in these files. BUT other people might do the same thing so we could do both fixes

hsogo:
Thank you for the reply.  I've fixed both.

PR #3275, opened 2020-11-09
RF: Set textbox editable on change to autoDraw, 

### LINKED ISSUE #3267: Textbox not editable if multiple textbox components used
Issue raised on forum https://discourse.psychopy.org/t/editable-textbox-not-responding/17786/5 if multiple textboxes are presented the second is not editable. See minimally working example here.
https://gitlab.pavlovia.org/lpxrh6/textboxbug2

TEParsons:
not on init, meaning it is only editable when on screen. Fixes #3267, overrides #3269.

### DIFF
--- a/psychopy/visual/basevisual.py
+++ b/psychopy/visual/basevisual.py
@@ -165,15 +165,17 @@ class MinimalStim(object):
             else:
                 toDraw.append(self)
                 toDrawDepths.append(self.depth)
+            # Add to editable list (if needed)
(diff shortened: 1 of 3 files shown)


TEParsons:
This fixes the bug described in this forum post: https://discourse.psychopy.org/t/editable-textbox-continues-to-appear-on-subsequent-routine-doesnt-end-in-current-routine/16752/7

as well as several similar problems with textboxes appearing when they're not supposed to

PR #3278, opened 2020-11-09
BF: Include relpath when compiling to js

TEParsons:


peircej:
Closing this in favour of  #3285
The fix to the paths should be applied at load time not at compile time (so it gets saved rather than always being incorrect in the psyexp file, this way will fix that). #3285 also fixes the issue at its source (when the files are imported into settings>online>resources box)

PR #3281, opened 2020-11-09
DOCS: Alert 4305 additional content

m-macaskill:
People don't necessarily remember how they disabled a component (e.g. this thread: https://discourse.psychopy.org/t/how-to-enable-a-disabled-component/17896), so now provide more explicit details here.


hsogo:
This causes an issue on non-English Windows. Please see #3274.

PR #3295, opened 2020-11-16
Update eyetracker.py

PanBartosz:
Added support for pupil diameter in mm.
As stated in official API fields LPUPILD and RPUPILD (which can be enabled by SEND_EYE_LEFT and SEND_EYE_RIGHT at init) contain pupil diameters in meters (sic!). These measures need to be converted and it is done by multiplying by 1000 (duh).
This code was tested with real hardware and it works.


peircej:
Thanks @PanBartosz 
@isolver would have a moment to check you're happy with this?

isolver:
Looks great. Thanks.

PR #3299, opened 2020-11-16
ENH: support for 'sqr' with specified dutycycle

rphlypo:
In the grating stim, using `tex='sqrXX'` where XX is an `int` will create a square grating with a duty cycle equivalent to XX%. The trick is to use a regular expression to capture the specific case and process the characters after 'sqr' as an `int`. The generation is based on the `tex='saw'`

### DIFF
--- a/psychopy/visual/basevisual.py
+++ b/psychopy/visual/basevisual.py
@@ -31,6 +31,7 @@ except ImportError:
 import copy
 import sys
 import os
+import re
 from psychopy import logging
 
 # tools must only be imported *after* event or MovieStim breaks on win32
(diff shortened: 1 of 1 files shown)


peircej:
Hey @rphlypo sorry I didn't get back to you about this PR - I just kept forgetting! The reason it wasn't an easy pull in is that it feels like we'll gradually end up with an unmanageable list of names going down this route. Also it means having to parse arguments that are built as strings in to the name, which is harder to work with while coding. Another possibility is that we could provide extra arguments to a dict, which we already do to provide extra args to maskParams. So we could have a texParams, for instance?

rphlypo:
Hey @peircej the extra dict sounds good, but it will probably be a hassle to keep all this well documented. If this kind of feature request remains marginal, no need to dive deeper into it. Thanks for having looked into this one.

mdcutone:
I agree with using something like `texParams`. This will make things a lot more flexible in other places too.

PR #3300, opened 2020-11-17
RF: Split "valType" into "valType" and "inputType", replace ParamCtrls nested if statements with discrete classes

TEParsons:
Major refactor, pulling in now to test Travis.


TEParsons:
Means parameters can have the following val types:
- str
- code
- num
- bool
- list
- file
- color (validates input against valid inputs to the Color class)

and the following input types:
- single (single line text)
- multi (multi line text)
- spin (numeric spinner, integers only)
- bool (tick box)
- choice (selection from list)
- file (includes a browse button)
- fileList (includes buttons to add/remove files from list)
- table (includes a browse button and an "open in excel / other table editor" button)
- color (includes button to open color picker)

Val type now purely describes what type of value to expect, meaning val types like "extendedStr" are no longer needed as this can be described by `valType=str` and `inputType=multi`

PR #3315, opened 2020-11-25
DOC update .psydat-related documentation in dataOutputs.rst

datalowe:
Replace old example (which was written for Python 2 and didn't wo…rk) with new examples for using .psydat files. Update the relevant documentation and make it more beginner-friendly, emphasizing the data retrieval/restoration aspect.

The former example for handling .psydat data doesn't work with the most recent versions of PsychoPy since it uses Python 2 commands (`print` as a statement, not a function) and attributes that ExperimentHandler instances no longer have (`datFile.data`). I added an example that introduces what is probably the use case most relevant to the widest audience, retrieving/restoring data not set to be saved in PsychoPy code/Builder. I wasn't sure what the expected output of the old example used to be or how to recreate it using the latest PsychoPy version, but I replaced it with an example making use of the `entries` attribute since this seemed the most suitable as a substitute.


peircej:
Thanks @datalowe 😄

PR #3325, opened 2020-11-27
Fix paths by having pathto(...) point to the right Sphinx documents

tpronk:
The links in the Documentation dropdown menu of the navbar used relative paths to HTML documents. This didn't work when the navbar was presented in a subdirectory (such as over [here](https://www.psychopy.org/online/index.html)). By replacing the link targets with pathto pointing to the right Sphinx documents, I resolved this issue.


peircej:
Thanks, yes, pathto is very definitely the way to have handled this - I think I copied across from the code without thinking from index.html (which can't use pathto because it isn't compiled by sphinx)

PR #3343, opened 2020-12-05
Remove encoding parameter from json.load for Python 3.9 compatibility.

### LINKED ISSUE #3340: encoding parameter causes crash when using Python 3.9
In Python 3.9, the `encoding` kwarg of `json.loads` is removed, see https://docs.python.org/dev/whatsnew/3.9.html#removed:

> The encoding parameter of json.loads() has been removed. As of Python 3.1, it was deprecated and ignored; using it has emitted a DeprecationWarning since Python 3.8. (Contributed by Inada Naoki in bpo-39377)

This causes a crash because the encoding parameter is passed to json_tricks in https://github.com/psychopy/psychopy/blob/master/psychopy/monitors/calibTools.py#L446
which forwards it to `json.loads` 

```python
  ...
  File "/Users/daniel/miniconda3/envs/os3.3-py39/lib/python3.9/site-packages/json_tricks/nonp.py", line 266, in load
    return loads(string, preserve_order=preserve_order, ignore_comments=ignore_comments, decompression=decompression,
  File "/Users/daniel/miniconda3/envs/os3.3-py39/lib/python3.9/site-packages/json_tricks/nonp.py", line 236, in loads
    return json_loads(string, object_pairs_hook=hook, **jsonkwargs)
  File "/Users/daniel/miniconda3/envs/os3.3-py39/lib/python3.9/json/__init__.py", line 359, in loads
    return cls(**kw).decode(s)
TypeError: __init__() got an unexpected keyword argument 'encoding'
```
I think this parameter can be safely removed for Python 3, since it is a dud anyway, but you may want to retain it for Python 2.

tirkarthi:
Fixes #3340

### DIFF
--- a/psychopy/monitors/calibTools.py
+++ b/psychopy/monitors/calibTools.py
@@ -442,8 +442,14 @@ class Monitor(object):
         else:
             if ext==".json":
                 with open(thisFileName, 'r') as thisFile:
-                    self.calibs = json_tricks.load(thisFile, ignore_comments=False,
-                                                   encoding='utf-8', preserve_order=False)
+                    if six.PY2:
+                        self.calibs = json_tricks.load(thisFile, ignore_comments=False,
+                                                       encoding='utf-8', preserve_order=False)
+                    else:
+                        # Passing encoding parameter to json.loads has been deprecated and
+                        # removed in Python 3.9
+                        self.calibs = json_tricks.load(thisFile, ignore_comments=False,
+                                                       preserve_order=False)
             else:
                 with open(thisFileName, 'rb') as thisFile:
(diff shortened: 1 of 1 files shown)


peircej (changes requested):
The Travis tests are failing due to a bug. You're using `six.Py2` but that isn't defined in this file (and isn't currently a dependency). We have our own constant, PY3 that you can use in similar (opposite) way:
```
from psychopy import constants
if constants.Py3:
    ...
```

mdcutone:
Closed, see https://github.com/psychopy/psychopy/pull/3470

Issue #3348, opened 2020-12-07
ioHub and DataStore on Linux: `RPC_ATTRIBUTE_ERROR` and `decoding str is not supported`


rphlypo:
Below is the output I have when running `psychopy/demos/coder/iohub/launchHub.py`, no idea about how to further deal with the error messages telling `decoding str is not supported` whenever trying to access the DataStore. Anyone with an idea on where to look into?

```bash
Running testWithNoKwargs Test:
pygame 2.0.0 (SDL 2.0.12, python 3.6.11)
Hello from the pygame community. https://www.pygame.org/contribute.html

 ** PRESS A KEY TO CONTINUE.....
gA Keyboard Event was Detected; exiting Test.
ioHub Server Process Completed With Code:  0
Test Result:  None

------------------------------------

Running testUsingPsychoPyMonitorConfig Test:
Display Psychopy Monitor Name:  testMonitor
Display Default Eye Distance:  570.0
Display Physical Dimensions:  {'width': 300.0, 'height': 168.75, 'unit_type': 'mm'}
ioHub Server Process Completed With Code:  0
Test Result:  None

------------------------------------

Running testEnabledDataStore Test:
RPC_ATTRIBUTE_ERROR
Traceback (most recent call last):
  File "/home/rphlypo/miniconda3/envs/psychopy/lib/python3.6/site-packages/psychopy/iohub/server.py", line 108, in handle
    result = getattr(self, unicode(callable_name, 'utf-8'))
TypeError: decoding str is not supported
RPC_ATTRIBUTE_ERROR
Traceback (most recent call last):
  File "/home/rphlypo/miniconda3/envs/psychopy/lib/python3.6/site-packages/psychopy/iohub/server.py", line 108, in handle
    result = getattr(self, unicode(callable_name, 'utf-8'))
TypeError: decoding str is not supported
Display Psychopy Monitor Name:  testMonitor
Display Default Eye Distance:  570.0
Display Physical Dimensions:  {'width': 300.0, 'height': 168.75, 'unit_type': 'mm'}
Experiment Metadata: 
{'code': 'gap_endo_que',
 'description': '',
 'experiment_id': 'C',
 'title': '',
 'total_sessions_to_run': 0,
 'version': ''}

Session Metadata: 
{'code': 'S_07_12_2020_16_18',
 'comments': '',
 'name': '',
 'session_id': 'C',
 'user_variables': {}}
ioHub Server Process Completed With Code:  0
Test Result:  None

[...]

PR #3350, opened 2020-12-07
BF: python 3 unicode manipulations unnecessary and non-iterable error fix…

rphlypo:
When solving the decoding problem with non-binary strings in unicode format, an error was handled by iterating over the non-iterable error (check whether "decoding str is not supported" in e). This has been solved by only looking at the string representation of e, which is iterable just as any string. This also solves Issue #3348

### DIFF
--- a/psychopy/iohub/server.py
+++ b/psychopy/iohub/server.py
@@ -107,13 +107,16 @@ class udpServer(DatagramServer):
             try:
                 result = getattr(self, unicode(callable_name, 'utf-8'))
             except TypeError as e:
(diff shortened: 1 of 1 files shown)


peircej:
Looks like the iterable one is my fault recently. Sorry about that! 🤭 

Would you mind adding prefixes to your commits. This one is a BF - it's a bug in released code so the commit needs cherry picking to the maintenance branch for 2020.2. The prefix is important in helping me find that

thanks for all your input lately! 😃

PR #3366, opened 2020-12-13
FF: Make pyo parsing of devices more robust

### LINKED ISSUE #3365: pyo: error when parsing devices
In the following method of the pyo backend, depending on the connected devices, it can fail to parse the device list:

https://github.com/psychopy/psychopy/blob/0b0d4f3c26584ecf720c44406dbdecbd69b75ea4/psychopy/sound/backend_pyo.py#L77-L89

Traceback:

```
  File "/home/elodie/.local/lib/python3.6/site-packages/psychopy/sound/backend_pyo.py", line 120, in get_input_devices
    devices = _query_devices()
  File "/home/elodie/.local/lib/python3.6/site-packages/psychopy/sound/backend_pyo.py", line 86, in _query_devices
    info[v.strip().split(": ")[0]] = v.strip().split(": ")[1]
IndexError: list index out of range
```

The method `pyo.pa_list_devices()` returns a string, which is then split by line and each line is parsed, splitted by commas.
Here's an example of device (on Ubuntu) which causes an error:

```
AUDIO devices:
0: IN, name: Logitech USB Headset: Audio (hw:1,0), host api index: 0, default sr: 44100 Hz, latency: 0.008707 s
0: OUT, name: Logitech USB Headset: Audio (hw:1,0), host api index: 0, default sr: 44100 Hz, latency: 0.008707 s
1: IN, name: sysdefault, host api index: 0, default sr: 48000 Hz, latency: 0.021333 s
1: OUT, name: sysdefault, host api index: 0, default sr: 48000 Hz, latency: 0.021333 s
2: OUT, name: front, host api index: 0, default sr: 44100 Hz, latency: 0.008707 s
3: OUT, name: surround40, host api index: 0, default sr: 44100 Hz, latency: 0.008707 s
4: OUT, name: surround51, host api index: 0, default sr: 44100 Hz, latency: 0.008707 s
5: OUT, name: surround71, host api index: 0, default sr: 44100 Hz, latency: 0.008707 s
6: OUT, name: hdmi, host api index: 0, default sr: 44100 Hz, latency: 0.008707 s
7: IN, name: default, host api index: 0, default sr: 48000 Hz, latency: 0.021333 s
7: OUT, name: default, host api index: 0, default sr: 48000 Hz, latency: 0.021333 s
8: OUT, name: dmix, host api index: 0, default sr: 48000 Hz, latency: 0.021333 s
9: IN, name: /dev/dsp, host api index: 1, default sr: 44100 Hz, latency: 0.008707 s
9: OUT, name: /dev/dsp, host api index: 1, default sr: 44100 Hz, latency: 0.008707 s
```

As the line `0: IN, name: Logitech USB Headset: Audio (hw:1,0), host api index: 0, default sr: 44100 Hz, latency: 0.008707 s` is split by commas, at some point, it is confused by `(hw:1,0)` in `name: Logitech USB Headset: Audio (hw:1,0)`.

I have no background on psychopy and pyo, but I wonder if `_query_devices` couldn't be replaced by [`pyo.pa_get_devices_infos()`](https://github.com/belangeo/pyo/blob/274ed61d6664d44d23416957e781b8afbb559589/src/engine/pyomodule.c#L139)?

guewen:
The methods to get input/output sound devices info in the pyo backend
rely on a parsing of the output of `pyo.pa_list_devices()`. It splits
the fields (name, host api index, latency, default sr) by splitting
on commas. This is fragile as the name can contain commas (e.g.
"Logitech USB Headset: Audio (hw:1,0)"), which leads to `IndexError:
list index out of range` error.

`pyo` provides a method [`pa_get_devices_infos()`](https://github.com/belangeo/pyo/blob/274ed61d6664d44d23416957e781b8afbb559589/src/engine/pyomodule.c#L139) that returns 2
dictionaries for input and output devices, almost exactly in the
expected format as returned by the current implementation of
`_query_devices()`. The only differences are in the fields values: no
units ("Hz", "s"), do not split the name to exclude the content after a
colon and the delay has many digits (vs 6 in the initial
implementation).

This commit switches to using `pa_get_devices_infos()` with a small
adaptation of the format to keep the same output (but the name) as the
previous implementation was doing.

:warning:  Behavior change:

The name of a sound device is now more precise, e.g.

Before: "Logitech USB Headset"
After: "Logitech USB Headset: Audio (hw:1,0)"

It makes the identifier more unique, as it contains the hardware position,
for example if we have several identical headsets connected.

Fixes #3365

### DIFF
--- a/psychopy/sound/backend_pyo.py
+++ b/psychopy/sound/backend_pyo.py
@@ -12,6 +12,7 @@ import sys
 import os
 import atexit
 import threading
+from itertools import chain
 from numpy import float64
 from psychopy import prefs, exceptions
 from psychopy import core, logging
@@ -74,57 +75,29 @@ def _bestDriver(devNames, devIDs):
         return None, None
 
 
-def _query_devices():
-    f = StringIO()
-    with redirect_stdout(f):
-        pyo.pa_list_devices()
-    s = f.getvalue().strip().split("\n")[1:]
-    devices = []
-    for device in s:
-        info = {}
-        for v in device.split(",")[1:]:
-            info[v.strip().split(": ")[0]] = v.strip().split(": ")[1]
-        info["id"], info["type"] = device.split(",")[0].split(": ")
-        devices.append(info)
-    return devices
-
-
 def get_devices_infos():
-    devices = _query_devices()
-    in_devices = {}
-    out_devices = {}
-    for device in devices:
-        param = {'host api index': device['host api index'],
-                 'latency': device['latency'],
-                 'default sr': device['default sr'],
-                 'name': device['name']}
-        if device['type'] == 'IN':
-            in_devices[int(device["id"])] = param
-        if device['type'] == 'OUT':
-            out_devices[int(device["id"])] = param
+    in_devices, out_devices = pyo.pa_get_devices_infos()
+    for index, device in chain(in_devices.items(), out_devices.items()):
+        device.update({
+            'default sr': '{} Hz'.format(device['default sr']),
+            'host api index': str(device['host api index']),
+            'latency': '{} s'.format(round(device['latency'], 6)),
+        })
     return (in_devices, out_devices)
 
 
 def get_output_devices():
-    devices = _query_devices()
-    names = []
-    ids = []
-    for device in devices:
-        if device['type'] == 'OUT':
-            names.append(device['name'])
-            ids.append(int(device["id"]))
-    return (names, ids)
+    _, out_devices = get_devices_infos()
+    return tuple(zip(*[
+        (device['name'], dev_id) for dev_id, device in out_devices.items()
+    ]))
 
 
 def get_input_devices():
-    devices = _query_devices()
-    names = []
-    ids = []
-    for device in devices:
-        if device['type'] == 'IN':
-            names.append(device['name'])
-            ids.append(int(device["id"]))
-    return (names, ids)
+    in_devices, _ = get_devices_infos()
+    return tuple(zip(*[
+        (device['name'], dev_id) for dev_id, device in in_devices.items()
+    ]))
 
 
 def getDevices(kind=None):


guewen:
I'm not sure if I should pick BF or FF, please let me know, I'll gladly update my commit.
Also, as a new contributor, I may miss context or use cases, in such case, please do not mind to tell me :)

peircej (approved):
Thanks @guewen 
The BF/FF distinction is interesting here. This is a fix to existing released code (so normally BF) but my guess is that is could involve a behaviour change - the name of the audio card being returned might end up being different in the new more robust method, and that might cause existing experiment to stop working (exp can't find the audio card it was told to use?). So technically this is a BF but I might squahs it to FF so that it doesn't get included until the next major release (2021.1)

peircej:
> Thanks @guewen
> The BF/FF distinction is interesting here. This is a fix to existing released code (so normally BF) but my guess is that is could involve a behaviour change - the name of the audio card being returned might end up being different in the new more robust method, and that might cause existing experiment to stop working (exp can't find the audio card it was told to use?). So technically this is a BF but I might squahs it to FF so that it doesn't get included until the next major release (2021.1)

The question is whether we think the returned names will exactly match

peircej:
@guewen how urgent is this as a release for you? Do you have others that depend on it or are you OK because fixed on your local install?

guewen:
Thanks for the fast review!



> The question is whether we think the returned names will exactly match

AFAIU, the name should be identical, because both in `pa_get_devices_infos()` and `pa_list_devices()`, it comes from the same place (http://files.portaudio.com/docs/v19-doxydocs/structPaDeviceInfo.html).

That being said, I can't pretend to know enough and I have no guarantee that the name can't change (also on other systems such as windows), so I'm 100% fine with taking the safe way and include this in the next major release only. 

> Do you have others that depend on it or are you OK because fixed on your local install?

I'm fine with keeping this on my local install until the next major release.

guewen:
BTW it's worth checking what is the desired behavior as we put it as FF.

When we have a name such as `"Logitech USB Headset: Audio (hw:1,0)"`, the previous implementation was keeping only `Logitech USB Headset`, but re-reading the code, it's hard to know if it was the intent or only a side-effect of the manual parsing on `:` to split the keys and values (sounds like the second one).
I could change to keep the whole name intact. But it's probably not worth having to reconfigure exp in the wild if keeping only the first part was not an issue.

peircej:
> BTW it's worth checking what is the desired behavior as we put it as FF.

Yes, indeed. Let's allow potential change and save it for 2021.1. So, regarding your Q the advantage of keeping all of the name `"Logitech USB Headset: Audio (hw:1,0)"` might be that I guess that's identifying it more completely/uniquely. In your example above it looks like it's identifying its hardware position so this would be unique if you had two Logitech headsets connected, whereas if we split on the first `:` we could get a non-unique name. So my inclination would be to keep the whole name intact

guewen:
I'm not sure if I should pick BF or FF, please let me know, I'll gladly update my commit.



> I guess that's identifying it more completely/uniquely. In your example above it looks like it's identifying its hardware position so this would be unique if you had two Logitech headsets connected, whereas if we split on the first `:` we could get a non-unique name. So my inclination would be to keep the whole name intact

Very good point! I amended my commit to keep the name intact (and changed to FF).

peircej:
Thanks @guewen 
You're the perfect contributor! 😊

PR #3367, opened 2020-12-14
FF: formatted string literal should not be used with _translate()

hsogo:
For example, 
```
newName = 'foo'
msg = _translate(f'{newName} is used. Try another one.')
```
PoEdit registers `'{newName} is used. Try another one.'` to translation catalog.
When this code is executed, it gives
```
msg = _translate('foo is used. Try another one.')
```
This string cannot be translated because it is not registered in the catalog.


peircej:
@TEParsons be aware of this. Translate needs to be run on the unformatted strings so don't use f-strings for anything where translation is required (e.g. in UI widgets). You can use older formatted string methods as long as the `format()` attribute is called *after* `_translate()`

PR #3370, opened 2020-12-15
NF: Add "disabled" button for Code components

TEParsons:


LunkRat:
I will use this feature. I was wishing it existed as I was managing disabled components in Builder. Thanks!

PR #3385, opened 2021-01-05
Input types

TEParsons:
Split valType into valType and inputType - making a distinction between "this is what kind of data the parameter produces" and "this is the kind of control needed to get data for this parameter". So valTypes "extendedStr" and "extendedCode" are now irrelevant - they are now valTypes "str" and "code" with inputType "multiple". Also adds classes extending wx controls for different inputTypes, making it much easier to add new kinds of controls.

valTypes:
- str: string
- code: code (verbatim)
- num: int or float
- bool: boolean
- list: list or tuple
- file: path to a file
- color: color spec string or Color object

inputTypes:
- single: a single line text control
- multi: a multi line text control
- spin: a numeric spinner
- choice: a wx.Choice control
- bool: a tickbox
- file: a single line text control with a File Open button
- fileList: a control for multiple files
- table: a file control with an Open In Excel button
- color: a single line text control with a color picker button

All component dialogs open and compile, but there are still some spacing issues and need to test more thoroughly before the next major release. Also discussed with Jon some extra things such as making inputTypes able to be switched between (e.g. deprecating $ and instead having the $ on a code control be a choice control with str and code as options) but this is a bigger change so may be better to wait for a summer release.


peircej:
Task linked: [CU-c51vk9 inputType](https://app.clickup.com/t/c51vk9)

Issue #3384, opened 2021-01-04
Error in Monitor Center when "Plot gamma" is clicked with no calibration data loaded


mdcutone:
Minor issue but raises the following exception:

```
Traceback (most recent call last):
  File "C:\Program Files\PsychoPy3\lib\site-packages\psychopy\monitors\MonitorCenter.py", line 1028, in plotGamma
    if lumsPre.any() != None:
AttributeError: 'NoneType' object has no attribute 'any'
```

mdcutone:
I guess it should just do nothing or be greyed out.

PR #3390, opened 2021-01-07
BF: 2021.x bugfixes

### LINKED ISSUE #3382: Prefs dialog: "module 'psychopy.sound' has no attribute 'getDevices'"
I tried adding `pygame` as the first choice in the list of audio libraries, since that was working well for me in an older version PsychoPy (with Python 2).

However, it doesn't seem to work on this machine, and now it's not possible to open the preferences dialog to undo the change:
```
  File "/home/omf/.local/lib/python3.8/site-packages/psychopy/app/_psychopyApp.py", line 767, in showPrefs
    prefsDlg = PreferencesDlg(app=self)
  File "/home/omf/.local/lib/python3.8/site-packages/psychopy/app/preferencesDlg.py", line 543, in __init__
    devnames = sorted(sound.getDevices('output'))
AttributeError: module 'psychopy.sound' has no attribute 'getDevices'
```

mdcutone:
Fixes #3382 #3384  and https://discourse.psychopy.org/t/psychopy-encountered-an-unhandled-error/18784


mdcutone:
@TEParsons Fixed the close buttons appearing on the color picker tabs.

mdcutone:
Just a note for @peircej and @TEParsons - Fixed the display flickering when moving the mouse and scrolling. Keep in mind that `GetContentScaleFactor` is not a good way of detecting if we are running on a retina display. Window and Linux also have HiDPI modes these days and double buffering is switched off on those platforms with the previous implementation. To determine if we are running on retina, also check if the platform is mac. Also, don't double buffer the whole frame as it was doing before, just individual panels as needed.

Issue #3391, opened 2021-01-07
A problem readme files if MarkdownIt not installed


peircej:
This came from the Navon demo task with MarkdownIt missing. I wonder if there was a removal of \n somewhere when markdown support was added?

![image](https://user-images.githubusercontent.com/380582/103944331-62ebdb80-512b-11eb-8d36-8f362ea3b87e.png)

TEParsons:
Ah I see the problem - if MarkdownIt is not installed, then it defaults to just printing the original text. I was thinking that would be the same as befpre, but it prints it as undecorated HTML, meaning no line breaks. I can solve this specific problem by a straight replacement of "/n" for "<br>", but could we add MarkdownIt to the project requirements?

TEParsons:
Didn't occur to me that html would be interpreted by GitHub in the comment :') I meant to say 
```
<br>
```

PR #3394, opened 2021-01-08
FF: Replace \n with HTML newline in README text when Markdown cannot be parsed

TEParsons:


TEParsons:
Fixes #3391

peircej:
Ah, I see, it's that we're now rendering as html even if MarkdownIt not found. OK, next possible issues: 
1. what happens if MarkDownIt not found and the text file contains other line endings (windows EOL is often \r\n instead of \n, old macOS used to use only \r). So on win the current replace will mean that a \n gets left in the text (and ignored during html render I guess?) and the others are all fine.
2. Next problem is what happens on save? If we replace \n with <br> and then they open the file in Notepad or similar what happens then? Possibly we need to detect the line-ending on file-load and on file-save we replace back to that file-end character
3. Actually, looking at the code it seems like, even if they do have markdownIt installed, the point at which the file gets saved will now save an html formatted file instead of the original markdown.

PR #3403, opened 2021-01-11
FF: Depreciate xlrd as it is no longer supporting xlsx files

TEParsons:
only use to open xls files and use openpyxl for everything else

### DIFF
--- a/psychopy/tests/test_data/test_utils.py
+++ b/psychopy/tests/test_data/test_utils.py
@@ -16,7 +16,7 @@ class Test_utilsClass:
     def test_importConditions(self):
         standard_files = []
         standard_files.append(join(fixturesPath, 'trialTypes.xlsx'))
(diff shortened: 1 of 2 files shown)


psychopy/tests/test_data/test_utils.py
```diff
@@ -16,7 +16,7 @@ class Test_utilsClass:
     def test_importConditions(self):
         standard_files = []
         standard_files.append(join(fixturesPath, 'trialTypes.xlsx'))
-        standard_files.append(join(fixturesPath, 'trialTypes.xls'))
+        #standard_files.append(join(fixturesPath, 'trialTypes.xls')) # xls is depreciated
```
peircej, line 19: I think this turned out not to be true - we misunderstood the pandas changelog to think that xls was no longer supported but actually the issue was that xlread is now only used for xls files (no longer for xlsx files) so xls is still fine

PR #3415, opened 2021-01-14
RF: Remove fillColorSpace and borderColorSpace

TEParsons:


peircej:
Do you have something in place so that experiments from older versions, that do have that parameter, still work?

TEParsons:
@peircej This only removes it from the param def, the classes for the objects will still accept them as inputs, it's just that new components won't have them (and they were never required as inputs so new components on old versions should also be fine)

peircej:
Yes, the param will still exist for the user's exp (because it gets added back as the Component loads) but it won't now have an effect on the drawing. So the experiment will stop working if they have, for instance, specified a line in RGB and a fill in another space. I suppose this is rare enough that we don't care? The user presumably does get the standard warning about "the experiment contains an unknown param and might not work as expected"

PR #3424, opened 2021-01-19
RF: Param order

TEParsons:
Pulling in the commits from reordering parameters piece by piece to find source of Travis stalling

### DIFF
--- a/psychopy/app/builder/dialogs/__init__.py
+++ b/psychopy/app/builder/dialogs/__init__.py
@@ -590,6 +590,7 @@ class _BaseParamsDlg(wx.Dialog):
                                                         currRow)
         currRow += 1
         # loop through the prescribed order (the most important?)
+        self.order = list(dict.fromkeys(self.order)) # Remove any duplicates, keep only the first instance
         for fieldName in self.order:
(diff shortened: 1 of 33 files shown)


TEParsons:
@peircej if this last push passes Travis then this will be alright to pull in, but the commit history is a mess because of my bit-by-bit figuring out where the problem is (it was in RatingScale btw), so could you do a squash merge with this as the commit message:

```
RF: Rather than each component overwriting its base .order, components now append to parent .order lists as necessary, meaning that e.g. all visual components will have colors in the same order. More generally, the order of parameters within components has been updated, along with the following fixes and enhancements:
- ENH: The default font for relevant components is now "Open Sans", as this is independent of locally installed fonts
- ENH: componsTemplate.txt has now been updated to reflect current state of component parameters, dramatically reducing spam in Travis
- ENH: "dollarSyntax" is now ignored when testing components to reduce Travis spam
- FF: RadialStim is no longer tested without shading in 3.6
- BF: Duplicates are removed from self.order when assembling a component dlg, so that the same parameter can be defined multiple times without breaking the dlg, making it easier to make new components without tripping up on base .order lists
```

PR #3430, opened 2021-01-20
BF: Textbox foreground color was not rendering correctly

TEParsons:


TEParsons:
This tree got kind of messy, reopening with a fresh branch

PR #3437, opened 2021-01-21
RF: Remove `kwargs` param from various classes.

mdcutone:
Here we're removing the `kwargs` argument from various classes. While this is usually a good design pattern, we find that this tends to confuse users since incorrectly specified arguments are silently soaked up causing unexpected behaviour. Now a hard error is raised.

To allow configuration of unique features available to only a sub-set of supported backends, a dictionary of options can be passed to the `bakendConf` parameter as so:

```
options = {'bpc': (8, 8, 8), 'depthBits': 24, 'stencilBits': 8}
win = visual.Window(winType='glfw', backendOptions=options)
```  

More work will be done to other classes which use the `kwargs` argument.


mdcutone:
Seems to be failing here:

```
sychopy/tests/test_app/test_builder/test_Experiment.py::TestExpt::test_xsd FAILED
```

PR #3457, opened 2021-01-25
BF: Fixes segfault when Window is set to just Runner, fixes #3112

### LINKED ISSUE #3112: Windows seg fault when defaultView set to Runner

TEParsons:


mdcutone:
I was wondering what was happening here. Thanks for the fix.

PR #3459, opened 2021-01-25
BF: Fix ElementArrayStim

tiborauer:
The previous version of ElementArrayStim fails when nElements is non-divisible with the number of colors.


TEParsons:
While discussing this bug we realised that iterating through a list of individual Color objects in the way I'd got it set up was severely impacting speed, so I've fixed this in a different way - utilising vectorised Color objects - so this fix will be overwritten. But thank you for flagging it up, it's this PR that drew our attention to the deeper problem!

PR #3464, opened 2021-01-26
FF: translation of choice labels is corrupted

hsogo:
Translation of choice items on the Preference Dialogs has been corrupted since 4e865a23514fe86f1817dc027707eade1c622c5a.
Item labels must be translated before creating wx.Choice object.
Because labels are localized, `SetStringSelection()` doesn't work.  Instead, get index of the selected item and use `SetSlection()`.


peircej:
Excellent, thanks. @TEParsons @mdcutone note the changes here for future reference (especially, not using f-strings in strings that need translations)

PR #3479, opened 2021-01-28
Fixes from flake8 - mostly corner cases but still often actual bugs

peircej:


peircej:
@mdcutone and @isolver you might be interested in this one. @isolver I split the iohub-relevant fixes into a separate commit 0e59717

mdcutone:
Is there a way to get a list of these bugs so we can work through them?

peircej:
> Is there a way to get a list of these bugs so we can work through them?

You mean the bugs that were giving rise to the changes? To get that you install flake8 and run (on the previous commit) a command like:
```
flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```

PR #3484, opened 2021-01-29
ENH: Builder demos

TEParsons:
"modernising" buider demos - rebuilding so that components contain all the correct params after the param rework, as well as replacing Text stim with TextBox stim


TEParsons:
To avoid merge conflicts I've applied all these changes whilst reordering demos instead

PR #3487, opened 2021-01-29
NF: Add Button component

TEParsons:


psychopy/experiment/components/button/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+        buff.writeIndentedLines(code % inits)
+
+    def writeRoutineEndCode(self, buff):
+        BaseVisualComponent.writeRoutineEndCode(self, buff)
+        if len(self.exp.flow._loopList):
+            currLoop = self.exp.flow._loopList[-1]  # last (outer-most) loop
+        else:
+            currLoop = self.exp._expHandler
+        name = self.params['name']
+        code = f"{currLoop.params['name']}.addData('{name}.rt', t)\n"
```
peircej, line 203: This looks like you're adding just the time of the end of the routine rather then the time the button was pressed. Maybe check the time code for mouse rt storage

psychopy/experiment/components/button/__init__.py
```diff
@@ -0,0 +1,210 @@
...
+            # Add indent to any linebreaks to account for the loop
+            callback = str(inits['callback']).replace("\n", "\n   ")
+        else:
+            callback = "pass"
+        if inits['forceEndRoutine'].val:
+            endRt = "continueRoutine = False"
+        else:
+            endRt = ""
+        if inits['oncePerClick'].val:
+            hammingStr = "   while %(name)s.isClicked:\n      pass\n"
```
peircej, line 185: hamming? For a sound that indicates the smoothing filter to prevent audio clicks. I haven't heard the term with button presses before

peircej (commented):
Not sure the time (RT) is being measured correctly here - we need to get that right

PR #3509, opened 2021-02-02
BF: Only add "editable" to Window's editables list if it is editable …

TEParsons:
…- otherwise non-editable Textboxes get set to editable

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -949,8 +949,8 @@ class Window(object):
         # Ignore if object is not editable
         if not hasattr(editable, "editable"):
             return
-        #if not editable.editable:
-        #    return
+        if not editable.editable:
+            return
         # If editable is already present do nothing
         eRef = False
         for ref in weakref.getweakrefs(editable):


TEParsons:
added @isolver as reviewer to make sure this won't contradict any fixes to the weakref problem

isolver:
Good catch. I'm realizing I did not really finish the changes I was making to the editable stim logic. For example, what happens if a textbox2 goes from editable=False to editable=True during the experiment? I do not think this is currently handled. 

I also did not remove the textbox2.hasFocus setter method, which is not longer needed or used because .hasFocus now just returns textbox2 instance == the window.currentEditable. 

I'll finish up /clean up the textbox2 editing and submit a seperate PR with the above things fixed. Thanks.

TEParsons:
Having a property which does that is still pretty handy though, it means we can set/check whether a textbox has focus by calling `textbox.hasFocus` rather than having to faff with the window, so I think just synonymising them rather than deleting one (as you've done) is the better way :)

isolver:
If a textbox instance is not == to the win.currentEditable, then it can not have focus, so it does not make sense IMO to have textbox storing something that can just be checked at runtime and does not need to be kept in sync. We still have .hasFous as a property, but internally it is just checking again win.currentEditable.

isolver:
I guess we could keep the setter, and if set from True -> False, then we could call window.nextEditable() in the setter, as a convience. 

My point above was I don't think we should keep the same logical value in two places in the code - i.e. window tells us what editable hasFocus (window.currentEditable) and the editables should use that instead of keeping track of it in a local var that needs to then be kept in sync with window.currentEditable.

isolver (commented):
Since this fix is also in PR #3512, can we close this PR assuming #3512 will be accepted?

PR #3512, opened 2021-02-02
BF: Misc. TextBox2 window focus fixes

isolver:
BF: Changing textbox.editable at runtime was not supported.
BF: Do not add editable to window editables list if editable.editable == False
RF: textbox2.hasFocus setter now updates window editable focus if necessary

### DIFF
--- a/psychopy/demos/coder/stimuli/textbox_editable.py
+++ b/psychopy/demos/coder/stimuli/textbox_editable.py
@@ -1,11 +1,10 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-import psychopy
 from psychopy import visual, core, event, logging
 import numpy as np
 
-logging.console.setLevel(logging.EXP)
+#logging.console.setLevel(logging.EXP)
 c = core.Clock()
 
 from psychopy.visual.textbox2 import TextBox2, allFonts
@@ -51,15 +50,24 @@ txt2.autoDraw=True
 
 clock = core.Clock()
 t=0
-while t<20:
+while t<30:
     t= clock.getTime()
 
     txt2.pos = (0.2*np.sin(t), 0.2*np.cos(t))
-    
+        
+    if t > 10 and t < 20:
+        if txt3.editable is False:
+            txt3.editable=True
+            txt3.autoDraw=True
+    elif t > 20:
+        if txt3.editable is True:
+            txt3.editable=False
+            txt3.autoDraw=False
+
     # Do not need to call draw() each flip() 
     # for txt1 or txt2 since autoDraw is enabled for them.
-        
-    txt3.draw()
+    if txt3.editable is False:
+        txt3.draw()
     
     if 'escape' in event.getKeys():
         core.quit()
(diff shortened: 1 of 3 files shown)


isolver:
Hi Todd, this is what I had in mind. It also implements the fix you put in [PR #3509](https://github.com/psychopy/psychopy/pull/3509). What do you think? 

I didn't add back a property setter for textbox2.hasFocus, but if you think it is needed I can do so and have it update window.currentEditable accordingly. 

Thanks.

isolver:
Yes, so now setting textbox2.hasFocus or textbox.editable should now affect window.currentEditable if needed. Opposite should also be true so they are kept in sync. 

Thinking on it more, I also now see why it could be good to be able to set textbox2.hasFocus as a way to set window.currentEditable, say based on a mouse press ;), so I added that setter back.

peircej:
I don't think the user should need to change the editable setting. The box is intended as an editable box - it is an input field and that is true whether or not the box is currently being drawn. Whether or not the editable box included in the list of editable objects tracked by the Window is a different matter - that should be determined by whether the box is editable AND being drawn

isolver:
@peircej:

> I don't think the user should need to change the editable setting.

The user does not need to, if a TextBox2 is created as editable=True, it is. Previously, if .editable was updated in textbox2 nothing was updated in the window editable list. That is fixed in this PR.

Or do you think textbox2.editable should be a read-only property that can not be set after the textbox2 is created in the script?

> Whether or not the editable box included in the list of editable objects tracked by the Window .... should be determined by whether the box is editable AND being drawn

It should be working like this now, but I'll double check.

isolver:
> Whether or not the editable box included in the list of editable objects tracked by the Window .... should be determined by whether the box is editable AND being drawn

Should have checked before commenting. Looks like if textbox2.autodraw = True, the textbox2 becomes editable even if editable = False when the textbox2 was created....

Looks like basevisual.py (line 171) autoDraw setter is always setting textbox2.editable to true if autodraw is set to true, regardless of initial textbox2.editable value. @peircej, Do you agree this is a bug? It should only be setting it to true if editable is also true, correct?

peircej:
Yes, those lines in the demo are a bit strange I think - they suggest to users that this should be needed whereas I think it would be unusual that users would be switching the editable attribute of a box mid-script

> Looks like basevisual.py (line 171) autoDraw setter is always setting textbox2.editable to true if autodraw is set to true, regardless of initial textbox2.editable value. @peircej, Do you agree this is a bug? [edit] autodraw should not need to set the value of textbox2.editable at all should it?

Yes, if a box is not editable and set to autodraw it should not become editable. It's just that the decision to add it to the editable list of the window should be based on the joint decision of something being editable AND being autodrawn

isolver:
Sorry, ignore last (deleted) comment, I was looking at wrong branch. ;(

As of this PR textbox2 is, from what I can tell / test, working as @peircej requested:

> Whether or not the editable box included in the list of editable objects tracked by the Window .... should be determined by > whether the box is editable AND being drawn

Even in the case of textbox2.editable changing value during the experiment.

isolver:
@peircej, the demo update was me testing that things worked. I'll revert to the original demo.

PR #3526, opened 2021-02-05
ENH: Modernising Coder demos

TEParsons:
Updating a few Coder demos to use TextBox2, also using the Knight character in a Google font to make the custom mouse demo a bit fun ;)


isolver:
Currently textbox2 is slower than TextStim when drawing text that does not change (3 times slower). Should we wait to update demos until we have TextBox2 performing better than TextStim?

Latest output from demos/coder/stimuli/compare_text_timing.py:

```
+ Draw Order: ['TextBox', 'TextBox2', 'TextStim']
+ Text Stim Char Length:         160

+ TextBox INIT Dur (sec):       1.0631777000089642
+ TextBox2 INIT Dur (sec):      0.12750420000520535
+ TextStim INIT Dur (sec):      0.37077660000068136

+ Text Change Flip Perc:        20.00
+ Total Flip Count:              600
+ Test Duration (secs):         4.187
+ FPS:                          143.291

+ Average Draw Call Durations (msec):

  Text Object           No Txt Change   Txt Change
  TextBox               0.265           1.779
  TextBox2              1.515           4.491
  TextStim              0.477           6.886
```

Issue #3539, opened 2021-02-08
colors.py demo exception


isolver:
While testing colors.py demo with the dev4 build, the following exception occurred:

```
Traceback (most recent call last):
  File ".psychopy\demos\coder\stimuli\colors.py", line 69, in <module>
    stim.draw()
  File ".\psychopy\visual\textbox2\textbox2.py", line 592, in draw
    self.box.setLineWidth(self.pallette['lineWidth']) # Use 1 as base if border width is none
  File ".\psychopy\visual\textbox2\textbox2.py", line 247, in pallette
    'fillColor': self._fillColor+0.1
  File ".\psychopy\colors.py", line 381, in __add__
    buffer.rgba = self.rgba + other
numpy.core._exceptions.UFuncTypeError: ufunc 'add' did not contain a loop with signature matching types (dtype('<U32'), dtype('<U32')) -> dtype('<U32')
```

I was pressing the different color buttons and typing in random text when it occurred.

I'm also not able to select the left or right most color types displayed in the demo using the mouse.

TEParsons:
Forgot to tag but this is fixed in #3541

PR #3541, opened 2021-02-08
BF/ENH: Fixes and improvements to Slider

TEParsons:


mdcutone:
Just tagging that this fixes #3539 so it closes the associated issue.

PR #3543, opened 2021-02-08
BF: Set opacity /after/ setting colors for shape stim

### LINKED ISSUE #3494: Setting 'opacity' in stim constuctor has no effect
For example,

```
circle = visual.Circle(
            win,
            pos=(0, 0),
            lineWidth=3.0,
            lineColor=[255,0,0],
            colorSpace='rgb255',
            fillColor=[255,0,0],
            radius=50,
            units='pix',
            opacity=0,
            interpolate=False)
circle.draw()
```

draws a solid red circle, even though `opacity` = 0. However setting opacity after stim is created works:

```
circle.setOpacity(0)
```

does hide the circle. Also occurs in Line stim, all should probably be checked.

TEParsons:
as Color objects have their own alpha values, if you set them after setting opacity then these override the object's opacity (whereas the object's opacity should override the Color objects'), fixes #3494

### DIFF
--- a/psychopy/visual/shape.py
+++ b/psychopy/visual/shape.py
@@ -126,7 +126,6 @@ class BaseShapeStim(BaseVisualStim, ColorMixin, ContainerMixin):
         super(BaseShapeStim, self).__init__(win, units=units,
                                             name=name, autoLog=False)
 
-        self.opacity = opacity
         self.pos = numpy.array(pos, float)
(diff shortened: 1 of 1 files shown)


peircej:
Is it possible this is causing the screenshots to fail. Maybe they were previously re-compiled in the /broken/ state?

PR #3546, opened 2021-02-08
Localization issues (fixing clash between hsogo translations and new colorPicker)

peircej:


hsogo:
Hi Jon, 
This fix includes new strings that need to be translated.  I think it's better to update translation after this pull request is merged to the master branch. Is that okay?

peircej:
Sounds good. I've pulled that in. The tests are currently not passing but that's for a different reason

PR #3548, opened 2021-02-09
BF: Previous opacity fix was causing all stimuli to default to opaque

TEParsons:
meaning that None became grey


peircej:
Why would we set the default opacity to be None rather than 1? Opaque seems like a good default, no?

TEParsons:
The problem is that opaque sets everything to opaque - even if its colour is set to None. None is a better default as it's essentially "not set", i.e. use the alpha value of the Color object. Default alpha on a Color is 1, so if no alphas are set then None opacity is functionally the same as 1 only it doesn't override None colour.

PR #3550, opened 2021-02-10
NF: Add new type of ParamCtrl for multiple choice and use this to add "adjustments" param to slider

TEParsons:
@peircej I've done this in two commits so that we can discuss the changes to Slider and cherry pick them out if they don't make sense, but following discussions with @tpronk I think this split is the way to go. Will raise at standups today so we can all chip in, just pushing out now to check Travis.


tpronk:
Cool :). Once we cleared this up I'll sync it up with the JS and adapt our
interactive tests.

PR #3558, opened 2021-02-11
BF: Forms with editable TextBox2 and iohub mouse setPosition

isolver:
BF: iohub mouse.setPosition was not working correctly
BF: Editable textbox2 elements within a form were not getting input because form autodraw needed to set textbox2 autodraw.


TEParsons (commented):
Seems to work!

PR #3560, opened 2021-02-12
RF/ENH: Changing Form styles to be more like how Slider styles are now

TEParsons:
i.e. text color, line color, etc. are controlled by color attributes and style controls the overall appearance of the form. I've added one style to get things started - makes it look like a legal pad


TEParsons:
Shelving until summer

PR #3568, opened 2021-02-15
BF: Fix label positioning on vertical Sliders

TEParsons:


TEParsons:
TextBox2 needs to use norm units before working with Slider - closing for now and will try to fix the ancorHoriz issue in Text instead

Issue #3567, opened 2021-02-14
Slider has no attribute opacity


RebeccaHirst:
The opacity of a slider does not change with the opacity parameter from builder (e.g. A slider drawn over a polygon with opacity set to .5). It looks as though the slider class no longer has the attribute opacity?

PR #3578, opened 2021-02-17
BF/ENH: Allow Slider opacity and colors to be set dynamically

TEParsons:


TEParsons:
Partially fixes #3567, but Slider labels still don't take opacity (can't figure out why - if you print self._foreColor.alpha in the draw function for Text stim it returns as expected, but the Text stim appears opaque regardless, but only for Slider labels...)

PR #3582, opened 2021-02-17
ENH: Make slider style Slider marker really thin, for two reasons: 

TEParsons:
1. makes it less visible when it goes off the edge (as it doesn't have a buffer like with scrollbar style), 2. make it look more different to scrollbar style


peircej:
Before I pull this in could you show a screenshot of what it now looks like?
(Are scientists going to like it if they've been using this to collect data?)

PR #3584, opened 2021-02-17
ENH: save variable "reliableFrame" to output for online experiments (…

RebeccaHirst:
…was the frame rate reliably measured or was 60Hz assumed?) adds a column to output file to save a binary variable staing if the frame rate was measured reliably or not for that experiment (i.e. was 60Hz assumed)


RebeccaHirst:
Closing this because it turns out frame estimation isn't actually implemented in psychoJS yet! https://psychopy.github.io/psychojs/core_Window.js.html#line142

peircej:
I think @sotiri is working on implementing that feature

Issue #3612, opened 2021-02-22
Look at improving TextBox2 draw performance.


isolver:
Try to improve the draw time performance of `psychopy.visual.TextBox2`, in particular when text has not changed since last flip(). Running the [demos/coder/stimuli/compare_text_timing.py](https://github.com/psychopy/psychopy/blob/release/psychopy/demos/coder/stimuli/compare_text_timing.py) demo gives these results on windows 10, psychopy v2021.1.0, nvidia 1060:

```
-------Text Draw Duration Test---------

+ Draw Order: ['TextBox', 'TextBox2', 'TextStim']
+ Text Stim Char Length:         160

+ TextBox INIT Dur (sec):       1.3761405000113882
+ TextBox2 INIT Dur (sec):      1.4610392000176944
+ TextStim INIT Dur (sec):      0.22892449999926612

+ Text Change Flip Perc:        20.00
+ Total Flip Count:              600
+ Test Duration (secs):         4.029
+ FPS:                          148.938

+ Average Draw Call Durations (msec):

  Text Object           No Txt Change   Txt Change
  TextBox               0.251           1.270
  TextBox2              1.247           4.235
  TextStim              0.472           7.119
```

TextBox2 appears to be 1.5 - 2 times _faster_ than TextStim when drawing changed text.
TextBox2 appears to be 2 - 3 times _slower_ than TextStim when redrawing unchanged text.
 
Ideally TextBox2 performance would be closer to that of TextBox, which is fast but limited to using monospace fonts.

isolver:
One option might be to use the same old school approach as the original TextBox stim but generalize it to work with any font. This is possible and the performance should end up being about the same as the current TextBox. Not saying this is the way to go, just pointing out it is an option.

isolver:
This maybe of relevence. https://github.com/psychopy/psychopy/pull/3624#issuecomment-786617692. Not sure how much time of textbox2 settext() is spent doing text parsing / layout vs. rebuilding the vert arrays needs by draw() based on the layout. That might influence where it is worth redoing the actual drawing code vs. optimizing the text parsing wrapping.

hsogo:
My code consists of following 4 parts.  I inserted `time.perf_counter()` to the code to to measure which part takes time.  

[...]

PR #3624, opened 2021-02-24
 ENH: Better line wrapping for textbox2

### LINKED ISSUE #3090: Support for Japanese (and other languages) text wrapping in TextBox2
TextBox2 doesn't wrap Japanese text.  I'm trying to support Japanese text wrapping. My code is pushed to https://github.com/hsogo/psychopy/tree/feature-textwrapping Current code may be inefficient and hard to read, but it seems to work.  I uploaded screen capture to https://youtu.be/3yQYdFr3a40

Basic idea is as follows:

1. Define characters that prohibited to place at the end of line and at the beginning of line.
2. Split text into small segments in which line should not be broken.
3. Calculate width of segments.
4. Concatenate segments within max line width.

This method has potential to support various languages by changing definition of "prohibited" characters.  For example, I defined characters prohibited at the end of line are defined as follows.  This rule allows to split a word at accented alphabets, but this behavior can be changed only adding accented alphabets here.  I think it is nice to prepare methods to get/set these characters.
```
            AVOID_EOL = "".join((
                    "({[`", # opening parenthesis and quotation marks
                    "（〔［｛〈《「『【",
                    "‘“〝≪",
                    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz", # alphabets
                    "0123456789", # numbers
                    ))
```
The current wrapping method of TextBox2 might be easier to use for western languages, so I prepared `customWrapping` flag to switch wrapping method (set False to use the current method).  This is hard-coded at present, though it should be changeable.
I'd appreciate any comments or suggestions.

hsogo:
Line wrapping of non-western language is improved.

See #3090


isolver:
closes #3090

hsogo:
Sorry for repeating commits.  Warnings are resolved.
By the way, without thinking deeply, I pulled the latest "release" branch before sending PR and sent a PR to "dev" branch.  Should I have pulled from "dev" branch?

isolver:
For this specific ENH, could you create the feature branch from Release
and then make the PR back to Release? This is localized enough change we
are treating it as a bug fix. ;) This way we can pull back into a release build
without worrying about other non production ready commits to dev getting
mixed in.

Normally an ENH should be done in a branch based on Dev and submitted as a
PR back to Dev.  A bug fix should be done in a branch based on Release and
submitted back to Release as a PR. 

That is my understanding of the new release / dev branch approach anyhow.

Thank you very much.

isolver:
Travis is probably failing because the text layout is slightly different than the old alg so the old and new screen grabs do not match up. Once the PR has been submitted to Release we can fix this up too.

Thanks again for your efforts.

hsogo:
Thanks @isolver,
I created this branch from previous "main" branch and never pulled "dev" branch.
So I'll close this PR and submit to "release" branch.

peircej:
@hsogo sorry, I'm going to disagree with @isolver on this one. There are a lot of lines changed here (including over 200 lines of TextBox2) and I'm *not* comfortable putting all those in as a quick bug fix in a patch release. I'm afraid I'd prefer we wait until the June release if that's OK, which means the dev branch is appropriate

Did you add the option to choose which layout engine is being used, or is this set of code changing all textBox2 objects to use the linebreak module?

isolver:
Sorry, @hsogo, I must have misunderstood my conversion with @peircej yesterday and got my branches crossed. :|

My understanding was that this would replace the existing textbox2 layout code since it works as fast as the old code, but is more generalizable to non western languages

peircej:
My understanding is that this does slow down layout time substantially, but the fact that involves many changed lines is a reason not to include it now. The decision around using release branch (include in next bugfix release) is basically:
- does this knowingly change behaviour of existing classes?
- is the risk assessment that this might reasonably introduce a new bug in existing code?

Bug-fix changes need to be small enough that we can reasonably confidently answer "no" to both questions

isolver:
Reopenning this PR to Dev. ;)

hsogo:
Thank you for comments.

Speed of  new code is much improved after replacing sqlite3 DB with dict object.  Layout time is about 0.9-1.2 times compared to the current algorism on my Windows10 and Ubuntu machines (python 3.6).

I haven't add an option for choosing layout engine yet because I had thought that it came up that the new code replace current textbox2 code in #3090.  I'm not against adding it.

hsogo:
`lineBreaking` option is added to select layout engine.  Specifying `'default'`, the current (2021.1.0) code is used.  Specifying `'uax14'`, new code is used.

I measured time to update text (an English text of 109 words, 623 characters) by the current and new code.  The new code is slower than the current code, but practically it would be usable at 60Hz, I think.
time

peircej:
Thanks @hsogo and sorry for the confusion. I had missed the piece where you showed it was now nearly as fast (using the dict method) and Sol corrected me on that. I still think it's right to hold this back for the next major release rather than including now, simply because of the number of lines added, and that gives us time to make a decision before that about whether it's worth having the extra argument for choosing which method to use (now I see the data I'd be inclined to say everything uses the new method for simplicity). So we'll pull this into dev in due course, once the we've created new test images for the test suite

Great job!

isolver:
I did some tests using an updated version of [compare_text_timing.py](https://github.com/isolver/psychopy/blob/hotfix-textbox2-linewrap/psychopy/demos/coder/stimuli/compare_text_timing.py) script. The branch it is in also has the changes @hsogo made.

This script tests the time to redraw unchanged text or time to settext and redraw the changed text. Ran the test:
 - using 600 letters (about 1/3 of a full screen @ 1920x1080)
 - randomly split into 'words' of 1 , 3, 5, 7, or 9 characters. 
 - once with TextBox2 set to `lineBreaking='default'` and then again with  `lineBreaking='aux14'`. 

Results (i7, nvidia1060):

#### 600 Letter Average [setText]+Draw Durations (msec)

 |  Text Object          |  Draw(unchanged)  |  setText()  |  Draw(changed)  |  setText()+draw()  |
 |:-------------------:|:---------------------:|:----------:|:-------------------:|:-------------------:|
 | TextBox                | 0.927            | 2.779            | 1.460           | 4.239 | 
 | TextBox2(default) | 1.562           | 6.371            | 1.794            | 8.165 | 
 | TextBox2(uax14)  | 1.586            | 10.919          | 1.864           | 12.783 |
 | TextStim               | 0.632            | 19.248          | 0.986           | 20.234 | 

So it looks like:

1. the latest uax14 mplementation is about 30% slower then the 'default' layout alg in textbox2.
2. the main difference in timing bewteen textbox and textbox2 is in the settext() time. @peircej, this might explain why textbox2 originally was closer in speed to textbox, but is now much slower when text is updated. Not sure how much time of textbox2 settext() is spent doing text parsing / layout vs. rebuilding the vert arrays needs by draw() based on the layout. That might influence if it is worth redoing the actual drawing code vs. optimizing the text parsing wrapping.

Therefore, we might not need to speed up the drawing part of textbox2 at all, we may just need to speed up the text parsing and wrapping part.

Based on this:

1. Glad we ended up bringing this into Dev. ;)
2. We should keep the 'lineBreaking' option for now.
3. We should probably revisit #3612.
4. We should open a new issue to look at optimizing the text parsing / layout. We may want to look at a c based implementation.

isolver:
It is interesting how the relative timing of the different textstim changes a bit for 80 character (1 - 2 line) strings:

#### 80 Letter Average [setText]+Draw Durations (msec)

 |  Text Object          |  Draw(unchanged)  |  setText()  |  Draw(changed)  |  setText()+draw()  |
 |:-------------------:|:---------------------:|:----------:|:-------------------:|:-------------------:|
 | TextBox                | 0.219           | 0.459             | 0.426           | 0.885 | 
 | TextBox2(default) |  1.270          | 0.971            | 1.442           | 2.413 | 
 | TextBox2(uax14)   | 1.295           | 1.678            | 1.487           | 3.164|
 | TextStim               |  0.457          | 3.697           | 0.579           | 4.276 | 

The original textbox drawing method really shines in this case.

PR #3631, opened 2021-02-24
BF: Remove redundant form colors

TEParsons:


peircej:
I wonder. Is this related to those warning messages that Sol was seeing about the fields not existing? Is this change going to cause a flood of those, because older versions of PsychoPy did actually have these fields and made no use of them?

PR #3638, opened 2021-02-25
ENH: Better line wrapping for textbox2

### LINKED ISSUE #3090: Support for Japanese (and other languages) text wrapping in TextBox2

hsogo:
Line wrapping of non-western language is improved.

Closes #3090


isolver:
closing, reopenning PR to DEV

PR #3640, opened 2021-02-25
BF: Compile Slider styles as list in JS

TEParsons:


peircej:
closing in favour of #3644
That one does the steps here but also
- combines tweaks with styles
- converts to the JS format ('radio'->'RADIO") etc

PR #3647, opened 2021-02-26
ENH: Simplify window names

TEParsons:
Users on the forum often list their version as just "PsychoPy3" - removing this from window titles removes an opportunity for confusion. Windows were also losing the version number when a file was open.

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -93,7 +93,7 @@ class MenuFrame(wx.Frame, themes.ThemeMixin):
     """A simple empty frame with a menubar, should be last frame closed on mac
     """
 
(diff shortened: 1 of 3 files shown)


peircej:
@hsogo for one day, would you like any part of the translated? I could imagine windows with something like
```
builderTitle = f"PsychoPy {_translate("Builder")} v{version}"
coderTitle = f"PsychoPy {_translate("Coder")} v{version}"
runnerTitle = f"PsychoPy {_translate("Runner")} v{version}"
```

hsogo:
I think that "Builder", "Coder" and "Runner" are the name of the applications and therefore need not be translated.  It's same as that application names such as Word, Chrome, Blender are not translated. :wink:

PR #3653, opened 2021-03-01
BF: Textbox startText should always be a string

TEParsons:
Was affecting TextBox when Set Each Repeat in Builder, as it was treating startText as None and throwing an error when trying to do string operations on it


peircej:
Should it be the same as 2 lines lower:
```
self.startText = text if text is not None else ""
```
so that we don't get "None" as start text?

PR #3656, opened 2021-03-02
BF: Mouse clickables no longer need encasing in [] 

### LINKED ISSUE #3651: AttributeError: 'list' object has no attribute 'contains'
I know, that this error was discussed and closed already, but I can't solve it anyway. (@peircej)

> This error message came up during the workshop when a participant running PsychoPy 2021.1 tried to create a clickable button without using the button component (the old `method). (@mdcutone)

That's my error message:

`Traceback (most recent call last):
  File "C:\Users\Admin\Desktop\posner\test_lastrun.py", line 157, in <module>
    if obj.contains(mouse):
AttributeError: 'list' object has no attribute 'contains'
##### Experiment ended. #####`

In my case there is only one picture ("image") and one click ("mouse"), but the code still crushes right after first click at any part of  screen. I can't code at all and I dont know english good enough to search through pythons' forums so I'm asking for help. If it's possible, describe anything in easy words.

Thanks a lot.

TEParsons:
as the conversion to list is already handled by the param itself.

Fixes #3651

### DIFF
--- a/psychopy/experiment/components/mouse/__init__.py
+++ b/psychopy/experiment/components/mouse/__init__.py
@@ -134,7 +134,7 @@ class MouseComponent(BaseComponent):
         code = (
             "# check if the mouse was inside our 'clickable' objects\n"
             "gotValidClick = False\n"
-            "for obj in [%(clickable)s]:\n"
+            "for obj in %(clickable)s:\n"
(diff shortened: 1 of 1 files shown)


m-macaskill:
Thanks for the quick response on this, Todd.

PR #3662, opened 2021-03-04
NF: Textbox right-to-left

### LINKED ISSUE #3648: LTR/RTL do nothing in Textbox
These haven't been implemented yet but are exposed in Builder

TEParsons:
Adds right-to-left option for Textbox formatting, as well as "auto" - which will figure out whether to use RTL or LTR based on the characters present (e.g. if any character is in the Hebrew unicode block, it will assume RTL)


TEParsons:
This is failing as a test needs updating - so I'll wait until #3663 is in as that touches the same test

m-macaskill:
Hi Todd, has taken me a while to get my local developer installation working again, so first time I've had a chance to test this. To my non-Arabic-reading eye, this looks like a great improvement upon `TextStim`, in that you have now provided for multi-line right-to-left text that correctly flows from top to bottom, and is neatly justified against the right-hand edge. As-is, this will solve a lot of display issues for many people using right-to-left languages.

However:
- it doesn't seem to be incorporating the bi-directional algorithm from the `bidi` module. e.g. if in the middle of a string of right-to-left text, one types Latin text like `1234` or `Todd`, those sub-strings are currently being displayed as `4321` or `ddoT`. i.e. displaying right-to-left text isn't as simple as displaying the stream of glyphs in reverse order - we need to cater for the very common case of a mixing of RTL and LTR substrings within an overall piece of text (e.g. in most RTL languages, numbers are read LTR). This is what the `bidi` algorithm provides - it can intelligently parse the string and be able to appropriately deal with sub-strings that run in the opposite direction to the main body. This is why it is dangerous to take the option of automatically inferring the direction of a block of text just because it contains a character from a RTL language. e.g. if a string is `'What is this character: א?'`, then the user probably intends that to be displayed LTR. If it is `'What does this word mean: قطة?'`, then it should be LTR with a bidirectional RTL component.
- we still need to incorporate the reshaping of Arabic text, so that characters are joined up to each other, rather than being presented in their isolated form. This is handled by the `arabic-reshaper` package.
- The order of applying the `bidi` and `arabic-reshaper` functions matters - probably best to consult the current `TextStim` code as a guide for that:
https://github.com/psychopy/psychopy/blob/2a5c02efeff910973cdb915528ad68fac3e485f0/psychopy/visual/text.py#L349
-  in `TextBox2`, though, we would need to reverse the output of the `bidi` `get_display()` function, because otherwise the overall string would be displayed backwards, as you are already taking care of drawing them in order from right to left.  

(Sorry if this is too much hand-holding for what is a feature in development, but I have to re-discover all of this stuff each time I encounter it, so am leaving these notes here for myself as much as anything).

Example Farsi string that includes some Latin text and numbers at the end:
در نسخه آخر سایکوپای، مشکل تبدیل متون انگلیسی به فارسی حل شد و از این پس شما قادر خواهید بود که در آزمایشات خود از متون فارسی استفاده کنید: TextStim 1234.

As a guide, below is output from `TextStim` (orange): this is correctly bi-directional (the Latin substring runs left-to-right) and the Arabic characters are re-shaped, but 
- the lines progress from bottom-to-top, and 
- are not right-aligned. 

Note that the `bidi` algorithm is pretty clever. It knows that the Arabic and Latin characters are strongly directional, but the punctuation adopts the dominant RTL style. So in the logical source string, the full stop neighbours the `4`, whereas in the display string, it neighbours the `T`: 

Screen Shot 2021-03-11 at 20 17 51

Output from `TextBox2` (white): lines progress correctly from  top-to-bottom, and the Arabic characters run right-to-left and are right-aligned, but
 - characters are not reshaped into their joined-up form, and 
 - the string is not bi-directional (the Latin text is shown in reverse, and the punctuation characters are not dealt with any differently either).

We need to aim to incorporate the good bits of `TextStim` (basically the `bidi` and `arabic-reshaper` output) into `TextBox2`, which has the proper overall layout.

Lastly, I've never actually tested the performance impact of the bi-directional algorithm - if it comes with a penalty, we may want to make it optional (as it is not strictly necessary for display of solely RTL text, as you are handling that yourself currently). And we don't have an option to use the bi-directional algorithm on predominantly left-to-right text, which could certainly be useful in linguistic tasks. So while revising `TextBox2`, perhaps it is time to revise all the language styles to be comprehensive and independent. In that case, they would be:

- Left-to-right
- Right-to-left
- Arabic
- Left-to-right (bi-directional).
- Right-to-left (bi-directional).
- Arabic (bi-directional)

If it doesn't come with much of a performance penalty, then we could just have the three language styles, and apply the bi-directional algorithm to all of them by default.

isolver (approved):

TEParsons:
@m-macaskill Sorry for the delay, I've finally had chance to look into `bidi`! It seems really good, the problem with editable textboxes vs static textstim is that `bidi` works by changing the character order - so running it every new character would be running the algorithm on already processed text. This is why I went the route of changing the glyph vertices rather than the character order - if the characters are direction agnostic (first-to-last rather than left-to-right or right-to-left), then we can just recalculate the vertices each draw instead. 

I think the solution is to store a direction agnostic representation, but have it be purely internal and separate from the text that's rendered. That way we can apply the `bidi` algorithm on this representation on render, without changing the order of the representation for the next keypress.

If this does have a time penalty (@isolver has tests for this) then we could maybe implement basic LTR, RTL and Arabic using vertices adjustment as in this PR and then have the default setting be `auto` which will use `bidi` to figure it out. So for most people who aren't too worried about precise timing they don't have to think about language direction, but for those who do care about timing they can manually set it for better performance. Which should cover most use cases, until someone comes along with an experiment comparing sub-millisecond timing on a typing task where users write in a mix of Hebrew, English and Arabic...

m-macaskill:
@TEParsons Hi Todd. I'm mainly approaching this from user requests about static, rather than editable text. I think we've hit a brick wall with the old `TextStim`, as we can't show multi-line RTL text there properly unless we were to make changes to `pyglet` itself.

You're right about `bidi` and the order of the string - unlike with `TextStim`, with `TextBox2`, we don't need to reverse the overall logical order of the characters, as you are taking care of drawing them in the correct order already. I think if we are to use `bidi` with `TextBox2`, then we would need to run it on the given string, and then immediately reverse the output. i.e. the logical order of the string would be the same as the source string, _except_ for any embedded substrings that should run in the opposite direction.

e.g. using `bidi`'s test convention of capitals representing RTL characters, then processing a string like this in the order it is typed:

    get_display(u'THE CAT IS brown.', upper_is_rtl=True)

would yield `'.brown SI TAC EHT'`. We would then immediately reverse that to `'THE CAT IS nworb.'`, which would be the new logical order for drawing the string in order from right to left. This is purely the display order for the string: the original typed response is the true logical order of the string and probably what needs to be maintained as a representation (e.g. for storing the response in the data file).

But bi-directional text is an edge case. The key thing you've already solved is correct display of multi-line right-to-left text, which `TextStim` can't do.. 

So the other thing we really need to add (and more important than bi-directional text I'd say), is the reshaping of Arabic text. At the moment, `TextBox2` should work for RTL languages like Hebrew, but is still broken for  Arabic/Farsi/Urdu, etc, which need that extra step of reshaping the characters to take account of their position within a word . Fortunately, `arabic-reshaper` takes care of that for us (although I think we still might have issues with harakat, the small optional vowel diacritic marks).

Again, there could be performance impacts here, as the reshaping does need to be applied as each character is typed (i.e. the same character looks different depending on whether it is at the start, middle or end of a word, and that can alter as new characters are typed). But again, what most users have been concerned about is the simple display of static, multi-line text, where the algorithm should only need to be applied once.

PR #3671, opened 2021-03-08
BF: dialog message is not translated

hsogo:
Reminder dialog when creating a new conditions file was not translated.
Don't use formatted string literal inside _translate() (Pleas see #3367).


hsogo:
By the way, while the parameter name is 'Conditions' on the parent dialog, the reminder dialog calls it 'conditionsFile'. It would be better to match them, I think.

PR #3673, opened 2021-03-09
RF: Refactoring the components panel

TEParsons:
Refactor of the components panel to:
- Make use of wx features (like WrapSizer) rather than doing things manually
- Index components in a way which makes it much easier to dynamically show/hide them
- Use general function stringtools.prettyname for labels rather than doing it manually
- Generally be more stable and readable

Appearance and function are largely unaffected, just three things different:
1. ~~Button labels aren't wrapping yet, there's nothing about the refactor stopping this I just need to reimplement it still~~ Done this now!
2. ~~Tooltips need moving to class attributes rather than module attributes, again nothing preventing this I just haven't done it yet, I've used docstrings as a placeholder~~ This too!
3. ~~I've replaced the panel buttons with static labels, buttons are still in sizers accorded to category so I can add them back in but they were kind of buggy so didn't want to spend time implementing them when labels may be better anyway.~~ Now reinstated, but without arrows.

Pulling to dev so this is non-urgent, just wanted to get this in at some point as I'm basing a standaloneRoutines branch off of this refactor to avoid conflicts.


TEParsons:
Update: I've reinstated category buttons, but I couldn't include the ᐊ and ᐁ from before. These are unique to plate buttons, but the way we were using plate buttons was kind of hacky - they're intended for context menus, not showing/hiding sizers within a panel. They're also an absolute nightmare to style, you have to find the protected class defining their init colours and change these then call updates on them... Whereas regular buttons can just use SetForegroundColour/SetBackgroundColour.

peircej:
With your changes to the Component structure, I'm guessing that means people's custom Components will break (because they will have their icons etc in the old location?)

TEParsons:
> With your changes to the Component structure, I'm guessing that means people's custom Components will break (because they will have their icons etc in the old location?)

Most recent commit addresses this - they'll need to haveiconFile, targets, tooltip and categories in the class def rather than module def for it to use these rather than the BaseComponent and BaseVisualComponent defaults, but it now won't crash :) Some may have already - before it wasn't consistent what was where

PR #3684, opened 2021-03-16
BF: Recognise if list param is a variable name

TEParsons:


TEParsons:
@m-macaskill this fixes the problem this user was having: https://discourse.psychopy.org/t/variable-name-in-allowed-keys-field-of-keyboard-component-causing-error/21332

peircej:
Note that the call to `valid_var_re` doesn't check if this *is* a variable name, only that it has valid syntax to be one. Is there a possible case where name of a key is actually intended but happens to be valid syntax for a variable name so gets the wrong assignment here, or is that only an issue in my imagination? 🤔

PR #3685, opened 2021-03-17
WIP: iohub updates

isolver:
RF: Started removing legacy sections of iohub code such as expRuntime class, ability of iohub to run as package outside of psychopy, and the util.clock.py and util.fix_encoding.py files
RF: Some cleaning up of eye tracker config files with an eye toward builder integration.
RF: __all__ iohub demos now use launchHubServer()
ADD: iohub.util.getDeviceNames(): return list of iohub eye tracker names (that launchHubServer takes as keys)
ADD: iohub.util.getDeviceDefaultConfig(): return the eye tracker device config with default values, filtered to show builder visible settings only.
ADD:  iohub.util.getDeviceParams(): returns eye tracker config settings similar to getDeviceDefaultConfig, but the value for each setting is a param dict containing valType, inputType, hint, etc...
ENH: eye tracker config file now supports builder_hides setting, which is used to list the device settings that should not be visible in Builder
ENH: iohub device setting can be static, for example manufacturer_name, so that it should be a disabled setting in builder, but still visible. 
ENH: gevent tasks scheduling resolution on windows supports sub msec again.

This has been retested with eyelink and gazepoint, as well as keyboard and mouse, so please pull in if possible.


isolver:
@mdcutone , can you please take a look at this and pull it in if it looks OK to you? I've tested it with the iohub coder demos and nothing is broken, so I think it is good to go into DEV.

Issue #3688, opened 2021-03-18
BF: sound files in a loop always get truncated to the duration of the first sound


yh-luo:
As title. This is an old bug that's still present in v2021.1.1, recently reported to me by my colleagues. Many forum posts were related to this bug, like [this](https://discourse.psychopy.org/t/audio-sample-cut-off/6885) and [this](https://discourse.psychopy.org/t/sound-ends-stops-before-it-should-audio-is-cut-short-in-builder-and-online/19416). The bug exists for both the sounddevice and the PTB backends.

The problem lies in that when call `setSound()` on a file, `_setSndFromFile` check if `self.stopTime > 0` and update `self.duration` according to it. If the sound instance is already initialized, `self.stopTime` is set to the duration of the sound. When `setSound()` is called for the second time (noted that this is the behavior in the builder when looping), the duration is always the duration of the first read-in sound file. See the example below for this problem.

**Example**
```python
In [1]: from psychopy import prefs
   ...: prefs.hardware['audioLib'] = ['sounddevice']
   ...: from psychopy import sound
   ...: import soundfile as sf
   ...: 
   ...: fname_1 = '1-1.wav'
   ...: fname_2 = '2-1.wav'
   ...: sound1 = sound.Sound(fname_1)
   ...: sound2 = sound.Sound(fname_2)
   ...: 

In [2]: # get durations
   ...: sound1.getDuration()
Out[2]: 8.405351473922902

In [3]: sound2.getDuration()
Out[3]: 10.048004535147392

In [5]: # setSound
   ...: sound1.setSound(fname_2)
   ...: sound1.getDuration()
   ...: 
Out[5]: 8.405351473922902  # should be 10.048004535147392!
```

I've proposed a simple fix #3689 that keeps the user-specified `stopTime` to fix the problem. It works for both backends.

**Tested on**

`sounddevice`

+ Linux (Fedora 33)
+ Windows 10

`PTB`
+ Windows 10 

I don't have a Mac to test, it would be nice if someone can test that!

pn2200:
I tested this change on a Mac with `sounddevice` backend and it appears to have resolved issues we were having with different sound files in a loop getting cut off or truncated.

peircej:
Nice one. Many thanks for tracking this down!
We'll have it in the 2021.1.4 release (target date not decided yet)

PR #3689, opened 2021-03-18
BF: fixed the bug that setSound() incorrectly truncate the sound file 

yh-luo:
Details in #3688 

* fix stopTime in sounddevice backend

* fix stopTime in ptb backend


mdcutone:
I wonder if this relates to another problem we've had. Cool if this fixes things.

PR #3694, opened 2021-03-20
Update setup.cfg so  psychtoolbox == 3.0.16

isolver:
Latest release of psychtoolbox causes a wrong argument type error. Using 3.0.16 works for now.


isolver:
merging to try and help self install problems for now.

Issue #3701, opened 2021-03-29
Indexing error in `data.TrialHandlerExt`


tomwallis:
I'm trying to use the `data.TrialHandlerExt` class to create unequally-balanced presentations of conditions. I receive an `IndexError` with the following basic configuration. Perhaps I am using this incorrectly?

```python
from psychopy import data

trials = data.TrialHandlerExt(
    [{"Name": "A", "weight": 2}, {"Name": "B", "weight": 1}], nReps=1
)

for trial in trials:
    print(trial)
```

Yields:

```shell
Traceback (most recent call last):
  File "wahrnehmung_seminar/tmp.py", line 9, in <module>
    for trial in trials:
  File "/Users/tsawallis/opt/miniconda3/envs/wahrnehmung_seminar/lib/python3.6/site-packages/psychopy/data/trial.py", line 1512, in __next__
    position=self.getNextTrialPosInDataHandler())
  File "/Users/tsawallis/opt/miniconda3/envs/wahrnehmung_seminar/lib/python3.6/site-packages/psychopy/data/base.py", line 556, in add
    self[thisType][position[0], int(position[1])] = value
  File "/Users/tsawallis/opt/miniconda3/envs/wahrnehmung_seminar/lib/python3.6/site-packages/numpy/ma/core.py", line 3386, in __setitem__
    _data[indx] = dval
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices

```

I'm running psychopy (`psychopy 2021.1.3 py36h79c6626_0 conda-forge`) in a python 3.6 conda environment.

Tagging @suddhasourav in case he's still about, since I believe he wrote the class.

Thanks!

suddhasourav:
Hi there @tomwallis!
I am sorry for the index error you encountered - you are not using the class incorrectly. I found the cause of the error: When I wrote the _TrialHandlerExt_ class for Psychopy 2 with the older numpy, numpy arrays could be indexed with floats. Example with my Psychopy v1.83.04 (win32):

```python
Python 2.7.3 (default, Apr 10 2012, 23:31:26) [MSC v.1500 32 bit (Intel)] on win32
import numpy as np
x = np.array([0,1,2,3])
x[2]
Out[4]: 2
x[2.0]
Out[5]: 2
x[2.9]
Out[6]: 2
```

[...]

PR #3705, opened 2021-03-29
BF: TrialHandlerExt indexing error

suddhasourav:
[Fixes issue #3701.](https://github.com/psychopy/psychopy/issues/3701)

Cause of the bug: TrialHandlerExt class was written with an older numpy version in which arrays could be indexed with floats. This operation now throws errors, as floats (e.g. 2.0) are not valid indices anymore.

Solution: np.sum() replaced with int(round(np.sum()))


peircej:
Thanks Suddha. By the way, you maybe know this already but, mentioning the the GH tag in the pull request comment creates a cross reference but it doesn't get an issue closed (I think). I think, to auto-close the issue you need to say `fixes GH-commit` in one of the commit messages

Issue #3672, opened 2021-03-08
Stop time of last component in routine is incorrect


pn2200:
I have an experiment displaying a ShapeStim for one second, then a TextStim for 4 seconds, in a loop five times. The output for the start and stop times of the TextStim is incorrect: the `.stopped` field has value `None` for some of the entries (sometimes 2 entries, sometimes 3), and the remainder are values before the `.started` field. When `TextStim.stopped` does have a value, then it is equal to `ShapeStim.started`.

The TextStim's `.started` field is before the end of the ShapeStim's `.stopped` field; it looks like the TextStim starts one or two frames before the ShapeStim ends, so they overlap visually when the experiment is run. In this case, it might be that the time is accurately recorded, but it does not reflect the desired experiment flow where the ShapeStim is displayed, followed by the TextStim.

I have attached the experiment file, the conditions file, and a sample output file. They were created using PsychoPy 2021.1.2.

[overlap.zip](https://github.com/psychopy/psychopy/files/6104399/overlap.zip)

pn2200:
I think this issue affects more components than just `TextStim`. I think that the stop times of the last component in any routine is not recorded correctly.

In this specific experiment, this issue occurs because the `routineTimer` becomes negative (`routineTimer.getTime()` returns a value less than 0) while the TextStim has a status of STARTED and `tThisFlipGlobal` is less than the TextStim's start time plus duration (`tThisFlipGlobal > text.tStartRefresh + 4.0-frameTolerance` is false), so code that stores the stop time for the component is never executed.

I don't know the right way to resolve this issue. Adding an else-statement on the while loop like the following seems to work, so I will create a pull request based on this:  

```
while continueRoutine and routineTimer.getTime() > 0:
    # Run loop
else:
    for thisComponent in trialComponents:
        if hasattr(thisComponent, 'status') and thisComponent.status == STARTED:
            # Stop component and record stopping time
            thisComponent.tStop = t
            thisComponent.frameNStop = frameN
            win.timeOnFlip(thisComponent, 'tStopRefresh')
    win.flip()
```

pn2200:
I also think that https://github.com/psychopy/psychopy/issues/2916 is a special case of this more general problem, so I think that the fix for that issue should be revisited.

peircej:
[...]

PR #3708, opened 2021-03-30
BF: For non-slip components, store the stop time when while loop ends

pn2200:
Related to issue #3672 .


peircej:
By adding a flip() call I think you're forcibly inserting a blank screen into the end of the routine (and a 16.7ms delay before next stimulus)

PR #3711, opened 2021-04-01
BF: Specify dtype when applying asarray to num/int params

TEParsons:
Fixes this user's problem:
https://discourse.psychopy.org/t/builder-inserts-useless-asarray-calls-in-py-code/21901/8


peircej:
I don't believe this fixes that user's problem. His issue is that numpy.asarray() for a float gives an ndarray with undefined length:

```
>>> import np
>>> len(asarray(3.0))
TypeError: len() of unsized object
```

I think we shouldn't be inserting asarray into the boilerplate code at all but I don't recall why it was being used before (or why it's only recently turned into a problem)

PR #3712, opened 2021-04-01
BF: Swap the Cancel and No buttons on message dlgs 

TEParsons:
for consistency with other dlgs


peircej:
I think this may be a platform-specific thing: Mac is one way and windows the other: https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/
but I agree we should be internally consistent

PR #3717, opened 2021-04-07
BF: Use github versions for update ctrl

TEParsons:


peircej:
mmm, I'm not sure about this. 2 things the file provides:
- control for us on when we push people to update (so we can choose NOT to encourage a x.x.0 release)
- the file includes information about what versions are "updatable" (meaning the dependencies haven't changed and just the psychopy.zip will suffice) and which are compatible (will updating cause a known breakage)

I'm aware we haven't updated that file recently (but I don't much mind) and we'll do it with upcoming 2021.1.3

I'm closing for now, pending being convinced otherwise ;-)

PR #3723, opened 2021-04-09
ENH: More consistent Classic theme icons

TEParsons:


TEParsons:
For things added since 2020.2 (Textbox, Form, Button, etc.) I'd just been using the dark theme icons (so they stand out on either background) but now I've downloaded the original icon database and picked out icons from there instead.

PR #3724, opened 2021-04-09
EN/FF: Various changes to the Microphone class

mdcutone:
Some fixes and changes as listed here:

* Removed `volume` setting from `Microphone` class. Doesn't work anyways but didn't raise errors originally.
* Added audio latency mode. Default value comes from prefs but the user can specify a value to override.
* Added logging to various events during the creation and use of the class.
* Some doc changes.


isolver:
In the warning:
`Text-to-speech recognition module not available` it might be good to give the name of the python package that needs to be installed, or tell them to `pip install SpeechRecognition`

isolver:
the `coder\input\latencyFromTone.py` demo no longer runs. 

```
  File "D:/DEV/my-code/psychopy/psychopy/demos/coder/input/latencyFromTone.py", line 38, in <module>
    names, idx = sound.backend.get_input_devices()
AttributeError: module 'psychopy.sound.backend_ptb' has no attribute 'get_input_devices'
```

Not sure if this dmeo should be fixed or changed to be a demo / test of the new recording functionality.

isolver:
Update: Using Audacity I can record audio at normal volume fine, so it seems that the latest mic update records with very low input levels. Maybe the volume code was doing something?

TEParsons:
Mine is a little quiet too - quieter than when recording from current dev branch

mdcutone:
Makes no sense why this would be happening. The volume code wasn't there before and it worked fine.

mdcutone:
Okay seemed to be a problem with the WASAPI driver not working well with audio latency classes other than zero.

PR #3726, opened 2021-04-12
ENH: Allow #/♭ characters in dict names for sharp/flat notes

TEParsons:


peircej:
Have you not introduced an indentation error here?

PR #3743, opened 2021-04-21
BF: allow granularity for slider in Form

RebeccaHirst:
granularity was not enabled for slider in forms, meaning responses were always on a continuous scale. Added optional column to the form template to enable granularity, if no granularity is provided (or there is no column header "granularity") use the default value of 0


peircej:
Is there documentation to go with this or was that already suggested and no matching feature?! 😉

PR #3749, opened 2021-04-22
NF: Five-colour model for Form component

TEParsons:


peircej:
I guess this is one we need to check with @thewhodidthis and @apitiot 
Guys, the idea is to move to a slightly richer color spec for Forms (without having tons of color params):
- add attributes to the Form Component for:
  -  fillColor (the fill of the Form)
  - borderColor (just the outer border of the form)
  - itemColor (for the item/question)
  - responseColor (all components of the response, e.g. slider and its labels)
  - markerColor (just for slider response objects)

itemColor, responseColor and MarkerColor(?) can be overridden by settings in the spreadsheet for the Form (ie. on a per-item level)

Is this going to be doable for psychoJS (by feature-freeze)?

thewhodidthis:
@peircej  Yes! Please is there a test .psyexp for this @tpronk @TEParsons?

TEParsons:
@thewhodidthis Just updated the BFI demo to use the new colours & pushed to this branch so you can use that to test :) Old forms also still open - "style" will override any colour settings with a pre-set style, and as style existed before it means that old experiments will look the same.

thewhodidthis:
@TEParsons Awesome thanks, x

PR #3766, opened 2021-05-01
ENH: Movie playback improvements

mdcutone:
This PR introduces some changes to MovieStim3 and VlcMovieStim.

* Pixel buffers are now used to contain video samples, allowing for video image data to be copied/streamed directly to the texture. This results in less GPU/CPU overhead than before and reduced studdering.
* Added `interpolate` property. Removed redundant texture parameter setters, now only updating when the value is set.
* Cleaned up documentation a bit.
* Reworked the VLC movie class substantially to greatly improve performance and reliability.


isolver:
Testing of MovieStim3 and VlcMovieStim 
----------------------------------------------

Used coder/stimuli/MovieStim.py and coder/stimuli/VlcMovieStim.py

Have not yet dug into why first frame is black in linux test as Jon reported. 

Windows 10
--------------

__MovieStim.py__  (uses MovieStim3)

- Video plays however last 0.5 seconds of video (maybe audio only?) repeats once. Playing original video using movie player does not show the issue.

__VlcMovieStim.py__
 
- Video usually plays fine, no repeating at end like moviestim3.
- did get one crash, but not sure how to reproduce: 
```
  File "_ctypes/callbacks.c", line 234, in 'calling callback function'
  File "D:\DEV\my-code\psychopy\psychopy\visual\vlcmoviestim.py", line 1190, in vlcLockCallback
    user_data, ctypes.POINTER(ctypes.py_object)).contents.value
ValueError: PyObject is NULL
```
- If no size is specified when creating video class, sometimes exp freezes and sometimes no video frames are displayed during playback, but audio plays. Above error may have also occurred when testing without a size specified.

macOS (latest)
----------------

__MovieStim.py__ (uses MovieStim3)

- Video plays however last 0.5 seconds of video (maybe audio only?) repeats once. Playing original video using movie player does not show the issue.

__VlcMovieStim.py__
 
- Video plays fine, no repeating at end like moviestim3.
- If no size is specified when creating video class, no video frames are displayed during playback, but audio plays.

Linux
------

Visually, same results as macOS:
 - MovieStim3 repeats last fration of second of video (maybe just audio part, hard to tell without diving in more). 
 - VlcMovieStim.py plays the video correctly, but has the same behaviour as macOS if the stim size is not provided (no video, plays audio). 

Also, since I did not have python-vlc installed when I first tries testing on Linux, this error was generated:

```
  File "/home/sol/Documents/GitHub/psychopy/psychopy/visual/vlcmoviestim.py", line 37, in <module>
    import vlc
ModuleNotFoundError: No module named 'vlc'

During handling of the above exception, another exception occurred:

  File "VlcMovieStim.py", line 41, in <module>
    mov = visual.VlcMovieStim(win, videopath,
  File "/home/sol/Documents/GitHub/psychopy/psychopy/contrib/lazy_import.py", line 119, in __call__
    obj = object.__getattribute__(self, '_resolve')()
  File "/home/sol/Documents/GitHub/psychopy/psychopy/contrib/lazy_import.py", line 88, in _resolve
    obj = factory(self, scope, name)
  File "/home/sol/Documents/GitHub/psychopy/psychopy/contrib/lazy_import.py", line 203, in _import
    module = __import__(module_python_path, scope, scope, [member], level=0)
  File "/home/sol/Documents/GitHub/psychopy/psychopy/visual/vlcmoviestim.py", line 39, in <module>
    if "wrong architecture" in err:
TypeError: argument of type 'ModuleNotFoundError' is not iterable
```

Not a big thing, but I wonder if vlcmoviestim.py should be checking if the import fails and giving a more informative error about what needs to be installed (python-vlc).

Finally, I ran the test locally and the first frame screen capture is also black even though the video appears to play. 

movFrame1_norm_local

Seem to get a black back buffer for the first three frames. When `frameN` is set to compare frame index 3, then the test passes, i.e.:

```
        for frameN in range(10):
            mov.draw()
            if frameN==3:
                utils.compareScreenshot('movFrame1_%s.png' %self.contextName,
                                        win, crit=10)
            win.flip()

```

mdcutone:
Thanks Sol. I've  seen the NULL object bug once on my system, might need to put a check in there to catch when that happens.

mdcutone:
@isolver I fixed most of the issues around the VLC movie playback class. Not sure why playback is repeating in MovieStim3.

jfkominsky:
@mdcutone I had someone contact me about this happening with the current build of PsychoPy with MoveiStim3 (in the context of PyHab). It seems to be codec-specific. I was able to solve the problem by re-encoding their videos with h.265 (HEVC) with AAC audio from h.264 with AAC audio. It may be an underlying issue with MoviePy and MP4s, see this potentially related bug: https://github.com/Zulko/moviepy/issues/1458

PR #3767, opened 2021-05-05
NF/RF: Add some basic shapes to ShapeStim

TEParsons:
Adds rectangle, triangle & circle to list of knownShapes, and allows equilateral shapes to be specified by setting vertices to be the number of sides (using the same function as PolygonStim). This makes the boilerplate for Polygon components a bit more simple without removing any functionality.


peircej:
That's a really long commit list. Did you accidentally submit PR to release instead of dev?

PR #3786, opened 2021-05-12
ENH: Enable scoring from ticks for choice/radio responses

RebeccaHirst:
If item type is choice/radio AND ticks are provided use the ticks as the response for output column form.response otherwise use the tick Lable (allows for easy negative scoring and scoring of choice responses with no code).

Note: Default values of ticks were previously (1, 2, 3, 4, 5, 6, 7) changed to None otherwise item[‘ticks’] is always true.


RebeccaHirst:
@TEParsons want to check with you that setting the default val of ticks to None should be OK (i.e. that shouldn't cause issues with slider right because slider already has it's own default ticks?)

PR #3790, opened 2021-05-13
[BF] Fix/bold italic in TextStim

bpinsard:
Bold/Italic is broken in TextStim, here are minimal changes to fix it.


peircej:
Thanks @bpinsard 
My guess is that this crept in at one point when the pyglet code was forced to change from `pyglet.text.Text` to `pyglet.text.Label` but I don't recall the details of why italic/bold wasn't (re)implemented then

By the way, for a bug fix like this could you:
- use the BF: prefix (and use FF: or similar ENH: for feature fixes or enhancements)
- base your branch on release and submit the pull request to release (rather than dev for both)

The release branch and the BF tag indicate that this should be in the next bug-fix release (which we do relatively often). Without them it will only appear in the next major release (twice per year)

PR #3801, opened 2021-05-17
ENH: Expose "font" of Form in builder

TEParsons:


TEParsons:
@RebeccaHirst it won't let me request you as a reviewer but would definitely like your feedback, as the person who works with Forms most :)

PR #3812, opened 2021-05-20
ENH: More info for component disabled alert

### LINKED ISSUE #3733: Feature request: Report the component name and routine of disabled components in Alert 4305
PsychoPy reports [an alert](https://psychopy.org/alerts/4305.html) when a component in the experiment is disabled. However, this alert doesn't tell you what component is disabled. It can be quite difficult to find the disabled component, aside from going through and manually checking every component. It would be a useful feature if this alert also included the name and routine of the disable component.

TEParsons:


peircej:
Yes! 😁 
We should do this more. Some experiments have a ton of warnings and it isn't clear that they're all coming from different objects. Ideally, for the Runner Alerts box these could actually be hyperlinks to the dialog box for the actual component rather than merely a text name

PR #3822, opened 2021-05-21
BF: Fix standalone routine disable logic

TEParsons:
I'd missed an if so was setting all routines as disabled if any one standalone routine was...


TEParsons:
(pending test)

PR #3826, opened 2021-05-24
BF: Add labels to legacy params

TEParsons:
In reference to ClickUp task CU-k980gd


peircej:
Task linked: [CU-k980gd Rendering Textbox Component color fields in dialogs is very broken](https://app.clickup.com/t/k980gd)

PR #3843, opened 2021-05-25
TEST: Add test for textbox typing, not just setting text

TEParsons:
, to better catch this error: https://discourse.psychopy.org/t/textbox2-crashes-when-korean-input-exceed-certain-length/22717/6


TEParsons:
The BF here fixes #3892

PR #3851, opened 2021-05-27
NF: Add Builder eyetracking demo

TEParsons:
Making this has highlighted a few bugs, so the fixes are / will be in this PR


isolver (commented):
Hi Todd, thanks for putting this together. Some initial suggestions and issues:

- Add a "Press any key to start calibration" screen at start of experiment.
- Calibration "Disable component" is set to True, should be False. 
- Setting Calibration "Disable component" to False does not change generated code (so no calibration runs).
- Calibration defaults should match agreed defaults.

I can do more testing once calibration is working again.

Thank you

PR #3878, opened 2021-06-02
ENH: Set beta tags in code rather than manually adding them to icons

TEParsons:


TEParsons:
@peircej This does change appearance slightly - beta tags won't appear in the Routine canvas, only the Component panel, is this okay for the book? I can add them to the Routine canvas, just have to make a point of requesting it, but it seems kind of better to only have them in the component panel?

PR #3879, opened 2021-06-02
BF: Reload Coder file on JS compile

TEParsons:


peircej:
But when I change a JS file and then hit the JS compile button the Coder view still doesn't update. Even if you manually try to open from Coder it doesn't fetch it. You have to close the file and open it again!

PR #3880, opened 2021-06-02
ENH: Button default start time should be 0

TEParsons:


peircej:
I think you changed that deliberately so that users would accidentally record a click in the next trial from the previous. Is that now handled by only detecting new clicks (as with the mouse) now?

Issue #3901, opened 2021-06-03
Keyboard: AttributeError: 'Keyboard' object has no attribute '_buffers'


RebeccaHirst:
Using a keyboard response with default parameters results in experiment closing with the following error: 

AttributeError: 'Keyboard' object has no attribute '_buffers'

Minimal demo here: https://pavlovia.org/lpxrh6/keyboard_bug_demo 

(all issues reported on dev branch)

isolver:
fixed in PR 3910

peircej:
additional fix in https://github.com/psychopy/psychopy/pull/3922

PR #3910, opened 2021-06-04
misc fixes

isolver:
FF: Fix pyglet missing events when win32MessagePump() is called by the psychopy process.
FF: Fix print warning when display unit type is None
RF: Removed unneeded import from keyboard.Keyboard
BF: keyboard.Keyboard .start / .stop when backend != ptb
FF: keyboard.Keyboard ._buffers only being created for first instance on Keyboard


isolver:
fixes #3901

PR #3914, opened 2021-06-04
ENH: Expose sample rate of Sound in Builder

TEParsons:


mdcutone:
We need a caveat that this will only work if PTB is running in exclusive mode (3).

TEParsons:
I'm alright to hold this back until 2022 if we don't want to add anything disruptive so close to release, @peircej ?

peircej:
Yes, I think it needs more thought, and implementing at the level of the sound module not the individual sound

PR #3922, opened 2021-06-05
BF: monkey-patch broken psychtoolbox function

peircej:
Docs for KbQueueCreate suggest that win_handle defaults to 0 so we use
that as default but since PTB 3.0.17 a value of 0 causes a crash.

Full details at
https://github.com/Psychtoolbox-3/Psychtoolbox-3/issues/723

We'l remove this hack (or set it to operate only on known-bad versions)
when we hear it's fixed upstream

### DIFF
--- a/psychopy/hardware/keyboard.py
+++ b/psychopy/hardware/keyboard.py
@@ -83,6 +83,15 @@ except ImportError as err:
 
 defaultBufferSize = 10000
 
+# monkey-patch bug in PTB keyboard where winHandle=0 is documented but crashes
+if havePTB and sys.platform=='win':
+    from psychtoolbox import PsychHID
+    # make a new function where we set default win_handle to be None instead of 0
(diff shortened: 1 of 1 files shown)


peircej:
@isolver here's that monkey-patch for broken ptb 3.0.17. Let me know if that solves it for you

psychopy/hardware/keyboard.py
```diff
@@ -83,6 +83,15 @@
 
 defaultBufferSize = 10000
 
+# monkey-patch bug in PTB keyboard where winHandle=0 is documented but crashes
+if havePTB and sys.platform=='win':
```
isolver: Typo in sys.platform check, change to:

`if havePTB and sys.platform=='win32':`

and it works great.

peircej: doh! thanks

PR #3929, opened 2021-06-07
ENH: Standardise order of Okay and Cancel buttons

TEParsons:
[UX Tik Tok](https://vm.tiktok.com/ZMec7vc7e/) has spoken.

(plus code to put them the wrong way round on Windows to match Windows being weird)


peircej:
FYI @TEParsons there is also a wx class to do this without creating each button at all https://wxpython.org/Phoenix/docs/html/wx.StdDialogButtonSizer.html

PR #3934, opened 2021-06-07
FF: MouseGaze support no button movement and escaping calibration

isolver:
FF: If the MouseGaze 'move' buttons list is empty ( [] ), or set to 'CONTINUOUS', or ['CONTINUOUS',...], then mousegaze eye position is updated with any mouse motion, not mouse drags. This is also now the default behaviour for the device, so will also be when using MouseGaze with any Coder demos.

FF: Pressing escape during Mousegaze calibration aborts the calibration and returns control to the experiment.


isolver:
Now that I use it, I have to also agree not needing any buttons is a better default movement setting. ;) 
+1 @peircej

isolver:
@TEParsons, added support for using CONTINUOUS in (or as) move button list to this PR, so now either [] or 'CONTINUOUS' can be used

PR #3941, opened 2021-06-08
FF: Audio transcriber fixes

mdcutone:
Fixed a few things here causing errors running the microphone transcription demo. Unfortunately, we still have a problem where the transcription is not being done and the demo experiment is just locking up. Not sure what's going on there.

Also fixed some documentation in the Microphone class.


peircej:
Could you send the PR to dev instead of release?

mdcutone:
Shoot, that was a mistake. Will do.

PR #3943, opened 2021-06-09
BF: Add PreloadJS script tag back in index.html template

### LINKED ISSUE #3942: Need bring back PreloadJS script tag
As it happens, trying to package this library in with the rest of our code creates more problems than it solves https://github.com/psychopy/psychojs/issues/379

thewhodidthis:
@peircej Closes #3942


tpronk:
I think it might be better to fix this in a different way. Discuss at standup?

PR #3956, opened 2021-06-09
BF: Replace backslashes in params by slashes

tpronk:
This bug-fix replaces backslashes in experiment parameters by slashes. This fixes the "missing resources in JS" issues, where resource keys in the resources list used slashes, but resource keys in the timeline used backslashes. I tested it by running [wdio_img](https://github.com/psychopy/psychojs_testing/tree/main/tests/wdio_img) online and offline.

Note that I added the conversion from backslashes to slashes after the if block that checks whether a valType is "file" or "table" (at line 204). I noticed that the file parameters in wdio_img weren't recognized as type "file", but as type "str".

### DIFF
--- a/psychopy/experiment/params.py
+++ b/psychopy/experiment/params.py
@@ -205,6 +205,7 @@ class Param(object):
                         # If param is a file of any kind, escape any \
                         val = re.sub(r"\\", r"\\\\", val)
                     val=re.sub("\n", "\\n", val) # Replace line breaks with escaped line break character
+                    val=re.sub("\\\\", "/", val) # Replace any backslashes by slashes
                     return repr(val)
             return repr(self.val)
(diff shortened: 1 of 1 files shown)


peircej:
Rather than changing one way and then changing the other, let's change it correctly the first time and then add your second catch after. Note better comment too so that future me doesn't think that looks silly and delete line 208

```
                        # If param is a file of any kind, escape any \
                        val = re.sub(r"\\", r"/", val)
                    val=re.sub("\n", "\\n", val)  # Replace line breaks with escaped line break character
                    val=re.sub("\\\\", "/", val)  # handle older exps where files were valType=str not file
```

tpronk:
I've applied your suggestion, but I see I targeted the "release" branch instead of "dev". One moment please...

PR #3957, opened 2021-06-09
Slashes everywhere

tpronk:
This bug-fix replaces backslashes in experiment parameters by slashes. This fixes the "missing resources in JS" issues, where resource keys in the resources list used slashes, but resource keys in the timeline used backslashes. I tested it by running [wdio_img](https://github.com/psychopy/psychojs_testing/tree/main/tests/wdio_img) online and offline.

Note that I added the conversion from backslashes to slashes after the if block that checks whether a valType is "file" or "table" (at line 204). I noticed that the file parameters in wdio_img weren't recognized as type "file", but as type "str".

NB, this PR is the same as [PR #3956](https://github.com/psychopy/psychopy/pull/3956), but now targeting dev instead of release.

### DIFF
--- a/psychopy/experiment/params.py
+++ b/psychopy/experiment/params.py
@@ -205,6 +205,7 @@ class Param(object):
                         # If param is a file of any kind, escape any \
                         val = re.sub(r"\\", r"\\\\", val)
                     val=re.sub("\n", "\\n", val) # Replace line breaks with escaped line break character
+                    val=re.sub("\\\\", "/", val) # Replace any backslashes by slashes
                     return repr(val)
             return repr(self.val)
         elif self.valType in ['code', 'extendedCode']:
--- a/psychopy/experiment/params.py
+++ b/psychopy/experiment/params.py
@@ -202,10 +202,10 @@ class Param(object):
                             # but for other targets that will raise an annoying error
                             val = val[1:]
                     if self.valType in ['file', 'table']:
-                        # If param is a file of any kind, escape any \
-                        val = re.sub(r"\\", r"\\\\", val)
+                        # If param is a file of any kind, replace \ by /
+                        val = re.sub(r"\\", r"/", val)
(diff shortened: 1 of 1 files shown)


TEParsons:
I think you're working off of an older branch - the merge conflict is with a commit I did some time ago which uses `pathlib.Path` to sanitize paths rather than manually replacing slashes

peircej:
@TEParsons there are a couple more cases that Thomas has found not handled but he'll put them in a separate PR based on up-to-date dev

tpronk:
TBH, I'm a bit confused as to what the right approach is. Let me pitch some code here to check if this is correct before filing a new PR. Starting at [line 209](https://github.com/psychopy/psychopy/blob/cfcdcc3babd35ddbd49592ad4cfab10b1f55f6e0/psychopy/experiment/params.py#L209) I've now got...
```
                    if self.valType in ['file', 'table']:
                        # If param is a file of any kind, use Path to make sure it's valid
                        val = str(Path(val))
                    val=re.sub("\n", "\\n", val) # Replace line breaks with escaped line break character
                    val = Path(val).as_posix() # Convert any backslashes to slashes
                    return repr(val)
```

peircej:
No, not that way. The idea is
1. When we're certain this is supposed to be a path then we can force to a path and then to posix.
2. After that do a simpler string replace for objects with valType='str' to handle older variants. We do not want to convert all entries with valType='str' into paths

So I believe this is what you want
```
                    if self.valType in ['file', 'table']:
                        # If param is a file of any kind, use Path to make sure it's valid
                        val = Path(val).as_posix()  # Convert to a valid path with / not \
                    val=re.sub("\n", "\\n", val)  # Replace line breaks with escaped line break character
                    val=re.sub("\\\\", "/", val)  # handle older exps where files were valType=str not file
                    return repr(val)
```

tpronk:
Thnx, I'll file a PR with this code

PR #3963, opened 2021-06-10
ENH: Update param font if has dollar label

TEParsons:


TEParsons:
Marked as ENH as it isn't technically broken as is, but not doing this font change is very noticeable and annoying

PR #3972, opened 2021-06-11
FF: Page titles in the Preference Dialog is not translated

hsogo:
These texts are not translated in the latest Release branch.
(I think this code is not released yet, but this fix is for Release branch.  Should I use BF tag in this case?)
image


peircej:
Thanks @hsogo 
We're building towards a new release (2021.2 hopefully next week) and, right now, the FF/BF distinction is rather blurred. Sorry we need to find a way to improve communication on this.

hsogo:
Oh, new release is coming soon?  I'll check for untranslated texts.

PR #3975, opened 2021-06-11
ENH: Classic themes should use old fonts for continuity

TEParsons:


TEParsons:
This isn't important but it's just the values it was before so very unlikely to cause problems. Don't mind whether this is pulled in or not, was just thinking of change-phobes who won't like the new font :p

PR #3976, opened 2021-06-11
ENH: Use Open Sans for README window

TEParsons:


peircej:
This is an ENH not a BF, but we're in feature freeze. How important is it and how certain are we that this change won't break something else? We know that font changes tend to mess things up on other platforms, right?

TEParsons:
Not hugely important but is very visible, I uninstalled open sans & disabled the private loading font code and all it did was use default Arial, no errors, so this should be safe

PR #3982, opened 2021-06-11
BF: Check disabled against bool method, not value

TEParsons:
was causing standalone routines to always compile as if disabled until field was checked and unchecked again


peircej:
Does `.params['disabled']` ever get stored as a string? Because if so we could have the additional issue that 
`bool("False") == True`

PR #3987, opened 2021-06-11
DOCS: redirect to resources on workshops page

RebeccaHirst:
Moving all materials presented in the "Resources" page to instead be hosted at workshops.psychopy.org Updating navbar to redirect there instead of internal page.


peircej:
@RebeccaHirst your pull request seems to remove the files FlowTop_CompLeft.png etc which we don't want to do! 😄 
Could you go back and check out the dev branch, pull from upstream dev and then start again creating a new branch and pull request from that? If it still suggests removing these files as part of the pull request could you ask git to put them back instead (something like "undo unstaged changes")?

PR #3989, opened 2021-06-14
BF: Use and translate callback in Button JS

TEParsons:


TEParsons:
@tpronk @thewhodidthis is this translating as expected? I tried it out with some basic JS and it seems fine to my untrained eye

PR #3992, opened 2021-06-14
FF: misc Builder related eye tracker issues

isolver:
FF: builder eye tracker validation target defaults should match calibration target defaults
FF: Fixed Tobii calibration settings being sent from Builder
FF: Eyelink handle case of Builder not saving hdf5 file (no edf file should be saved either).


isolver:
@TEParsons, even after the validation gfx defaults are set to match the calibration defaults, the validation target looks quite different then the calibration target when using Builder generated code. Is this because of how the validation target stim you make is created / drawn? If so, could it be made to operate / appear the more like the calibration target does?

PR #3997, opened 2021-06-14
BF: Use different labels for transcription backends

TEParsons:
Means that e.g. AZURE can be printed even though the value in the dlg is "Microsoft Azure"


psychopy/demos/builder/Experiments/wordNaming/wordNaming.psyexp
```diff
@@ -163,7 +163,7 @@
         <Param val="2.5" valType="code" updates="constant" name="stopVal"/>
         <Param val="False" valType="bool" updates="None" name="syncScreenRefresh"/>
         <Param val="True" valType="bool" updates="None" name="transcribe"/>
-        <Param val="BUILT-IN" valType="code" updates="None" name="transcribeBackend"/>
+        <Param val="Buit-In" valType="code" updates="None" name="transcribeBackend"/>
```
peircej: typo: missing an l

PR #3999, opened 2021-06-14
EN/RF/FF: Transcription improvments

mdcutone:
This PR introduces the following changes:

- Adds a `TranscriptionResult` class, this is returned by transcribe. Gives us breathing room to add more features in the future, also deals with error states.
- API key is loaded from an environment variable or preferences in not given the `transcribe()`. Looks for the environment variable `PSYCHOPY_TRANSCR_KEY_<API_NAME>`.
- Key loading is done on import now, not every time the transcription is run.
- Moves transcription to it's own module, this will allow us to expand this feature in the future without building it within audioclip. Still retains the `transcribe` method on audioclip.
- Removed `rawResp` for now since the new `TranscriptionResult` takes it's place.

Things that need to be done:

- [x] Transcription doc pages for new classes and functions.
- [x] JavaScript boiler plate using the new transcription results object.
- [ ] Confidence values are not reported, might need to parse raw responses.
- [x] CMU Sphinx has a wierd way of specifying expected words, currently fails if you try. See below...

Pocket Sphinx requires expected words/phrases to be specified with confidences like so:

```
words = [('hello', 0.5), ('bye': 0.5)] 
```

Where each value is a tuple where the first value is the expected phrase and the seconds is the confidence/sensitivity level. This functions differently than other API calls. We need come up with a way to deal with this difference since it can be confusing to users. Any ideas @peircej?


mdcutone:
Updated, now multiple transcription API keys can be used.

PR #4000, opened 2021-06-14
BF: Syntax error in boiler plate for saving data from mic

RebeccaHirst:


RebeccaHirst:
closing because this fix does not work when the mic is in a loop

PR #4011, opened 2021-06-15
Merge dev release

peircej:
This includes a merge conflict with translations and the new StartStopCtrl

@TEParsons @hsogo maybe check the few lines around in case I did something daft?
https://github.com/psychopy/psychopy/commit/f957026ef21b50847d9cd16824ae23202e19904e#diff-7bcdb89ea5b3ec5cae09c8cf2877e7c044c8daf263b60dcf1758fcb485fd3e68L409


TEParsons:
Looks reasonable - so the merge conflict was just setting allowedVals to be the val if not found? I think you've done that right

hsogo:
I also think this will work fine.

peircej:
Thanks guys

PR #4014, opened 2021-06-15
BF: eye tracker demo units change for look time

RebeccaHirst:
Change units of look time for eye tracker demo from ms to seconds


TEParsons:
Beat you to it ;) #4013

PR #4016, opened 2021-06-15
BF: Checking mic component

TEParsons:


tpronk:
Hey @TEParsons, just checking: does this also work if the mic component is not in a loop?

PR #4032, opened 2021-06-15
BF: Some additional transcriber fixes.

mdcutone:
Some additional fixes which came up in testing, also some other changes requested.

- Engines `sphinx` and `built-in` are aliases for each other.
- Fixed issue when None is passed to `expectedWords`.
- Removed empty words from returned value, sometimes seen with sphinx when specifying `expectedWords`.


peircej:
I also don't get why sphinx seems to return more than one copy of a word. Like you tell it to expect ['red', 'gree'n, 'blue'] and it suggests it found ['red', 'red', 'blue', 'red']. What does that even mean?! 😦

PR #4036, opened 2021-06-15
BF: fonts under macOS (not sure what range of versions) were tiny

peircej:
The params of dialog boxes were tiny without this

Simply scaling up the font for macOS on retina. Not sure if we need it
elsewhere yet


TEParsons (approved):
Weird that font sizes aren't consistent across devices, but this seems a good solution

PR #4039, opened 2021-06-16
BF: Add some name aliases for Built-In

TEParsons:


peircej:
At the moment we convert to a bunch of options here and then transcribe.py has a bunch of `transcriberEngineValues` and then it also has a dict of `_recognizers`. It's a mess of lookup-relookup

peircej:
Given that it all comes from a dropdown menu in builder we should be able to get this right once and stay right

peircej:
I'll pull for now but I think it needs a cleanup (and you know how rarely I say that! 😉 )

PR #4043, opened 2021-06-16
BF: Use google cloud locally, to line up with online

TEParsons:


peircej:
closing in favour of #4051

Issue #3995, opened 2021-06-14
Gazepoint calibration / validation from Builder


isolver:
In Dev branch, when gazepoint is used from Builder, ET calibration runs, but the resulting gaze data seems __very__ poor compared to the same setup when one of the Coder demos is run with gazepoint. Only thing I can think it that there is some difference in settings being provided by Builder vs. those used by coder examples and that is causing an issue. 

MouseGaze, Tobii and EyeLink run fine, so the issue seems specific to GazePoint being used in Builder

isolver:
Seems to have been fixed by https://github.com/psychopy/psychopy/pull/4060.
Closing.

PR #4060, opened 2021-06-17
BF: iohub GazePoint issues

isolver:
BF: fixed further iohub bytes vs. str issue in standalone installer 
BF: validation procedure was failing when using gazepoint


isolver:
fixes #3995

PR #4064, opened 2021-06-18
FF: Transcriber fixes

mdcutone:
Couple of changes here to get the voice stroop demo working on my end.

- Use `getRecording` in `bank` to prevent passing samples to the recognizer during a recording. `getRecording` raises an exception.
- Added `stop()` to microphone boilerplate before `bank`. We're passing samples to the recognizer *while* it was still recording them.
- Updated documentation to fix some inaccuracies.
- Speech-to-text API key loader properly handles empty prefs as `None`.  
- Put Google back in as a transcriber option. Might want to use longer names to better delineate plain Google and Google Cloud. See `psychopy.sound.transcribe.transcribeEngineValues` for my suggestions for Builder.
- Fixed `expectedWords` default value in `transcribe()` function.

### DIFF
--- a/psychopy/experiment/components/microphone/__init__.py
+++ b/psychopy/experiment/components/microphone/__init__.py
@@ -37,7 +37,8 @@ onlineTranscribers = {
     "Microsoft Azure": "AZURE",
 }
 localTranscribers = {
-    "Google": "googleCloud",
+    "Google": "google",
+    "Google Cloud": "googleCloud",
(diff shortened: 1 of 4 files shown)


TEParsons:
Just pulled this down and tried it out, running `stroopVoice` I'm still getting "red green blue blue red blue green" etc. spam with mic set to Built-in, and blank output when I try `Feature Demos -> microphone` with it set to Google

PR #4068, opened 2021-06-21
BF: Fixes to mic transcription

TEParsons:
- Use stroop demo
- Expose Built-in in Builder
- Clearer wording on transcriber alerts
- No auto-substitution of transcribers (gets confusing)
- Load credentials as string if given a file path
- Tooltip with instructions on using : to specify confidence lvls
- Stroop demo has `:100` after each keyword by default (otherwise oversensitive, meaning keyword spam)


peircej:
@hsogo sorry, this does actually alter one of the messages - we've added some more detail to the tooltip about setting "sensitivity" for the expectedWords param

PR #4071, opened 2021-06-21
RF: Transcriber changes

mdcutone:
This PR changes how transcription operates under-the-hood, using the official Google Cloud API library instead of SpeechRecognition. I removed some options from the components dialog, only Pocket Sphinx and Google Cloud are supported. You may need to update demos to only use those options.

To use Google Cloud, set the property `appKeyGoogleCloud` in prefs to point to your key.json file. Select "Google" in the microphone component to use that transcriber. You can no longer specify a key through the `transcriber()` functions directly, only through environment variables and prefs.

### DIFF
--- a/psychopy/app/preferencesDlg.py
+++ b/psychopy/app/preferencesDlg.py
@@ -326,8 +326,9 @@ class PrefPropGrid(wx.Panel):
         if section not in self.sections.keys():
             self.sections[section] = []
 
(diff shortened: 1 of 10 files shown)


peircej:
Why do we need `appKeyGoogleCloud` as well as `transcrKeyGoogleCloud`? Aren't they the same thing?

PR #4079, opened 2021-06-22
Translation tweak for Japanese

peircej:
Sorry @hsogo we made a v minor change to one of the prefs - we removed `transcrKeyGoogle` and `transcrKeyGoogleCloud` in favour of `appKeyGoogleCloud` (because the one key actually does more than transcription if desried)

The way I've handled it here is that, for labelling purposes in the immediate release, it actually translates the old phrase so your previous translation will show up anyway. If you do want to update that translation then you can.

Also your own prefs will presumably have the 2 old keys being loaded from the .psychopy3 appdata folder but those won't be translated. Never mind - other users won't be seeing those.


hsogo:
@peircej Thank you for letting me know. This tweak works fine under Japanese environment, but old label is shown under English environment.
image
In addition, hint text for this item has also changed so that translation have to be updated.  I'll fix them and send pull request immediately.

PR #4083, opened 2021-06-25
BF: Slider start value was being overwritten by `.reset()` call

### LINKED ISSUE #4063: Slider start value doesn't affect .value, .rating or .markerPos
Even if Slider component in Builder is given a start value, it still starts off as None (so you still have to do `(slider.markerPos or 0)`)

TEParsons:
fixes #4063 in Python

@thewhodidthis does the JS slider have a reset function? Is this something we should implement online too?


thewhodidthis:
@TEParsons Yes, looks like it does, https://github.com/psychopy/psychojs/blob/main/src/visual/Slider.js#L250

PR #4084, opened 2021-06-25
NF: Initial commit of new Python/JS REPL

mdcutone:
This PR introduces a new control which replaces PyShell. This control supports running interpreters independent from the one used by the main application. This is much safer than before as typing a command like `while 1:0` into the current shell will crash the application, something that may happen in teaching environments. Python's built-in interactive help system is also supported by this widget.

You can also run other REPL interfaces such the one for node.js (again useful for teaching) and even system terminals. Still rough but will be fully fleshed out before the next feature release.


peircej:
@mdcutone FYI I merged to dev, not release

PR #4092, opened 2021-06-30
Multichannel sound option

jundurraga:
Dear psychopy coders, 
I have made some tiny modifications to allow for multichannel sounds with ASIO drivers. 
This is useful in situations with more than two simultaneous sounds are needed, e.g. sending sound and triggers from a soundcard, or quadraphonic signals, etcetera. 
I have tested this only using the test example I also included in this commit. If this requires a more rigorous test, I would be happy to try it with some of your test scripts. 
Cheers,
Jaime

### DIFF
--- a/psychopy/tests/test_sound/test_multichannel_asio.py
+++ b/psychopy/tests/test_sound/test_multichannel_asio.py
@@ -1,8 +1,7 @@
-from psychopy import prefs
-prefs.general['audioLib'] = ['sounddevice']
-from psychopy import sound
+from psychopy import sound, prefs
 import numpy as np
 from psychopy import core  # import some libraries from PsychoPy
+prefs.general['audioLib'] = ['sounddevice']
 
 fs = 44100
 my_asio_device = 'ASIO Fireface USB'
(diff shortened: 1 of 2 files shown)


psychopy/tests/test_sound/test_multichannel_asio.py
```diff
@@ -1,8 +1,7 @@
-from psychopy import prefs
-prefs.general['audioLib'] = ['sounddevice']
```
peircej: This actually needs to be here - in this test we're deliberately changing the sound lib in preferences which needs to be done before importing sound (otherwise the sound will be imported thinking it's a different preferred lib) so on this occasion you need to just override PEP8. You might be able to do `# noqa` as a comment on that line to stop you editor complaining about it

peircej (changes requested):
Thanks Jaime,

I can see the direction you're taking here and it sounds sensible although I certainly notice one incorrect change caused by overzealous PEP8 corrections. I'd also like @mdcutone to check I'm not missing anything silly. Unfortunately the sound libs don't get tested on the automatic test suite because the virtual machines being used don't have audio cards so we have to be particularly cautious on audio changes.

For your commit messages, could I draw your attention to the docs here: https://www.psychopy.org/developers/repository.html
Basically, we like to distinguish between changes that are very minor and can be included in a point release (i.e. where we're certain they won't break anything) versus changes that are new features or changes in behaviour because we tend to hold those for a future release. In your case, to be fair, it's harder to know but I think I'd probably call this an Enhancement (rather than a bug fix) and hold it back for 2021.3.0 rather than 2021.2.1

peircej:
Just to be clear, I think we can live with the slightly incorrect commit names, but I'd like you to revert the import order with prefs setting in the test. Thanks for contributing though - great to see people getting involved! 🥳

psychopy/tests/test_sound/test_multichannel_asio.py
jundurraga: Thanks, I have now read some of the documentation. I performed a new pull request on dev, with the changes you suggested. Thanks!

PR #4094, opened 2021-07-01
FF: Make REPL console a RichTextCtrl so it accepts themes

TEParsons:


mdcutone (approved):
Yes you can pull this in now. Just sets the theme for the REPL.

PR #4097, opened 2021-07-01
TESTS: App unit test fixes

mdcutone:
Some fixes to the application unit tests. Adds a timeout to the experiment settings dialog which prevented the suite from advancing. Also adds the `quitApp()` function and testing mode to `startApp()`.


papr:
Hi, this PR, more specifically https://github.com/psychopy/psychopy/commit/599d16d7f142b5873bdc7b8c06381de806ad06f0, causes the settings dialogue to close automatically when starting the UI via `python -m psychopy.app.psychopyApp`. Based on the commit message, this is only intended in test mode.

I can contribute a PR to fix this. My questions would be:

1. Should `DlgExperimentProperties` be responsible for checking if it is in test mode? Or should `BuilderFrame.setExperimentSettings` take care of this?
1. What is the best way to check if the app is in test mode from within the `DlgExperimentProperties`?

Also note, that `BuilderFrame.setExperimentSettings` has a timeout argument which is not passed along to `DlgExperimentProperties`. I am not sure if this is intentional.

Edit: Looks like "test mode" is set implicitly by setting the timeout argument in `BuilderFrame.setExperimentSettings`. Therefore, forwarding the argument to `DlgExperimentProperties` solves the issue. See #4167

PR #4101, opened 2021-07-05
RF: Move eyetracking to its own module under psychopy

TEParsons:
Eyetracking being all over the place makes the API harder to navigate, but it's possible to move all Eyetracking stuff to its own module without breaking imports. Can move this to hardware.eyetracking if needed.


TEParsons:
@RebeccaHirst this addresses what we were talking about, would mean you could get all the eye tracker stuff (calibration, validation, roi and record) under `psychopy.eyetracking` rather than being spread across `psychopy.hardware.eyetracker`, `psychopy.visual`  and `psychopy.iohub`

isolver (commented):
I think we should close this PR for the time being. We can do an api review after 2022.0.x. Generally, I like the idea of moving visual related iohub code to psychopy.visual, but I also think we should be using the iohub eyetracker device / api directly in code generation as much as possible so that Builder code gen is somewhat similar to what the Coder eye tracking examples use. In other words, I don't think we want two Coder level eye tracking api's, both using the iohub eye tracker device. It is hard enough maintaining one. ;)

PR #4105, opened 2021-07-05
DOCS: Eye tracker validation docs

RebeccaHirst:
@isolver @TEParsons this is a basic placeholder for the builder eye tracker validation component. I would like to expand on the info under the data tab (perhaps with a screenshot of the results) but I believe this only applies to cases where an eye tracker is plugged in (not to mousegaze) - is that correct?


isolver:
@RebeccaHirst, you can use MouseGaze in Validation, just move the mouse to each target point during the validation. I wiggle it a bit so that you see a little spread of the mouse samples around each target point (see attached example). If you did want to show an image, the nice thing about using MouseGaze is that we are not making any implicit / unintended claim about acccuracy of an actual eye tracker, which is often very dependent on the eye tracker setup. 

validation_09_07_2021_06_54

PR #4117, opened 2021-07-08
BF: Routine never ends in PsychoJS

bollaking:
If sound with duration less than 0.5 seconds …the routine never stops. 

To reproduce problem  play sound with 800 hertz for 0.4 seconds

Warning: I DONT UNDERSTAND THE PURPOSE OF THIS CODE
This change fixes my problem of the routine to never ending because status was never getting set to FINISHED.

There was a problem with the indentation which might have something to do with the problem.  I have fixed the indentation problem, and moved the status = FINISHED out of the "if" block.

### DIFF
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -197,10 +197,10 @@ class SoundComponent(BaseComponent):
         else:
             # sounds with stop values
             self.writeStopTestCodeJS(buff)
-            code = ("if (%(stopVal)s > 0.5) {"
+            code = ("if (%(stopVal)s > 0.5) {\n"
                     "  %(name)s.stop();  // stop the sound (if longer than duration)\n"
-                    "  %(name)s.status = PsychoJS.Status.FINISHED;\n"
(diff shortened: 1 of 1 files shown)


peircej:
I *think* we can do away with that if statement altogether now. There was once a problem around very brief sounds being stopped before they'd really started if they were short and I think this code is there for that purpose but I want to do some more thorough testing of the issue before pulling this in

thanks

bollaking:
OK 
FYI This is the javascript code (with the poor indentaiont) that is created for a sound component with duration 0.4 seconds
```
frameRemains = 1.0 + 0.4 - psychoJS.window.monitorFramePeriod * 0.75;  // most of one frame period left
    if (sound_1.status === PsychoJS.Status.STARTED && t >= frameRemains) {
      if (0.4 > 0.5) {  sound_1.stop();  // stop the sound (if longer than duration)
        sound_1.status = PsychoJS.Status.FINISHED;
      }
    }
```
As you can see `sound_1.stop()` and `sound_1.status = PsychoJS.Status.FINISHED` will never get run.

PR #4130, opened 2021-07-12
TEST: One test for units on all relevant stim

TEParsons:


peircej:
That's an awful lot of commits for release trunk, including RFs. Was it meant for `dev`?

TEParsons:
It's because it's based on other branches, so it's got all their commits too. Should be fine to pull into Dev though, and might be safer

PR #4136, opened 2021-07-14
BF: Have QuestHandler with method "mode" use first value of tuplet returned by QuestObject

tpronk:
Using the QuestHandler with method "mode" gave an error. That's because the QuestObject returned a tuple with two values, of which the first value is the mode of the posterior PDF that we are after. See [this line](https://github.com/psychopy/psychopy/blob/6118a09794404de17a7a2842fcf258822f5bbc94/psychopy/contrib/quest.py#L206). This bug-fix ensures that the mode returned by the QuestObject is indeed used.

### DIFF
--- a/psychopy/data/staircase.py
+++ b/psychopy/data/staircase.py
@@ -969,7 +969,7 @@ class QuestHandler(StairHandler):
         if self.method == 'mean':
             self._questNextIntensity = self._quest.mean()
         elif self.method == 'mode':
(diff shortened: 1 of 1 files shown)


tpronk:
Added a tiny enhancement; noticed in my test that all of the parameters are exposed, except for pThreshold, so I exposed that one too

peircej:
Thanks @tpronk 
How strange that mean returns one value and mode returns 2! 🤔 😆

PR #4139, opened 2021-07-18
DOC: Fix deprecation documentation of ShapeStim color parameter

sueskind:
The `color` parameter for the constructor of the `psychopy.visual.shape.ShapeStim` class was wrongly not documented as deprecated. However, this parameter is simply ignored which caused a weird bug for me, thinking I could use this as a shorthand for `lineColor` and `fillColor`.

### DIFF
--- a/psychopy/visual/shape.py
+++ b/psychopy/visual/shape.py
@@ -532,9 +532,7 @@ class ShapeStim(BaseShapeStim):
         Enable auto drawing. When `True`, the stimulus will be drawn every
         frame without the need to explicitly call the
         :py:meth:`~psychopy.visual.ShapeStim.draw` method.
-    color : array_like, str, :class:`~psychopy.colors.Color` or None
-        Sets both the initial `lineColor` and `fillColor` of the shape.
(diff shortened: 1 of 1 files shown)


sueskind:
Apologies, I forgot to label this pull request as **DOC**.

peircej:
Thanks @sueskind but I actually thought color *is* supposed to operate as a synonym for applying changes to both line and fill. I'm sure that was discussed at one point and I thought it was implemented 🤔 
@TEParsons do you know anything about this? Could it have gotten lost in your refactoring or was it never implemented in the first place?

TEParsons:
I think that's just on init, changing `.color` just refers to the ColorMixin function... Could overload it though, I'll give it a go tomorrow and see if it breaks anything.

TEParsons:
My mistake, it already is overloaded! I just didn't see as it's in BaseShapeStim rather than ShapeStim. So yes, you're right, setting `ShapeStim.color` sets both `ShapeStim.fillColor` and `ShapeStim.borderColor`.

peircej:
OK @sueskind I'm closing this for now as not needed but let us know if there's any further query

PR #4141, opened 2021-07-19
BF: Update window after color is changed

TEParsons:
Otherwise the color doesn't take

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -2651,6 +2651,12 @@ class Window(object):
             return getattr(self._color, self.colorSpace)
     @color.setter
     def color(self, value):
+        # Get old color for reference
+        if hasattr(self, "_color"):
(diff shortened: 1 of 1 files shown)


peircej:
This is something that's highlighted in the dos actually: https://www.psychopy.org/api/visual/window.html#psychopy.visual.Window.color

The problem with the solution you've made here is that setting the color mid-frame will delete preceding rendering calls:
```
stim1.draw()
win.color = 'red'
stim2.draw()
```
only stim2 will be visible. One could more simply call `win.clearBuffer()` (a bit more efficient than calling `win._setupGL()` and that solution is also documented but has the same issue that it will obscure all rendering.

I've wondered about removing it as an option altogether, so you can only ever set the color at window creation, or We could potentially just raise an alert to warn people that using win.color is probably not what they want. But I think the solution here isn't ideal

PR #4143, opened 2021-07-19
BF: Fixes to button timing

TEParsons:


peircej:
@TEParsons What was the symptom being fixed here? (for the purposes of the changelog)

TEParsons:
Button timings were being recorded relative to when button.__init__ was called when set to "relative to button onset", rather than being relative to when the button was presented

Issue #4138, opened 2021-07-18
Microphone: using en-GB raises error


RebeccaHirst:
using en-GB raises error when using mic transcription in builder even though the hover prompt tells users to use en-GB.

  File "C:\Users\44797\Desktop\tests\mic\untitled_lastrun.py", line 333, in <module>
    language='en-GB', expectedWords=None
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\sound\microphone.py", line 929, in bank
    engine=engine, **kwargs)
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\sound\audioclip.py", line 756, in transcribe
    config=config)
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\sound\transcribe.py", line 345, in transcribe
    config=config)
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\sound\transcribe.py", line 490, in recognizeSphinx
    msg = (f"Language `{config['language']}` is not installed for "
KeyError: 'language'

mdcutone:
Thanks for pointing this out, will fix.

PR #4149, opened 2021-07-20
BF: docs should advise the use of en-US not en-GB (en-GB is not in th…

peircej:
…e box)

Fixes issue #4138


peircej:
@hsogo This commit includes a minor change that will need updating in the translation index

PR #4165, opened 2021-07-26
Quest JS outputs from Builder

peircej:
This branch largely comes from 
https://github.com/psychopy/psychopy/pull/4147
but with the cryptography error hopefully fixed


TEParsons:
I've been pushing commits to #4147, might be worth rebasing to check for merge conflicts?

PR #4167, opened 2021-07-26
FF: Only auto-close experiment settings dialogue in test mode

papr:
Context: https://github.com/psychopy/psychopy/pull/4097#issuecomment-886768714

It looks like the experiment settings dialogue should only auto-close during testing. This PR forwards the timeout argument set during testing to `DlgExperimentProperties` which is responsible for handling the timeout.


peircej:
Thanks @papr 
We'll work out what's going on with that bug fix on the xml schema (only on ubuntu)

PR #4172, opened 2021-07-27
TEST: Use root of current file for temp dir

TEParsons:
No effect on tests running online, but when run locally means they still work even if you're using e.g. PyCharm running off of C:// and your psychopy repo in F:// (which is how mine is)


peircej:
Doesn't work on ubuntu because no permission to write to root folder

peircej:
The current failure in the tests has, I think, been fixed already. Does the branch need updating from dev to pass?

PR #4179, opened 2021-07-27
ENH/BF: `Rift` class

mdcutone:
This PR fixes bugs in the `Rift` class introduced in the previous version. Also updates things to use PsychXR 0.2.4 as the current version in use is very outdated.

- Explicitly specifies colorspace when clearing a VR view buffer.
- Removes the `shutdown()` command to prevent memory access violations.
- Some modifications to work with PsychXR 0.2.4 which uses the latest version of the Oculus SDK (23.0).
- Tweaked the head-tracking demo a bit.

@peircej Don't pull this in until PsychXR 0.2.4rc3.post0 is put up on PyPI. That should happen by tomorrow at the latest.


mdcutone:
Updated the packages on PyPI for PsychXR to version `0.2.4rc3.post0`. That will be the minimum requirement for using the Rift class here.

PR #4186, opened 2021-07-29
BF: fixes Minolta recognition error in python 3.9 (maybe other versions of python 3.x are also affected)

guillermoaguilar:
There was a type error that prevented the Minolta class to get initialized. Error was:

ERROR     Couldn't initialize photometer LS100: can't concat str to bytes

It turns out the error was due to different variable types (bytes or string). The commit fixes the inconsistency. 

I'm not sure if this error occurs exclusively in python 3.9.  I tried to install psychopy in a conda environment with python 3.6 (as recommended by the documentation) but I failed to install wxpython (which is whole another story).

### DIFF
--- a/psychopy/hardware/minolta.py
+++ b/psychopy/hardware/minolta.py
@@ -223,24 +223,24 @@ class LS100(object):
         timeout for a response.
         """
         if message[-2:] != '\r\n':
-            message += '\r\n'  # append a newline if necess
+            message += b'\r\n'  # append a newline if necess
 
         # flush the read buffer first
         # read as many chars as are in the buffer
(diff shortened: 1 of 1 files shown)


peircej:
Many thanks for picking this up and fixing it @guillermoaguilar  The old wrong-way-round error for encode/decode! 😆 

I've followed it up with a further commit in #4212
The idea there is that we delay the use of bytes as long as possible to reduce erroneous comparisons with strings. Convert to bytes *just before* sending the message and convert back to strings *as soon as possible*. It was giving rise a a further bug that was probably harmless, where we were adding `\r\n` to every message by an erroneous comparison

thanks again

guillermoaguilar:
You're welcome!

PR #4187, opened 2021-07-29
NF: Intial work on new plugin system

mdcutone:
This plugin system is based on `PluginBase` rather than `distutils` or `setuptools` like before. This should allow us to use plugins without needing to add/install packages to the Python environment. PsychoPy looks for plugins in directories specified by the user in preferences. Plugins can be either a Python file, folder, or zip file. 

We still need to work out a formal specification for plugins so developers can extend PsychoPy. Plugins can be used to create third-party builder components, add window backends, etc.

### DIFF
--- a/psychopy/plugins/__init__.py
+++ b/psychopy/plugins/__init__.py
@@ -47,7 +47,7 @@ import psychopy.experiment.components as components
 # package dotted path which loaded plugins appear
 PLUGIN_PACKAGE = 'psychopy.plugins'
 # prefix for temporary directories for plugins extracted from Zip files
-PLUGIN_TEMP_FILE_PREFIX = 'psychopy_plugin_'
+PLUGIN_TEMP_FILE_PREFIX = 'psychopy_plugin'
 # Keep track of plugins that have been loaded. Keys are plugin names and values
 # are their entry point mappings.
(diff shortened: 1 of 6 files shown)


papr:
@mdcutone Hi! Just wanted to let you know that we (Pupil Labs) have had good experiences with such an approach in [our own software](https://github.com/psychopy/psychopy/pull/4187). Especially, since it also works for stand-alone apps. 👍🏻

In our software, nearly every functionality is build as a plugin now. This makes it easy for users to turn on/off functionality as they need it.

From what we have learned, it can make sense to have different plugin bases, depending on their purpose. In PsychoPy, this could mean e.g. to have a plugin base for frontend components as well as a separate base for plugins running in the ioHub backend. The latter would be especially helpful for people (like us) integrating new hardware without having to wait for the next major PsychoPy release.

I am looking forward to using this new plugin api! We will make sure to adopt it as early as possible in order to give feedback.

mdcutone:
Hello @papr,

Your feedback will be greatly appreciated!

Right now the specification for our plugin system is very similar to that of Blender's.  The plugin system will cover both adding features to the Builder/UI and the PsychoPy library itself.

PR #4196, opened 2021-08-04
BF: Set color object contrast on set contrs for ElementArrayStim, fixes #3816

### LINKED ISSUE #3816: ElementArrayStim is ignoring contrast parameter
`visual.ElementArrayStim` is ignoring the `contrs` parameter, and it is not possible to set the contrast of the visual stimuli. 

Example:

```python
from numpy.random import rand
from psychopy.visual import Window, ElementArrayStim
from psychopy.event import waitKeys

win = Window()
stim = ElementArrayStim(win, sfs=3, sizes=.2, contrs=rand(100))
stim.draw()
win.flip()
waitKeys(keyList=["space"])
```

image

Indeed, it looks like the code where `contrs` took effect was removed recently:

```
$ git diff -U10 2020.2.10 release -- psychopy/visual/elementarray.py
```
```diff

     def updateElementColors(self):
         """Create a new array of self._RGBAs based on self.rgbs.

         Not needed by the user (simple call setColors())

         For element arrays the self.rgbs values correspond to one
         element so this function also converts them to be one for
         each vertex of each element.
         """
         N = self.nElements
-        self._RGBAs = numpy.zeros([N, 4], 'd')
-        if self.colorSpace in ('rgb', 'dkl', 'lms', 'hsv'):
-            # these spaces are 0-centred
-            self._RGBAs[:, 0:3] = (self.rgbs[:, :] *
-                self.contrs[:].reshape([N, 1]).repeat(3, 1) / 2 + 0.5)
-        else:
-            self._RGBAs[:, 0:3] = (self.rgbs *
-                self.contrs[:].reshape([N, 1]).repeat(3, 1) / 255.0)
-
-        self._RGBAs[:, -1] = self.opacities.reshape([N, ])
-        # repeat for the 4 vertices in the grid
-        self._RGBAs = self._RGBAs.reshape([N, 1, 4]).repeat(4, 1)
-
+        _RGBAs = numpy.zeros([len(self.verticesPix), 4], 'd')
+        _RGBAs[:,:] = self._colors.render('rgba1')
+        _RGBAs[:, -1] = self.opacities.reshape([N, ])
+        self._RGBAs = _RGBAs.reshape([len(self.verticesPix), 1, 4]).repeat(4, 1)
         self._needColorUpdate = False
```

If you search around that file, I don't think `self.contrs`  gets used anywhere, it is just set and then ignored.

It looks like maybe this is a mistake too, since `ElementArrayStim` does not have a `contrast` property:

```diff
+    @colors.setter
+    def colors(self, value):
+        # Create blank array of colors
+        self._colors = Color(value, self.colorSpace, self.contrast)
+        self._needColorUpdate = True
```

TEParsons:


TEParsons:
@peircej New commit uses that function now :) I kept the bit about extending rgb though, as that's needed to maintain old functionality. Previously, if the colour was `(1, 1, 1)` and contrast as an Nx1 array, it would just copy the color loads of times to match the length of the contrast array so they can be broadcast together.

PR #4199, opened 2021-08-05
NF/BF: Tools to parse strings for either variables of arguments

TEParsons:
Also implements these in RatingScale, fixes #1227

PR #4200, opened 2021-08-05
NF/BF: Use string tools to parse for variables

TEParsons:
Cherry picks the relevant commits from #4199 which are safe to pull into release rather than dev


peircej:
Task linked: [CU-d11gzt variable / function names in code bring up uneccessary warning](https://app.clickup.com/t/d11gzt)

PR #4204, opened 2021-08-06
DOC: Add documentation for contributing to the test suite

TEParsons:
@RebeccaHirst


peircej (changes requested):
Underlines in Sphinx (restructured text) should be at least the length of the text or they raise a warning, and we're trying to cut down the warnings in the sphinx build so that we spot the more important ones. Could you extend all the underlines please? :-)

PR #4206, opened 2021-08-06
NF: Added paste options to component context menu

### LINKED ISSUE #3798: Right-click on a Component allows you to copy but not Paste
To paste a Routine requires the Experiment menu.

Ideally:
- the context menu for a Component should have a Paste Above and Paste Below option
- the context menu when clicking on the background should paste to the bottom of the Routine (probably too much faff to try and work out where the mouse is, but this is a good start)

mdcutone:
This PR resolves #3798 by adding 'paste above' and 'paste below' options to the component context menu.


TEParsons (changes requested):
Works well! One thing though: Rather than `copy (component)` and `paste above (component)`, could we have `copy 'component'` and `paste 'component' above`? I think that would make it clearer that you're pasting `component` above what's currently clicked on, not pasting above `component`.

mdcutone:
> 
> 
> Works well! One thing though: Rather than `copy (component)` and `paste above (component)`, could we have `copy 'component'` and `paste 'component' above`? I think that would make it clearer that you're pasting `component` above what's currently clicked on, not pasting above `component`.

Not really easy to do since "paste above" and "paste below" are keys used to generate the context menu. Also the grammar for translations may break.

PR #4209, opened 2021-08-09
BF: Fix #1983, small nElements in ElementArray causing index errors

### LINKED ISSUE #1983: ElementArrayStim randomly fails to generate positions when nElements is too low
This issue exists for me on both 1.90.03 and 3.0.0b7 (2 & 3), not sure if it's a bug, not supported behavior, or I'm doing something wrong. I'm on Mac OS 10.11.6.

Here is a minimum example:

```
from psychopy import visual

window = visual.Window()

for i in range(1000):
    dots = visual.ElementArrayStim(window,nElements=5)
```

Trackback:

```
  File "/Users/Colin/Desktop/debug_crash.py", line 6, in <module>
    dots = visual.ElementArrayStim(window,nElements=5)
  File "/Applications/PsychoPy3.app/Contents/Resources/lib/python3.6/psychopy/visual/elementarray.py", line 159, in __init__
    self._updateVertices()
  File "/Applications/PsychoPy3.app/Contents/Resources/lib/python3.6/psychopy/visual/elementarray.py", line 616, in _updateVertices
    units=self.units, win=self.win)
  File "/Applications/PsychoPy3.app/Contents/Resources/lib/python3.6/psychopy/tools/monitorunittools.py", line 95, in convertToPix
    return unit2pixFunc(vertices, pos, win)
  File "/Applications/PsychoPy3.app/Contents/Resources/lib/python3.6/psychopy/tools/monitorunittools.py", line 58, in _norm2pix
    return (pos + vertices) * win.size / 4.0
ValueError: operands could not be broadcast together with shapes (3,2) (20,2)
```

Changing `nElements` to `20` works as expected. I poked around a bit and obviously ElementArrayStim is [expecting a larger number of objects](https://github.com/psychopy/psychopy/blob/c8f7fe70a8db6e47d38b8800456a69ab8d3725df/psychopy/visual/elementarray.py#L257), but it's not clear to me if a smaller number is a valid value.

TEParsons:


peircej:
I've reverted this after all in 8d9472d
I believe this change might alter the number/density of dots in the display. That's a behaviour change so isn't something we should be doing in a point release. Before proceeding with this we need to investigate the effect on dot density with respect to what number of dots are requested

PR #4211, opened 2021-08-10
NF: Layout classes

### LINKED ISSUE #1296: Add ___2norm and norm2___ conversion functions to monitorunittools
This was recently referenced on the forum:
http://discourse.psychopy.org/t/ratingscale-from-norm-to-cm-units/1398

Would it make sense to add these functions, particularly since some stimuli only accept norm units?

TEParsons:
Adds classes:
- Vector: An Nx2 or Nx3 array which knows its values in every unit space and can be accessed via `.pix`, `.norm`, `.height`, etc.
- Size: Identical to Vector, different name for convenience
- Position: Identical to Vector, different name for convenience
- Vertices: `numpy.ndarray` with additional methods to render vertices relative to either a sized object or size & position values

and implements these classes in `psychopy.visual` objects via the BaseVisual class methods


peircej:
Task linked: [CU-jvacg0 Integrate Vector class](https://app.clickup.com/t/jvacg0)

PR #4212, opened 2021-08-10
BF: further fix to minolta str/bytes handling

peircej:
There was an error with always adding `\r\n` to every string in `sendMessage` even if it had that there (due to a str==bytes comparison always being False).

In this PR `LS100.sendMessage()` is adapted to handle unicode inputs as well as bytes, and always returns a unicode string.
With that done, the messages are then better sent as unicode objects for less confusion


peircej:
@guillermoaguilar I don't actually have an LS110 to hand anymore (which is why that bug apparently survived for 4 years!). Would you have a chance to test my updated fix? Many thanks

guillermoaguilar:
I just tested the code at the last commit of the release branch (66432b203035bd45e7524a5e7dc0d9fa82a9b330)

**It works!** The Minolta is recognized and reads outs luminance values correctly.

PR #4216, opened 2021-08-11
NF: Add support for Pupil Core eye tracker

romanroibu:
This PR adds support for [Pupil Core](https://pupil-labs.com/products/core/) eye tracker.


papr:
Hi @isolver 

With the current version of this PR, we can run the `psychopy/demos/builder/Feature Demos/eyetracking/eyetracking.psyexp` successfully (assuming a successful external calibration and AOI/surface setup). Eye samples are created and passed to the foreground process successfully.

Unfortunately, we seem to have an issue with saving the samples to disk. We enable the "Save HDF5 file" option, the file is being created, and it includes tables for `/data_collection/events/eyetracker/MonocularEyeSampleEvent` and `/data_collection/events/eyetracker/BinocularEyeSampleEvent`. Unfortunately, these tables are empty. When we select the mouse point eye tracker, the tables are filled as expected. In both cases, the file contains mouse and keyboard data as expected.

The configuration sets `save_events: True`. Is there anything else we need to do to enable the automatic saving of eye samples?

isolver:
I think the issue is that the eye sample `status` field must be a `uint8`, not a string, for monoc and binoc events.

romanroibu:
Hi @isolver,

I've tried replacing the `status` value for both monocular and binocular events with `0`, but unfortunately this didn't solve the issue.

Are there other reasons why the data wouldn't be written to the HDF5 that we could've overlooked?

isolver:
Do you get any error messages, or just no eye data in the hdf5 file ? Thank
you

papr:
I do not see any explicit error message in the PsychoPy runner or the terminal from which I have launched builder. The only related message is:
> Alert 4510: A Pupil Labs eye tracker has been configured, but no calibration routine is present.

We do not run an explicit calibration sequence yet since it is not implemented yet. Instead, we assume the tracker to be calibrated already.

isolver:
Briefly reviewing your code, looks like you have the config file and eye tracker class setup correctly to generate samples. This leads me back to the issue likely being the sample format, but I maybe wrong. 

Just as a test, I hacked the `MouseGaze` implementation to try saving string sample status values and then ran the demo; I get this type of error printed for each monoc sample that tries to be saved:
```
Error saving event: [1, 1, 0, 327, 51, 13.653315900010057, 13.653315900010057, 13.653315900010057, 0, 0, 0, 22, -0.208984375, 0.1123046875, 0, 0, 0, 0, 0, 0, 0, 0, 5, 77, 0, 0, 0, 0, 0, 0, 0, 'string_status']
  File "D:\DEV\my-code\psychopy\psychopy\iohub\datastore\__init__.py", line 381, in _handleEvent
    np_array = np.array([tuple(event), ], dtype=eventClass.NUMPY_DTYPE)
ValueError: invalid literal for int() with base 10: 'string_status'
```
So I would have expected you to get the same error. Can you confirm you were not getting any errors when setting the status field to a string?
  
Could you send me one of the hdf5 files that is created when using the pupil core eye tracker (that is missing the sample events)? I want to see if there are any hints in the experiment log event table as to what might be going on.

Could your also add a `print2err(sample_field_list)` to your `_monocular_eye_sample_from_gaze_3d` and associated binocular sample creation function to print the sample list being returned by the function and give that output for a sample so we can check it's values as well?

Thanks again

papr:
[ppr_eyetracking_2021_Aug_24_1612.zip](https://github.com/psychopy/psychopy/files/7049328/ppr_eyetracking_2021_Aug_24_1612.zip)

papr:
I build a script that simulates our realtime network API by replaying a recording. When started, it waits for `setRecordingState(True)` before starting to publish/replay the recorded data.

```
Usage: simulate_ipc_backend.py [OPTIONS] CAPTURE_RECORDING_LOC

Options:
  --repeat
  --pupil-remote-port INTEGER
  --pub-socket-port INTEGER
  --help                       Show this message and exit.
```
Resources:
- [Replay script](https://gist.github.com/papr/49f58a894364dd94b23c53e6bc6929d0) (Requires Python >=3.6.1)
    - `pip install -r requirements.txt` to install requirements
    - `python simulate_ipc_backend.py <path to unzipped recording>`
- PsychoPy experiment: [eyetracking.psyexp.zip](https://github.com/psychopy/psychopy/files/7059782/eyetracking.psyexp.zip)
- Minimal Capture recording for replay: [minimal_recording.zip](https://github.com/psychopy/psychopy/files/7059787/minimal_recording.zip)

papr:
This is the (partial) output of the print2err output (https://github.com/psychopy/psychopy/pull/4216/commits/84a7eb0207cbbf01fedfd4cc1f2eb9880ccf6ec6): https://pastebin.com/05Gy7LJW

Given that there are no "Error saving event[s]", I am wondering if these samples are reaching the iohub storage class at all...

isolver:
@papr, thanks for creating the simulator, that should help me track down what is going on in the pupil core implementation that is causing iohub to not save hdf5 sample events. I may not have time to look closely at this until next week. 

> Given that there are no "Error saving event[s]", I am wondering if these samples are reaching the iohub storage class at all...

Agreed, I must be missing something...

Thanks again

isolver:
OK, the reason why samples were not being saved to the hdf5 is because pupil core is currently always returning False for the device `isReportingEvents()` method. You can fix this by changing your `setRecordingState` implementation to return:

```
        return EyeTrackerDevice.enableEventReporting(self, self._actively_recording)
```

instead of `return self._actively_recording` as it does now.

With that change only, I get the expected error about status being a string:

```
  File "D:\DEV\my-code\psychopy\psychopy\iohub\datastore\__init__.py", line 381, in _handleEvent
    np_array = np.array([tuple(event), ], dtype=eventClass.NUMPY_DTYPE)
ValueError: invalid literal for int() with base 10: 'pye3d 0.1.1 real-time --- pye3d 0.1.1 real-time'
```

If I then change the pupil core `parse_data.py` functions to return 0 for the status,the hdf5 file is created with binocular samples in it. 

As a general iohub functionality note, it does seem inconsistent that streaming of events is not impacted by the state of `device.isReportingEvents()` but saving to the hdf5 file silently does nothing if it is False. I should improve on this.

Thanks again for creating the simulated link to your tracker; made debugging the situation much easier.

isolver:
I noticed in the `pupil_labs_debug.py` script, you call both:

```
tracker.enableEventReporting(True)
```

and

```
# Check for and print any eye tracker events received...
tracker.setRecordingState(True)
```

Only one is required. The two methods are expected to do the same thing (we should probably deprecate one of them). The current demo's and Builder generated code all use `setRecordingState`.

```
tracker.setRecordingState(True)
```

Also you can enable saving of hdf5 files for the `pupil_labs_debug.py` by changing the `launchHubServer` call to include an experiment code and optional session code, which tells the function you want an iohub experiment results file (hdf5):

```
io = launchHubServer(experiment_code='plabs_test', **iohub_config)
``` 

An option to use pupil_core should be added to the existing Coder demo's as well at some point (atleast simple.py and gcCursor I think).

papr:
@isolver Thank you very much, the recording works now. The next step for us will be to confirm the temporal accuracy of the recording. We will run a simple PLR experiment where we compare the stimulus onset timestamps with the recorded pupil diameter data. If you have further idea on how to validate temporal correctness please let us know.

We have two further questions and some feedback recording the recording state/event reporting functions.

1. How to minimize the frontend PsychoPy window from the ioHub backend?
In order to build initial support for our own calibration, we would like to minimize the PsychoPy (frontend) window and let our external software open a fullscreen window. Once the calibration was successful or failed, we would like to reopen the PsychoPy window. What would be the best way to accomplish this? From our understanding, the calibration is handled in the ioHhub backend and it is not clear to us how one gets control over the frontend window displaying the actual experiment.

2. Cleanup on experiment abort?
Currently, we have some cleanup functions that are called as part of the `EyeTracker.__del__()` method. But we have noticed that these functions are not called when the experiment is aborted. From my understanding, there is no guarantee for `__del__` functions to be called. Therefore, we would like to implement an explicit cleanup callback. Does PsychoPy support/implement something like that?

3. Feedback event reporting + recording state
From our perspective, it makes perfect sense to have these two functions. But they should be conceptually different:
    - event reporting
        - A flag that indicates the ioHub server if it should be processing device events (or even call `_poll()`).
        - Alternatively, this flag could indicate if the events need to be pushed to the frontend. I am not sure about the current implemention. If it pushes all events to the frontend by default, then this flag might make sense. If events are only forwarded if a proxy device is created in the frontend, then it might not be of use.
    - recording state
        - A flag indicating if the events should be written to the hdf5 file. With the current implementation, the flag would provide more fine grained control over which events are stored to disk.
        - Alternatively, this function could be responsible for starting/stopping recordings in the external eye tracking software.

isolver:
@papr, the time stamp check you suggest sounds like a reasonable sanity test not requiring extra test equipment.    

 > 1. How to minimize the frontend PsychoPy window from the ioHub backend?

The short answer is that from the iohub process you can't. To minimize the experiment window it needs to be done from the experiment process. I've found that this is only actually necessary on Windows: macOS and Linux seem OK with openning a second full screen window on top of another and then closing it properly. When using Builder generated code, the extra code needed to minimize the psychopy window and then restore it is added automatically. In a Coder script you can:

```
from psychopy.iohub.util import hideWindow, showWindow
```

and use the functions like, where `win` is the psychopy experiment Window:

```
# Minimize the PsychoPy window if needed
hideWindow(win)

# Display calibration gfx window and run calibration.
result = tracker.runSetupProcedure()
print("Calibration returned: ", result)

# Maximize the PsychoPy window if needed
showWindow(win)
```

> 2. Cleanup on experiment abort?

You can stop the iohub server by using `io.quit()` or `io.shutdown()`, where `io` is the connection to the iohub server. This results in each iohub devices `_close()` method getting called before the server is stopped.

However, I do not think this is not being done when a Builder experiment is aborted. This is something we should fix / improve.  When a Builder experiment is aborted, I think psychopy `core.quit()` is called. `core.quit()` does not however tell iohub to shutdown like it probably should. 

Is this a big issue for your implementation or is it something you can work around for the short term atleast?

papr:
@isolver Thanks for your input. We will give the suggestions in 1) a try and report how well it works for us. Re 2) thanks for mentioning `_close()`. We will move our cleanup into this function. I will also have a look at possibilities to have that executed even if `core.quit()` is called without changing too much of the code.

isolver:
@papr, I can fix `core.quit()` not calling iohub `quit()` next week. I think something like the following should be added to `core.quit()`:

```
from psychopy.iohub.client import ioHubConnection

if ioHubConnection.ACTIVE_CONNECTION:
    ioHubConnection.ACTIVE_CONNECTION.quit()
```

papr:
We will give this a try locally and if this works we can prepare a PR 👍🏻

papr:
Works as intended when using the simulator 👍🏻 https://github.com/psychopy/psychopy/pull/4258

papr:
Temporal correctness sanity check seems to have worked well.

[temporal-correctness.zip](https://github.com/psychopy/psychopy/files/7086079/temporal-correctness.zip) includes
- experiment code (2 trials with two sequences: 5 second black background + 5 second white background)
- recorded PsychoPy data
- Jupyter Notebook for the generation of the graph below

I have not quantified the delay between background color change and the actual PLR. Some delay is expected.

pupil_size_during_black_white_stimulus

I have also implemented a "pupillometry only" mode that does not require calibration or surface/aoi setups. Makes the collection of this kind of data much easier.

peircej (changes requested):
One tiny thing - the ioHubConnection class is only actually needed in the core.quit method and I'd prefer that it then only gets imported in that function. The reasoning here is just that it reduces the amount of stuff being loaded and visible int he core submodule.

Otherwise, well done and thanks for all your work on this! 👏

psychopy/demos/coder/iohub/eyetracking/pupil_labs_debug.py
```diff
@@ -0,0 +1,48 @@
...
+tracker = io.devices.tracker
+
+# run eyetracker calibration
+# r = tracker.runSetupProcedure()
+
+tracker.setConnectionState(True)
+tracker.enableEventReporting(True)
+
+# Check for and print any eye tracker events received...
+tracker.setRecordingState(True)
```
isolver: `tracker.enableEventReporting(True)` and `tracker.setRecordingState(True)` do the same thing, so only one is needed. Since `tracker.setRecordingState(False)` is used later in the script, maybe remove the `tracker.enableEventReporting(True)`  call.

isolver:
@TEParsons, do you mind reviewing the `psychopy/experiment/*` related changes in this PR when you have a chance? Thank you

psychopy/iohub/devices/eyetracker/hw/pupil_labs/pupil_core/pupil_remote.py
```diff
@@ -0,0 +1,139 @@
...
+# Copyright (C) 2012-2020 iSolver Software Solutions (C) 2021 Open Science Tools Ltd.
+# Distributed under the terms of the GNU General Public License (GPL).
+import abc
+import contextlib
+import logging
+import time
+import weakref
+
+import msgpack
+import zmq
```
isolver: iohub uses `gevent`, so can this be switched to:
```
import zmq.green as zmq
```
?

peircej:
Task linked: [CU-1k15fxq Review Pupil Core eyetracker support PR](https://app.clickup.com/t/1k15fxq)

psychopy/experiment/components/apriltag/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+        """Context manager which restores image params on yield"""
+
+        # Restore the image UI config params
+        for key, param in self.__hidden_params.items():
+            self.params[key] = param
+
+        try:
+            # Validate and update the image path param
+            image_path = self._image_path
+            assert path.isfile(image_path), image_path
```
TEParsons: It's generally better not to use `assert` in the compile stage as it will appear as an unhandled error - if the user inserts an invalid value, the ImageStim handles it when the experiment runs

psychopy/experiment/components/apriltag/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+
+class AprilTagComponent(ImageComponent):
+    """An event class for presenting image-based AprilTag stimuli"""
+
+    categories = ['Stimuli']
+    targets = ['PsychoPy', 'PsychoJS']
+    iconFile = Path(__file__).parent / 'apriltag.png'
+    tooltip = _translate('AprilTag: present AprilTag image')
+
+    _apriltag_images_dir = Path(__file__).parent / 'apriltag_images'
```
TEParsons: It would be better if this was visually distinct from Image components - maybe a Pupil Core logo to the bottom right of the image symbol? Similar to how the Eyetracking components each have a different icon next to the eye. Happy to send over the raw vector files for any icons you need, or I could whip something up for you if you'd prefer.

TEParsons: Should there be a folder under experiment->components->apriltag? There's nothing in the repo so I just get "could not find file" - does it need adding to git or is it something users would need to manually add?

psychopy/experiment/components/apriltag/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+
+from contextlib import contextmanager
+from os import path
+from pathlib import Path
+from psychopy.experiment.components import image
+from psychopy.experiment.components.image import ImageComponent, Param, getInitVals
+from psychopy.localization import _translate
+
+
+class AprilTagComponent(ImageComponent):
```
TEParsons: I'm not sure we need a component for this - if I'm interpreting it correct, all this does is assemble the path of a specific image file from a set, then presents it in an image component? Adding a component would put it in the components panel for everyone - I'm not sure this is a good idea for such a specific use case.

Instead, why not add a function to the library which is something like this:
```suggestion
def getAprilTagImage(family, id, root=Path(__file__).parent):
    """
    Get the path to an AprilTag image from its family and id.
    """
    # Map family tags to image prefices
    _apriltag_family_name_to_image_prefix = {
        "tag16h5": "tag16_05_",
        "tag25h9": "tag25_09_",
        "tag36h11": "tag36_11_",
        "tagCircle21h7": "tag21_07_",
        "tagCircle49h12": "tag49_12_",
        "tagCustom48h12": "tag48_12_",
        "tagStandard41h12": "tag41_12_",
        "tagStandard52h13": "tag52_13_",
    }
    # Make sure we have a valid tag name
    assert family in _apriltag_family_name_to_image_prefix, _translate(f"Invalid family name, should be one of: {list(_apriltag_family_name_to_image_prefix)}")
    # Make sure root is a path
    root = Path(root)
    # Construct path to image
    return str(
        root + _apriltag_family_name_to_image_prefix[family] + str(id).zfill(5)
    )
```

You could put it in its own folder (maybe within plugins? or maybe in the iohub sub-module?) alongside the images, then just instruct users to set their Image components' values to:
```
$plugins.apriltag.getAprilTagImage(desiredFamily, desiredId)
```
(or wherever the function ends up being)

That way we don't clutter up the components panel with hyper-specific components, while retaining an easy way to use the AprilTag stimuli fresh out of the box.

TEParsons (dismissed):
The Settings stuff all looks great, but I don't think the component is a good idea - it would be better to just have a function which constructs the image paths and instruct users to use that function in a regular Image component.

psychopy/demos/coder/iohub/eyetracking/pupil_labs_debug.py
papr: In our implementation, they do not do the same but `setRecordingState` also calls `enableEventReporting`. This is why I agree with removing `enableEventReporting`.

papr:
Hi everyone,

As discussed, I have implemented `runSetupProcedure` for the Pupil Core integration. It starts the calibration and waits for either success, failure, or user abort.

Unfortunately, I was not able to test the success case as I am developing on a remote machine right now. It would be great if you could give this a try.

We do not have a demo using a PsychoPy window yet but it is likely that it would need the additional code suggested [here](https://github.com/psychopy/psychopy/pull/4216/#issuecomment-908272085).

Otherwise, I have merged dev into the integration branch and have reapplied the deferred import changes in core.py. I did not make any changes to the AprilTag code as I do not know how one would integrate that as a flow component.

Thanks again for helping us with the integration! Please let me know if I can help with anything else.

Best regards,
Pablo

peircej:
Hi there, we're keen to pull this in for the new release (feature freeze was actually the other day but we can be a bit flexible with features that don't touch other code). We want to have a period of testing though before the 22.1.0 release at the end of January though. Maybe at this point we could bring in the bulk of the work and save the AprilTags for the following feature release (22.2.0) which would be in the summer?

papr:
Hi, sounds good to me. I will remove the Apriltag changes tomorrow morning.

papr:
Dear PsychoPy team,

I realize that my recent changes come very late and I apologize for that. Somehow, I lost track of my todo of removing the Apriltags. I hope I am not too late for the release, though.

I tested the branch as much as home-office possible. I encountered an issue with how we were generating the ioHub settings when compiling the experiment. I was able to fix it.

I also noticed that the current `dev` branch lists `None` twice as a selectable eye tracking device in the settings. But this seems to be unrelated to the changes of this PR.

Please let me know if there is anything else for me to help with.

isolver:
@papr, your timing is perfect. I'll give these changes a try next week and let you know if I run into any issues. 

> I also noticed that the current dev branch lists None twice as a selectable eye tracking device in the settings. But this seems to be unrelated to the changes of this PR.

I noticed that as well this week; @TEParsons, any ideas?

Thanks again

papr:
@isolver Thank you very much! 🙏🏻 I will keep my schedule/todo list for next week as free as possible such that I can fix anything that might come short-notice.

isolver:
Hi @papr, I can run the Builder Feature Demos->eyetracking project with pupil-core. It runs the (external) calibration and then starts the experiment. No sample data was saved to the hdf5 file for me though. Do you have the same issue? I do not see any errors reported. Thank you

papr:
@isolver Hey, so far so good I guess! Did the experiment work as expected otherwise, i.e. did you get realtime gaze to interact with the experiment stimulus?

If not, I would like to confirm that the setup:
1. Added apriltag markers around the screen (no duplicated marker ids/pattern)
2. Started Pupil Capture
3. Enabled the surface tracker
4. Added a surface based on the markers from step 1
5. Aligned the surface corners with the screen corners as good as possible
6. Renamed the surface to the `surface_name` of the eye tracking project settings (default: `psychopy_iohub_surface`)
7. Calibration was successful

isolver:
@papr, I missed / did not know aboyut steps 1 and 3. ;( I'll try to get it properly setup first thing tomrrow morning and get back to you. thanks again

papr:
Yeah, we should be adding documentation. I will try to work on that tomorrow morning.

In the meantime, you can record data with the `pupillometry_only` setting set to `True` (see eye tracking runtime settings)? With this configuration, no calibration or marker setup is necessary.

isolver:
> Yeah, we should be adding documentation. 

It would be good to have a page for pupil core to go with [the other supported trackers](https://psychopy.org/api/iohub/device/eyetracker.html). I still need to do a doc update for iohub in general, the focus on end users using / needing to know about .yaml files is no longer really the case, for all the coder demo's python dicts are now used for device configuration.

It would be also be great to update the psychopy/demos/coder/iohub/eyetracking/simple.py Coder demo and add pupil-core support.

isolver:
> In the meantime, you can record data with the pupillometry_only setting set to True? With this configuration, no calibration or marker setup is necessary.

With pupillometry_only setting to True I see Monocular Samples from both eyes, but no Binocular Samples. Is this expected? In case it mattters, my camera setup was probably not ideal for the test.  

I have not setup the april tags yet so have not tested the calibrated mode.

isolver:
@papr, is this the full process I need to follow to [setup the physical april tags](https://docs.pupil-labs.com/core/software/pupil-capture/#surface-tracking)? 

[This doc suggests simple screen calibration should be possible without april tags](https://docs.pupil-labs.com/core/software/pupil-capture/#screen-marker-calibration-choreography), but I guess we want to require a different method by default?

thanks again

papr:
> It would be good to have a page for pupil core to go with the other supported trackers. I still need to do a doc update for iohub in general, the focus on end users using / needing to know about .yaml files is no longer really the case, for all the coder demo's python dicts are now used for device configuration.

Yes, this is what we are working towards. We will be adding high level docs about Core and the setup process, too. We will add the technical docs, too, for now to be consistent with the remaining documentation.

We would also like to add (2) images. How should we be adding them?
1. Hosting them ourselves and linking to them
2. Adding them to the git repository (if yes, should git lfs be used?) 

> With pupillometry_only setting to True I see Monocular Samples from both eyes, but no Binocular Samples

Yes, this is expected since the combination of both streams is part of the gaze mapping in Pupil Capture.

> is this the full process I need to follow to setup the physical april tags?

Yes, it is. We will be more explicit about it in our psychopy docs and add a reference image.

> This doc suggests simple screen calibration should be possible without april tags, but I guess we want to require a different method by default?

Yes, while Capture's calibration does not require apriltags, they are required to map gaze from scene camera coordinates to screen coordinates.

isolver:
@papr, the marker setup / surface creation & editing is really slick and robust - very cool. I created a surface called 'psychopy_screen'in the Pupil Capture app, changed the surface name in the psychopy pupil-core settings to match, and everything seems to work. I got reasonable gaze positioning during the experiment and both mono and binoc samples saved to the file. 

Nice work.

papr:
@isolver This is great to hear! How do you want us to proceed with the documentation? How easy would it be for us to change the docs (a) after merging this PR and (b) after the PsychoPy release?

isolver (approved):

isolver:
> How easy would it be for us to change the docs (a) after merging this PR and (b) after the PsychoPy release?

I think online docs can be updated much more easily, not just when we do a binary release, but @peircej can confirm. 

I think we can pull in this PR today or tomorrow if you want @papr, you can always submit another PR with doc updates when you have time. thoughts?

papr:
I would be very happy to get this merged! :) I would not be able to submit a finalized version of the documentation today. But if we could get the documentation merged in the coming days, that would be great.

peircej (approved):

papr:
Thank you very much for your time to review our integration! We will follow up with a dedicated documentation PR in the coming days and are looking forward to the release otherwise! Let us know if there is anything else we can do :)

PR #4218, opened 2021-08-11
BF: Monitor Center hangs when calling "Plot gamma" after linearization

guillermoaguilar:
The bug occurs when trying to plot the gamma functions measured *after* calibration / linearization. At that moment the Monitor Center hangs with 

```
  File "/home/guille/anaconda3/envs/psychopy/lib/python3.9/site-packages/psychopy/monitors/MonitorCenter.py", line 1067, in plotGamma
    if lumsPost != None:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

This bug also relates to the discussion in:
https://discourse.psychopy.org/t/monitor-calibration-using-colorcal-ii-mk-questions/16364/4

That bug has been already fixed (line 1033). This is now the same bug but for the variable `lumsPost` instead of `lumsPre` (line 1067)

### DIFF
--- a/psychopy/monitors/MonitorCenter.py
+++ b/psychopy/monitors/MonitorCenter.py
@@ -1064,7 +1064,7 @@ class MainFrame(wx.Frame):
 
             lumsPost = self.currentMon.getLumsPost()
             levelsPost = self.currentMon.getLevelsPost()
-        if lumsPost != None:
+        if lumsPost.any() != None:
             for gun in range(4):  # includes lum,r,g,b
                 lums = lumsPost[gun, :]
                 gamma = gammaGrid[gun, 2]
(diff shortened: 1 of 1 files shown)


peircej:
Actually that isn't quite the right fix. On this occasion what we need is to check whether the entire thing is None, not whether its elements are None. That used to work fine but numpy changed its policy for handling it. What you want is:
```
if lumsPost is not None:
```
Do you want to update (so you still get credit for chasing down the issue 😄 )?

guillermoaguilar:
You're right, I didn't think it thoroughly. With my changes the call to Plot Gamma actually crashed for the case `lumsPost` *is* `None`, which is actually the moment _after_ being measured the gamma functions once but _before_ having them checked.

Now I tested the `Plot Gamma` before, during and after measuring the gamma functions, it works fine!

peircej:
Cool, thanks for the contribution!

PR #4228, opened 2021-08-16
ENH: Add util.randchoice to JS transpiler

TEParsons:
to work with: https://github.com/psychopy/psychojs/pull/457

### DIFF
--- a/psychopy/experiment/py2js_transpiler.py
+++ b/psychopy/experiment/py2js_transpiler.py
@@ -76,7 +76,7 @@ class pythonTransformer(ast.NodeTransformer):
 
     # operation from the math python module or builtin operations that are available
     # in util/Util.js:
-    utilOperations = ['sum', 'average', 'randint', 'range', 'sort', 'shuffle']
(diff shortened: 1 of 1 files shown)


tpronk:
Hi! Cool dual Python JS commit :). Got one suggestion: perhaps bypass `shuffle` and only use the random generator to index the array? Would be a bit more efficient if you've got big arrays, since `shuffle` iterates over every element, while we only need one.

TEParsons:
Could do yes :) I can't imagine it making a huge difference as `randchoice` tends to be a convenience function that people use on relatively small lists, but certainly no reason not to make it as efficient as it can be!

tpronk:
Coolio! One last request if I may. I see that `random.choice` throws an error if you pass it an empty array (i.e., `random.choice([])`). Let's have the JS version? throw an error too? PsychoJS errors add quite a lot of context info, see [this line](https://github.com/psychopy/psychojs/blob/569432b5c53ce8236b8dfac99d293ddfba6a9a4e/src/util/Util.js#L866) for an example.

Issue #3734, opened 2021-04-15
Backslash in ImageStim image path causes unknown resource error


JLC827:
If backslashes are used instead of forward slashes in image resource paths PsychoPy reports an "unknown resource" error. Forward slashes are used for file paths on Unix, but Windows allows forward and backslashes interchangeably, so it's not unreasonable for a user to not notice this difference. I'm not sure whether to best way to solve this problem would be to allow backslashes for file paths, or to report an error to the user if a backslash is present in a file path.

peircej:
We really need to get people using forward slashes so that if they want to run online it still works too. The question is whether to automatically/silently replace with forward slashes or to warn user verbally

PR #4229, opened 2021-08-17
BF: Don't sanitise pathlike str params if canBePath=False

### LINKED ISSUE #3476: BUG: \n in f string inserts new line in JS code
Using an fstring in a textstim e.g.

$f"I would like a \n line break here"

Will result in a linebreak in the exported .js code.

### LINKED ISSUE #4259: Dollar sign turns code into text
#https://discourse.psychopy.org/t/bug-dollar-sign-turns-code-into-text/24659

In PsychoPy 2021.2.3 (using Javascript):

The dollar sign ($) as code indicator in any TextBox’s Text property appears to be foiled by another dollar sign.

So this works: `$Math.E+'%'` generates `2.72%` as expected.
But this does not: `$Math.E+'$'` generates `Math.E+'$'` instead of `2.72$`.
This also behaves strangely: `$Math.E+'\$'` generates `2.72\$` instead of `2.72$`.

Screenshot from 2021-08-23 18-43-01

TEParsons:


TEParsons:
Fixes #3734

PR #4232, opened 2021-08-18
BF: Colors as a list should be code, not str

TEParsons:


psychopy/experiment/params.py
```diff
@@ -235,6 +235,15 @@ def __str__(self):
                 return valJS
             else:
                 return val
+        elif self.valType == 'color':
+            valid, val = self.dollarSyntax()
```
mdcutone: Throw out unused `valid` like this: `_, val = self.dollarSyntax()` to keep LGTM happy.

PR #4234, opened 2021-08-18
FF: Fix closing ioDevice dict for eyetracker in SettingsComponent

romanroibu:


TEParsons (approved):
So if I'm reading this right: Rather than closing each dict in its if statement, add code to close it after, regardless of what's chosen? Seems reasonable - given there'll always be a dict to close.

PR #4236, opened 2021-08-19
EN: Log if a component fails to load for any reason

### LINKED ISSUE #2396: Exception during component registration seems to silently get swallowed
I just fetched `master` and started Builder, and the Pump Component seems to have disappeared from the Components panel. It ought to appear under `Custom`.

mdcutone:
Fixes #2396 by logging an error when a component fails to load.

### DIFF
--- a/psychopy/experiment/components/__init__.py
+++ b/psychopy/experiment/components/__init__.py
@@ -159,7 +159,7 @@ def getComponents(folder=None, fetchIcons=True):
         except ImportError:
             logging.error(
                 'Failed to load component package `{}`. Does it have a '
-                '`__init__.py`.`'.format(cmpfile))
+                '`__init__.py`?'.format(cmpfile))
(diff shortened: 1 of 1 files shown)


psychopy/experiment/components/__init__.py
```diff
@@ -157,7 +157,11 @@ def getComponents(folder=None, fetchIcons=True):
         try:
             module = import_module(explicit_rel_path, package=pkg)
         except ImportError:
+            logging.error(
+                'Failed to load component package `{}`. Does it have a '
+                '`__init__.py`.`'.format(cmpfile))
```
peircej: Really minor edit, but we've got non-matching ` here and no question mark. I think this line should be:
```
'`__init__.py`.?'.format(cmpfile))
```

mdcutone: Okay I'll fix it.

Issue #2479, opened 2019-05-24
Escape key only works in pyglet backend


mirh:
I mean, the toggle in the basic tab of experiment settings.

peircej:
You could maybe explore and see what is causing the issue. The source code is all there for you.

mdcutone:
Are these features missing from the GLFW backend?

mirh:
Yes

PR #4238, opened 2021-08-19
NF: Get `setFullScr` working correctly for glfw and pyglet backends

mdcutone:
This PR implements for `seFullScr` for the Pyglet and GLFW backends. Currently that method is non-existent for GLFW and broken for Pyglet. 

Resolves part of issue #2479.


peircej (changes requested):
Could we add a usage of this to the test suite? Even if we don't include a screenshot/grab it would be good to flush the code

PR #4244, opened 2021-08-23
NF: Layout classes

### LINKED ISSUE #1296: Add ___2norm and norm2___ conversion functions to monitorunittools

### LINKED ISSUE #2381: TextStim.verticesPix throws TypeError: unsupported operands type(s) for *: 'attributeSetter' and 'float'
It seems that there is a problem with accessing the `verticesPix` attribute of a `TextStim` object.

```python
from psychopy import visual
win = visual.Window()
text = visual.TextStim(win, text="Hello, World!")
print(text.verticesPix)

  File "C:\Anaconda3\envs\psypy3\lib\site-packages\psychopy\visual\basevisual.py", line 477, in verticesPix
    self._updateVertices()
  File "C:\Anaconda3\envs\psypy3\lib\site-packages\psychopy\visual\basevisual.py", line 516, in _updateVertices
    verts = numpy.dot(self.size * verts * flip, self._rotationMatrix)
TypeError: unsupported operand type(s) for *: 'attributeSetter' and 'float'
```
It seems that the `.size` attribute isn't resolved correctly, something to do with how it is wrapped in the `attributeSetter` class.

If you go ahead and draw + flip the text, the `.verticesPix` attribute up and vanishes completely:
```python
from psychopy import visual
win = visual.Window()
text = visual.TextStim(win, text="Hello, World!")
text.draw()
win.flip()
print(text.verticesPix)

  File "C:\Anaconda3\envs\psypy3\lib\site-packages\psychopy\visual\basevisual.py", line 478, in verticesPix
    return self.__dict__['verticesPix']
KeyError: 'verticesPix'
```
I'm thinking something related to inheritance is going awry.

All this raises a larger question: am I doing it wrong? My goal is to find out where boundaries of the text are on the screen. I know about the `.boundingBox` property, but the API docs give me the impression that this is just the width and height of the text in pixels. Is this impression correct? And is there a correct and consistent way to find boundaries that works for all stimuli types?

TEParsons:
Adds classes:

Vector: An Nx2 or Nx3 array which knows its values in every unit space and can be accessed via .pix, .norm, .height, etc.
Size: Identical to Vector, different name for convenience
Position: Identical to Vector, different name for convenience
Vertices: numpy.ndarray with additional methods to render vertices relative to either a sized object or size & position values
and implements these classes in psychopy.visual objects via the BaseVisual class methods


peircej:
Task linked: [CU-ktb1ue Textbox stretching characters in norm](https://app.clickup.com/t/ktb1ue)

mdcutone (approved):
Looks okay to me.

PR #4264, opened 2021-09-06
BF: Slider start val shouldn't count as a response

TEParsons:
Counting the start val as a response meant that, when start val was specified and slider had "force end routine", the routine would end immediately.


peircej:
Task linked: [CU-1dv9702 Slider bug: crashes when value entered in starting val](https://app.clickup.com/t/1dv9702)

peircej:
I think PR #4278 is the better way to handle this. Maybe just check that it does actually fix the symptom but certainly it has the approach that I prefer (to use the existing distinction between markerPos and rating)

PR #4266, opened 2021-09-06
BF: Texture resolution needs to be code, not num, so it stays an int

TEParsons:


peircej:
Task linked: [CU-1e0xyex Bug with dropdown options for EnvelopeGrating](https://app.clickup.com/t/1e0xyex)

PR #4272, opened 2021-09-10
BF: Coder's "Find" fails when non-ASCII characters are included

hsogo:
Here attaches a sample file for the issue.
[min_sample.zip](https://github.com/psychopy/psychopy/files/7142138/min_sample.zip)

Please open min_sample.py in this Zip file with Coder and search for a string "target".  The first three "target" (Line 3, 6, 9) are properly found, but "target" after non-ASCII characters (Line 12) are found at wrong position. 
fig1

The cause of this issue is that `SetSelection()` doesn't recognize a non-ASCII multi-byte character as a single character.  For example, `'日'` is thee byte character in UTF-8 (`E697A5`) and treated as three characters by `SetSelection()`.  I tried to find a way to make `SetSelection()` recognize multi-byte characters, but that seemed quite hard.  So I tried to convert text to Bytes string before running `find()`.  This solution worked fine on my Japanese Windows10 and Ubuntu PCs.


mdcutone:
Thank you @hsogo! Kept hitting this issue intermittently and never figured out why.

PR #4278, opened 2021-09-17
BF: setting startValue for Slider should only change markerPos not rating

peircej:
Slider.rating should only be set when the participant *has made a rating*
It indicates a response. MarkerPos indicates the current location of the
marker before a rating has occurred


peircej:
Task linked: [CU-1eqrcvx Slider: if rating of 0 given slider.rating cannot be used to set the onset of something using conditional](https://app.clickup.com/t/1eqrcvx)

TEParsons (approved):

peircej:
Task linked: [CU-1dv9702 Slider bug: crashes when value entered in starting val](https://app.clickup.com/t/1dv9702)

PR #4279, opened 2021-09-21
BF: iohub eyelink start / stop record blocking events

isolver:
When eyelink start / stop record is called, it blocks for 0.5 seconds. This was causing iohub to miss any kb events that occurred during the 0.5 seconds because it is too long of a time period to service the callback generated by pyWinHook.

This PR makes the blocking call to eyelink start / stop record run in a seperate thread on iohub so other events can be serviced.


isolver:
@mdcutone If this can be pulled into Dev that would probably be better.

isolver:
closing. recreated new PR

PR #4287, opened 2021-09-28
WIP: iohub and event.getKeys should support same key mappings

isolver:
ENH: Add 'use_mapping' setting to iohub keyboard device. 'original' and 'psychopy' are supported, with psychopy being the default.
ENH: When 'use_mapping'=='psychopy' match iohub and event.getKeys() key mappings. Mainly numpad and some punctuation keys have been updated. For example, space key is ' ' in original mapping and 'space' in psychopy mapping.

NOTE: This means by default old iohub experiments must be updated to expect new key mappings or keyboard device must be set to use original mappings when moving to this release.

### DIFF
--- a/psychopy/demos/coder/iohub/keyboard.py
+++ b/psychopy/demos/coder/iohub/keyboard.py
@@ -10,9 +10,12 @@ from __future__ import absolute_import, division, print_function
 
 from builtins import str
 from psychopy import core, visual, event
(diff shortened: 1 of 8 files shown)


isolver:
when ready, this should be pulled into Dev, not Release.

isolver:
closing; will recreate

PR #4290, opened 2021-09-29
RF: Update Builder eye tracking demos and ROI.py

isolver:
RF: Update Builder eye tracking demos to use `eyetracker.getPos()` for GC stim instead of ETRecord component.
RF: ROI.py: Handle case of tracker returning error code (number) from getPos(). Returned value must be list or tuple to be considered valid.
RF: ROI.py: rename ROI `.tracker` attribute to be `.device` since an ROI with `device=None` creates an `event.Mouse` class to use.

### DIFF
--- a/psychopy/visual/roi.py
+++ b/psychopy/visual/roi.py
@@ -11,15 +11,15 @@ class ROI(ShapeStim):
     Parameters
     ----------
     win : :class:`~psychopy.visual.Window`
-        Window which eyetracking input will be relative to. The stimulus instance will
+        Window which device position input will be relative to. The stimulus instance will
         allocate its required resources using that Windows context. In many
         cases, a stimulus instance cannot be drawn on different windows
         unless those windows share the same OpenGL context, which permits
         resources to be shared between them.
     name : str
         Optional name of the ROI for logging.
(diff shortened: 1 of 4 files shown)


psychopy/visual/roi.py
```diff
@@ -85,19 +85,19 @@ def numLooks(self):
...
-        elif hasattr(self.tracker, "getPosition"):
-            pos = self.tracker.getPosition()
-        else:
-            # If there's no position functions, assume False
-            return False
-        if pos is None:
-            # If there's no eye data (e.g. during a blink) assume False
+        try:
+            # Get current device position
+            pos = self.device.getPos()
```
TEParsons, line 90: Is there any instance where we'll need the `if hasattr(self.device, "getPosition")` check or was that just because the device called it `getPosition`?

TEParsons (approved):
Looks good! I can't remember why I added the check for a `getPosition` method so it may well not be needed, in which case you're right to remove it.

psychopy/visual/roi.py
isolver: I checked and the eye tracker and mouse both have a getPos method already so i think we are good.

PR #4295, opened 2021-10-01
RF: Improve how subprocesses are spawned and managed

mdcutone:
This PR introduces the `psychopy.app.jobs.Job` class and some other changes which simplifies spawning sub-processes off the main UI process, mainly for running local scripts. Right now, the `Job` class has been applied to running local Python scripts from Runner. 

This is part of a larger project to improve parallelism in the UI.

### DIFF
--- a/psychopy/app/jobs.py
+++ b/psychopy/app/jobs.py
@@ -473,17 +473,35 @@ class Job:
     #         return None
     #
     #     return self._process.OutputStream
-    #
-    # @property
-    # def isInputAvailable(self):
-    #     """Check if there are bytes available to be read from the input stream
-    #     (`bool`).
-    #     """
-    #     if self._process is None:
-    #         return False
-    #
-    #     return self._process.IsInputAvailable()
-    #
+
+    @property
+    def isInputAvailable(self):
+        """Check if there are bytes available to be read from the input stream
+        (`bool`).
+        """
+        if self._process is None:
+            return False
+
+        return self._stdoutReader.isAvailable
+
+    def getInputData(self):
+        """Get any new data which has shown up on the input pipe (stdout of the
+        subprocess).
+
+        Returns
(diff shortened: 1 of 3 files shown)


TEParsons (changes requested):
Are there changes to how paths are handled? When I try to run any of my demos, they error out before even loading the .py file because one folder in the path is called "My Drive" so it misinterprets the path as "E://My"

mdcutone:
> 
> 
> Are there changes to how paths are handled? When I try to run any of my demos, they error out before even loading the .py file because one folder in the path is called "My Drive" so it misinterprets the path as "E://My"

I'll see if I can figure this out. Don't recall changing any path stuff, but I might have messed up handling them somehow.

peircej:
One path issue is that I was setting popen explicitly to use the file folder as the current working directory for running the study which doesn't seem to be done in this system?

mdcutone:
> One path issue is that I was setting popen explicitly to use the file folder as the current working directory for running the study which doesn't seem to be done in this system?

Might be due to spaces in the paths like Todd suggested. I just put quotes around those before passing it to the `Job` constructor.

TEParsons (approved):
Spaces in filename bug seems to be fixed, no longer erroring when I try to run stuff, now seems to be working fine!

peircej:
This looks good. Do you think a test to the test suite that launches a process by this? So we could at least flush the code in the test suite and catch typos? Ideally we'd create a test process that generates expected outputs and check we receive them but just making sure a process can be launched and completed seems useful

peircej (commented):
This looks good. Do you think a test to the test suite that launches a process by this? So we could at least flush the code in the test suite and catch typos? Ideally we'd create a test process that generates expected outputs and check we receive them but just making sure a process can be launched and completed seems useful

PR #4296, opened 2021-10-03
**DOC**: Showing a numpy array as an image with ImageStim

tomwallis:
This doc is being added as a ["Recipe"](https://psychopy.org/recipes/index.html) for showing an image on screen. Showing an image from a file (e.g. PNG) is straightforward, but showing an image from a numpy array has a couple of gotchas, and confuses a few people (see e.g. [here](https://discourse.psychopy.org/t/display-a-colored-image-with-pixels-stored-as-numpy-array/4494/7), [here](https://discourse.psychopy.org/t/displaying-images-stored-as-numpy-array/2886/10), [here](https://discourse.psychopy.org/t/imagestim-and-gratingstim-do-not-display-an-image-that-is-stored-as-numpy-array/15368) and [here](https://discourse.psychopy.org/t/grayscale-images-as-numpy-arrays-display-as-blue-for-values-of-exactly-1/17054)).

This doc starts to demo that by loading an image, converting it to a numpy array, then showing.

I'm using the image `PsychoPy2_screenshot.png` that's part of the repo as a demo. Happy for feedback whether there's a better demo image to use. I also use the [`img_as_float` function from `skimage`](https://scikit-image.org/docs/stable/api/skimage.util.html?highlight=img_as_float#skimage.util.img_as_float) since PsychoPy's internal [`image2array`](https://github.com/psychopy/psychopy/blob/6722d365aca0ea1f76c3b0d91f9930f8976f5951/psychopy/tools/imagetools.py#L41) function is comparatively limited.

Sorry about all the old junk merge commits; I'm not sure how to squash them.


docs/source/recipes/showImages.py
```diff
@@ -0,0 +1,77 @@
+from pathlib import Path
+
+import numpy as np
+from PIL import Image
+from psychopy import core, event, visual
+from skimage import img_as_float
```
mdcutone: @peircej I don't think we have `scikit` as a requirement. I think demos should work with the packages we ship with standalone. @tomwallis we may need to use builtin numpy and Pillow functions to convert the image to an array. Perhaps something like this ...

```
im = Image.open('myPic.png')  # Pillow
arr = numpy.array(im, order='C')  # convert to array, must be C contiguous for OpenGL 
arr /= 255.  # normalize
```

tomwallis: Thanks for the suggestion @mdcutone. I removed the dependency on `skimage` and added a few more docs. I also thought it would make sense to add some more information to the docstring of `ImageStim` itself.

PR #4298, opened 2021-10-06
Updated ROI 'look away' logic 

isolver:
BF: If ROI ends routine using 'look away', eyes must be outside of roi for 'Min. Look Time' seconds before routine ends. ROI 'look at' routine end logic is unchanged.
BF: MouseGaze .getPos() was returning 0,0 until first mouse event occurred after starting to record, now it gets current mouse position and uses that if no events have been received yet.


TEParsons (approved):
This is a change of behaviour, but one I think is more reasonable and in line with what users assume anyway

PR #4299, opened 2021-10-07
BF: Calculate slider horiz/vert by combination of ori and size

### LINKED ISSUE #3775: BUG: orientation parameter of slider does not function locally
Setting the orientation parameter of slider to 90 does not create a vertical slider in 2021.1.4.

TEParsons:
Uses trigonometry to work out the total width/height of a tight box around the Slider (its extent) and returns horiz according to the proportions of that box, rather than the Slider itself.

Reviewers:
@mdcutone to check that `extent` does what it says it's doing
@apitiot to confirm that this is (or can/should be) consistent with JS
@RebeccaHirst to confirm that this is intuitive to users


peircej:
Task linked: [CU-m3bhhv Slider should use *either* ori or size to set its orientation](https://app.clickup.com/t/m3bhhv)

PR #4303, opened 2021-10-11
BF: Builder crashes if custom categories are defined in custom compon…

hsogo:
…ents
When `componentsFolders` is set to load custom components and custom categories are defined in the custom components, Builder window crashes immediately after opening.  I confirmed that this issue happens in 2021.2.0 and later.
The value of `componentsFolders` must be passed to `getAllCategories()` to deal with custom categories.

By the way, newly added `getAllStandaloneRoutines()` doesn't have `folderList` option.  Does this mean that custom standalone routines will not be supported?

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2483,7 +2483,8 @@ class ComponentsPanel(scrolledpanel.ScrolledPanel):
         del self.components['SettingsComponent']
         self.routines = experiment.getAllStandaloneRoutines()
         # Get categories
-        self.categories = getAllCategories()
+        self.categories = getAllCategories(
+            self.app.prefs.builder['componentsFolders'])
         for name, rt in self.routines.items():
             for cat in rt.categories:
                 if cat not in self.categories:


mdcutone:
I'm not sure we wanted to remove custom component folders, perhaps the function gets the list of folders from the config file now?

hsogo:
I run bisect and found that the issue originated from this commit.
7d2e126c84e95d869144179fb96edccbe84d5dbc

Before this commit, `self.components`  had been generated by `getAllComponents(self.app.prefs.builder['componentsFolders'])`.
Then, `self.categories` had been generated by picking up categories used in `self.components`.
As a result, custom categories defined in the custom components had been included in `self.categories`.

On the other hand, after this commit, `self.categories` is generated by calling `getAllCategories()` without passing `self.app.prefs.builder['componentsFolders']`. So, categories defined in the custom components are not found.
Because `self.app.prefs.builder['componentsFolders']` is passed to `getAllComponents()` even after this commit, I don't think this commit intended to remove support for custom component folders,

peircej:
Indeed, I don't think there was a deliberate aim to remove support for custom components/categories. I strongly suspect that use case just got overlooked when the new function was added by @TEParsons 

I guess we need to add a test to the suite, to make sure that a custom component, with a custom category doesn't break the app in future iterations.

PR #4307, opened 2021-10-13
BF: Iohub wintab demo update

isolver:
BF: `coder/iohub/wintab` demo was not updated to pass PsychoPy window to `launchHubServer`, causing position info to be in wrong coord space.
RF: some auto pep-8 format cleanup


TEParsons (approved):
Without a wintab device I can't test this, but the code looks fine as far as I can see :)

peircej:
Was ScreenPositionValidation not needed any more?

isolver:
No, that was left over code from when I ported the demo from OpenHandWrite.

PR #4308, opened 2021-10-13
BF: Allow start/stop controls to be targeted by depends

TEParsons:
Adds attributes to target the full start/stop control objects, so these can be referred to in depends as "start" and "stop" as if they were params in and of themselves


isolver (commented):
Thanks for doing this so quickly. Looks good. Does line 423 need a `return`?:
` return all(ctrl.IsShown() for ctrl in self.ctrls.values())`

PR #4309, opened 2021-10-13
BF: Ensure that allowedKeys always compiles as a list

TEParsons:


peircej:
Task linked: [CU-1h26bkx List params not compiling properly](https://app.clickup.com/t/1h26bkx)

peircej:
I think the test needs to include an experiment file that currently fails

TEParsons:
@peircej I've implemented the more general solution which covers all list params now, so taking this off draft and marking as ready for review :)

PR #4315, opened 2021-10-15
BF / ENH: Fix .setData() and add U3 'register' param to Parallel out component

isolver:
BF: U3.setData() was not working
ENH: When U3 is selected device for Parallel Out component, show a 'Register' param that lets user select whether U3 EIO or FIO should be used to write data byte to. Default is EIO. Register param is not visible when parallel port address or USB2TTL8 is selected.
DOC: Update labjacks.py and parallelout.rst docs

### DIFF
--- a/docs/source/builder/components/parallelout.rst
+++ b/docs/source/builder/components/parallelout.rst
@@ -3,7 +3,7 @@
 Parallel Port Out Component
 ---------------------------------
 
-This component allows you to send triggers to a parallel port or to a LabJack device.
+This component allows you to send triggers to a parallel port, USB2TTL8, or LabJack U3 device.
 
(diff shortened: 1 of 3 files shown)


isolver:
One outstanding thing that would be nice to clean up is the warning you get if using a U3 or USB2TTL8:

```
2.3572     WARNING     psychopy.parallel has been imported but no parallel port driver found. 
Install either inpout32, inpoutx64 or dlportio
```

I think this is because of line 44 in `parallelOut/__init__.py`:

```self.exp.requirePsychopyLibs(['parallel'])```

Is there a way to add / remove this requirment based on the Port Address selected in the component?

isolver:
Switched to draft until U3 fix is added to PR.

PR #4320, opened 2021-10-20
RF: Remove `None` color from colorpicker

mdcutone:
Removes the transparent color preset from the color picker. Since we don't allow users to set transparency by setting the alpha channel directly, it doesn't make sense to have transparent presets.


TEParsons (changes requested):
"None" needs to be a named colour so that setting the colour to None doesn't require a colorSpace, but I don't think it necessarily needs to be `(0, 0, 0)` - so could just replace it with None so that in the colour picker it outputs None

mdcutone:
Changed things to avoid modifying data in the colors module. Tests should pass now.

TEParsons (approved):

PR #4324, opened 2021-10-25
ENH: Visually represent stop values with no start value

TEParsons:
Useful for new workings of etRecord

PR #4330, opened 2021-10-29
BF: Interface fixes to et record

TEParsons:


peircej:
Task linked: [CU-1gb776t ENH: Builder ET Record Component](https://app.clickup.com/t/1gb776t)

isolver (changes requested):
We agreed at yesterday's standup that we would use start of 0 for Stop only so that the bar is drawn from start of routine to when recoring will stop. So I think that needs to be reverted back to `self.ctrls["startVal"].Value = "0"`

TEParsons:
> We agreed at yesterday's standup that we would use start of 0 for Stop only so that the bar is drawn from start of routine to when recoring will stop. So I think that needs to be reverted back to `self.ctrls["startVal"].Value = "0"`

I added some code in #4324 so that when we have a set stop & blank start, it draws the bar from just before 0 to indicate an open start. The question is whether to draw this when start is blank and stop is set as a duration... Do we draw it as if stop was a time and start was 0?

isolver:
> I added some code in #4324 so that when we have a set stop & blank start, it draws the bar from just before 0 to indicate an open start.

I don't think that is working as intended then, here I had it set to stop at .5 seconds, but bar is drawn to  4.x seconds.
ETrecord_stop_time

TEParsons:
> > I added some code in #4324 so that when we have a set stop & blank start, it draws the bar from just before 0 to indicate an open start.
> 
> I don't think that is working as intended then, here I had it set to stop at .5 seconds, but bar is drawn to 4.x seconds. ETrecord_stop_time

Ahh, I need to add the adjustment on to the duration... Will make that change now :)

isolver (approved):
OK, this makes sense. Thanks

peircej:
Task linked: [CU-1ja98yz If stop type is duration and start time is blank, treat as 0](https://app.clickup.com/t/1ja98yz)

PR #4336, opened 2021-11-01
BF: Only treat `var` as a variable declaration if it has a space after

TEParsons:
Means that components with `var` in their name are no longer compiled as blank, as they were mistaken for variable declarations previously


peircej:
Task linked: [CU-1fu7e8j using component that contains word "variable" (e.g. "variableFix") will induce JS syntax error](https://app.clickup.com/t/1fu7e8j)

PR #4345, opened 2021-11-05
RF: Split ColorMixin into Fore-/Fill-/Border- ColorMixin

TEParsons:
Looks like a bigger change than it is! There's a lot of movement but minimal new code added - just copy and pasting stuff in different places


peircej:
Task linked: [CU-1jvejpc Non-applicable color methods](https://app.clickup.com/t/1jvejpc)

PR #4348, opened 2021-11-08
NF: Callback assignement for window events

mdcutone:
This PR adds callbacks for window move and resize events. These are intended to notify iohub of these events. Some other changes were made to clean up the backend code a bit. 

Adding a callback is easy:

```
def moveCallback(win, x, y):  # callback function, gets window object and new X,Y coords
    print(x, y)

myWindow.backend.onMoveCallback = moveCallback  # called every time the window is moved, frequency depends on the backend
```

### DIFF
--- a/psychopy/visual/backends/_base.py
+++ b/psychopy/visual/backends/_base.py
@@ -25,9 +25,9 @@ class BaseBackend(ABC):
     """The backend abstract base class that defines all the core low-level
     functions required by a :class:`~psychopy.visual.Window` class.
 
-    Such functions as the ability to create an OpenGL context, process events,
-    and flip the window. Sub-classes of this function must implement the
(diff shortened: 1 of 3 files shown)


isolver:
@mdcutone, I think we need the callback to include the `win._hw_handle`. So something like:

```
self._onMoveCallback(self.win._hw_handle, posX, posY)
```

since the registered / assigned callback will not know which window moved without it.

Thanks

mdcutone:
> @mdcutone, I think we need the callback to include the `win._hw_handle`. So something like:
> 
> ```
> self._onMoveCallback(self.win._hw_handle, posX, posY)
> ```
> 
> since the registered / assigned callback will not know which window moved without it.
> 
> Thanks

D'oh, Forgot to push out those commits.

PR #4362, opened 2021-11-11
BF: Modification of parameter with default

DimitriPapadopoulos:
Fix series of the same LGTM.com error.

The default value of keyword arguments is mutable and is eventually mutated, changing the default value in future function calls. Perhaps it's currently not an issue in some cases, but it is a Python anti-pattern that may cause hard to detect problems in the future.

See for example [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/).


DimitriPapadopoulos:
Not sure about `RTBoxEnable` though. Is it OK for `RTBoxEnable` to accept a mutable argument that is later mutated? Should it perform a deep copy instead?

PR #4375, opened 2021-11-16
BF/TEST: Always respect dollar syntax in color params

TEParsons:


isolver (approved):
Fixes the error in the stroop and eyetracking Builder demo's; did not test others.

PR #4380, opened 2021-11-17
BF: Translate bools in a dictCtrl differently for PsychoJS

### LINKED ISSUE #4280: Transpiler: True in expInfo not translated to true in JS
If the user uses "True" in the expInfo to generate a checkbox in the startup that produced an error online because `True` is not translated to `true` in Javascript.

TEParsons:
Fixes #4280 and PsychoJS/464


peircej:
Task linked: [CU-1f934kb Use of boolean values in expInfo online](https://app.clickup.com/t/1f934kb)

PR #4387, opened 2021-11-22
ENH: Layout textbox with alignment and anchor

TEParsons:
Meaning unit conversion is handled in a stable way, fixing the character stretching bug in `norm` units


peircej:
Task linked: [CU-ktb1ue Textbox stretching characters in norm](https://app.clickup.com/t/ktb1ue)

peircej:
Task linked: [CU-1e0za0b Layout TextBox2 with text alignment](https://app.clickup.com/t/1e0za0b)

psychopy/visual/textbox2/textbox2.py
```diff
@@ -1262,27 +1256,37 @@ def vertices(self):
...
-        # lastChar = [bottLeft, topLeft, **bottRight**, **topRight**]
+        # Get vertices of caret based on characters and index
         ii = self.index
         if textbox.vertices.shape[0] == 0:
-            verts = textbox._getStartingVertices() / textbox._pixelScaling
-            verts[:,1] = verts[:,1]
-            verts[:,0] = verts[:,0] + float(textbox._anchorOffsetX)
+            # If there are no chars, put caret at start position (determined by alignment)
+            if textbox.alignment[1] == "bottom":
+                bottom = min(textbox.contentBox._vertices.pix[:, 1])
```
peircej, line 1264: Do we need to do a min() here? Aren't all the chars on this row going to have the same vertical position?

psychopy/visual/textbox2/textbox2.py
```diff
@@ -1262,27 +1256,37 @@ def vertices(self):
...
+            elif textbox.alignment[1] == "center":
+                bottom = (min(textbox.contentBox._vertices.pix[:, 1]) + max(textbox.contentBox._vertices.pix[:, 1]) - textbox.glFont.ascender - textbox.glFont.descender) / 2
```
peircej, line 1266: Do we need min() and max()? Won't the min() value always be the first char and max() always be the last?

peircej (commented):
At a first pass this looks good. I wonder if we need as many uses of min/max (which have to traverse an entire list to make their calculations) and it would still be worth running some timing tests to see whether the laying out of text has gotten slower but broadly this looks great

psychopy/visual/textbox2/textbox2.py
TEParsons: The vertices of `.contentBox` aren't character vertices, `.contentBox` is the inner rect (aka without padding), so its vertices are just the 4 corners of a rectangle. If it's reliable that the verts are in the same order I could just take one value, but I figured `min` would be safer, especially as there's only 4 values to search. Could also do `textbox.contentBox._pos.pix[1] - textbox.contentBox._size.pix[1]/2` but it seemed a bit wordy :P

isolver:
`psychopy/demos/coder/stimuli/compare_text_timing.py` can be used to compare TextStim and TextBox2 performance. Does not look like these changes impact draw time compared to current release branch.

Results from this branch:

```
-------Text Draw Duration Test---------

+ Draw Order: ['TextBox', 'TextBox2', 'TextStim']	
+ Text Stim Char Length:	 160
+ Text Change Flip Perc:	20.00
+ Total Flip Count:		 600
+ Test Duration (secs):		10.022
+ FPS:				59.869

+ Average Draw Call Durations (msec):

  Text Object		No Txt Change	Txt Change
  TextBox		0.837		2.126
  TextBox2		2.118		5.730
  TextStim		1.013		8.179

```

PR #4388, opened 2021-11-22
BF: Fix layout bug in TargetStim

TEParsons:


isolver (approved):

psychopy/visual/target.py
```diff
@@ -92,11 +92,11 @@ def scale(self, newScale):
 
     @property
     def radius(self):
-        return sum(self._size)/2
+        return sum(self.size)/2
 
     @radius.setter
     def radius(self, value):
-        self._size = (value*2, value*2)
+        setattr(self._size, self.units, (value*2, value*2))
```
isolver, line 99: The crash is fixed, but it does not seem like setting the radius works anymore / it does not change. Do you mind testing using the eyetracking Builder demo? Maybe setting the base `visual.Circle` radius is no longer working?

isolver (changes requested):
The crash is fixed, but it does not seem like setting the radius works anymore / it does not change. Do you mind testing using the eyetracking Builder demo?

psychopy/visual/target.py
TEParsons: I think the problem may be specific to calibration - the target stim itself seems to have a settable radius as `TestTarget.test_radius` is passing and if I make a minimal example it also works, e.g. this:
```
from psychopy import visual

win = visual.Window(units="height")
stim = visual.TargetStim(win, radius=0.5, innerRadius=0.25)
stim.radius = 0.3
stim.draw()
win.flip()
win.screenshot.save("C:\\Users\\Work\\Downloads\\test_radius.png")
```
makes this:
test_radius

TEParsons: Ah, I think I see the problem! It's in `visual.Polygon` - it treats its radius as being relative to its size

PR #4390, opened 2021-11-23
ENH: Adding call to end_experiment function

bollaking:
At the end of the experiment if there is an emotiv object that has the end_experiment function psychojs will call the function to enable clean handling of end of experiment in hosting site.

Note to maintainers this does involve a change the core code. Code checks for the presence of the function before calling it so should be safe

### DIFF
--- a/psychopy/experiment/components/emotiv_record/__init__.py
+++ b/psychopy/experiment/components/emotiv_record/__init__.py
@@ -69,3 +69,16 @@ class EmotivRecordingComponent(BaseComponent):  # or (VisualComponent)
         )
         buff.writeIndentedLines(code)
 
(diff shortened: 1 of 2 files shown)


peircej:
Thanks @bollaking 
It would be great if you could prefix the commit message rather than the Pull Request title e.g. with ENH/BF so that we know easily whether this should be ported to the current release stream (BF commits only) or should be heading for the next major release instead (anything that might possibly break stuff). The reason this is needed on the commit more than the Pull Request is that I search through the commit history for relevant commits on a terminal (the names of PRs don't show up there).
cheers, jon

bollaking:
@peircej Noted.

PR #4401, opened 2021-11-30
ENH: Layout Slider using layout classes

TEParsons:
- Fixes problems with laying out labels with vertical sliders - as we no longer need to guess and offset the label width, we can instead just use TextBox2 components with an anchor property.
- Create and layout sub-elements of Slider in different functions, meaning extant elements can be laid out again, so size/pos become settable on the fly


peircej:
Task linked: [CU-1n012p5 Make Slider size/pos settable with layout classes so that Form lays out correctly](https://app.clickup.com/t/1n012p5)

psychopy/layout.py
```diff
@@ -80,9 +80,11 @@ def validate(self, value, units):
         if (value == None).any() or numpy.isnan(value).any():
             win = Vector((1, 1), units="norm", win=self.win)
             if len(value.shape) == 1:
-                value[0] = getattr(win, units)[0]
-                value[1] = getattr(win, units)[1]
+                value[value == None] = getattr(win, units)[value == None]
```
mdcutone, line 83: Wasn't this changed in another PR? Might cause a merge conflict.

mdcutone (changes requested):
Bit of déjà vu for some of the commits here with PR #4407

TEParsons:
> Bit of déjà vu for some of the commits here with PR #4407

Yep, 4407 is based on this branch - couldn't fix Forms until Slider was working.

PR #4402, opened 2021-11-30
NF: Set focus on output console when experiment ends

mdcutone:
This PR automatically sets focus to an output window when the experiment ends. Uses the following rules to determine which window to use.

1. If the experiment was launched from runner, the runner window is focused.
2. If the experiment was lunched from coder, the coder output view gets focus.
3. If no output window is present, or coder was closed at some point while the experiment is running, open a runner window and focus on the dialog.

~Rule #3 could be changed to open/raise a runner window then set focus.~

### DIFF
--- a/psychopy/app/runner/scriptProcess.py
+++ b/psychopy/app/runner/scriptProcess.py
@@ -36,6 +36,24 @@ class ScriptProcess:
         self.app = app  # reference to the app
         self.scriptProcess = None  # reference to the `Job` object
         self._processEndTime = None  # time the process ends
+        self._focusOnExit = 'runner'
+
+    @property
+    def focusOnExit(self):
+        """Which output to focus on when the script exits (`str`)?
+        """
+        return self._focusOnExit
+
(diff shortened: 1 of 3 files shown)


psychopy/app/runner/scriptProcess.py
```diff
@@ -232,6 +254,32 @@ def _onTerminateCallback(self, pid, exitCode):
...
 
+        def _focusOnOutput(win):
+            """Subroutine to focus on a given output window."""
+            win.Show()
+            win.Raise()
+            win.Iconize(False)
+
+        # set focus to output window
+        if self.app is not None:
+            if self.focusOnExit == 'coder' and hasattr(self.app, 'coder'):
```
TEParsons, line 265: I'd quite like to take this further at some point - and have `.builder`, `.coder` and `.runner` as a `@property` which returns the window if open, or opens the window if not.

TEParsons (approved):

PR #4403, opened 2021-12-02
ENH: Builder saving messages to hdf5 and other related changes

isolver:
ENH: If iohub is running and saving a hdf5 file, Builder now saves visual stim start time messages to hdf5
ENH: Added `saveEventReport` function to `psychopy.iohub.datastore.util`
BF: P3 str / bytes issues in `psychopy.iohub.datastore.util`
ENH: Added iohub hdf5 reader to gcCursor demo that uses experiment conditions table
RF: Switched datastore coder demo to use new `saveEventReport` function
ENH: Added `cacheMessageEvents()` and `sendMessageEvents()` and to iohub connection class. `sendMessageEvents()` also sends any cached messages.


psychopy/experiment/components/_base.py
```diff
@@ -808,6 +809,29 @@ def writeFrameCodeJS(self, buff):
...
+        BaseComponent.writeRoutineEndCode(self, buff)
+        saveHdf5 = self.exp.settings.params['Save hdf5 file'].val
+        if saveHdf5:
+            name = self.params['name']
+            comp_onset_tag = "%s.started" % name
+            if 'saveStartStop' in self.params and self.params['saveStartStop'].val:
+                if self.params['syncScreenRefresh'].val:
+                        code = (
+                            f"if ioServer:\n"
+                            f"    ioServer.cacheMessageEvent('{comp_onset_tag}', sec_time={name}.tStartRefresh)\n"
```
TEParsons: Adding tabs to boilerplate like this can cause an indentation error in some cases, it's safer to do it like this:
```suggestion
                code = (
                        f"if ioServer:\n"
                )
                buff.writeIndentedLines(code)
                buff.setIndentLevel(1, relative=True)
                if self.params['syncScreenRefresh'].val:
                        code = (
                                f"ioServer.cacheMessageEvent('{comp_onset_tag}', sec_time={name}.tStartRefresh)\n"
                        )
                else:
                        code = (
                                f"ioServer.cacheMessageEvent('{comp_onset_tag}', sec_time={name}.tStart)\n"
                        )
                buff.writeIndentedLines(code)
                buff.setIndentLevel(-1, relative=True)
```
as `buff.setIndentLevel` will figure out how much to indent by based on user's prefs rather than assuming an indent is 4 spaces.

(though there's loads of instances where we've already done indentation wrong, particularly in older components)

PR #4404, opened 2021-12-02
BF: Apply units to textbox sub-components on set

TEParsons:


psychopy/visual/textbox2/textbox2.py
```diff
@@ -281,6 +281,22 @@ def font(self, fontName, italic=False, bold=False):
                     size=self.letterHeightPix,
                     bold=self.bold, italic=self.italic)
 
+    @property
+    def units(self):
+        return WindowMixin.units.fget(self)
+
+    @units.setter
+    def units(self, value):
+        WindowMixin.units.fset(self, value)
```
mdcutone, line 290: Can we call `super().units.fget()` here instead? Or this this a static method?

TEParsons: `super().units` returns `None`, I don't like using `super()` in general because if you call it on a property it returns the value rather than the object with fget/fset methods

mdcutone (approved):
Seems fine for now. Let's make sure the tests cover this stuff since it hasn't seen widespread use yet.

PR #4405, opened 2021-12-03
BF: Store all clicks if mouse is set to save "on click"

TEParsons:


mdcutone (approved):
Going to assume this works, probably should consider using `format` for this stuff in the future. :)

PR #4406, opened 2021-12-03
BF: Fix positioning of tight bounding box on TextBox2

TEParsons:


psychopy/visual/textbox2/textbox2.py
```diff
@@ -1009,21 +1009,22 @@ def _updateVertices(self):
...
         # tight bounding box
-        if self.vertices.shape[0] < 1:  # editable box with no letters?
-            self.boundingBox.size = 0, 0
-            self.boundingBox.pos = self.pos
-        else:
-            L = self.vertices[:, 0].min()
-            R = self.vertices[:, 0].max()
-            B = self.vertices[:, 1].min()
-            T = self.vertices[:, 1].max()
+        if hasattr(self._vertices, self.units) and self.vertices.shape[0] >= 1:
```
mdcutone, line 1012: `hasattr(self._vertices, self.units)` seems off here, I thought the signature `hasattr(object: Any, name: str) -> bool`

mdcutone (changes requested):
Take a look at the comment, not sure if the syntax is valid there.

psychopy/visual/textbox2/textbox2.py
TEParsons: I think this is right? `self._vertices` is a vertices object, `hasattr` checks that it has the attribute corresponding to the current units, to avoid an attribute error later.

PR #4407, opened 2021-12-03
BF: Layout Form using layout classes

TEParsons:


peircej:
Task linked: [CU-1n012p5 Make Slider size/pos settable with layout classes so that Form lays out correctly](https://app.clickup.com/t/1n012p5)

psychopy/tools/arraytools.py
```diff
@@ -219,6 +219,27 @@ def array2pointer(arr, dtype=None):
         ctypes.POINTER(numpy.ctypeslib.as_ctypes_type(dtype)))
 
 
+def snapto(x, points):
+    """
+    Snap values in array x to their closest equivalent in an array of target values.
+    """
+
+    # Force values to 1d numpy arrays, though keep track of original shape of x
+    x = numpy.array(x)
```
mdcutone: Use `asarray` to prevent copying the array if already a numpy array. You might not need to reshape anything if input arrays have the same shape like `Nx2` or `Nx3`.

psychopy/tools/arraytools.py
```diff
@@ -219,6 +219,27 @@ def array2pointer(arr, dtype=None):
...
+    x1d = x.reshape((-1, 1))
+    points = numpy.array(points).reshape((1, -1))
+    # Get differences
+    deltas = numpy.abs(x1d - points)
```
mdcutone, line 254: This may work without reshaping given numpys broadcasting rules. Just do shape checks before to ensure they are the same. Checkout how `mathtools` handles inputs with different shapes.

psychopy/tools/arraytools.py
```diff
@@ -219,6 +219,27 @@ def array2pointer(arr, dtype=None):
         ctypes.POINTER(numpy.ctypeslib.as_ctypes_type(dtype)))
 
 
+def snapto(x, points):
```
mdcutone, line 222: How does this work? Please provide and example in the docstring.

mdcutone (changes requested):
Snap function needs some documentation. Not sure what the use case is for and what the expected output is suppose to be. The routine in that function could be improved to be more efficient and robust.

psychopy/tools/arraytools.py
TEParsons: The problem is that I expect them to be different shapes, so the reshaping is necessary. Essentially I need to convert it to 1d to get a 2d matrix of deltas, but once those deltas are converted to linear indices and indexed into `points` we can reshape the resulting array to match the original dimensions of x.

psychopy/tools/arraytools.py
TEParsons: Can do :)

TEParsons:
> Snap function needs some documentation. Not sure what the use case is for and what the expected output is suppose to be. The routine in that function could be improved to be more efficient and robust.

The snap function is basically just for snapping slider labels to ticks, use case is when there's more ticks than labels. Previously it was done in the Slider creation method, but I thought it might be useful to have it as a more general function.

psychopy/tools/arraytools.py
mdcutone: Thanks for the documentation, much clearer what I'm looking at.

mdcutone (approved):
Looks okay for now, could pull this in once the tests pass.

PR #4411, opened 2021-12-07
RF: Changes to `Window.getActualFrameRate()`

mdcutone:
Redirects logging of `waitBlanking` attribute setting to DEBUG rather than EXP. Added EXP level log messages that report when `getActualFrameRate()` is called and always reports the measured frame rate when the process is complete. This should clear up the output logs a bit. Minor changes added to clean up the code a bit while I was in there.


psychopy/visual/window.py
```diff
@@ -3105,36 +3105,53 @@ def getActualFrameRate(self, nIdentical=10, nMaxFrames=100,
...
         if nIdentical > nMaxFrames:
-            raise ValueError('nIdentical must be equal to or '
-                             'less than nMaxFrames')
+            raise ValueError(
+                'Parameter `nIdentical` must be equal to or less than '
+                '`nMaxFrames`')
+
+        # log that we're measuring the frame rate now
+        msg = "{}: Attempting to measure frame rate of screen ({:d}) ..."
+        logging.exp(msg.format(self.name, self.screen))
```
peircej: Should only make a logging call if `Window.autolog` is True

psychopy/visual/window.py
```diff
@@ -3105,36 +3105,53 @@ def getActualFrameRate(self, nIdentical=10, nMaxFrames=100,
...
+
+        # Only log changing `recordFrameIntervals` if differs from the user's
+        # settings. Disable recording for the warmup period.
         recordFrmIntsOrig = self.recordFrameIntervals
-        # run warm-ups
-        self.recordFrameIntervals = False
+        self.setRecordFrameIntervals(False, log=logging.DEBUG)
```
peircej: The `log` arg is a Bool indicating whether or not to override the autolog` setting for this Window, so setting it to `logging.DEBUG` has no effect

peircej: I suppose we could change it so it *does* the level of logging, as you've tried here, but there would be quite a lot of lines to refactor then

peircej (changes requested):
Could you change the Window logging call to be dependent on win.autolog setting. The comment about log being a bool is OK. It currently won't be having any impact but your changes here could be left and we could make the functional change later

PR #4419, opened 2021-12-13
BF: Fix layout bug in TargetStim

TEParsons:
Second attempt at #4388 but on a clean branch

This time, fixes the layout bug in TargetStim AND replaces circle stim in calibration graphics with an actual TargetStim object


TEParsons:
@isolver I can only test the calibration changes in MouseGaze, but given that the code was broadly similar before and I've made the same changes in each I don't see why it wouldn't work in all if it works in MouseGaze! Plus it's definitely broken in dev now, so can't get any broken-er ;) 

If you're happy with this PR at a glance we can pull in now and you can test in the other devices when you're back :)

isolver:
Will do. I may not get to it until after xmas given how things are going.
Happy holidays

PR #4424, opened 2021-12-15
Fix all other typos found by codespell

DimitriPapadopoulos:
Note that this fixes both code comments and user-visible text, because fixing only user-visible text and more importantly avoiding typos in the long term is a difficult task otherwise - at least with [codespell](https://github.com/codespell-project/codespell) which doesn't make a difference between code, code comments, documentation or user-visible output.

If the above issue or the extent of the changes are a problem, please tell me, I may be able to split into different commits.

Please also note that the following changes **break the API** by modifying property names. Not sure how stable that part of the API is supposed to be. If that's a problem, please do tell me, I can revert these specific changes and later propose a more complex change that would preserve backward compatibility:
* `accelleration` → `acceleration`
  https://github.com/psychopy/psychopy/blob/35521338dfc30e147719c91d5fb0d02c27ca8b5a/psychopy/iohub/client/wintab.py#L142-L143
* `pallette` → `palette`
  https://github.com/psychopy/psychopy/blob/35521338dfc30e147719c91d5fb0d02c27ca8b5a/psychopy/visual/textbox2/textbox2.py#L253-L254


psychopy/iohub/client/wintab.py
```diff
@@ -140,8 +140,8 @@ def velocity(self):
         return self._velocity
 
     @property
-    def accelleration(self):
```
TEParsons, line 143: This one is a bit scary - spellchecking the docs is all good, but this has the potential to actually break something. It's better to create an alias rather than changing the existing method, so users can use either.

psychopy/visual/textbox2/textbox2.py
```diff
@@ -769,13 +769,13 @@ def _getStartingVertices(self):
     def draw(self):
         """Draw the text to the back buffer"""
         # Border width
-        self.box.setLineWidth(self.pallette['lineWidth']) # Use 1 as base if border width is none
```
TEParsons: Ditto my previous comment - not sure about changing names of actual things used in code

psychopy/iohub/client/wintab.py
DimitriPapadopoulos: I agree. See my initial comment:
> Please also note that the following changes break the API by modifying property names. Not sure how stable that part of the API is supposed to be. If that's a problem, please do tell me, I can revert these specific changes and later propose a more complex change that would preserve backwards compatibility:

psychopy/visual/textbox2/textbox2.py
DimitriPapadopoulos: I agree. See my initial comment:
> Please also note that the following changes break the API by modifying property names. Not sure how stable that part of the API is supposed to be. If that's a problem, please do tell me, I can revert these specific changes and later propose a more complex change that would preserve backwards compatibility:

TEParsons (changes requested):
I'm happy with the doc string changes but I think we should leave any actual attribute/method names as is, at least until we can have a deeper look at their use. So roll those back and I'll pull this in and then look at those other typos after :)

psychopy/iohub/client/wintab.py
DimitriPapadopoulos: Additionally, as a proof that the change is intrusive and error-prone, I had forgot to change `self._accelleration`!

Such a change should go in its own PR and make sure the code is backwards compatible. I'll do that after this PR.

DimitriPapadopoulos:
Should be OK now. Will open another PR for `pallette` and `accelleration`.

hsogo:
After this fix, I can't update .po file with PoEdit (version 3.0.1 on Windows).
Here is a screenshot of the error message.

I'll try to find what is wrong.

DimitriPapadopoulos:
Mmmh... I need to regenerate the .po files.

DimitriPapadopoulos:
I reverted `messages.pot`; `fa_IR/LC_MESSAGE/messages.po` and `ja_JP/LC_MESSAGE/messages.po` to their previous state in #4434. They should never have been changed, sorry about that. I have checked that the reverted files can be loaded in poedit.

That said, the localization [`readme.txt`](https://github.com/psychopy/psychopy/blob/release/psychopy/localization/readme.txt) clearly needs to be updated, I have opened a different ticket #4435 with some of the issues I had with it.

PR #4425, opened 2021-12-16
Fix typos found by codespell in actual code

DimitriPapadopoulos:
1. Fix private variables:
   - `_accelleration` → `_acceleration`
   - `_pallette` → `_palette`
2. Modify API:
   - Keep existing properties, with typos, but deprecate them.
   - Create new properties, without typos.
   - Modify code/examples to use new properties.

Could add [deprecation](https://deprecation.readthedocs.io/) as a dependency: better and standard way to deprecate the function.

### DIFF
--- a/psychopy/iohub/client/wintab.py
+++ b/psychopy/iohub/client/wintab.py
@@ -68,7 +68,7 @@ class PenSampleEvent(ioEvent):
             if efvalue >= 0:
                 setattr(self, '_' + efname, ioe_array[efvalue])
         self._velocity = 0.0
-        self._accelleration = 0.0
+        self._acceleration = 0.0
 
     @property
     def x(self):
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
By the way, palettes without/with focus differ at initialization:
https://github.com/psychopy/psychopy/blob/d87f467c7488126d91ed71a59e4962bbe2158448/psychopy/visual/textbox2/textbox2.py#L255-L266
but not in the setter function:
https://github.com/psychopy/psychopy/blob/d87f467c7488126d91ed71a59e4962bbe2158448/psychopy/visual/textbox2/textbox2.py#L271-L274

Is that on purpose?

TEParsons:
Good spot! Honestly pallette shouldn't even be settable - the setter doesn't even do anything as the value is overwritten next time the getter is called. I'll just remove the setter and have it be read only.

TEParsons (approved):
All looks good! I'll remove the palette/pallette setter in a different PR once this is in

PR #4430, opened 2021-12-17
FF: Don't compute stdev on frame intervals if degrees of freedom is invalid

mdcutone:
Fixes an issue where the variability calculation for frame intervals is being computed with inadequate degrees of freedom.


peircej:
Ah, sorry @mdcutone I fixed that one (slightly differently) here:
https://github.com/psychopy/psychopy/pull/4431

PR #4432, opened 2021-12-17
RF: get rid of pump warning if pyqmix not installed

peircej:
pyqmix is a soft dependency but the Pump Component expects it to be there
so that it can check syringeTypes.

Hard-coded the syringeTypes and inserted a check that they're correct so
that the pyqmix lib no longer spits annoying warnings to users that don't
need it (most users)

### DIFF
--- a/psychopy/hardware/qmix.py
+++ b/psychopy/hardware/qmix.py
@@ -16,21 +16,43 @@ system is described in the following publication:
     Behavior Research Methods. doi: 10.3758/s13428-018-1145-1
 
 """
+from psychopy import prefs, logging
 
-
-from psychopy import prefs
-from pyqmix import config, QmixBus, QmixPump
-import pyqmix.pump
-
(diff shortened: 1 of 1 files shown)


psychopy/hardware/qmix.py
```diff
@@ -16,21 +16,43 @@
...
 bus = None
-pumps = []  # To keep track of all instantiated pumps.
+pumps = []  # To keep track of all instantiated pumps
+# syringeTypes could be fetched
+syringeTypes = ['25 mL glass', '50 mL glass']
+
+
+def _checkSyringeTypes():
+    """Check that the hard-coded and lib-defined syringes match"""
+    """NB the reason for this check:
```
mdcutone, line 35: Concatenate these doc strings.

mdcutone (commented):
Just clean up the doc string and this should be fine to merge. We all should get into he habit of just-in-time module imports to reduce load time and memory footprint where performance isn't too important.

PR #4434, opened 2021-12-17
Revert translation changes from #4424 / 9d626c0

DimitriPapadopoulos:
Revert template and `fa_IR`/`ja_JP` to their previous state, they should not have been changed.

### DIFF
--- a/psychopy/app/locale/fa_IR/LC_MESSAGE/messages.po
+++ b/psychopy/app/locale/fa_IR/LC_MESSAGE/messages.po
@@ -754,7 +754,7 @@ msgstr ""
 #: app/builder/dialogs/__init__.py:1094
 #: app/builder/dialogs/dlgsConditions.py:304
 #: app/builder/dialogs/dlgsConditions.py:579 app/builder/validators.py:110
-msgid "Name must be alphanumeric or _, no spaces"
+msgid "Name must be alpha-numeric or _, no spaces"
 msgstr ""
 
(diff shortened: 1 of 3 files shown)


hsogo:
Thank you for your quick response.
I found that the error is caused by change on line 8790 of ja\_JP/MESSAGE/messages.po:
```
 #, fuzzy
 #~| msgid "Data filename"
-#~ msgid "Data file delimiter"
+#~ msgid "Data file delimeter"
 #~ msgstr "データファイル名"
```
This entry seems "translation memory" created by PoEdit and it can be safely deleted.  Removing this entry, PoEdit can update the file.

hsogo:
I tried updating fa_IR/MESSAGE/messages.po from PoEdit.  It can be updated without any error even after applying #4424.  I don't think it should be reverted.  Shall I create a pull request of my fix on ja_JP/MESSAGE/messages.po?

DimitriPapadopoulos:
I still think it's best to revert the previous changes, including other files `psychopy/app/locale/fa_IR/LC_MESSAGE/messages.po` and template `psychopy/localization/messages.pot`.

Then you can update `psychopy/app/locale/fa_IR/LC_MESSAGE/messages.po`. It will need to be updated anyway.

hsogo:
I've fixed only translation memory in ja\_JP/MESSAGE/messages.po. Translation memory is a set of old unused translations that is used by PoEdit to suggest translation.  I haven't modified other changes (i.e. correcting typos) you made on ja\_JP/MESSAGE/messages.po.

peircej:
Closing in favour of #4439

PR #4437, opened 2021-12-18
Update messages.pot

DimitriPapadopoulos:
Updated using [Babel](http://babel.pocoo.org/), as was the case with the previous version:
```
cd psychopy/localization
pybabel extract --no-default-keywords -k _translate --no-wrap --project=PsychoPy --version=`cat ../../version` --sort-by-file -o messages.pot ..
```

### DIFF
--- a/psychopy/localization/messages.pot
+++ b/psychopy/localization/messages.pot
@@ -1,2081 +1,1842 @@
 # Translations template for PsychoPy.
-# Copyright (C) 2019 ORGANIZATION
+# Copyright (C) 2021 ORGANIZATION
 # This file is distributed under the same license as the PsychoPy project.
-# FIRST AUTHOR <EMAIL@ADDRESS>, 2019.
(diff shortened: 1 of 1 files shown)


peircej:
@hsogo Just checking this with you, @DimitriPapadopoulos has done a bunch of work finding typos in the project, some of which were in strings to be translated. This is us hopefully getting that work brought in for (re)translation. Is this done right for your workflow?

DimitriPapadopoulos:
The message.pot file is a template, to be used when starting a new translation for a new language. It does not interfere with translations of existing languages. For existing languages, poedit updates the relevant \*.po  file directly from the \*.py source files.

hsogo:
As @DimitriPapadopoulos says, updating .pot file doesn't affect translations of exisitng languages.  All corrections of typos made by #4424 are included in #4439.  I have no trouble in updating Japanese translation now.

PR #4462, opened 2022-01-04
ENH: Changes to resourceManager JS boilerplate

TEParsons:


apitiot:
There is a little mistake on line 147:
"%(name)s.status = FINISHED;\n"
should be:
"%(name)s.status = PsychoJS.Status.FINISHED;\n"

Otherwise, I believe we are good to go.

PR #4465, opened 2022-01-05
BF: Don't clear keypresses after checking for esc

TEParsons:
as it's now the same backend as experiment keys


isolver:
closing, since this is no longer an issue after pulling in #4467

PR #4466, opened 2022-01-05
ENH: Add additional template routines

RebeccaHirst:
Added a few template routines for the domain "trials":

* Fixed interval
* Random interval
* feedback for accuracy with key responses
*  feedback for rts with key responses
* feedback for accuracy with mouse/touch responses
*  feedback for rts with mouse/touch responses

Note that because feedback routines are dependant on other routines, I added a `try` statement and if there is no routine with the required response component the text displays a warning that this is required.


RebeccaHirst:
Also note that the method of rounding RTs in the feedback routine is due to an issue with the transpiler (Math.round() in JS rounds to the nearest integer whereas round() in python rounds to decimal place) so this method makes it easy to run online too (probably can fix the rounding in the transpiler at some point)

PR #4469, opened 2022-01-06
NF: Add default image as a psychopy constant

TEParsons:
If an experiment has any Image components, then `psychopy.constants.DEFAULT_IMAGE` gets added to the list of imports at the beginning, meaning you can use the image in a component by setting the value of image to `$DEFAULT_IMAGE`.


peircej:
Task linked: [CU-1tn16g4 Add default image for Image component](https://app.clickup.com/t/1tn16g4)

peircej:
I prefer the approach I've taken in https://github.com/psychopy/psychopy/pull/4470 where we search the app/resources folder as an additional location when looking for an image (after searching the root of the experiment).

PR #4471, opened 2022-01-06
ENH:10 second countdown routine

suelynnmah:
Under misc template, this is a 10 second countdown routine.


RebeccaHirst (commented):
@suelynnmah couple of things here:

1. the countdown text only lasts 1 second in duration (I guess because of the current bug?) 
2. the routine name has "var_" at the beginning of it, can you change it's name to "countdown_10secs"

PR #4474, opened 2022-01-06
BF: Ensure ResourceManager does nothing locally

TEParsons:
Even if a component does not write any code, its name is still included in the list of components within a routine, so there needs to be *something* there in order to prevent the experiment from crashing. The safest way to do this is to just add a class for objects not implemented locally yet, meaning a placeholder object is created which can accept & return attribute values. Running an experiment with a ResourceManagerComponent in will raise an alert.


peircej:
Task linked: [CU-1tn26rg Resource manager causes builder experiments to crash](https://app.clickup.com/t/1tn26rg)

PR #4477, opened 2022-01-07
DOCS: remove how to recipies from psychopy.org and move to workshops.psychopy.org

RebeccaHirst:
Moving all tutorials to workshops.psychopy.org so that tutorials are in one place and so that they are move easily editable and maintainable by the science team

https://workshops.psychopy.org/tutorials/index.html


peircej:
Closing for now, since the docs can now be automatically regenerated from a push to the `release` branch of the repo. We can revisit later if it's still needed

PR #4480, opened 2022-01-07
BF: Copy template on adding, don't add template

TEParsons:
Adding the template itself rather than a copy means that editing a new routine changes the template - and all routines from the same template (including blank)


peircej:
Task linked: [CU-1tuh2nb component copying over from previously made routine in new routine](https://app.clickup.com/t/1tuh2nb)

PR #4486, opened 2022-01-07
BF: New routine name should default to "routine_1" not "var_1"

TEParsons:


peircej:
Task linked: [CU-1tn1nk3 new routines are called "var" rather than "routine_1" etc. by default](https://app.clickup.com/t/1tn1nk3)

PR #4495, opened 2022-01-12
BF: Fix layout issues and crashes around the custom colorpicker

### LINKED ISSUE #4331: Crash upon second use of color picker
When attempting to choose the color of text stimuli on macOS, I have experienced a recurring bug that causes PsychoPy to crash reliably on either the first or second use of the color picker.

The most reliable reproduction method I have found is as follows:

- Create a new project
- Create a text stimulus
- Go to the Appearance tab
- Click on the color picker 
- Select a color on the right
- Click Insert (sometimes crashes here)
- Click OK or click on the color picker again, and it will crash here.

I have attached a macOS crash log:
[PsychoPy Crash Log.txt](https://github.com/psychopy/psychopy/files/7443013/PsychoPy.Crash.Log.txt)

mdcutone:
This PR completely rewrites the colorpicker dialog to fix various issues that lead to it crashing on MacOS and producing layout problems on GTK. Some quality of life improvements have also found their way in.

* No more segfaults on MacOS (fixes #4331).
* Fixed some GTK layout issues.
* Hex input is a little nicer to use.
* Context flags can be set to disable Copy and Insert separately.
* If only Insert is specified, it is renamed to OK to make it more consistent with other dialogs. Nicer when used to pick colors in component dialogs.
* More consistent spacing between pages.
* Listbox is used for preset colors now, more consistent across platforms, less things to go wrong.

@TEParsons Check if the output format for color values is appropriately formatted for components, please :)


TEParsons (commented):
Hex doesn't actually need quotes around it in Builder - the param can detect that it's a hex color and adds them automatically. Color objects should handle superfluous quotes, which I'll add in regardless, but at present they break it.

PR #4520, opened 2022-01-19
ENH: Supply filename from experiment settings in JS

TEParsons:


peircej:
@apitiot some corner-case/considerations here:
- what to do if the user puts in something that makes no sense online (e.g. someone might have used `"g://myStudy"` if they want to save data to a google drive
- what if they've not used a folder called "data"? Add that part of the path server-side?
- what if they've used `"../../myFolder/filename"`? Sanitize server-side to make sure they aren't trying to save outside their permitted zone? Or just leave it to the permissions to crash at file-save-time?

In reality, these are likely to be rare events. The default path name is just 
"data/<expName>_<participant>_<datestr>" and Todd has prepended with "./" but just alerting that there may be need to sanitise

peircej:
I'm going to pull this one in for now so we can test as part of the release build. I believe the changes that are likely to be needed are going to be changes on server side (to sanitise bad filename requests) rather than on the script side

PR #4524, opened 2022-01-20
ENH: Replace \ with / in file controls on set as well as on compile

TEParsons:


peircej:
Belt and braces! :-) 
Hopefully we won't now come up with a problem of people being unable to include `\` in their strings

PR #4526, opened 2022-01-21
ENH: Enable Slider units in Builder now that they work

TEParsons:


peircej:
This is generating an error `AttributeError: 'TargetStim' object has no attribute '_size'`

PR #4567, opened 2022-02-01
BF: Remove resources from experiment start in JS if they're handled by a Resource Manager elsewhere

TEParsons:


peircej:
Cool, next we need a similar test for Comps that use a Static Comp for resource loading

PR #4568, opened 2022-02-01
FF: Remove timer from subprocess pipe polling loop

mdcutone:
This PR removes the timer from the polling thread which is likely causing output to be missed. Likely not needed here since the threads are blocked waiting for input on `read()` calls. Also we make sure threads are joined when the process ends.

### DIFF
--- a/psychopy/app/jobs.py
+++ b/psychopy/app/jobs.py
@@ -67,7 +67,7 @@ KILL_ACCESS_DENIED = wx.KILL_ACCESS_DENIED
 KILL_NO_PROCESS = wx.KILL_NO_PROCESS
 KILL_ERROR = wx.KILL_ERROR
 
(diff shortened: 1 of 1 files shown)


psychopy/app/jobs.py
```diff
@@ -588,9 +597,13 @@ def onTerminate(self, exitCode):
...
         self.parent.Bind(wx.EVT_IDLE, None)
 
+        self._readPipes()  # read remaining data
+
         # stop the pipe reader threads now
         self._stdoutReader.stop()
-        self._stderrReader.stop()
+        self._stdoutReader.stop()
+        self._stderrReader.join(timeout=1)
+        self._stderrReader.join(timeout=1)
```
peircej, line 606: I don't get it. You call `stdoutReader.stop()` twice and then `stderrReader.join()` twice. Shouldn't each pipe stop once, and then each pipe joins once?

PR #4587, opened 2022-02-05
DOCS: update static comp docs to consider online studies

RebeccaHirst:
Updating statics component docs to consider online studies. 

Notes:

-  At the moment I think for online studies static component overrides resources loaded at the beginning - is that correct? if no I'll remove the note on this page.
- These docs contain a ref to resourceManager that will only work when the PR for resource manager gets pulled in https://github.com/psychopy/psychopy/pull/4586


RebeccaHirst:
wrong branch

PR #4595, opened 2022-02-07
BF: allow the routine to end when resource download is complete

peircej:


peircej:
@RebeccaHirst you can see the changes required here - I haven't had time to test this yet directly! Hopefully we can pull it to dev branch and you can check it

PR #4603, opened 2022-02-08
BF: Check state for first-push via project attributes specifically

TEParsons:
otherwise second sync call will also call firstPush, causing errors if you've already started collecting data


peircej:
Task linked: [CU-1x1kt45 Error when syncing to pavlovia after data has been gathered in online study "remote contains work that you do not have locally"](https://app.clickup.com/t/1x1kt45)

PR #4618, opened 2022-02-14
BF: Still define mouse.x and mouse.y when "on valid click"

TEParsons:
Unable to test all the way as the experiment I made to test it with gives "404 forbidden" when running for some reason - I'm guessing this isn't related to the boilerplate. Science team, if you could try it out with the experiment which initially showed the bug then we can pull in once it's confirmed working :)


peircej:
Task linked: [CU-1y3vdf4 Save data on valid click causes TypeError in online experiment](https://app.clickup.com/t/1y3vdf4)

suelynnmah:
I've checked and it works

PR #4639, opened 2022-02-18
Update translations

hsogo:
New strings are translated.

By the way, I doubt that tooltip for "delete invalid control" button may be wrong.  When the button is "x", the tooltip says "This parameter will not be deleted until you click Okay. Click this button to revert the deletion and keep the parameter". I think that this message should be shown when the button is "⟲".  When the button is "⟲", no tooltip is shown.   (test environment: WinPython 3.8.10 x64, PsyhcoPy 2022.1.0, wxPython 4.3.20210620).


hsogo:
I think that line 211 of app/builder/dialogs/paramCtrls.py should be `self.undoBtn.SetToolTipString()`.  In addition, `SetToolTip()` should be used instead of `SetToolTipString()`.  May I fix them?

TEParsons:
Good spot! Yep, the tooltip should be on the undo button rather than the delete button. Will put in a PR now.

peircej:
Using ToolTip instead of ToolTipString was something I had actually fixed in 2f4a784 and aafe9f2e6a6 which I'm now updating to merge with your additions above. No need for you to do more with them I think

peircej:
Thanks hsogo! :-) 
This will be in 2021.1.1 which we'll hopefully upload today

PR #4651, opened 2022-02-22
NF/RF: `BaseAuiFrame` class

mdcutone:
This PR introduces the `BaseAuiFrame` class, which is now the base class for `CoderFrame` and `BuilderFrame`. The purpose of this is to ensure consistency of AUI configurations between different parts of the application UI. Classes for AUI panes are going to be added at some point to simplify their creation and placement.

As PsychoPy grows, the UI will increase in complexity and these changes will help us leverage AUI better to meet those needs.

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -148,6 +148,9 @@ class BuilderFrame(BaseAuiFrame, ThemeMixin):
         self.readmeFrame = None
         self.generateScript = generateScript
 
+        # default window title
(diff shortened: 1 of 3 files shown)


TEParsons:
Would it be worth refactoring this class to be generic and in line with wx's general stuff and submitting a PR to wx itself too?

mdcutone:
> Would it be worth refactoring this class to be generic and in line with wx's general stuff and submitting a PR to wx itself too?

We can make it more generic in the future. Right now I'm focused on consistency between UI frames within our application, decoupling application code from the UI code, and increasing readability.

PR #4657, opened 2022-02-24
BF: Use setAttribute in size/pos setters so that autoLog is respected

TEParsons:


TEParsons:
wrt logging `layout.Size` and `layout.Position` objects: When converted to a string, they look like this:
```
<psychopy.layout.Size: [100. 100.]px>
```

PR #4661, opened 2022-02-25
Update procedure.py

matinpf:
"self.sample_type" does not make sense, Also I got an error when I used it with the pupil_lab eye tracker.


isolver (approved):

isolver:
Thanks for finding this, which will be an issue for any tracker when using binocular samples. Fix looks good to go.

peircej:
OK, thanks @matinpf and for checking @isolver 
That's pulled in and I think we'll be having a release of 2022.1.2 next week

PR #4675, opened 2022-03-07
ENH: Use layout classes to do unit conversions for Mouse

TEParsons:


psychopy/hardware/__init__.py
```diff
@@ -12,7 +12,7 @@
 except ImportError:
     from collections import Iterable
 
-__all__ = ['forp', 'cedrus', 'minolta', 'pr', 'crs', 'iolab', 'eyetracker']
+__all__ = ['forp', 'cedrus', 'minolta', 'pr', 'crs', 'iolab', 'eyetracker', 'mouse']
```
mdcutone: keyboard too? Might be a small enough fix for this PR to include.

TEParsons: Sure, good shout

PR #4677, opened 2022-03-08
BF: allGuns=False option in monitors.getLumSeries

dekuenstle:
For allGuns=False, the *allGuns* instead of *guns* variable was defined and an error was raised when *guns* was used.

### DIFF
--- a/psychopy/monitors/getLumSeries.py
+++ b/psychopy/monitors/getLumSeries.py
@@ -120,11 +120,10 @@ def getLumSeries(lumLevels=8,
     else:
         toTest = numpy.asarray(lumLevels)
 
-    # FIXME - this is a bit confusing here, `guns` might not be defined!
     if allGuns:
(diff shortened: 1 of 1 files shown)


mdcutone (approved):
Looks good, thanks for the fix.

dekuenstle:
The current release branch has two similar but not identical getLumSeries functions in [getLumSeries.py](https://github.com/psychopy/psychopy/blob/release/psychopy/monitors/getLumSeries.py) and [calibTools.py](https://github.com/psychopy/psychopy/blob/release/psychopy/monitors/calibTools.py). The function in calibTools.py is exposed by the monitor module and used by the MonitorCenter. 

Both functions would need this fix, but I am wondering if it wouldn't be better to simply delete the duplicate in getLumSeries.py. What do you think @mdcutone?

dekuenstle:
@peircej unfortunately this merge will not completely fix the issue. Please see my comment 40 minutes ago.

Issue #4678, opened 2022-03-08
Additional photometer (Gamma Scientific S470)


dekuenstle:
Hi, we have a working photometer class implementation for the gamma scientific S470 . 
Is there interest in supporting additional light measurement devices in psychopy, i.e. would you appreciate a pull request?

Cheers
David

mdcutone:
Yes, you may submit PRs to expand hardware support.

PR #4680, opened 2022-03-08
NF: Add gamma scientific flexOpt photometer

dekuenstle:
#4678


dekuenstle:
@mdcutone I just added a unit test and tested the device in the monitor center. I think it is complete from my side.

Issue #840, opened 2015-02-19
BufferImageStim set pos has no effect


jeremygray:

jeremygray:
I suspect that ._needsUpdate or .needsVertexUpdate is being ignored. Doing the same thing as is done for ImageStim did not work, however. 

There is a simple work-around, fortunately. Just convert the BufferImageStim to a normal ImageStim, and use that instead, like this:

temp_screenshot = visual.BufferImageStim(win, stim=stimlist, rect=rect)
screenshot = visual.ImageStim(win, temp_screenshot.image)

PR #4695, opened 2022-03-14
TEST: Add base units tests for ImageStim

TEParsons:


TEParsons:
@mdcutone I tried to do the same for BufferImageStim too to test whether #840 was still an issue, but couldn't get it working - for some reason it always ended up with the image being massive. It's got me stumped, could you take a look? Here's the code:
```
class TestBufferImage(_TestUnitsMixin):
    def setup(self):
        self.win = visual.Window()
        self.stim = visual.SimpleImageStim(self.win, str(Path(TESTS_DATA_PATH) / 'testimage.jpg'))
        self.obj = visual.BufferImageStim(
            self.win,
            stim=[self.stim],
            rect=(-1, 1, 1, -1)
        )

```

PR #4707, opened 2022-03-17
NF: Add stop/start/restart buttons to Coder shell

TEParsons:


mdcutone:
Can't seem to make this start. Crashes during initialization of the UI.

peircej:
In a similar vein, for the Output view(s) it feels like we need a "clear" button to get rid of previous messages

TEParsons:
@mdcutone do you mean the app doesn't load, or just that the buttons don't work? Because on the office mac I'm finding that the buttons just print "Keyboard Interrupt" rather than restarting/stopping the console, but they do load and the app doesn't crash

PR #4708, opened 2022-03-17
BF: Fix loading movie stimuli during the static period

mdcutone:
Fixes crash when running a Builder experiment which loads a movie during a static period.

### DIFF
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -498,6 +498,47 @@ class BaseComponent:
                 if stopVal in ['', None, -1, 'None']:
                     stopVal = '-1'
                 buff.writeIndented(f"{compName}.setSound({params['sound']}, secs={stopVal}){endStr}\n")
+            elif paramName == 'movie':
+                # we're going to do this for now ...
+                if params['units'].val == 'from exp settings':
+                    unitsStr = "units=''"
+                else:
+                    unitsStr = "units=%(units)s" % params
+
+                if params['backend'].val == 'moviepy':
+                    code = ("%s = visual.MovieStim3(\n" % params['name'] +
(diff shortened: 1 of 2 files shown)


psychopy/experiment/components/_base.py
```diff
@@ -498,6 +498,47 @@ def writeParamUpdate(self, buff, compName, paramName, val, updateType,
...
                 buff.writeIndented(f"{compName}.setSound({params['sound']}, secs={stopVal}){endStr}\n")
+            elif paramName == 'movie':
+                # we're going to do this for now ...
+                if params['units'].val == 'from exp settings':
+                    unitsStr = "units=''"
+                else:
+                    unitsStr = "units=%(units)s" % params
+
+                if params['backend'].val == 'moviepy':
+                    code = ("%s = visual.MovieStim3(\n" % params['name'] +
```
TEParsons, line 509: Using `%s` is a bit hard to follow - it's better to do `"%(name)s = visual.MovieStim3(\n"` and then at the end of the whole thing just do `% params`. For things like `unitsStr` you can either add it on manually or use an f string with `{}` so it doesn't conflict with the modulus formatting.

psychopy/experiment/components/_base.py
```diff
@@ -498,6 +498,47 @@ def writeParamUpdate(self, buff, compName, paramName, val, updateType,
...
+                            "    win=win, name='%s', %s,\n" % (
```
TEParsons, line 510: In general I'd advise against including indentation in the string - it's better to use `buff.setIndentLevel(+1, relative=True)` (we've done manual indentation in loads of places elsewhere but it's bad and someday I wanna do a sweep to get rid of it all)

PR #4727, opened 2022-03-27
BF: Fix to Updates  dialog box border display issue

### LINKED ISSUE #4726: Psychopy Updates dialog elements are not fully displayed
- OS:windows 10 21H2
- Screen Resolution: 3000*2000
- Version:2022.1.2

Due to the unreasonable layout, the input box of "Browse file" is not fully displayed. As shown below:
Snipaste_2022-03-27_10-55-49

May be related to this border parameter:https://github.com/psychopy/psychopy/blob/a779a71db4d031668f2ac2f15a893a1eceae5809/psychopy/app/connections/updates.py#L317

shun2wang:
Fixed https://github.com/psychopy/psychopy/issues/4726 and tested it

Before:
Snipaste_2022-03-27_10-55-49

After this PR:
Snipaste_2022-03-27_11-56-42

### DIFF
--- a/psychopy/app/connections/updates.py
+++ b/psychopy/app/connections/updates.py
@@ -314,7 +314,7 @@ class InstallUpdateDialog(wx.Dialog):
             btns = [self.cancelBtn, self.installBtn]
         btnSizer.Add(btns[0], 0, flag=wx.LEFT, border=5)
         btnSizer.Add(btns[1], 0, flag=wx.LEFT, border=5)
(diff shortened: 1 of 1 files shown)


TEParsons:
Good catch! I think this solution works as the extra padding on the buttons gives space for the file control to be displayed, but there's actually some buttons which are still cut off. I'll pull this in once tests pass and then add a further fix to make the whole thing display.

PR #4730, opened 2022-03-28
BF: Increase size of update dlg to allow for buttons

### LINKED ISSUE #4726: Psychopy Updates dialog elements are not fully displayed

TEParsons:
builds on #4727 to fully fix #4726


shun2wang:
@TEParsons Thank you! I did ignore that there is also an install button below, I tested this PR and it looks great!
Snipaste_2022-03-28_21-33-34

By the way, there may need another minor fix, is it possible to make the file Entry (text and text boxes and browse button) all grayed out by default when auto-update is checked, in current, if you switch back and forth between these two buttons, the file input will be set disabled. like this:
image

shun2wang (approved):

PR #4733, opened 2022-03-29
PKG: unify `python-vlc` dependency definition

NicolasT:
The dependency on `python-vlc` is (somewhat) duplicated in `setup.cfg`.
This causes the Poetry package manager to bail out when adding PsychoPy
as a dependency (likely due to a bug in Poetry, see
python-poetry/poetry#5378).

This commit simplifies/unifies the `python-vlc` dependency definitions,
retaining the semantics of the original ones.

See: https://github.com/python-poetry/poetry/issues/5378
See: 8692b9c6dfc8fc0a573b7bd5fe40f8815873023e
See: https://github.com/psychopy/psychopy/pull/4613
See: d5c366e4c92064ba199d26d002246999e44526ef
See: https://github.com/psychopy/psychopy/pull/4059

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,7 @@ install_requires =
     google-cloud-speech
     googleapis-common-protos
     # Platform-specific dependencies.
-    python-vlc <= 3.0.11115; platform_system == "Windows"
+    python-vlc == 3.0.11115; platform_system == "Windows"
     python-vlc >= 3.0.12118; platform_system != "Windows"
     javascripthon; python_version >= "3.5"
     pyglet >= 1.5; platform_system == "Darwin"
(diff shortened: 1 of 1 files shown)


peircej:
Thanks. I'll just let the tests complete and then pull in but I can see this is right. 👍 
I was actually planning on creating a poetry installer/toml file but hadn't got round to working out where we might get bitten (particularly in having to support pip as well as poetry and the implications for our standalone binary distribution)

anyway, thanks again

NicolasT:
> Thanks. I'll just let the tests complete and then pull in but I can see this is right. +1

Not sure how this patch can impact code coverage in testing :thinking: so :shrug: 

> I was actually planning on creating a poetry installer/toml file but hadn't got round to working out where we might get bitten (particularly in having to support pip as well as poetry and the implications for our standalone binary distribution)

Poetry is more of a project/dependency management and package building/publishing tool, not really an 'installer' per se. A project built (into a wheel or sdist) using Poetry can be installed using (recent versions of) `pip` (from PyPI or elsewhere): a project using Poetry for its development doesn't enforce the use of Poetry on a consumer of the project. One difference would be that a Poetry project would, by default, not provide a `requirements.txt` in its repo, though one such file can be generated e.g., as part of a release project, from the Poetry lock-file (`poetry export ...`, from memory), and then shipped as one of the release artefacts for those who rely on `requirements.txt`.

PR #4734, opened 2022-03-30
BF: Check for emotiv parameter names and relabel them so they will no…

bollaking:
EmotivMarkingComponent does not have set{ParamNames) function and there is no need to call them. Previously I checked for the name but that is silly cos it can be changed, so I have relabelled all the params with a prefix "emotiv_" and then if those parameter are found they are just ignored


peircej:
Hi @bollaking just checking with you because this seems like a behaviour change. What will happen to existing experiments that use this component? My guess is that, at best, they'll lose their current settings. That being the case we would make this a commit to `dev` branch and it would get included in the next major release (summer) rather than the upcoming bug-fix release

bollaking:
@peircej Yes I think you are correct, the old markers would need to be deleted and re-added.  I see you have already changed the merge target so I gather there is nothing for me to do.

PR #4736, opened 2022-03-30
JS gratings enabled in Builder boilerplate

peircej:


psychopy/experiment/components/grating/__init__.py
```diff
@@ -160,3 +160,48 @@ def writeInitCode(self, buff):
...
+            if val is True:
+                inits[paramName] = 'true'
+            elif val is False:
+                inits[paramName] = 'false'
+            elif val in [None, 'None', 'none', '', 'sin']:
+                inits[paramName].valType = 'code'
+                inits[paramName].val = 'undefined'
+
+        code = ("{inits[name]} = new visual.GratingStim({{\n"
+                "  win : psychoJS.window,\n"
```
TEParsons, line 180: As we write new boilerplate, I think we should try to be consistent in using `buff.setIndentLevel` rather than manually indenting, and using `%(key)s` with `+` signs to add in the `unitsStr` rather than `.format` with multiple dicts

psychopy/experiment/components/grating/__init__.py
```diff
@@ -160,3 +160,48 @@ def writeInitCode(self, buff):
...
+            unitsStr = "units : undefined, "
+        else:
+            unitsStr = "units : %(units)s, " % self.params
+
+        # replace variable params with defaults
+        inits = getInitVals(self.params, 'PsychoJS')
+
+        for paramName in inits:
+            val = inits[paramName].val
+            if val is True:
```
TEParsons: Is this not handled by `Param`? I feel like it should be - a `bool` param should return in lower case if target is JS, and a `code` param should be run through the transpiler

TEParsons (changes requested):

psychopy/experiment/components/grating/__init__.py
peircej: I'm not certain. I feel like belt and braces is fine

psychopy/experiment/components/grating/__init__.py
peircej: I think that requires a discussion where we look at the visuals of the different options and agree together. For now, I think this is easier to read

TEParsons (approved):

PR #4740, opened 2022-03-31
BF: Update component experiment refs on paste

TEParsons:
Otherwise, copying from one experiment to another causes an error on compile as the component still points to its original experiment


peircej:
Task linked: [CU-2hrnq67 Copying and pasteing a routine from one experiment to another then prevents a sync to pavlovia](https://app.clickup.com/t/2hrnq67)

PR #4742, opened 2022-03-31
BF: formItems.xltx had been corrupted by google sheets

peircej:
It was no longer able to open in Excel. Has now been fixed with a version exported from Excel itself


peircej:
Task linked: [CU-25ek6kk Trying to open form template from builder raises "cannot open file" error](https://app.clickup.com/t/25ek6kk)

PR #4746, opened 2022-04-01
BF: if mic/transcription libs let's still offer sound components

peircej:
Beforehand, if pocketsphinx failed to load (which is a bit finicky anyway)
then the Sound component would not show up in Builder


peircej:
closing in favour of #4747 pulling in to release trunk

PR #4755, opened 2022-04-05
BF: Mouse coordinate mapping on retina displays

### LINKED ISSUE #4731: Mouse click inaccurate position
I'm new to PsychoPy. I'm using the 2022.1.2 version on a Mac with Retina Display - Monterey.

I'm testing the simplest experiment - a mouse click over an image should play audio and then clicking another image should terminate the trial. The first image is at the (0,0) and i'm using a star from the polygon options, and the second one is at (-0.4, 0). The position of both images is correct when displayed; however, clicking over the star does not play the audio. The audio plays when the upper right corner is clicked. So the mouse event is not getting the same position as the monitor sets. The (0,0) for the mouse is moved to the upper right corner.

### LINKED ISSUE #4751: Slider and button not picking up software-triggered mouse click event
Hello,

We are using pynput (also tried with pyautogui) to input mouse clicks and keyboard presses (instead of physical mouse clicks and keyboard presses). The psychopy experiment responds to the software-triggered keyboard presses, however it does not respond to the software-triggered mouse input. The experiment is responsive to physical keyboard/mouse-triggered events, and the software-triggered events work outside of the psychopy experiment window (on the desktop or in other apps). 

Have you run into this before? 
Where is the code relevant to input events so I may dig into why this might be happening? (I believe the psychopy mouse event code is built on pygame, though I am not entirely sure). 

Thank you!

macOS Big Sur 11.2.3 and Windows 10
Psychopy v2021.2.3
Python 3.8

mdcutone:
Fixes the issue that mouse coordinates are incorrectly mapped to window coordinates on macOS. Tested on the office Mac here without a retina display (same problem as before), need someone to check on a real retina display if the fix works there too. 

Furthermore, the fix should apply to HiDPI displays on other platforms too since we're using the new `setContentScaleFactor()` method which generalizes mapping between window and framebuffer coordinates in cases where they differ in size.


TEParsons:
Unfortunately this doesn't work on retina :( When I run this experiment:
[untitled.zip](https://github.com/psychopy/psychopy/files/8426653/untitled.zip)
 the text is only True when the mouse is in the top right corner

mdcutone:
I think I needed to apply these changes to the pyglet backend directly since it handles coordinates differently and overrides the underlying conversion methods of the base class. Hope it works now!

mdcutone:
Converted to a draft, don't pull in until we confirm this fixes the issue across different Mac configurations (i.e. retina/non-retina).

TEParsons:
It works on my retina screen! I'll figure out what this textbox glyph test bug is and submit a separate PR for it

PR #4759, opened 2022-04-06
BF: Fix `FileNotFoundError` when running tasks

mdcutone:
If a Builder task is passed to runner without saving first, the task list item created uses a relative path instead of absolute which causes an unhandled exception to occur when running the file. This patch prevents this from happening by ensuring absolute paths are used when building the command for `Job`. 

I also added a dialog box and log messages to catch errors trying to start a subprocess, made some changes to ensure the UI state is correct after such an error occurs to maintain app stability.


TEParsons:
How is a Builder experiment passed to the Runner without saving? It should prompt the user to save when the "Send To Runner" button is clicked, and seems to when I try

TEParsons (approved):

PR #4760, opened 2022-04-07
BF: Ori was calculated in relative terms, needs to be absolute

TEParsons:
Was found by this user:
https://discourse.psychopy.org/t/changing-orientation-of-polygon/28459

Essentially, when calculating `verticesPix`, the rotation matrix was applied to the *relative* vertices, meaning they can't exceed the original stimulus bounds, when it needs to be applied to the *absolute* vertices so that corners can move outside the original bounds.


peircej:
@TEParsons it looks like this was based on dev but is PR to release? I can't pull in to release given the changes that come with it. Do you want to make a new PR for release based on that branch?

TEParsons:
Ah my mistake, I'll cherry pick the relevant commits and redo it from release!

PR #4788, opened 2022-04-21
RF: Alias Textbox.startText with Textbox.placeholder for JS parity

TEParsons:
Could just as easily be pulled into dev if we want to be more cautious

Ideally, should be matched with an alias in PsychoJS to allow `startText` in place of `placeholder`


peircej:
Task linked: [CU-22ybvdr Textbox: Parity between JS and Py](https://app.clickup.com/t/22ybvdr)

PR #4795, opened 2022-04-27
BF: Copied routines skip the builder rename routine method, so do comp updates in base routine method

TEParsons:


peircej:
Task linked: [CU-28nfp9a Error syncing project to pavlovia after routine has been copied and pasted](https://app.clickup.com/t/28nfp9a)

PR #4816, opened 2022-05-05
NF: Serial port component

TEParsons:


peircej:
Task linked: [CU-1yffw5a Component for serial port triggers (like parallel port triggers)](https://app.clickup.com/t/1yffw5a)

PR #4820, opened 2022-05-06
RF: Rather than wrapping serial, use it directly so users can learn from boilerplate

TEParsons:


TEParsons:
As we already discussed this in standups yesterday I'm just gonna pull in, it's passing tests and I'm keen to get it in for the science team to tire kick!

PR #4821, opened 2022-05-06
BF: ShapeStim was not setting opacity when opacity is 0

TEParsons:
Because `bool(0)` is `False`, when what we want really is to compare it against None


peircej:
Task linked: [CU-29ck2z5 Polygons visible when opacity field is 0](https://app.clickup.com/t/29ck2z5)

PR #4822, opened 2022-05-06
BF: When removing a static component, we need to remove references to it

TEParsons:


peircej:
Task linked: [CU-29qzwe6 Disabling a static component AND the component it's loading causes error on compile](https://app.clickup.com/t/29qzwe6)

TEParsons:
[untitled.zip](https://github.com/psychopy/psychopy/files/8641495/untitled.zip)

PR #4823, opened 2022-05-06
ENH: Generate a random participant id and supply it as default to expInfo participant field

TEParsons:


peircej:
Task linked: [CU-1zx854f Ability to randomly generate unique participant identifier](https://app.clickup.com/t/1zx854f)

PR #4832, opened 2022-05-11
BF: Standalone routines were missing a copy method

TEParsons:


peircej:
Task linked: [CU-29j2qek Cannot copy and paste an eyetracker calibration routine (has no property)](https://app.clickup.com/t/29j2qek)

PR #4835, opened 2022-05-12
TEST: Mark TestLoops tests as needing sound, fixes #1243

### LINKED ISSUE #1243: Error with the test_loopBlocks unit test
The test_loopBlocks unit test in tests/test_app/test_builder/test_Experiment.py requires an audio API, but the test is not marked "@pytest.mark.needs_sound".

TEParsons:


peircej:
I think this will cut down the coverage of the tests - it means testLoops wont be run on the virtual machines of GitHub because they don't support sound. Could we instead have a separate test for testSoundLoops so that we can test (non-sound) loops everywhere?

peircej:
Actually, from what I can tell this has been fixed elsewhere (by not requiring sound while testing loops so it doesn't fail on systems with no sound?!). So rather than fixing the label it appears we fixed the test at some point. The fix used here means that the coverage goes down, but from what I can tell the bug is actually no longer an issue

PR #4839, opened 2022-05-17
ENH: Save on click as Mouse default

TEParsons:
Default value for when to save mouse state should be "on click" rather than "final" (as per ClickUp request)


peircej:
Task linked: [CU-1k14v7e Mouse default save on state - change default to on click rather than final](https://app.clickup.com/t/1k14v7e)

PR #4840, opened 2022-05-18
FF: choice labels are not translated

hsogo:
I found that choice labels are not translated after c8cd3802ed567d4e6946d89d305ba418a8122285.
I tried to re-enable translation based on this commit, but I came to think that the old code would be more simple to perform bidirectional conversion between internal values and translated labels.  So I reverted the code and modified it to utilize `label` parameter.  I confirmed that "Audio latency priority" on the Experiment settings dialog is properly presented.


peircej:
@TEParsons there's effectively a reminder not to use Get/SetStringSelection in our code for any control that might be getting localised - the exact string won't exist in the other language

Issue #4818, opened 2022-05-05
Crash on Startup when loading Pavlovia project (404 Project Not Found)


frankpapenmeier:
Problem:
PsychoPy 2022.1.3 does not open .psyexp files for projects that have been pushed to Pavlovia before. I am running PsychoPy on Windows 11, but given the solution to this bug (see below), I assume that the issue should occur on all platforms.

Error Log:
This is the stacktrace from the "last_app_load.log" file when trying to open a respective psyexp file:

```
Traceback (most recent call last):
  File "C:\Program Files\PsychoPy\lib\site-packages\gitlab\exceptions.py", line 304, in wrapped_f
    return f(*args, **kwargs)
  File "C:\Program Files\PsychoPy\lib\site-packages\gitlab\mixins.py", line 112, in get
    server_data = self.gitlab.http_get(path, **kwargs)
  File "C:\Program Files\PsychoPy\lib\site-packages\gitlab\client.py", line 662, in http_get
    result = self.http_request(
  File "C:\Program Files\PsychoPy\lib\site-packages\gitlab\client.py", line 628, in http_request
    raise gitlab.exceptions.GitlabHttpError(
gitlab.exceptions.GitlabHttpError: 404: 404 Project Not Found

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\Program Files\PsychoPy\Lib\site-packages\psychopy\app\psychopyApp.py", line 102, in <module>
    main()
  File "C:\Program Files\PsychoPy\Lib\site-packages\psychopy\app\psychopyApp.py", line 98, in main
    start_app()
  File "C:\Program Files\PsychoPy\Lib\site-packages\psychopy\app\psychopyApp.py", line 26, in start_app
    app = startApp(showSplash=showSplash)
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\app\__init__.py", line 85, in startApp
    _psychopyApp = PsychoPyApp(
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\app\_psychopyApp.py", line 215, in __init__
    self.onInit(testMode=testMode, **kwargs)
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\app\_psychopyApp.py", line 386, in onInit
    self.showBuilder(fileList=exps)
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\app\_psychopyApp.py", line 599, in showBuilder
[...]

PR #4843, opened 2022-05-20
BF: Reinstate original getProject method

TEParsons:
Now uses Pavlovia API to get project id from namespace/name rather than gitlab


TEParsons:
Should fix #4818

Issue #4845, opened 2022-05-23
Possible unexpected and/or unintended behaviour of different keyboard backends for continuous keypresses with getKeys(waitRelease=False, clear=False) (ioHub, PsychToolBox, Pyglet) (e.g. missing key.duration in ioHub)


lboel:
Hi,
I am currently working on a consistent method (online/offline) to move objects as long as a key is pressed. 
Using a straight forward approach by checking for an existing duration on a key in a "Store-All" keyboard component in the builder to check for keyPresses and keyReleases. The logic would be:
As long as the last pressed key has no duration, it's likely not released yet. 
To get a fast impression here is a screenshot of the code component:
![image](https://user-images.githubusercontent.com/8046474/169765933-e4d19005-f429-424d-b709-e8227bff1833.png)

See the whole demo file here: [testKeyPressReleaseIohub.zip](https://github.com/psychopy/psychopy/files/8752451/testKeyPressReleaseIohub.zip)

This approach works fine and as expected (as documented) in PsychoJS:

https://user-images.githubusercontent.com/8046474/169765518-7e35fee4-f2b1-4701-99ab-a603a49c8506.mp4

tDown is returned on KeyPress, key duration is added then the Key is released. As expected

The documentation of Keyboard.getKeys (https://psychopy.org/_modules/psychopy/hardware/keyboard.html#Keyboard.getKeys) states: 

`     waitRelease: bool (default True):
            If True then we won't report any "incomplete" keypress but all
            presses will then be given a `duration`. If False then all
            keys will be presses will be returned, but only those with a
            corresponding release will contain a `duration` value (others will
            have `duration=None`
`
But the current iohub.keyboard usage in hardware.keyboard results (at least on a Windows machine) in no durations at all for iohub backend in the same project. That id likely caused by an unexpected behaviour of the underlying iohub function "getPresses" here https://github.com/psychopy/psychopy/blob/9398a71beefa2cef30baaca7c1b5df34708ebfef/psychopy/hardware/keyboard.py#L309

The same code results here in no durations at all: 
https://user-images.githubusercontent.com/8046474/169767671-63f2f21e-94e3-4f8a-b3b8-aceb19ba46fb.mp4

For Iohub I will open a PR for a proposed fix.  

In addition Psychtoolbox and Pyglet Keyboard backend produce two different unexpected outcomes in this simple demo:
PsychToolBox: 
Is likely returning some Dict and not a series of pressed buttons and there is something wrong with tDown

[...]

PR #4846, opened 2022-05-23
BF: fix waitRelease=False behaviour for iohubKeyboard in hardware.keyboard

lboel:
The documentation of Keyboard.getKeys (https://psychopy.org/_modules/psychopy/hardware/keyboard.html#Keyboard.getKeys) states: 

`     waitRelease: bool (default True)

            If True then we won't report any "incomplete" keypress but all
            presses will then be given a `duration`. If False then all
            keys will be presses will be returned, but only those with a
            corresponding release will contain a `duration` value (others will
            have `duration=None`
`
But the current iohub.keyboard usage in hardware.keyboard results (at least on a Windows machine) in no durations at all for iohub backend in the same project. That likely caused by an unexpected behaviour of the underlying iohub function "getPresses" here https://github.com/psychopy/psychopy/blob/9398a71beefa2cef30baaca7c1b5df34708ebfef/psychopy/hardware/keyboard.py#L309

See #4845 for more information on this.
Update:
I tested the PR on Windows-10-10.0.19044-SP0 with PsychoPy 2022.1.3 (pyglet (1.4.11) & psychtoolbox (3.0.18.2))


isolver:
Thanks for pointing this out. I implemented a fix for the keyboard.Keyboard iohub backend that does not require making 2 RPC calls to get the keys so it should be more efficient. My version of the fix is [here](https://github.com/isolver/psychopy/commit/905ca378c01ee2c0db8b076254e0e29c6efff7e6). @lboel, let me know what you think.

lboel:
You`re right. Multiple calls of getRelease/Pressed should be avoided. I just wasnt about the getKeys-Behaviour. I ve just notived that KeyPress()-Class-Code-Property could actually be filled with the corresponding Char-Code (iohub can provide the charcode and I dont see why this should be kept as "None" if the hardware.keyboard.KeyPress-Class supports it.

isolver (approved):

isolver:
Looks great, thanks again.

peircej:
We've now discovered that this PR has caused a rather worse unexpected behaviour - it means that Builder experiments are now treating both key-ups and key-downs as responses. So a response occurs twice!

If the issue was a lack of a duration value being returned, I think we can live with that better than reporting key presses as 2 events so we're looking at reverting the switch from `getKeys()` back to `getPresses()` somewhat urgently!

lboel:
Hey @peircej 
Oh, no! Thats really odd. Just had a look into it with the new release.  
My original approach was to use both (release,press) functions separately and really only changing the waitRelease Behaviour, but as isolver pointed out, this was to ressource heavy, in terms of multiple function calls.
This may be an option: https://github.com/psychopy/psychopy/pull/4846/commits/10fd1cfb9ba8b7a416b5759d9576a86169573c52
Looking further in the reverse function approach it turned out, that:

```python
for idk, k in enumerate(all_key_events):
                    if hasattr(k, 'pressEventID'):
                        released_press_events.append(k)
                        key_events.append(k)
                    else:
                        if not any(k.key == x.key for x in key_events):
                            key_events.append(k)
```
on https://github.com/psychopy/psychopy/blob/5d79806a5be72871262dabccf2eb83a7415e2522/psychopy/hardware/keyboard.py#L318 could do the trick and removing the 2 events one press behaviour. 

After that i noticed that the Keyboard._iohubKeyboard.getKeys(clear=True) only works if explicitly called with waitRelease=False - key_resp.getKeys(clear=True,waitRelease=False). 

If I call something like key_resp.getKeys(clear=True,waitRelease=False) at the beginning of the trial my demo works as expected:

bexp

If I just call key_resp.getKeys(clear=True) (that should clear all buttons as well, the pressed events remain, although Keyboard._iohubKeyboard.getKeys(clear=True) should have removed them:

bexp2

But still haven`t figured out why...

Sorry for the inconvenience. Still want to use python/psychojs for a wheel turning task so duration returning (null if still pressed and with time if released) would be greatly needed for continuous key responses. We may find a robust solution to get the desired and documented behaviour somehow. 
Here the used demo file:
[keyreleaseTests.zip](https://github.com/psychopy/psychopy/files/9018405/keyreleaseTests.zip)

isolver:
Hi @lboel,
I think we fixed the issue @peircej raised above with this PR https://github.com/psychopy/psychopy/pull/4990, which has now been merged with `release`. The issue was that the iohub backend logic was including release only events since the last call to `keyboard.clearEvents()` in the keys returned by `keyboard.getKeys(...)`, instead of only press or press+released event combo's since the last call to `keyboard.clearEvents()`.
This fix is compatible with the fix for the original issue you raised, but let me know if you are still seeing issues. 
Thanks again

lboel:
Great! Thank you! @isolver

PR #4867, opened 2022-05-30
BF: fixed long wait times causing windows to mark psychopy as unrespo…

### LINKED ISSUE #2995: core.wait doesn't dispatch_events during long waits causing problems in Windows
Per [https://discourse.psychopy.org/t/screen-stucks-when-using-core-wait-function/9914/2](https://discourse.psychopy.org/t/screen-stucks-when-using-core-wait-function/9914/2)

Long calls to core.wait spend too much time in `time.sleep()` causing Windows to treat the window as non-responsive.

Suggested fix in clock.py, check for events to dispatch during the non-hogCPU period every (e.g.) .0998 s. Code example below (different from link) adds a _hogCPUactivity function to help and in the case of sub-second wait times is approximately unchanged.
```
def _hogCPUactivity():
    """Helper function for :func:`~.psychopy.core.wait
       
       Handles window event if needed or returns otherwise.
    """
    
    from . import core
    
    if not (core.havePyglet and core.checkPygletDuringWait):
        return
    # let's see if pyglet collected any event in meantime
    try:
        # this takes focus away from command line terminal window:
        if parse_version(pyglet.version) < parse_version('1.2'):
            # events for sounds/video should run independently of wait()
            pyglet.media.dispatch_events()
    except AttributeError:
        # see http://www.pyglet.org/doc/api/pyglet.media-module.html#dispatch_events
        # Deprecated: Since pyglet 1.1, Player objects schedule themselves
        # on the default clock automatically. Applications should not call
        # pyglet.media.dispatch_events().
        pass
    for winWeakRef in core.openWindows:
        win = winWeakRef()
        if (win.winType == "pyglet" and
                hasattr(win.winHandle, "dispatch_events")):
            win.winHandle.dispatch_events()  # pump events
            
def wait(secs, hogCPUperiod=0.2):
    """Wait for a given time period.

    If secs=10 and hogCPU=0.2 then for 9.8s python's time.sleep function
    will be used, which is not especially precise, but allows the cpu to
    perform housekeeping. In the final hogCPUperiod the more precise
    method of constantly polling the clock is used for greater precision.

    If you want to obtain key-presses during the wait, be sure to use
    pyglet and to hogCPU for the entire time, and then call
    :func:`psychopy.event.getKeys()` after calling
    :func:`~.psychopy.core.wait()`

    If you want to suppress checking for pyglet events during the wait,
    do this once::

        core.checkPygletDuringWait = False

    and from then on you can do::

        core.wait(sec)

    This will preserve terminal-window focus during command line usage.
    """
    
    t0 = getTime()
    
    # initial relaxed period, using sleep (better for system resources etc)
    if secs > hogCPUperiod:
        #Waiting for more than a second - break a little under every .1 s 
        if secs > 1:
            delta = .0998
            deltas = int((secs-hogCPUperiod)/delta) - 2
            for brief_interval in range(deltas):
                time.sleep(delta)
                _hogCPUactivity()

        #Waiting for less than a second, wait for the specified time 
        else:
            time.sleep(secs-hogCPUperiod)
    
    # hog the cpu, checking time
    while (getTime() - t0) < secs:
        _hogCPUactivity()


```

maqadri:
…nsive

Separated the "hogCPU" function as repeatedly call it every tenth of a second, allowing events to dispatched at a reasonable rate

fixes GH-2995


maqadri:
Fixed the continue - made sure this one runs.

peircej:
Thanks @maqadri 
I've asked @mdcutone to take a look too but I *think* this could be simplified and still do what you'r aiming for. 

I'm suggesting:
- rename your `hogCPU` to be `dispatchEvents` (nothing in the function itself is hogging the CPU - it's just the fact that we're timing in a tight loop)
- simplify the first part, before hogging. The point is that, at that part of the code the timing is not precise anyway, because we're a long way still from the deadline
- also, I've heard reports from other places that allowing very brief time.sleep period even during the tight loop actually improves timing by allow time for the OS to do housecleaning. I haven't tested myself but it seems plausible. The issue is that on win32 (prior to Python 3.11) time.sleep will sleep for 1ms even if you ask it to sleep for less. So not sure what we want to do about that

Anyway, the timing code from the above ideas results in this simpler version:

```
if (secs - hogCPUperiod) > 0.1:  # we have at least 100ms left so can be relaxed)
    _dispatchEvents()
    time.sleep(0.01) # sleep for 10ms

# hog cpu for this period
while (getTime() - t0) < secs:
    _dispatchEvents()
    time.sleep(0.00001) # 0.1ms so OS doesn't complain
    ```

PR #4871, opened 2022-06-01
Feature phidgets

maqadri:
Phidget relays are used by a few different labs for controlling operant chamber events. A component would be easier than replicating code components across experiments.


psychopy/experiment/components/Phidgets/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+                 reversedRelay=False,
+                 syncToScreen=True):
+
+        super(PhidgetRelayComponent, self).__init__(
+            exp, parentName, name,
+            startType=startType, startVal=startVal,
+            stopType=stopType, stopVal=stopVal,
+            startEstim=startEstim, durationEstim=durationEstim)
+
+        self.type = 'PhidgetOutputRelays'
```
peircej, line 49: If the component class is called `PhidgetRelayComponent` the .type needs to be `PhidgetRelay`. This is used at times to work out what Component class we need to call

docs/source/builder/components/phidgets.rst
```diff
@@ -0,0 +1,48 @@
+.. _phidgetRelayComponent:
+
+Phidget Relay Component
+--------------
```
peircej, line 4: in ReST format underlines should be (at least) the length of the characters that make the title (it compiles but generates a warning in sphinx, which we're trying to reduce)

psychopy/hardware/phidgets.py
```diff
@@ -0,0 +1,88 @@
...
+                digitalOutput.setChannel(channel)
+                digitalOutput.openWaitForAttachment(100)
+                all_phidgets[all_phidgets_code] = digitalOutput
+            else:
+                digitalOutput = all_phidgets[all_phidgets_code]
+
+            self.phidget_outputs.append(digitalOutput)
+
+    def __del__(self):
+        self.openRelay()
```
peircej, line 63: Should this be `closeRelay()`? It seems weird to open when deleting.

psychopy/hardware/phidgets.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+#Future - add analog input, digital input, digital output (5V) via 8/8/8 Interfacekit
+
+from psychopy import prefs, logging
+from psychopy import core
+
+all_phidgets = {}
+
+class phidgetOutputComponent:
```
peircej, line 20: Capital P please, as in PhidgetOutputComponent. Our classes start with a capital, methods start with lower case

peircej (changes requested):
This looks great. Thanks for your contribution! I haven't tested it as I don't have access to the hardware but it looks good, based on simply reading. At the moment the tests don't pass because of the self.type being incorrect. Here I've included comments directly related to specific code but there are further points for discussion in the pull-request "Conversation"

peircej:
Regarding the logo, have you spoken to Phidget themselves about this to check they're OK with using it on something they didn't write/test themselves? I note you've stated this is unrelated to them in the docs. 

Also regarding the icon, I note that the Component/class you've provided here is specifically for the Relay Phidget but the icon is for Phidgets in general. It feels like we need some way to identify this as the Relay so that in future other Phidgets could be added.

FYI in future we're planning to move to a plugin model for features like this that aren't going to be used widely. We haven't yet settled on how to make those plugins searchable. and installable. We'll get in touch when we do that, about migrating PhidgetRelayComponent into an independent plugin.

peircej:
If you want to have a verbal (video?) chat about any of this ping me an email

peircej:
I'm closing this for now. I like the idea of supporting Phidgets, and with the new plugins options this will be easier in the future

maqadri:
Thanks for the feedback; responses below. Sorry about the silence -- bad semester.

I've gone through and packaged this as a plugin, available via pip as psychopy-phidgets. Further feedback appreciated, though not sure of your interest in the peripheral plugin packages.

I realized that the current app version does not load component automatically (even though it has the "start up plugins" key in the preferences) -- do you have an idea of when the next version will be available or how I can forward-proof sharing this plugin?

- updated self.type to match the name sans the Component
- renamed phidgetOutputRelay to PhidgetOutputRelay
- updated rst file - not sure of its value if it's no longer part of the psychopy distribution
- "open" relays are the default state (i.e., in an unpowered relay, the state is "open")

On that last point, more philosophically, opening/closing relays is a super weird thing altogether in the philosophy of the builder - I think a future setup would have a more configurable three-part system - routine start value (closed, open, don't change), routine end value (same), and component action (close-open or open-close). I would need to figure out the dropdown/checkbox setup for the preferences/dialogs.

PR #4905, opened 2022-06-13
BF: Normalise field size before refreshing dot stim vertices

### LINKED ISSUE #4902: DotStim size parameter does not correspond to displayed size
The size attribute of a DotStim does not correspond to its actual displayed size. For example, below are two polygons each set to have a size of (2.5, 2.5) deg, and a DotStim notionally of the same size and units:

dotstim_size_issue_25

This is not a diameter/radius confusion, as doubling the size of the polygons still does not match the extend of the DotStim field (see https://discourse.psychopy.org/t/dotstim-fieldsize-argument-has-unclear-units/29520/3 )

When using other units, such as `height` rather than `deg`, the displayed size does not seem to lawfully correspond to the specified values (see the initial post in the Discourse thread above).

***This then raises questions about whether the specified dot speeds are correct, if the stimulus doesn't know its own angular extent.***

Below is another example, where all three stimuli were set to have sizes of (0.5, 0.5) in height units. In this case, the DotStim is now markedly smaller than the reference polygons:
dotstim_size_issue_05_height

TEParsons:
fixes #4902

Also added test to catch this in future

### DIFF
--- a/psychopy/tests/test_visual/test_dots.py
+++ b/psychopy/tests/test_visual/test_dots.py
@@ -1,12 +1,130 @@
 import numpy as np
-from psychopy import visual
+from psychopy import visual, colors
 
 
 class TestDots:
 
     @classmethod
     def setup(self):
-        self.win = visual.Window([128, 128], pos=[50,50], allowGUI=False, autoLog=False)
+        self.win = visual.Window([128, 128], monitor="testMonitor", pos=[50,50], allowGUI=False, autoLog=False)
+
+    def test_fieldSize(self):
+        """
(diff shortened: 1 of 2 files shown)


TEParsons:
@m-macaskill are there any other combinations of fieldSize/units which you've come across causing problems? If so I'll add them to `cases`

m-macaskill:
That's a really clever test - I wasn't sure how this could be done but that is a quite elegant solution.

I've only tried `height` and `deg` units, but they are probably sufficient in terms of covering both relative an absolute coordinates. If something goes wrong, either or both should be affected. Cheers.

PR #4907, opened 2022-06-13
ENH: Add switch user button for Pavlovia

TEParsons:
Needed a quicker way to do this while testing a bug so thought I may as well add it as a button, seeing as we already have all the necessary functions (log out, log in, get saved users, etc.) it's trivial to implement


TEParsons:
image

PR #4908, opened 2022-06-13
BF: Paste component at end of routine rather than index -1

TEParsons:


peircej:
Task linked: [CU-29j55zn When a component is copied and pasted, it is placed second from the bottom rather than at the bottom of the routine](https://app.clickup.com/t/29j55zn)

PR #4912, opened 2022-06-13
BF: Fixed transformation of mouse coordinates on Retina displays

mdcutone:
Not sure if something changed but mouse coordinates are off when using Retina displays. This patch should fix the issue, but needs testing on a Retina monitor. Works fine on Macs hooked up to a regular display, so nothing should break on those hardware setups and Windows/Linux.


peircej:
No, this wasn't the issue. The issue was with the reported mouse position, not the scaling of stimuli. Those were fine.

The real fix turned out to be an error where scaling was applied as if using glfw backend but when we were actually using pyglet https://github.com/psychopy/psychopy/pull/4924

PR #4915, opened 2022-06-14
BF: Show asterisk on Code component if EITHER code has contents

TEParsons:


peircej:
Task linked: [CU-m3bdm4 If there is code in the Code Component JS box (but not in the Py) still count that as "code present" and open that tab by default](https://app.clickup.com/t/m3bdm4)

PR #4919, opened 2022-06-14
BF: State visibility explicitly when sending query to Pavlovia API

TEParsons:
This isn't needed for other filters as these don't affect whether or not authentication needs to be done, so the server can afford to assume blank = no filter for these, but visibility needs to be treated as a special case as showing private projects means authenticating, which is a more significant undertaking, computationally.


peircej:
Task linked: [CU-2er6yau Cannot find own private projects when searching from Builder. Issue first appeared early 2022.](https://app.clickup.com/t/2er6yau)

PR #4929, opened 2022-06-15
BF: Specify data type when setting status/keywords on Pavlovia project

TEParsons:


peircej:
Task linked: [CU-1z67491 Changing project status from inactive to running in pavlovia UI doesn't "work"](https://app.clickup.com/t/1z67491)

PR #4931, opened 2022-06-15
BF: Hide mouse on request, but disable option when windowed

TEParsons:


peircej:
Task linked: [CU-2d4mj55 mouse set visible in experiment does not alter visibility of mouse](https://app.clickup.com/t/2d4mj55)

peircej:
Task linked: [CU-2gbhpzp Mouse cursor is visible even with Experiment settings > Screen > Show mouse deselected](https://app.clickup.com/t/2gbhpzp)

peircej:
I thought about this some more and I wonder if we should only hide the setting, rather than also changing its value to True when in windowed mode. The thing is, if you switch to window-mode to do some debugging and then switch back it's annoying to have to change the setting. Let's just hide the setting which can be implicit for not hiding the mouse.

Issue #4828, opened 2022-05-10
Apparent off-by-one (or more) timing error in Builder scripts


m-macaskill:
Builder components can record their`.started` and `.stopped` attributes in the datafile, using an absolute time in seconds. In this minimal example, the first text stimulus `t_0` has a specified onset time of 0 s relative to the start of the trial, with subsequent text stimuli having onsets at 2 and 4 s:

<img width="662" alt="Screen Shot 2022-05-10 at 19 42 18" src="https://user-images.githubusercontent.com/931027/167575773-0cff245b-d4fb-4494-a249-ae830ba51da5.png">

Over two separate runs of five trials, subtracting the times of onset between the three stimuli gives the following relative onset times (which should be 2.0, 4.0, and 2.0):

trial | t2-t0 | t4-t0 | t4-t2
--|-- | -- | --
0 | 1.9551 | 3.9537 | 1.9986
1 | 1.9833 | 3.9839 | 2.0006
2 | 1.9831 | 3.9840 | 2.0009
3 | 1.9831 | 3.9834 | 2.0003
4 | 1.9836 | 3.9842 | 2.0006

trial | t2-t0 | t4-t0 | t4-t2
-- | -- | -- | --
0 | 1.9658 | 3.9659 | 2.0001
1 | 1.9841 | 3.9834 | 1.9993
2 | 1.9655 | 3.9837 | 2.0182
3 | 1.9859 | 3.9836 | 1.9977
4 | 1.9678 | 3.9682 | 2.0004

(1) The intervals between the onset of stimuli `t_4` and `t_2` (the third column) are effectively exactly 2.0 seconds, as intended.
(2) The intervals between the onset of `t_0` and the later stimuli are shorter than intended, by 1, 2 or even more frames.
(3) When subtracting the time of onset of `t_4` from its time of onset on the previous trial, it was short by approximately 1 frame. i.e. this is an empirical measure of the total routine duration, including transitions between one routine and the next, and should be 5.0 seconds in this case. The shortened duration means the timing error would tend to accumulate over loop iterations. This might particularly affect fMRI tasks, and other studies which require multiple consecutive trials on a fixed schedule.

The error in relative onset times seems to apply only with respect to stimuli appearing at t = 0, as the relative times between the later stimuli are correct. This would be consistent with the onset of the first stimulus being delayed (e.g. a dropped frame). But that wouldn't be consistent with the overall duration of the trial being apparently shortened by a frame.

[...]

PR #4933, opened 2022-06-16
BF: non-slip timing applied to all Component times rather than the Routine endpoint

peircej:
Previous implementation of non-slip timing was to check a timer at the end of the Routine. That works fine if things typically over-run but fails to control the issue if stimuli are tending to under-run because they will have terminated before the check has a chance to have any impact.

This commit revamps that logic so that it can be applied to every stimulus start/stop time rather than just the end time of the Routine

New implementation:
- switch from a count-down timer to a regular (count-up) Clock for `routineTimer`
- at the end of each Routine we perform some form of reset on the `routineTimer`, ready for the next Routine:
  - if the current Routine had a known duration (non-slip safe) then simply subtract that expected duration so that the routineTimer now shows zero +/- any under/over-shoot
  - if the current Routine was not non-slip safe then we may as well do `routineTimer.reset(0)` so we don't try to account for any offsets

This is actually simpler than beforehand, where we had different clocks for each Routine. Now there's just one and a simple rule for its reset


peircej:
fixes GH-4828

PR #4957, opened 2022-06-20
Fix setting color has no effect

jakobchwastek:
Only setting the color parameter which should set both the 'fillColor' and 'lineColor' has no effect since 'Polygon' parameter 'fillColor' is 'white'. BaseShapeStim then would draw the default white color instead of the color handed over to Polygon as argument.

### DIFF
--- a/psychopy/visual/polygon.py
+++ b/psychopy/visual/polygon.py
@@ -137,6 +137,10 @@ class Polygon(BaseShapeStim):
         self.__dict__['edges'] = edges
         self.radius = np.asarray(radius)
         self._calcVertices()
+        
+        # Set fillColor to False if color is used
+        if color is not None:
+            fillColor = False
 
         super(Polygon, self).__init__(
             win,


TEParsons:
I can see the logic here - in theory, BaseShapeStim should handle this kind of stuff, but as you point out Polygon having a default fillColor of "white" kind of messes with that as undefined fillColor is no longer `False`... I wonder if a better solution is maybe to have the default fillColor for Polygon also be `False`, and instead have the end of the `if else` statement handling fillColor in BaseShapeStim set it to "white" rather than `None`; as setting fillColor to `False` when color is defined means that color overrides fillColor, which is the opposite of how all other BaseShapeStim-derived classes work.

TEParsons:
It's not often I'll admit JS does it better, but distinguishing between `None` (aka transparent) and `undefined` would actually be very useful here :')

peircej:
Thanks so much @jakobchwastek and sorry for delay pulling this in but you're quite right about the bug and the fix :-) 
We're looking at whether we can make this more intuitive than it currently is so that this won't confuse people (including us) in the future, and that took some time but I think it's time to pull in your fix as a good start and work through any refactoring separately in the future

thanks again

PR #4963, opened 2022-06-22
FF/NF: Multi-threaded camera for realtime video recording

mdcutone:
PR adds various improvements increase the real-time performance of `psychopy.hardware.Camera`, done by multi-threading video file writing and rendering. This allows for video streams to be recorded from live feeds while the previous video is still being rendered. The camera can now start/stop recordings very quickly without disrupting the flow of the experiment. The video rendering step is also threaded, so the `Camera.save()` command doesn't block. The most recent clip saved will appear on `Camera.lastClip` once the rendering process is complete.


mdcutone:
Made the adjustment for `RECORDING` -> `STARTED` for the purpose of Builder.

PR #4968, opened 2022-06-22
build(deps): removed redundant version constraints

ErikBjare:
Also removed not needed pathlib (needed for <3.4, but psychopy currently targets >3.6)

Found it as was looking into the crazy amount of time it to poetry to generate a lockfile for eeg-notebooks: https://github.com/NeuroTechX/eeg-notebooks/pull/22

No idea if it actually impacts the time taken for lockfile generation, though.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -75,17 +75,16 @@ install_requires =
     # Platform-specific dependencies.
     python-vlc == 3.0.11115; platform_system == "Windows"
     python-vlc >= 3.0.12118; platform_system != "Windows"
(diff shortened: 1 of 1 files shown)


ErikBjare:
Feel free to squash the title into "FF: removed redundant version constraints for dependencies"

peircej:
Regarding poetry, we should be providing a more explicit release listing of the versions to use so it doesn't need to compute anything at all, but we haven't got around to the mechanics of that yet

PR #4972, opened 2022-06-23
FF: Fix performance drop with growing text buffer size in Coder shell

mdcutone:
Fixes an issues where `_applyThemes` grinds the console to a halt when dealing with large buffers. Fonts can spontaneously change though but there is no way around this in the current implementation. Might make the shell based off `StyledTextCtrl` eventually to get around font styling performance drops.


TEParsons (approved):
Code looks fine and pulling down it seems to style just fine

TEParsons:
Is this one for the first bugfix release? If so it's safe to pull into release rather than dev now as Jon has merged them

PR #4976, opened 2022-06-24
BF: Set markerPos AFTER rating when resetting slider

TEParsons:
Otherwise Slider resets with no value regardless of startVal


peircej:
Task linked: [CU-2htajg9 Start value doesn't show on slider (locally)](https://app.clickup.com/t/2htajg9)

PR #4979, opened 2022-06-27
BF: Recreate components on copy rather than pickle

TEParsons:
Pickling components was causing some lag when doing it a lot as it meant deep copying the entire experiment for the `.exp` attribute. Instead, we can recreate the component & assign params like we do when a psyexp is loaded, supplying the new experiment context rather than creating from deep copy.


peircej:
Task linked: [CU-2f9bd5u Copying and Pasting Routines/Components eventually causes PsychoPy App to become very laggy or non responsive](https://app.clickup.com/t/2f9bd5u)

PR #4988, opened 2022-06-29
NF: PIP interface

mdcutone:
Testing build to see if it's possible to install PIP packages from within PsychoPy. You can access the command prompt by going to Coder > Tools > PIP Install... 

You can enter commands like `pip install numpy` or `pip uninstall -- yes numpy`. When doing an uninstall, make sure `--yes` is specified or you cannot exit the subprocess and the whole app will stall.

### DIFF
--- a/psychopy/app/plugin_manager/pipManager.py
+++ b/psychopy/app/plugin_manager/pipManager.py
@@ -1,0 +1,65 @@
+import wx
+import wx.richtext
+
+from psychopy.app.themes import handlers, fonts
+from psychopy.localization import _translate
+
+
+class PIPManagerDlg(wx.Dialog, handlers.ThemeMixin):
+    """
+    Interface for interacting with PIP within the standalone PsychoPy environment.
+    """
+
+    def __init__(self):
+        wx.Dialog.__init__(self, None, title="PIP installer",
+                           size=(480, 240),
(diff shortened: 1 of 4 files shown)


psychopy/app/plugin_manager/pipManager.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class PIPManagerDlg(wx.Dialog, handlers.ThemeMixin):
+    """
+    Interface for interacting with PIP within the standalone PsychoPy environment.
+    """
+
+    def __init__(self):
+        wx.Dialog.__init__(self, None, title="PIP installer",
+                           size=(480, 240),
```
TEParsons, line 18: Could do with being a bit taller - 480x480 works well imo.

psychopy/app/plugin_manager/pipManager.py
```diff
@@ -0,0 +1,81 @@
...
+                           style=wx.DEFAULT_DIALOG_STYLE | wx.DIALOG_NO_PARENT)
```
TEParsons, line 19: Could you add  `| wx.RESIZE_BORDER` to this?

TEParsons (changes requested):
Looks great! Just needs to be a bit taller I think, and resizeable

PR #4989, opened 2022-06-29
BF: iohub keyboard.Keyboard backend getKeys(waitReleasee=False)

isolver:
keyboard.Keyboard iohub backend getKeys bug was introduced that made getKeys(waitReleasee=False) return a key even if it was pressed prior to the last keyboard.clear(). Now iohub backend only returns keys that have been pressed, or pressed+released, since last clear.


isolver:
targeted wrong branch...closed

PR #4990, opened 2022-06-29
BF: keyboard.Keyboard iohub backend getKeys bug

isolver:
keyboard.Keyboard iohub backend getKeys bug was introduced that made getKeys(waitReleasee=False) return a key even if it was pressed prior to the last keyboard.clear(). Now iohub backend only returns keys that have been pressed, or pressed+released, since last clear.

### DIFF
--- a/psychopy/hardware/keyboard.py
+++ b/psychopy/hardware/keyboard.py
@@ -311,18 +311,23 @@ class Keyboard:
             else:
                 key_events = []
                 released_press_evt_ids = []
-
                 all_key_events = Keyboard._iohubKeyboard.getKeys(keys=watchForKeys, clear=clear)
-                all_key_events.reverse()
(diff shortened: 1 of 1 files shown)


psychopy/hardware/keyboard.py
```diff
@@ -392,6 +397,7 @@ def waitKeys(self, maxWait=float('inf'), keyList=None, waitRelease=True,
 
     def clearEvents(self, eventType=None):
         """"""
+        print("clearEvents")
```
mdcutone: Remove this.

mdcutone (changes requested):
Works fine with the test experiment. Behaves like expected. Just remove the stray `print()` statement before merging.

mdcutone (approved):

PR #4991, opened 2022-06-29
BF: Fixed path appending in `getComponents`

mdcutone:
Fixes an issue where we try to append a path to a non-existent value in `getComponents()`. Unless we're doing something weird here by monkey-patching the `os` module with `sys`, I'm sure the desired behaviour is what I changed it to. Surprised we didn't see an error or some other ill effect.


peircej:
I'm moving this to `dev` unless we identify some bad effect that it caused. i.e this is a very small refactor if there wasn't any bug

PR #5001, opened 2022-07-06
ENH: Add support for user defined iohub calibration target graphic

isolver:
Added support for user defined iohub calibration target graphic when using eyelink or MouseGaze trackers. Other eye trackers will be added in future PR's.

For Builder, see the demos/builder/Feature Demos/eyetracking_custom_cal project. The Builder ET Calibration Routine is replaced by a snippet of custom code so that the custom target can be defined.

In general, to setup iohub eye tracker calibration to use a custom target stim update the eye tracker calibration settings as follows:

1. Set eyetracker calibration:target_type to CUSTOM
2. Add to calibration:target_attributes:
```
        target_attributes:
            custom:
                module_name: name of module that class is defined in
                class_name: name of class to use as calibration stim
                class_kwargs: dict of any kwargs to use when creating the stim. The win kwarg is added by iohub at runtime.
```
for example, to use a movie as the cal / val target for an eyelink calibration / validation:

```
    calibration:
        target_type: CUSTOM
        target_attributes:
            custom:
                module_name: psychopy.visual
                class_name: MovieStim3
                class_kwargs:
                    filename: mymovie.mov
                    loop: True
```

For class_kwargs `image` and `filename`, path can be absolute or relative to the experiment working directory.

TODO:
- [x] create Builder example using custom code component to demo custom calibration target
- [x] Add MouseGaze support
- [x] remove test .mov from simple.py demo folder
- [x] revert simple.py demo


isolver:
@TEParsons , do you mind looking over the visual.TargetStim changes? The idea is to start moving away from using .inner and .outer in calibration code (still keep it internally for backwards compat) and instead use the stim .size type attribues so more calibration logic can be shared between different stim types during calibration. Thank you

psychopy/visual/target.py
```diff
@@ -138,17 +130,23 @@ def lineWidth(self, value):
...
+        TODO: Deprecate use of radius in favor of using .size
+        :return:
+        """
         return self.outerRadius
 
     @radius.setter
     def radius(self, value):
         # Set outer radius
         self.outerRadius = value
-        # Set inner radius to maintain scale
```
TEParsons, line 147: So rather than maintaining the ratio between inner an outer, setting radius sets the outer radius and setting innerRadius sets the inner? I guess this is fine, but what's the reason for the change? It does mean that if you wanted to expand the whole thing (maintaining ratio) it's harder to do

isolver: The purpose of the my original target stim was to never change the inner circle radius when the target expands / contracts, only the outer circle radius. This was very intentional. So this is more in line with the original intent of the stim and keeps it operating as before.

> I guess this is fine, but what's the reason for the change?

Switching to use `size` makes the calibration code for different stim types much more sharable / the same as possible across stim types.

> It does mean that if you wanted to expand the whole thing (maintaining ratio) it's harder to do

This was not possible before during calibration / validation either. Now if a user wants that type of behaviour they can do it using an image or any custom stim they make.

TEParsons: Cool, all seems reasonable :) In that case consider my review approved

TEParsons (approved):

PR #5004, opened 2022-07-11
BF: If routine force ended, reset clock rather than subtracting expected dur

TEParsons:


peircej:
Task linked: [CU-2k758du Mouse doesn't end routine (stimuli are "undrawn" but routine progresses only after stimuli duration)](https://app.clickup.com/t/2k758du)

PR #5007, opened 2022-07-12
NF: Add FrameAnimation class for animating images

TEParsons:


TEParsons:
Closing this as it'll make more sense in an "Animation Tools" plugin once plugins are a thing

PR #5008, opened 2022-07-12
RF: Improved `clock.wait()` routine with periodic event dispatching 

mdcutone:
Allows for event dispatching during long intervals bit also improves the precision of the `wait()` function. Credit to @maqadri for helping solve this issue (original PR: https://github.com/psychopy/psychopy/pull/4867)


peircej:
Fixes GH-2995

PR #5009, opened 2022-07-13
BF: choice labels are not translated

hsogo:
This issue was solved by #4840, but was reverted by a8b119a88a53e6598be6a7ead356de088c3afa61.
In the `ChoiceCtrl.__init__()` function, we have to build and translate choice labels.  Both of these tasks were done at the same time in the previous codes, which caused troubles in translation when the codes for building the label list were modified.  In this fix, I separated building and translating labels.  I hope this fix will prevent troubles in the translation of choice labels.


peircej:
sorry that got excluded by the merge conflict. Thanks for tracking and resubmitting. This will be in 2022.2.2 (imminent due to an error using non-slip timing with force-end-trial events)

PR #5010, opened 2022-07-13
BF: ROI status wasn't being set to STARTED as intended

TEParsons:


peircej:
Task linked: [CU-2jvmz43 Eyetracker ROI started/finished flags in boilerplate](https://app.clickup.com/t/2jvmz43)

peircej:
What was the symptom of this from the user perspective? Did the ROI never end?

PR #5017, opened 2022-07-18
BF: pip install error for conda envs

davidmwatson:
`setup.cfg` was changed to remove redundant version constraints ( #4968 ), but `setup.py` still contains reference to old name for pyqt5 package during check for conda environments. This results in an error when attempting to pip install in a conda environment. Updating the name in `setup.py` seems to fix the issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ if 'CONDA_PREFIX' in os.environ:
 # `pyqt` package should be installed via conda instead
 # cf. https://github.com/ContinuumIO/anaconda-issues/issues/1554
 if 'CONDA_PREFIX' in os.environ:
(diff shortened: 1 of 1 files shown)


peircej:
@davidmwatson thanks. I've noticed that your fix refers to changing a name in setup.cfg but what you've done is remove the unnecessary version requirement. Jus tchecking there wasn't an additional change that was missed from the pull request?

davidmwatson:
Yes, the `setup.cfg` was changed previously by someone else to remove the version requirement, but this meant it didn't match the name in the `setup.py` file anymore, which resulted in an error when trying to run the setup. I've just updated `setup.py` so that it matches the details in `setup.cfg`.

PR #5018, opened 2022-07-18
BF: Handle stop time of None for Camera

TEParsons:


peircej:
Task linked: [CU-2k75f21 if camera has no duration only first frame of movie is saved](https://app.clickup.com/t/2k75f21)

PR #5020, opened 2022-07-18
NF: Add CheckBoxStim, will allow bool ctrls in Form

TEParsons:


TEParsons:
Next stop: Drop down controls. Then: The world!

PR #5022, opened 2022-07-19
NF: Add a drop down control

TEParsons:
Uses a ButtonStim for the persistent ctrl, and a Slider for the context menu. Also adds a new style to Slider to facilitate this, which can be used by itself.


TEParsons:
Minimal example:
```
from psychopy import visual, event

win = visual.Window(units="height")
obj = visual.DropDownCtrl(win, pos=(0, 0.3), size=(0.4, 0.1), choices=("one", "two", "three"))

while not event.getKeys(["escape"]):
    obj.draw()
    win.flip()
```

PR #5024, opened 2022-07-20
Hotfix linebreak class

alexander-pastukhov:
## Issue: very slow import when _debugging_.
Problem file: `psychopy/tools/linebreak_class.py` used in `linebreak.py`
Current implementation stores all linebreak codes in a massive dictionary (300K+ lines). This works for normal running (~0.6 s on my setup) but slows initialization (imports) dramatically when debugging (~60 s). The slow start up times when debugging are general issue replicated both on Windows and Macs. 

## Solution
Codes and classes are stored in a CSV, which is read and converted to dictionary. 
Impact on the rest of the code: none.
Impact on start up (import) times: in both cases on my setup ~1 s (slower than original for normal execution but no slow down for debugging). 

## How the timing was tested
import time
start = time.time()
from psychopy.tools.linebreak import line_break
print(time.time() - start)


peircej:
Do you actually need uax14 linebreaks in your code?
I'd suggest we move the import statement to inside the 
```
        elif self._lineBreaking == 'uax14':
            from psychopy.tools.linebreak import get_breakable_points, break_units
```

then it won't take time for most users at all

alexander-pastukhov:
```
elif self._lineBreaking == 'uax14':
     from psychopy.tools.linebreak import get_breakable_points, break_units
```
Importing inside the method is so out-of-the-box, it did not occur to me but you are right, it is only needed within the method's scope, so would work. Still be a pain to debug for people who _do_ need uax14 linebreaks. I'll close this pull request and create a new one with your solution. Thanks!

PR #5025, opened 2022-07-20
ENH: Give designer control over when keypresses register

### LINKED ISSUE #4863: Textbox component receiving input from Keyboard component in separate routine in prior trial
I put this in discussion but so far no comments, this seems like a bug to me.

**OS (e.g. Win10):** Win10
**PsychoPy version (e.g. 1.84.x):** 2022.1.3 and 2022.1.4
**Standard Standalone? (y/n) If not then what?:** Standard Standalone
**What are you trying to achieve?:**
I have 2 routines in the trial, the first has a textbox component to ask for text input from the participant. I’m detecting when they hit the Enter key to continue onto the next routine in the trial which has a keyboard response component to ask for a specific key response. However, when the experiment iterates onto the next trial, the response for the keyboard object in the previous trial appears in the textbox.

**What did you try to make it work?:**
I’ve tried explicitly setting the textbox text to blank, and calling textbox.reset in the Begin Routine section of a code component. My textbox component has its Text set to “set every repeat”.  I've tried explicitly clearing keyboard buffers.

**What specifically went wrong when you tried that?:**
Neither of the above worked. I seem unable to prevent the response made to the keyboard object from appearing in the textbox on the next iteration.

**Test file here:**
[testTextbox.zip](https://github.com/psychopy/psychopy/files/8794890/testTextbox.zip)

TEParsons:


TEParsons:
While this isn't the best way to fix the issue it was originally intended to fix, this is generally very useful - designers do want to control when a keypress is registered

PR #5027, opened 2022-07-21
ENH: Improve `Color.render` performance for the case when contrast is 1

lkeegan:
This makes `draw()` faster, e.g. here are snakeviz profiles for 100x `Circle.draw()` calls:

Without this PR:

image

With this PR:

image

Code used for above profiles:
```
from psychopy.visual.window import Window
from psychopy.visual.circle import Circle
import cProfile

window = Window(fullscr=True, units="height")
circle = Circle(window, radius=0.1, lineColor="black", fillColor="red")

with cProfile.Profile() as pr:
    for i in range(100):
        circle.draw()
        window.flip()
pr.dump_stats("profile.prof")
```

### DIFF
--- a/psychopy/colors.py
+++ b/psychopy/colors.py
@@ -376,6 +376,8 @@ class Color:
         """
         if space not in colorSpaces:
             raise ValueError(f"{space} is not a valid color space")
+        if self.contrast == 1:
+            return getattr(self, space)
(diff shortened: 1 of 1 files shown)


TEParsons (approved):
Good shout, I can see why this would speed things up!

PR #5031, opened 2022-07-22
BF: Enable Coder save button via EnableTool rather than Enable

TEParsons:
.Enable doesn't seem to work for toolbar tools


peircej:
Task linked: [CU-2mn095j Save icon doesn't work in coder (shortcut keys and save as does)](https://app.clickup.com/t/2mn095j)

PR #5032, opened 2022-07-22
ENH: Fix granularity as 1 for radio slider

TEParsons:
Note: Had to move `style` param to Basic so it's on the same page as `granularity`, otherwise depends check doesn't register


peircej:
Task linked: [CU-2htajmd Can move marker between radio buttons when slider is RADIO (online)](https://app.clickup.com/t/2htajmd)

PR #5038, opened 2022-07-22
RF: import linebreak only for "aux14"

alexander-pastukhov:
**Problem**: psychopy/tools/linebreak_class.py used in linebreak.py
Current implementation stores all linebreak codes in a massive dictionary (300K+ lines). This works for normal running (~0.6 s on my setup) but slows initialization (imports) dramatically when debugging (~60 s). The slow start up times when debugging are general issue replicated both on Windows and Macs.

**Solution:** as suggested by @peircej, import inside the method only if line breaking is "aux14". Solves problem for everyone who uses "default" line breaks.

### DIFF
--- a/psychopy/visual/textbox2/textbox2.py
+++ b/psychopy/visual/textbox2/textbox2.py
@@ -29,8 +29,6 @@ from .. import shaders
 from ..rect import Rect
 from ... import core, alerts, layout
 
-from psychopy.tools.linebreak import get_breakable_points, break_units
-
 allFonts = FontManager()
 
 # compile global shader programs later (when we're certain a GL context exists)
@@ -790,6 +788,8 @@ class TextBox2(BaseVisualStim, ContainerMixin, ColorMixin):
             self._lineLenChars.append(charsThisLine)
 
         elif self._lineBreaking == 'uax14':
+            # not needed for regular line breaks but slows down debugging
+            from psychopy.tools.linebreak import get_breakable_points, break_units
 
             # get a list of line-breakable points according to UAX#14
             breakable_points = list(get_breakable_points(self._text))


peircej:
We could, of course, use both methods I guess. Western language users don't need anything to happen and users of uax14 could get a csv file (we'd need to cache it so we don't reload the file on each call) so their debugger doesn't do so much work 

Longer term, it feels like we could do something to get rid (or at least cut down) this massive file - there's a lot of redundancy in there, like hundreds of thousands of entries for the same thing where we might simply specify a range instead. But we'd need to check whether that was slower on each get() call than the simple dict lookup

@hsogo any thoughts?

peircej:
but I'll pull this for now

alexander-pastukhov:
@peircej Yes, we can reduce it to under a couple of thousands of sequences. That was my first idea because it is indeed very redundant. But, perhaps, this could be a way to construct the dictionary (once), instead of using if-elif in the function itself. This way we can keep fast look up but cut the file size by two orders of magnitude. What do you think?

hsogo:
I'm sorry I left this code for a long time.  I haven't tested how fast (slow) if-elif implementation is.  I'll try it.

hsogo:
I wrote a function that converts codes to line-breaking classes by a long chain of if-sentences.  The code was 7002 lines long.  I measured the time of detecting breakable points of an English text (576 characters) and a Japanese text (625 characters).  The values are the average of 1000 runs (on Windows 11 PC, i9-9820X, python 3.9.2).

| text | dict-based(current) | if-based |
|:---:|---:|---:|
| English | 1.98ms | 2.62ms |
| Japanese | 2.32ms | 28.12ms|

If-based fonction was so slow with the Japanese text because plenty of if-sentences must be processed until codes for Japanese characters are found.  The performance for the English text was much better (because codes of English characters are at the top of the Unicode code list), but still slower than the dict-based method.

Next, in line with @alexander-pastukhov, I wrote a module that generates the dict object when imported.  The code was 5562 lines long.  It must be as quick as the current code within the margin of error.

| text | dict-based(current) | dict-based(new) |
|:---:|---:|---:|
| English | 1.98ms | 1.88ms |
| Japanese | 2.32ms | 2.57ms|

The result was that the new code was slightly slower for the Japanese text.  I compared the current and the new dict object and noticed that the current dict object contains fewer entries (333243) than the new one (351115).  This may affected the speed.  I think that the current dict object contains fewer entries because it is based on uniseg-0.7.1 which was released in 2015.  New characters have been added to the Unicode since that time.

In conclusion, I think we can replace the current `linebreaking_class` code with the new code that generates the dict object when importing.  Can I send a pull request?

PR #5043, opened 2022-07-26
ENH: Translations: Updated Chinese simplified language

shun2wang:


shun2wang:
@peircej  Chinese translations stage 2 is ready for merged

PR #5047, opened 2022-07-26
NF: Scrollable textbox

TEParsons:
Adds `overflow` argument to textbox, which can be "visible", "hidden", or "scroll" (same as CSS). Set to draft while I check with the JS team about scroll behaviour online.


peircej:
Task linked: [CU-2htak2d text in editable textbox able to exceed the bounds of the textbox (locally)](https://app.clickup.com/t/2htak2d)

peircej:
Task linked: [CU-hv3cjn Py: When text overflows the height of textbox, it continues outside of borders. JS: stays within textbox, scrollbar appears. Py follow JS](https://app.clickup.com/t/hv3cjn)

PR #5048, opened 2022-07-27
BF: Setting autodraw on Brush should set autodraw on constituent shapes

TEParsons:
Previously, when Builder calls `setAutoDraw(False)`, it was only setting autoDraw on the brush itself, not on the ShapeStim objects which make up the brush stroke.


peircej:
Task linked: [CU-2mn1e3t Brush needs call to "brush.reset()" at end of routine to undraw the drawing](https://app.clickup.com/t/2mn1e3t)

PR #5062, opened 2022-08-03
RF: Replace file browser toolbar with a panel and context tools

TEParsons:
This may fix the intermittent aui error we've been getting, as using a wx.Toolbar for anything other than a frame is generally inadvisible, but even if it doesn't I think it's a more intuitive way for the file browser to function - more in line with other software.

Before:
image
After:
image


peircej:
Task linked: [CU-2f9cve0 AUI frame manager error](https://app.clickup.com/t/2f9cve0)

PR #5064, opened 2022-08-04
TEST: Update TestTarget artefacts to reflect changes made in #5001

TEParsons:
Resizing TargetStim no longer resizes the inner circle as discussed in #5001, meaning the test suite exemplars were wrong and so tests were failing. Tagging @isolver to make sure the new exemplars look as expected


isolver (approved):
Looks good. Thanks

PR #5067, opened 2022-08-04
BF: When slider is categorical, snap to nearest tick

TEParsons:
Safer than setting granularity to 1 as designers do use radio sliders with custom ticks


peircej:
Task linked: [CU-2p8ajdy Snap ratings to ticks when Slider is categorical](https://app.clickup.com/t/2p8ajdy)

PR #5069, opened 2022-08-08
ENH: Rework FileListCtrl to give more control over values

TEParsons:
Allows for:
- Custom string
- $ syntax so values can be set as code
- Set each repeat on ResourceManager
- Control over whether value is a file or folder

How it looks: 
image


peircej:
Task linked: [CU-1x1kp35 Resource manager: Would be useful if resources could be set through spreadsheet](https://app.clickup.com/t/1x1kp35)

PR #5070, opened 2022-08-08
BF: Handle when value in DictCtrl has a syntax error

TEParsons:


peircej:
Task linked: [CU-2nqae4e Making a syntax error in experiment settings then makes experiment settings impossible to open again (error raised) 2022.2](https://app.clickup.com/t/2nqae4e)

PR #5078, opened 2022-08-11
TextBox2 Color Support

ollie-d:
An enhancement and set of bugfixes for TextBox2.

pr_textbox2

The main enhancement is giving users to add custom colors using a new tag, `<c>`. The format is: `<c=#ffffff>X</c>`, where the color information is given in 6-digit hex format. 'c' was used rather than 'color' to keep raw text as small as possible, however, an alias between 'c' and 'color' can be easily added. The current implementation allows each individual character to be colored independently. The default font color will be used when a color is not explicitly defined.

A further enhancement is the ability for users to apply multiple styles simultaneously (currently `<b>`, `<i>`, `<c=>`). The previous code had some logic that seemed to promote simultaneous `<b><i>`, however, it was not fully functional. Also, there were various bugs in the previous style implementation that have been fixed/changed. For example, previously, `<i><i>X</i></i>` was equivalent to `<b>X</b>`. These issues are no longer present since the `_style` list has been replaced with a new class `Style()`. This is a collection of lists that store individual style information per-character. 

Text is still editable and behaves exactly as it did before. All styles are inherited from the character to the left of the caret (even though I think in some cases it should inherit from the right as well). Only one line was modified in the caret-modifying functions for simplicity.

Using a modified version of `compare_text_timing.py`, I found that the changing text is now slightly slower _<1ms_. However, I think there's a lot of room for improvement by using numpy arrays. Also there are currently regex calls for color which may be eventually replaced with a more efficient parser.

I wasn't sure how to correctly perform tests for other scripts such as Arabic, but I'd be happy to follow any established procedures and make changes/additional pull requests as needed.


psychopy/visual/textbox2/textbox2.py
```diff
@@ -559,63 +562,71 @@ def alignment(self, alignment):
...
+
+        # Handle starting color tag
+        color_values = re.findall(re_hex_pattern, text)
+
+        # Only execute if color codes are found to save a regex call
+        if len(color_values) > 0:
+            text = re.sub(re_hex_pattern, codes['COLOR_START'], text)
+            
+            # Convert color_values from markdown+hex to RGBA (0, 1) for _layout
+            color_values = [Color(v[3:-1], 'hex').render('rgba1') for v in color_values]
```
TEParsons: I think it's likely that this is the source of the slower times you found in your tests - once a Color object is 'render'ed, it caches the value so that future render calls in the same space just return the already calculated value (until an attribute of the Color object has changed). This means you can do:
```
col = Color("white", 'rgb')
for n in range(1000):
    col.render("rgba1")
```
and the time taken will just be first calculation time + the time taken to index an array * 1000, rather than calculation time * 1000. Creating a new Color object means creating a new cache, so:
```
for n in range(1000):
    col = Color("white", 'rgb')
    col.render("rgba1")
```
will be a lot slower. I have an idea for how to do this more efficiently so I'll put in a PR to your branch :)

ollie-d: Thanks for looking at the PR and for your suggestion!

I did originally have a different solution, but didn't find that it was any faster/slower during the aforementioned test:

```
color_values_ = []
for v in color_values:
    h = v[3:-1] # hex string (with or without #)
    h = h.strip('#')
    
    # Map from hex -> RGB (0,255) -> RGBA (0, 1). Alpha always == 1.0
    r = int(h[0:2], 16) / 255.
    g = int(h[2:4], 16) / 255.
    b = int(h[4:6], 16) / 255.
    a = 1. # value doesn't seem to affect rendering in _layout
    color_values_.append((r, g, b, a))
color_values = color_values_
```

Although it's possible that even this is slow (although I did try to optimize this by getting rid of any unnecessary loops)

TEParsons: I don't think I have the permissions to submit a PR to your branch, but this is what I was thinking: https://github.com/TEParsons/psychopy/tree/textbox2_colors

So once you've got a color specified once in text, you cache the object handle so that when the same color is specified again you're getting the rgba value from the same object, meaning it won't calculate it again

ollie-d: I've added you as a direct collaborator, hopefully that fixes it. I can also manually implement your changes if that's more convenient, just let me know (although I will be heading off to bed shortly since it's 2:30)

Your solution makes sense to me and it looks good! I can test/merge it when I wake up later today

TEParsons: That would be fab, thanks! I've submitted a PR now :)

ollie-d:
Accepted the PR and merged dev into my branch. Everything appears to be fine. Thanks again for looking at the PR and contributing to the solution :)

ollie-d:
Hi, just wanted to follow up on this PR. Please let me know if there's anything I need to do on my end :)

TEParsons:
Nothing for now, reviewing PRs is currently at a bit of a road block as there's a packaging issue with the tests - once we've got them passing again in the main dev/release branches we'll be able to review all the PRs that are building up :)

TEParsons:
@ollie-d could you rebase from current dev? The Mac test at least is fixed now

ollie-d:
@TEParsons I synced and rebased. Hopefully everything is still good!

TEParsons:
@ollie-d I've submitted another PR to your branch to handle merge conflicts - the changes look pretty crazy but it's just because of merging current dev

ollie-d:
Merged 👍

ollie-d:
Any updates on this? I'll be giving a talk proposing my PhD thesis work in order to advance to candidacy on October 20th and since one of my studies involves translating a large body of experiments from Presentation -> PsychoPy I will be talking about this feature/PR. 

Currently I'm stating that it is likely to be accepted given our back and forth, hopefully that's okay, but figured I'd check in :)

PR #5080, opened 2022-08-12
BF: Fix allowedLabels 

TEParsons:
Translating the value rather than the label in a choice ctrl meant that, if the value has a translation stored, the translated value would replace the label even if the label is something very different.


hsogo:
I think that the key and the value in `_labels` are generally the same.  The only exception I know is "Audio latency priority" in the experiment settings dialog.  This case has no problem because the labels are translated in advance at lines 668-672 of psychopy/app/preferenceDlg.py (Please let me know if there are other exceptions).

Anyway, if there is a possibility that labels that have different key-label pairs in the future, this fix may prevent potential troubles.

PR #5082, opened 2022-08-12
BF: Code translation was redefining experiment-level variables

TEParsons:
e.g. `var continueRoutine;` doesn't need to be at the start of a Code component which uses `continueRoutine`. Previously this was handled by checking whether the first line was a var def, but this fails if the code is preceeded by e.g. a multiline comment

(includes commits from release - fix needed more changes than I was comfortable pulling into release so I changed branch)


peircej:
Task linked: [CU-1fu7eng Using a multiline comment in python will cause errors in JS side by redefining any variables reset/created](https://app.clickup.com/t/1fu7eng)

PR #5083, opened 2022-08-12
ENH: Allow timestamps to include timezone (if requested), fixes #5075

### LINKED ISSUE #5075: Timezones should be stored in the date by default
## Problem description

Currently, the experiment info contains the date of the experiment without a timezone. This would create problems when labs use the the same script across different timezones.

## Suggested solution

`utils.getDateStr()` should store the [local timezone](https://dateutil.readthedocs.io/en/stable/examples.html#tzlocal-examples) (or even better UTC - but this would require some awareness in the users).

TEParsons:
I'm wary of changing the default format as it means introducing variability in filename structure between this and the next version, however giving designers control over the format and having the current format as the default value seems a reasonable solution to the problem of timezone differences within international labs without breaking anyone's analysis pipeline. This way if the designer needs their files marking by timezone they can just add a `%z` to the format string


TEParsons:
#5075

TEParsons:
Reverting adding the Builder param following discussion with the team - it adds potentially confusing clutter to the Experiment Settings dialog for something easily solved by a Code component:
```
thisExp.addData("timezone", utils.getDateStr("%Z%z"))
```

PR #5099, opened 2022-08-26
BF: Code translation was redefining experiment-level variables 

TEParsons:


peircej:
Task linked: [CU-1hkj2mc Defining a variable in the before experiment tab will produce an "undefined" error in Javascript](https://app.clickup.com/t/1hkj2mc)

PR #5111, opened 2022-09-07
BF: Get Pavlovia session via Builder and Coder on access so it's always the current session

TEParsons:
(includes #5109 as it was super annoying trying to debug with avatars broken)


peircej:
Task linked: [CU-2gbntrd Open psychopy when logged into one account, log out of that account and into another, try to push a project to pavlovia](https://app.clickup.com/t/2gbntrd)

PR #5113, opened 2022-09-08
ENH: Split words longer than entire line in textbox

TEParsons:


peircej:
Task linked: [CU-2htak2d text in editable textbox able to exceed the bounds of the textbox (locally)](https://app.clickup.com/t/2htak2d)

PR #5115, opened 2022-09-08
precise value setter for sound;

lightest:
When "set every repeat" is ticked JS version of experiment creates new instance of Sound in *routineBegin function. Combined with loops this causes continuous re-instantiation an eventually depletes resources.

Bug described here:
https://discourse.psychopy.org/t/repeated-sound-plays-inconsistently-after-a-few-minutes/30374/2

PsychoJS portion of the fix:
https://github.com/psychopy/psychojs/pull/527

cid: CU-3h645y3

### DIFF
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -128,13 +128,7 @@ class SoundComponent(BaseComponent):
             stopVal = -1
 
         if self.params['sound'].updates == 'set every repeat':
-            buff.writeIndented("%s = new sound.Sound({\n"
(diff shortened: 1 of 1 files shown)


peircej:
Task linked: [CU-3h645y3 Set every repeat for sound depletes resources when combined with loops.](https://app.clickup.com/t/3h645y3)

peircej:
@lightest this sounds fine to me, assuming that sounds in psychojs can be set this way (which I'm guessing you've tested! 😆 So yes, you can convert from draft to PR and we can merge.

PR #5122, opened 2022-09-09
Docs: add Shun Wang to translators

peircej:
@shun2wang is this OK as the way to insert your name in the About>Translators?

Or maybe we should be writing Japanese/Chinese in kanji!? @hsogo

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -1004,7 +1004,8 @@ class PsychoPyApp(wx.App, handlers.ThemeMixin):
         devNames.sort()
 
         intNames = [
-            'Hiroyuki Sogo'
+            'Hiroyuki Sogo (Japanese)'
+            'Shun Wang (Chinese)'
         ]
         intNames.sort()
 


shun2wang:
>is this OK as the way to insert your name in the...

that's OK, Thanks. and i added Chinese name for Shun Wang (王顺)

hsogo:
Hi Jon and Shun, I'm a bit confused.  Jon, do you mean to write "Hiroyuki Sogo (日本語)" instead of "Hiroyuki Sogo (Japanese)"?   Shun, following your suggestion, how should we write the language being translated to? "Shun Wang (王顺) (Chinese)" for example?

peircej:
Yes, sorry, I was thinking that it's worth referring to the language that was translated and possibly, in your case, we could add something about being the lead of translation project in general because you've done a lot more here than "just" providing the Japanese labels. I don't have strong opinions in any direction, just throwing out ideas of how best to display your contributions

hsogo:
@peircej  I also don't have strong opinions about that.  But I feel that my contributions to multilingual support are represented by being listed as one of the developers. So I don't think special notation is necessary within the About>Translators section.

peircej:
OK, well I've pulled this in for now, with the language names in English, I'll leave to you whether it's preferred to switch them to be in the language they represent

PR #5128, opened 2022-09-13
NF: Add "placeholder" text to TextBox

TEParsons:
Works the same as a HTML textarea, whereby the placeholder text is shown when the TextBox is both editable and empty, and the placeholder text isn't included in the value of the TextBox when e.g. saving data.


TEParsons:
Pulling this in as we agreed on it in standups already, if I remember right we were just waiting on tests completing?

PR #5142, opened 2022-09-21
BF: Fix `Window.resetViewport()` method 

### LINKED ISSUE #5135: visual.Window.resetViewport() error
When I run

```
from psychopy import visual  
win = visual.Window()  
win.resetViewport()  
```
with psychopy 2022.1.4 and python 3.8.13 on a Win10 laptop with integrated graphics, I get an error
```
File:  "psychopy\visual\window.py", line 1710, in resetViewport
    self.scissor = self.viewport = self.frameBufferSize
TypeError: this function takes 4 arguments (2 given)
```

It seems that the line should be `self.scissor = self.viewport = [0,0] + self.frameBufferSize` to correctly reset the viewport.

Edit: I didn't see the frameBufferSize is a numpy array, so the fix should be `self.scissor = self.viewport = [0,0] + self.frameBufferSize.tolist()`

mdcutone:
Fixes #5135 and adds a test to validate the fix.

### DIFF
--- a/psychopy/tests/test_visual/test_all_stimuli.py
+++ b/psychopy/tests/test_visual/test_all_stimuli.py
@@ -55,6 +55,28 @@ class Test_Window():
         self.win.callOnFlip(assertThisIs2, 2)
         self.win.flip()
 
+    def test_resetViewport(self):
+        # Check if the `Window.resetViewport()` method works correctly. Not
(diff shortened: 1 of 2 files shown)


mdcutone:
Tests are all passing, pulling this in.

PR #5143, opened 2022-09-21
updated shelf counterbalancing record documentation

suelynnmah:


RebeccaHirst:
Not pulling this in just yet because I think some more tweaks are needed to clarify how this is making the docs clearer. These changes might include the below which we discussed:

- to update the demo to be one that used a record with a name other than "groups" e.g. "mygroups"
- then update the screenshot to show that
- clearly define in the text what "groups" in the code comp is referring to

PR #5153, opened 2022-09-28
BF: Make sure depth is supplied, used and functions as intended

TEParsons:


peircej:
Task linked: [CU-2vee9xu layers/depth incorrect with textbox](https://app.clickup.com/t/2vee9xu)

PR #5158, opened 2022-10-02
NF: Preliminary work on plugin system

mdcutone:
PR adds the foundation for plugin support. This is likely going to require more than one PR to get fully working.


mdcutone:
Got to the point where we can load plugins by specifying their package names directly in the config. Components in plugins are loading just fine.

PR #5166, opened 2022-10-11
BF: Added addTime() to class CountdownTimer

Catyre:
Current behavior of `CountdownTimer.addTime()` subtracts time from the timer because it inherits from `Clock.addTime()`, which has its time counting up.  This pull request adds an `addTime()` specific to `CountdownTimer` which increases the time to match the behavior expected from the example in the docs (contrast [this](https://discourse.psychopy.org/t/countdowntimer-addtime-seems-to-be-subtracting-time-instead/31250) behavior, and the example usage [here](https://psychopy.org/api/core.html#psychopy.core.CountdownTimer)).

This is my first pull request for both psychopy and in general.  Please let me know if I'm just misunderstanding the purpose of `addTime()`, or, if I'm not, if there is a better way to fix this.

(also fixed a typo in `Clock.addTime()`'s example usage)

### DIFF
--- a/psychopy/clock.py
+++ b/psychopy/clock.py
@@ -241,7 +241,7 @@ class CountdownTimer(Clock):
             while countdownTimer.getTime() > 0:
                 # do something
         """
-        
+
         self._timeAtLastReset += t
 
     def reset(self, t=None):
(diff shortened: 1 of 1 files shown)


peircej:
Thanks. This is confusing stuff, I know. I think it's right to add this and it will be in the next major release. I won't put it in the next bug-fix release because it is a behaviour change

thanks for contributing! 🙏

PR #5167, opened 2022-10-11
FF: Remove redundant calls to `_createTexture` during `GratingStim` init

mdcutone:
Keeps `_createTexture` from being called multiple times to no real effect during `GratingStim` initialization. Removes the attribute setter since `_needsUpdate` is set as `True` at the end of the init routine, which will cause the update to happen anyways later on rendering the `self.tex = tex` call pointless.


mdcutone:
Closing because #5174 fixes this too.

PR #5170, opened 2022-10-12
BF: Create one serial object per device, not per component

TEParsons:


TEParsons:
Closing as this is now a plugin so will pull into the plugin instead

Issue #4831, opened 2022-05-10
Possible Bug - Risk of Data Loss!: 2022.1.x URL Parameter not included in filename


shabkr:
I think I have encountered a bug with the interaction between URL parameters and datafile saving in 2022.1.1 and 2022.1.3 of PsychoPy/JS. I have additionally posted about this issue on the [PsychoPy Discourse](https://discourse.psychopy.org/t/possible-bug-2022-1-1-2022-1-3-url-parameter-not-included-in-filename/29080)

**OS:** Windows 10
**Coder or Builder?:** Builder
**PsychoPy or PsychoJS?:** PsychoJS
**Offline or Online:** Both
**PsychoPy version:** 2022.1.x

**What are you trying to achieve?**
I use URL parameters to transfer information from my participant recruitment site (e.g. [Prolific](https://www.prolific.co/)) to my Pavlovia-hosted experiment. One of the URL parameters is the participant number, which I use to name the file with the format ```participant_expName_Time.csv```. 

**Experiment Set-up**
-  Experiment Settings > Data > Data Filename says 
```u'data/%s_%s_%s' % (expInfo['participant'], expName, expInfo['date'])```
- provide link with URL parameter 
(e.g., ```https://run.pavlovia.org/USER/EXPERIMENT/?participant=1```)

**Previous Behaviour - tested on version 2021.2.3**
- participant does not have to input a participant number
- upon experiment completion or interruption:
    - datafile is saved at ```data/1_expName_TIMESTAMP.csv```
    - datafile contains column ```participant``` and each cell includes the URL parameter provided participant number ```1```

**Current Behaviour - tested on versions 2022.1.1 and 2022.1.3**
- participant does not have to input a participant number
- upon experiment completion or interruption:
    - datafile is saved at ```data/expName_TIMESTAMP.csv``` **note how the participant number is missing from the data filename**
    - datafile contains column ```participant``` and each cell includes the URL parameter provided participant number ```1```

**What did you already try?**
- This behaviour can be reproduced offline using the Run Local JS option if you append 
```/?participant=1``` to the LocalHost URL
data is saved as ```_expName_TIMESTAMP.csv```
- If the participant number is filled manually during the dialog box (offline or online) the filename is correctly saved as ```1_expName_Time.csv```
- I have replicated this behaviour on two computers running version 2022.1.1 (current behaviour) and one computer running versions 2021.2.3 (previous behaviour) and 2022.1.3 (current behaviour).

[...]

PR #5171, opened 2022-10-12
BF: Fix order of dataFileName and addInfoFromUrl (#4831)

shabkr:
This fixes a bug where participant ids provided via URL parameters are not being added to the data filename of experiments run online (#4831). dataFileName assignment should come after addInfoFromUrl so that URL parameters (such as participant id) can be added to the filename.


TEParsons (approved):

TEParsons:
Thanks for this, took a minute for me to work backwards and see what we got wrong but I see it now and your fix is exactly right. I'll pull in now, thanks for contributing!

PR #5179, opened 2022-10-24
RF/BF: Always substitute PsychoJS names

TEParsons:
Moves the code to substitute PsychoJS names from the expression2js function (which is only called on single-line parameters, not e.g. Code components) to the base-level name node visitor, which is always called.

Also converted names dict to basic Python syntax for readability (i.e. so future contributors won't have the same reaction I had of "wait, we're always indexing this dict, even if the index isn't in its keys?!")


peircej:
Task linked: [CU-2zj0zy3 round() does not transpile correctly to JS](https://app.clickup.com/t/2zj0zy3)

Issue #4748, opened 2022-04-04
VSCode debugging very slow when importing psychopy.visual


nicoa47:
On several machines, we noted that debugging the following is very slow:

```
import time
start = time.time()

import psychopy
from psychopy import visual

end = time.time()
print(end - start)
```
> 145.16652703285217  
(note that I have a very low-end machine right now. normally it is about a minute - but still, very long!)

compared to:  

```
import time
start = time.time()

import psychopy
# from psychopy import visual

end = time.time()
print(end - start)
```
> 0.6649966239929199
[environment.txt](https://github.com/psychopy/psychopy/files/8406609/environment.txt)

Attached is the environment file we use, in case it is due to any certain package interaction. This file is now very long, since I didn't manage yet to setup an env file with only psychopy quickly. I can provide this later if needed.

mdcutone:
The `visual` package has a lot of dependencies but components in are mostly lazy loaded so this is a bit surprising. I'll try running your examples locally and see if the timings are comparable.

paretech:
I'm also seeing load times for psychopy.visual on the order of 55 seconds when run from VSCode debug vs 2.5 seconds from REPL.

## Environment
- Windows 10 21H2 (19044.1706)
- Python 3.8.5
- Visual Studio 1.67.2
- PsychoPy 2022.1.4

## Minimum Code to Reproduce
``` Python
import time
start = time.time()

import psychopy.visual

end = time.time()
print(end - start)
```

## Observed Output
I included the run command that VSCode uses in the event useful to debug.
```sh
c:; cd 'c:\Users\user\source\repos\project'; & 'c:\Users\user\source\repos\project\venv\Scripts\python.exe' 'c:\Users\user\.vscode\extensions\ms-python.python-2022.6.3\pythonFiles\lib\python\debugpy\launcher' '51189' '--' 'c:\Users\user\source\repos\project\project\main.py'
pygame 2.1.2 (SDL 2.0.18, Python 3.8.5)
Hello from the pygame community. https://www.pygame.org/contribute.html
55.13599967956543
```

alexander-pastukhov:
[...]

PR #5185, opened 2022-10-27
ENH: better loading of UAX14 linebreak table

peircej:
Improvements:
- smaller data file (<300kb rather than 5Mb)
- uses the original file from Unicode.org preserving the provenance
- updated(?) to newest file. Some slight changes but very minor (see below)
- Python3.10 (but only 3.10) has a limit on pines for parsing py files which the previous method broke
- more universal fix for the issues #4748 and #5038 which now won't be an issue at all (previous fix ameliorated but didn't fix)

Worse:
- read time is 70ms now instead of 60ms before

Changed codes (due to Unicode 15 instead of Unicode 14?):

For char=1885: orig=AL, official=CM
... (see full list in c4d1a506d28f0c37742104698a480eab5d3d1541 message)
For char=1f6c0: orig=ID, official=EB


peircej:
@hsogo @alexander-pastukhov this may be of interest
It's an update to the import of UAX14 linebreaks. 1 issue was that it was slowing down debuggers (VS Code and PyCharm) because they were trying to inspect all the lines of the .py file. That's now fixed because the data are moved to a txt file. 
The txt file is much smaller now but includes comments and info about where it came from. Load time is very marginally worse than importing as python code but 70ms instead of 60ms so no big deal

Hopefully it all still works as before! I don't know what version of Unicode was used to create the previous file but this one is using Unicode 15 and there are some slight differences. I'm _guessing_ the previous one was using an older unicode spec.

Any issue here that you see?
thanks

hsogo (commented):
Hi, using LineBreak.txt from Unicode.org to build the dict object is a smart idea. As you thought, changes in codes is bacause the original UniSeg package was based on older version of Unicode.

[Edit: Sorry, I reported a hyphenation issue in the first post, but it was due to my carelessness (I forgot to add `lineBreaking='uax14'`!)]

PR #5186, opened 2022-10-27
BF: Default shape colors by class

TEParsons:
The problem was that BaseShapeStim assumed that `color=False` and `fillColor=False` meant `fillColor=None`, regardless of what subclass of BaseShapeStim was being used. This meant that we either had to make the default fillColor for Rect, Circle, etc. `False` (which made it transparent) or have the shape's `color` attribute be unsettable, as `fillColor` was never undefined. 

The solution is to end the set of if statements in BaseShapeStim at a point where colors are set to a default value defined at the class level - meaning subclasses can override this default without affecting the defined/undefined logic.


peircej:
Task linked: [CU-2z0vux2 Alternate solution for color/fillColor conflict in Polygon](https://app.clickup.com/t/2z0vux2)

PR #5187, opened 2022-10-28
ENH: Panorama should be in beta

TEParsons:


TEParsons:
I'm just gonna pull this in as it's inconsequential

PR #5191, opened 2022-11-02
ENH: Add alert for when column in conditions file clashes with namespace

TEParsons:


peircej:
Task linked: [CU-2zc9vc9 Using "sound" as column header causes media player error online](https://app.clickup.com/t/2zc9vc9)

PR #5202, opened 2022-11-09
BF: Use namespace/name id rather than numeric for knownProjects

TEParsons:


peircej:
Task linked: [CU-327udb9 404 project not found when trying to sync from psychopy](https://app.clickup.com/t/327udb9)

PR #5210, opened 2022-11-14
BF: Register when routine is force ended even if frame loop never starts

TEParsons:


peircej:
Task linked: [CU-32k8g5f continueRoutine = False failing?](https://app.clickup.com/t/32k8g5f)

PR #5221, opened 2022-11-21
BF: Cherrypick changes from dev to fix pavlovia syncing

TEParsons:
Changes:
- Access Pavlovia projects via namespace/name string rather than numeric ID, unless numeric ID is all we have
- Don't request info from Pavlovia if we have it locally
- Handle genuine project not found errors (i.e. where the project has actually been deleted) via dedicated dialog rather than unhandled error
- Handle avatar too big with dedicated dialog rather than unhandled error


TEParsons:
There's a few commits from dev squashed in to one here, but I believe Git should handle that and recognise that the changes are identical when we merge release into dev? If not I can handle merge conflicts

PR #5227, opened 2022-11-24
BF: Fixed audio backends piling up audio when `play()` is called multiple times

mdcutone:
Adds a status flag check to the beginning of the `play()` routine for multiple backends. Prevents audio samples from accumulating if `play()` is called multiple times.


TEParsons (changes requested):
I don't think we should be using status for this - Sound should have its own internal marker to indicate playing/paused/stopped/etc. which is independent of Builder flags. Builder already sets the status to STARTED once it's begun, so the `self.status = STARTED` lines aren't needed - instead it should be like `self._isPlaying = True` or something, that it can query internally independently of Builder.

TEParsons (approved):

PR #5228, opened 2022-11-25
BF: Cache frame rate after first get

TEParsons:
avoids redundant calls slowing down MovieStim3


peircej:
closing in favour of #5248

PR #5264, opened 2022-12-19
RF: Load Surveys library via JS rather than changing index.html

TEParsons:


TEParsons:
Think this is safe to pull in now, we were waiting on tests passing if I remember right?

PR #5271, opened 2022-12-21
NF: Added `installPackage`

mdcutone:
Adds function to `pkgtools.installPackage()` which interfaces with `pip` to installed packages. By default, calling this function will add packages to the main PsychoPy user directory under `packages`.

Use as follows:

```
# install package
import psychopy.tools.pkgtools as pkg
pkg.installPackage('psychopy-crs')

# upgrade package
pkg.installPackage('psychopy-crs', upgrade=True)
```

### DIFF
--- a/psychopy/tools/pkgtools.py
+++ b/psychopy/tools/pkgtools.py
@@ -11,14 +11,20 @@
 __all__ = [
     'getDistributions',
     'addDistribution',
+    'installPackage',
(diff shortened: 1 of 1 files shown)


TEParsons:
Thanks! You'll need to merge in dev and push again for tests to pass

TEParsons (approved):

PR #5289, opened 2023-01-10
FF: Install packages via pkgtools

TEParsons:
Means they're installed to the user folder rather than directly to the Python environment


peircej:
Task linked: [CU-862j1cgwr Use pkgtools for installing packages / plugins](https://app.clickup.com/t/862j1cgwr)

PR #5317, opened 2023-01-25
FF: If Mic is given infinite duration, substitute max possible dur and print alert

TEParsons:


peircej:
Task linked: [CU-862j4kd18 Substitute no stop time for max stop time in Mic](https://app.clickup.com/t/862j4kd18)

PR #5321, opened 2023-01-26
FF: Fixes to surveys from science team feedback

TEParsons:


peircej:
Task linked: [CU-862j3k102 When using json "unable to save a response for a survey that is not active" (but survey is active)](https://app.clickup.com/t/862j3k102)

PR #5327, opened 2023-01-30
BF: Set movie each frame in JS rather than recreating

TEParsons:
Code was a solution to a problem which no longer exists (movie not being settable after creation in JS) and creates errors (e.g. setting via Static component) so best to just remove it


peircej:
Task linked: [CU-862hzvt94 Error when using static period to set movie: TypeError: Cannot set properties of undefined (setting 'status')](https://app.clickup.com/t/862hzvt94)

PR #5332, opened 2023-01-31
FF: Handle when correct/clickable stimuli is a string / list of strings

TEParsons:


peircej:
Task linked: [CU-862j5fq0e cannot use textstim as a clickable stimulus with mouse](https://app.clickup.com/t/862j5fq0e)

PR #5333, opened 2023-01-31
FF: Don't include surveyId in resources if survey is using a JSON model

TEParsons:


peircej:
Task linked: [CU-862j3k102 When using json "unable to save a response for a survey that is not active" (but survey is active)](https://app.clickup.com/t/862j3k102)

PR #5335, opened 2023-02-01
FF: Don't include var declarations in Code components as they're already handled

TEParsons:


peircej:
Task linked: [CU-862j5mt0j 2023.1.0 Auto->JS](https://app.clickup.com/t/862j5mt0j)

PR #5334, opened 2023-01-31
FF: Fixes to plugins dlg

TEParsons:

PR #5337, opened 2023-02-01
FF: Force reload of `pkg_resources` on scan to avoid using stale path list

mdcutone:
Fixes packages not showing up after installation. Should make them loadable right after install (in theory).


TEParsons:
Changes merged into #5334

Issue #5345, opened 2023-02-02
Overwriting existing files by "Save-As" for untitled files raises ValueError and crashes session


EitanHemed:
Hi all
I'm working on a Windows 11, using Psychopy 2022.2.4 in Coder view. 

When trying to overwrite an existing file with an untitled file (i.e., unsaved draft in coder), the following error occured and i was forced to exit PsychoPy. 
I replicated the issue a couple of times:
- When either the draft is empty or has some content.
- When either the to-be-overwritten file is empty or has some content.
- For .txt and .py extensions. 

```
Traceback (most recent call last):
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\app\coder\coder.py", line 2016, in onIdle
    if not self.expectedModTime(self.currentDoc):
  File "C:\Program Files\PsychoPy\lib\site-packages\psychopy\app\coder\coder.py", line 2434, in expectedModTime
    if abs(float(actualModTime) - float(expectedModTime)) > 1:
ValueError: could not convert string to float: 'Thu Feb  2 21:33:28 2023'

```

![image](https://user-images.githubusercontent.com/37670372/216433992-ec2edba5-c03e-4cf9-8728-8b6fcfdd4f2c.png)

---

Steps to reproduce:
1. Create some file (empty or not), e.g. hi.txt, and close it. 
2. Open up coder and start a new untitled draft (ctrl + N).
3. Try to overwrite the existing file using Save-as (ctrl + S). 

---

Some more context:

The error occurs here: 
https://github.com/psychopy/psychopy/blob/release/psychopy/app/coder/coder.py#L2433-L2435

It seems to be due to `float(expectedModTime)` expecting something which can be cast to a float. However, the value of `expectedModTime` is a string timestamp (`time.ctime()`), assigned in the following lines:
https://github.com/psychopy/psychopy/blob/release/psychopy/app/coder/coder.py#L2335-L24336

PR #5346, opened 2023-02-02
Changed assignment of self.currentDoc.fileModTime from string to float

EitanHemed:
This change aims to solve the bug described under issue #5345.

- Run all tests in suite. The 2 failed tests are unrelated to the changes. 
- Tested manually by editing the source code of a separate, standalone installation (Version 2022.2.4).

### DIFF
--- a/psychopy/app/coder/coder.py
+++ b/psychopy/app/coder/coder.py
@@ -2343,7 +2343,7 @@ class CoderFrame(BaseAuiFrame, handlers.ThemeMixin):
                     n += 1
 
                 # create modification time for in memory document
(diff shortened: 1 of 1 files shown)


peircej:
Thanks Eitan, this looks good. Just going to check the tests pass on macos but I'm sure they will.

For future, it would be great if you could prefix your commit message with `BF:` for bug fix (where the bug is released in the wild) versus `FF:` for fixes to features that are not yet released. More info at https://psychopy.org/developers/repository.html#psychopy-commit-messages

thanks for finding the bug :-)

EitanHemed:
@peircej happy to help! Sorry for missing the commit flag policy.

PR #5348, opened 2023-02-03
BF: Default values for static component

TEParsons:


peircej:
Task linked: [CU-862j676rw When using Static period error occurs if no value given to start time (which is the default)](https://app.clickup.com/t/862j676rw)

PR #5351, opened 2023-02-03
ENH: Show plugin install output as a panel

TEParsons:


peircej:
Doesn't work well as a very narrow window. Let's deal with better rendering of this after release. The version where it brings up a new dialog is at least a good start

PR #5355, opened 2023-02-03
BF: changed PyQt style to get around the weird greyed-out OK btn bug

peircej:
PyQt has this weird bug where the OK button of the expInfo dlg is greyed out (only in version >5.15.0)

Using Fusion style instead makes the button nicer, but in this the focus isn't set to the first input field so we have to set that manually as well


peircej:
New appearance:
image

Old appearance:
image

PR #5358, opened 2023-02-03
BF: Fix `MovieStim` stalling during sequential playback

mdcutone:
Fixes an issue where `setMovie()` wasn't stopping and cleaning up previous streams when called again.


peircej:
closing this for now, having pulled in e00094da2c6 already

PR #5364, opened 2023-02-07
NF: pyproject.toml 

EitanHemed:
Adds a pyproject.toml file (as an optional future replacement for setup.py). 

Tested validity of the on a development installation by removing setup.py and running all tests locally.


peircej:
My bad, didn't spot that this broke the test suite due to a new failure to package correctly with the pyproject added. I've reverted this for now in https://github.com/psychopy/psychopy/pull/5426 while we explore this (and other possibilities) further

peircej:
Incidentally @EitanHemed I note you've gone with vanilla pyproject.toml here, rather than using poetry. My instinct, since it seems to have risen in popularity, was to use poetry but do you have thoughts about that?

EitanHemed:
@peircej sorry for this! 

TBH, there was no special motivation here to avoid using poetry. If you're OK with this, i can look into it, but it will take me a little while to get around to this.

peircej:
No worries. I'm actually now looking at [pdm](https://pdm.fming.dev/) (as well as poetry) to see what's the best route to take

PR #5373, opened 2023-02-10
BF: allow saving non-numeric data to excel format for TrialHandlerExt

suddhasourav:
Trying to save non-numeric data (e.g., strings) to excel files via the saveAsExcel method of the TrialHandlerExt class was throwing the following error: "AttributeError: 'numpy.ndarray' object has no attribute 'mask'"

Reason: non-numeric data do not use masked arrays in Psychopy, but rather object arrays.

Solution: for non-numeric data, create a code block respecting the convention for these data type, e.g., use '--' for missing values instead of masks.

### DIFF
--- a/psychopy/data/trial.py
+++ b/psychopy/data/trial.py
@@ -1669,30 +1669,51 @@ class TrialHandlerExt(TrialHandler):
             if self.trialWeights is None:
                 thisData = self.data[dataType]
             else:
-                resizedData = np.ma.masked_array(
-                    np.zeros((len(self.trialList),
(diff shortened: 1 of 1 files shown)


TEParsons:
Thanks, all looks good! I'll pull in but if you could put in another PR adding some comments for future reference to make it clear what's going on then that would be fab.

suddhasourav:
Hi @TEParsons! Thanks, just to be sure: we're talking about code comments documenting the steps a bit more in detail?

TEParsons:
Yep, just thinking about future contributors :)

PR #5385, opened 2023-02-15
BF: Set movie opacity AFTER color

TEParsons:
Setting opacity sets the alpha channel of each color, if you set opacity then set a new color (and that new color has an alpha channel, as is the case most of the time with movies as the default is "white" aka `(1, 1, 1, 1)`), then the effects of changing the opacity are overwritten.


mdcutone (approved):
Yes, this works.

PR #5387, opened 2023-02-16
Update translation

hsogo:
I'm working on translating 2023.1.0.  It has almost finished, but I found some issues.

### The tooltip of "Altitude" in the Panorama component is wrong.
It says "**Vertical** look position, ranging from -1 **(fully left)** to 1 **(fully right)**".  Is it a typo?

### Message on the Plugin dialog seems wrong (I'm not sure).
I tried installing the Cedrus plugin.  The message says "Installation failed. See above for info", while pip output says "Successfully installed".  Actually, the package seems to be installed successfuly.
image

### Can I edit app/colorpicker/ui.py?
Though I have to edit this file to translate the Color picker dialog, the comment at the top of this file says "Do not edit this file".

```
###########################################################################
## Python code generated with wxFormBuilder (version 3.10.0-4761b0c)
## http://www.wxformbuilder.org/
##
## PLEASE DO *NOT* EDIT THIS FILE!
###########################################################################
```
Can I directly edit this file?

### Labels of some "default" wxPython menus and buttons are not translated with wxPython 4.2.0.
As far as I noticed, buttons created by `CreateStdDialogButtonSizer`, context menu on `app.coder.repl.ConsoleTextCtrl`, and the TIP dialog created by `CreateFileTipProvider` are not translated with wxPython 4.2.0. They look fine with wxPython 4.1.1.

### app.coder.PsychopyPyShell class is no longer used.
This class was used to translate a popup menu of `wx.py.shell.Shell` object, which was used in the "Shell" pane in the Coder window.  In the recent version, this was replaced `app.coder.repl.PythonREPLCtrl`.  I think that this class can be safely deleted.


hsogo:
I noticed that PoEdit cannot find strings appended to `_localized`  in Plugins.  This means that translation of these strings is not included in the catalog file (messages.po).  

I tried adding local translator to the component.  For example, I installed Cedrus plugin and added following code to psychopy_cedrus/cedrusBox/\_\_init\_\_.py. Then, create "locale" folder in psychopy_cedrus/cedrusBox and create messages.mo files for English and Japanese. In messages.mo files, only translation of "Device number" and "Use box timer" are defined.

```
import gettext
from psychopy.localization import getID
_, lang = getID()
localTranslation = gettext.GNUTranslations(open(Path(__file__).parent/'locale'/lang/'messages.mo','rb'))
_localTranslate = localTranslation.gettext

# only use _localized values for label values, nothing functional:
_localized.update({'deviceNumber': _localTranslate('Device number'),
                   'useBoxTimer': _localTranslate("Use box timer")})
```
In my environment (Japanese Windows 11, python 3.8.10), this works fine. Changing language settings in the Preference Dialog of the PsychoPy app, strings are shown appropriately.

I'm not sure whether this is the best way, but it would be better than adding a list of `_translate` for plugins somewhere in the PsychoPy app (e.g. psychpy/app/builder/localizedStrings.py), I think.

TEParsons:
> The tooltip of "Altitude" in the Panorama component is wrong.

Oops, good catch, thanks! 

> Message on the Plugin dialog seems wrong (I'm not sure).

We had a fair bit of trouble with this during development. The "failed" message means that, when we query whether the package is installed, it returned "False". This is most likely just a timing thing - if the install button is gone afterwards then the plugin is installed and that function has returned True, it probably just wasn't finished when it was checking for the message in the dialog. @mdcutone is working on making this process update live and return output step by step so hopefully once he's got that working we'll be able to get that message returning more accurately.

> Can I edit app/colorpicker/ui.py?

I think that message is auto-generated by the tool used to build the dialog. I've definitely already ignored it and made changes so I think the sanctity of wx Form Builder is already broken ;P So go ahead!

> Labels of some "default" wxPython menus and buttons are not translated with wxPython 4.2.0.

Ah... We use CreateStdButtonSizer rather than creating buttons manually as it means the order of Okay and Cancel always matches the OS, so that function is very helpful. The question I suppose is whether this is something in wx that's actually *broken* or just something that's incompatible with PsychoPy specifically.

> app.coder.PsychopyPyShell class is no longer used.

Good spot, I'll remove it and try to test thoroughly, but always good to get the bloat down!

peircej:
@hsogo you can actually fully delete that warning about wxformbuilder creating the file - this won't be regenerated from the builder anyway

PR #5389, opened 2023-02-16
ENH: Implement feedback from #5387

TEParsons:


hsogo (commented):
I think that line 13 (`import wx.py`) and lines 76-80 (see below) are also unused.
```
try:  # needed for wx.py shell
    import code
    haveCode = True
except Exception:
    haveCode = False
```

PR #5397, opened 2023-02-20
BF: Fix GH workflow install dependencies on Linux

shun2wang:
Some dependencies packages had no longer available for Ubuntu-latest(22.04 LTS).
Not sure if this will break the environment for unit testing but this will let workflow continue to unit tests steps.  [logs](https://github.com/shun2wang/psychopy/actions/runs/4221398747/jobs/7328642335) on my test branch.

If we want to use a specific version of the dependencies, we may need to specify the installation in other ways, such as lowering the operating system image version(Ubuntu 20.04 LTS) or compiling the installation from other installation sources.


peircej:
thanks shun2wang. This is a good start! :-)

PR #5402, opened 2023-02-22
FF: Make `PanoramicImageStim` work with `useFBO=True`

mdcutone:
This PR refactors `PanoramicImageStim` to be based around `ImageStim`. This allows the texture to be rendered with an FBO enabled and provides better adherence with PsychoPy's conventions for loading and rendering images.


mdcutone:
Closing since changes are being pushed into dev from release.

PR #5404, opened 2023-02-24
NF: Split parts of experiments into their own methods

TEParsons:
Allows for parts of an experiment to be called from a super-experiment, supplying it with expInfo, thisExp, win, etc. rather than relying on it being created at the script level each time.

### DIFF
--- a/psychopy/core.py
+++ b/psychopy/core.py
@@ -219,6 +219,17 @@ def getFromNames(names, namespace=None):
     return objs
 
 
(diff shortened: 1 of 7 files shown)


TEParsons:
Example code and psyexp it came from:
[stroop.zip](https://github.com/psychopy/psychopy/files/10823290/stroop.zip)
(`stroop.py` is stroop from 2023.1.0, `stroop_modular.py` is from this branch)

PR #5406, opened 2023-02-24
BF: Correctly ignore blank columns

wader:
When skipping blank columns the index for the field name might end up incorrect cause a index of of range exception.

Example table:
```
| |a| |b|
|-------|
|1|2|3|4|
```

Would end up with:

Currently index 0 then index 2 will be remove in the already modified rangeCols list:

fieldNames = ["a", "b"]
rangeCols = [1,2]

With change:

fieldNames = ["a", "b"]
rangeCols = [0,1]

### DIFF
--- a/psychopy/data/utils.py
+++ b/psychopy/data/utils.py
@@ -396,7 +396,7 @@ def importConditions(fileName, returnFieldNames=False, selection=""):
 
         # get parameter names from the first row header
         fieldNames = []
-        rangeCols = list(range(nCols))
+        rangeCols = []
         for colN in range(nCols):
             if parse_version(openpyxl.__version__) < parse_version('2.0'):
                 fieldName = ws.cell(_getExcelCellName(col=colN, row=0)).value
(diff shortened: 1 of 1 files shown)


wader:
Hi, i'm very new psychopy development. I was only able to test this in the builder, for some reason the local web runner does not find my generated files, returns 404, when running the dev version.

peircej:
Hmm, interesting. So we're currently skipping the name without skipping the data. That does sound unwise, but I'm wondering if allowing empty headers at all is the problem. Like, what is the "expected behaviour" from a user if a column has no header?

wader:
Hi, i'm not a user myself, i helped someone that encountered this issue and it took quite a while to figure that it was a blank column. I'm think if there is no field name you can't refer to the data so should just be skipped? in this case some empty columns were used for comments/scratch data and whatnot so maybe it is nice to allow it?

peircej:
I've pulled this for now as I don't have a better idea, but I still wonder if we should simply say "no" when we detect it

wader:
👍 yeah either way works i think, have no strong opinion. if it should show an error a very descriptive message would be good helping to find the affected column, maybe can mention the name of the non-broken column next to it etc?

PR #5412, opened 2023-03-01
NF: Session class to run multiple experiments from

TEParsons:


psychopy/sessions.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Parameters
+        ==========
+        stem : str
+            Stem of the experiment file - in other words, the name of the experiment.
+        """
+        # Run the setupInputs method
+        self.inputs = self.experiments[stem].setupInputs(expInfo=self.expInfo, win=self.win)
+
+    def runExperiment(self, stem):
```
peircej: @TEParsons I'm pretty certain we'll need to receive `expInfo` here to be available in the running of trials

peircej: self.runExp ( stem='stroop', expInfo={'stimTime':0.1} )

PR #5437, opened 2023-03-09
TESTS: use psychopy._vmTesting instead of travisCI

peircej:
We aren't testing on Travis any more so checking for VM (GitHub) is more useful


peircej:
To try and solve the issue in #5437 with pygame sound not being available on GitHub actions, I'm trying to make the test suite recognise it's using GitHub and then switch to sounddevice (where we built in some protections so we can "mock" the sounddevice to a fair extent)

PR #5454, opened 2023-03-15
RF: Substitute default stim file paths in pathToString instead

TEParsons:


peircej:
@TEParsons make sure that the order is:
- check if the image exists FIRST
- if not THEN look for it in our defaults set

PR #5465, opened 2023-03-22
BF: Handle is_file on invalid path

TEParsons:


TEParsons:
Gonna go ahead and pull this in, the bug is very annoying and the fix is simple

PR #5467, opened 2023-03-23
NF: Pluggable speech-to-text and `AudioClip` improvements

mdcutone:
This PR will make the speech-to-text system pluggable. It will also add new audio processing features to `AudioClip` which help format audio data for use with speech-to-text APIs.


mdcutone:
Added OpenAI Whisper support and made it accessible via the `transcribe` function. Needs a little more work to make it accessible in Builder dialogs.

PR #5502, opened 2023-04-03
TEST: Add libsdl2-dev library for test

shun2wang:
This fix for ffpyplayer build install error like [this](https://github.com/psychopy/psychopy/actions/runs/4585125899/jobs/8097070403).

### DIFF
--- a/.github/workflows/pytests.yaml
+++ b/.github/workflows/pytests.yaml
@@ -36,7 +36,7 @@ jobs:
     - name: Install PsychoPy and dependencies
       run: |
         # for pocketsphinx we need this adapted package:
-        brew install swig vlc portaudio portmidi liblo
+        brew install swig vlc portaudio portmidi liblo sdl2
         # install optional components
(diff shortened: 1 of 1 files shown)


peircej:
We shouldn't _normally_ need to build ffpyplayer. They provide a [large set of wheels](https://pypi.org/project/ffpyplayer/#files)

I'm guessing what happened is that they released the tar.gz of a new version before uploading the wheels and this PR test was unlucky to run in the intervening period. That _used_ to get handled by `pip install XXXXX --prefer-binary` which would search for an older wheel rather install a newer tar.gz file. Looks like this is a side-effect of 'updating' our process to use [pdm](https://pdm.fming.dev/). Pdm isn't aware of any `--prefer-binary` so when it generates the lock file it uses newset versions throughout. Then (I guess) pip receives the lock file and obeys, installing specific versions instead of performing the preference for binaries

peircej:
Anyway, although it _shouldn't_ be needed nearly all the time, I suppose it's fine to do this as a "belt and braces" approach

PR #5520, opened 2023-04-13
BF: Multihead fixes for X11

### LINKED ISSUE #5407: Ubuntu20 Dual-Screen / Zaphodhead: pyglet.gl.ContextException: Could not create GL context - workaround available
While writing a PsychoPy script it wasn’t possible to choose our second monitor for stimulus presentation. We ran into the ‘pyglet.gl.ContextException: Could not create GL context’ and unfortunately none of the solutions proposed for earlier reports [1] of similar issues helped.

Our system comprises a stimulus PC with two monitors. These are set up as experimenter and stimulus displays via separate X windows, configured via X11 (we used PsychToolbox XOrgConfCreator). Both monitors are connected to an AMD Radeon 550X GPU in Zaphodhead mode. OS is Ubuntu 20.04.5 LTS.

What solved the problem for us in the end was to create an OpenGL context at the beginning of the script like this:

`import pyglet`
`display = pyglet.canvas.Display(x_screen=1)`
`screen = display.get_default_screen()`
`template = pyglet.gl.Config()`
`config = screen.get_best_config(template)`
`context = config.create_context(None)`
`window = pyglet.window.Window(context=context)`

We are not sure how to proceed to integrate this into PsychoPy, but hope this report is useful to fix the issue directly.
[1] https://github.com/mmatl/pyrender/issues/13

mdcutone:
Fixes #5407 using the provided workaround for multihead Linux systems running X11. 

Needs additional testing on systems which use such configurations. Stuff works normally on my Linux machine after applying these changes.

### DIFF
--- a/psychopy/visual/backends/pygletbackend.py
+++ b/psychopy/visual/backends/pygletbackend.py
@@ -15,6 +15,7 @@ and initialize an instance using the attributes of the Window.
 
 import sys
 import os
+import platform
 import numpy as np
 
 import psychopy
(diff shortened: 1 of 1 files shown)


geigermartin:
This fix unfortunately isn’t working for us. 
The workaround proposed in #5407 also broke with the fix.
While initially only using Spyder, I did some further testing in VSCode, and the issue is consistent across both.

PR #5524, opened 2023-04-14
ENH: Continue dragging even if mouse moves outside of shape

TEParsons:
Otherwise dragging abruptly stops if you move the mouse too fast, as it moves outside the shape before the shape has had chance to adjust its position


peircej:
Task linked: [CU-862jmmga2 fix: Account for Mouse moving faster than Shape can](https://app.clickup.com/t/862jmmga2)

PR #5528, opened 2023-04-18
FF: Fix typo in Session

TEParsons:
Commit was lost in merge conflict


peircej:
@TEParsons I guess we needed this one urgently but it looks like it broke things with iohub in the tests because iohub doesn't want to be open for multiple processes I think

PR #5533, opened 2023-04-19
BF: Allow ioHub to be initialised multiple times without error

TEParsons:


TEParsons:
This *should* fix the test error in #5528, seems to locally at least. If tests pass I'll pull in as this is needed urgently.

PR #5541, opened 2023-04-21
BF: Include `isFinished` attribute in PTB backend

### LINKED ISSUE #5537: AttributeError: 'SoundPTB' object has no attribute 'isFinished'
In the latest release: v2023.1.2 , the default sound library is PTB, and when using the builder sound component to run a WAV file, you get the following error message:

`AttributeError: 'SoundPTB' object has no attribute 'isFinished'`

This is tested on Windows 10, and after uninstalling and re-installing to double check.

Doesn't complain about ".status" or ".isPlaying" attributes that is also generated upon running the builder script.   I've not yet tested in v2023 versions prior. I can confirm that it is working fine in v2022.1.3 (which is the version we're fixed on for this academic year).

**EDIT:** Just tested in v2023.1.1 and no errors.

mdcutone:
PR adds the `isFinished` attribute to the PTB audio backend. Closes #5537

### DIFF
--- a/psychopy/sound/backend_ptb.py
+++ b/psychopy/sound/backend_ptb.py
@@ -548,6 +548,7 @@ class SoundPTB(_SoundBase):
             logTime = None
         self.track.start(repetitions=loops, when=when)
         self._isPlaying = True
+        self._isFinished = False
         # time.sleep(0.)
         if log and self.autoLog:
             logging.exp(u"Sound %s started" % (self.name), obj=self, t=logTime)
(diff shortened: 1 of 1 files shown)


psychopy/sound/backend_ptb.py
```diff
@@ -332,13 +332,19 @@ def __init__(self, value="C", secs=0.5, octave=4, stereo=-1,
...
+        self._isFinished = False
         self.status = NOT_STARTED
 
     @property
     def isPlaying(self):
         """`True` if the audio playback is ongoing."""
         return self._isPlaying
 
+    @property
+    def isFinsihed(self):
```
asoskic, line 344: Typo in the name of the definition (isFinsihed ). Makes the code break

peircej: Thanks @asoskic 🤦 😆

peircej: https://github.com/psychopy/psychopy/pull/5627

Issue #4412, opened 2021-12-07
Required "Experiment info" fields


richlv:
According to https://discourse.psychopy.org/t/require-filling-in-experiment-info-fields/7014 , it is not currently possible to mark fields as required for the locally run experiments.
It does seem to be possible for the online experiments: https://discourse.psychopy.org/t/demographic-form-at-the-beginning-of-online-experiments/9773 .
Would be great to have such a capability also for the local experiments.

jessicabrinegar:
@TEParsons Has this issue been addressed by any chance?

TEParsons:
Not yet no, but it's on the list!

ainazay:
@TEParsons Hi, I'm trying to work on this feature. I was wondering if you'd know which folder and where exactly the code is located for the online experiments. I was looking to adapt it for the local experiments (are they also located in the same location?) to make the fields there, required, too. Thanks!

PR #5543, opened 2023-04-22
ENH: Allow expInfo fields to be marked as required (*)

ainazay:
This is the qtgui.py file, updated with a new function called "validate," in response to the issue: 

https://github.com/psychopy/psychopy/issues/4412


TEParsons:
I can see what you were trying to do with `validate` but from the looks of it the function was never called and contained some references to attributes which didn't exist. I've pushed some changes which implement a similar idea, hope that's okay!

Issue #5386, opened 2023-02-16
Duplicate info


HadeidM:
There is no way to specify if the CSV data file will or will not contain attributes such as thisN, thisIndex, thisRepN. This causes duplicate info to be specified in the csv data file. You guys should add an argument to either trialHandler or ExperimentHandler that takes care of this in the dataFrame level. Thanks

![image](https://user-images.githubusercontent.com/71160126/219304166-3eb02a00-8e4e-4261-a838-fd72ad984721.png)

HadeidM:
Hello! I added a pull request for this issue: https://github.com/psychopy/psychopy/pull/5544

peircej:
Hi there, sorry for the delay handling this and thanks to both you and @nicolecuthbert for your contributions. 

I've been struggling to decide if this is a good idea to have columns that might change. The 2 columns you point to store information about the trial number in this repeat versus the trial number in total. You're right that if participants only run 1 repeat of each condition then there is duplication but as soon as they run a second repeat the columns become different. If we remove one of the columns either (a) as a setting or (b) automatically based on the apparent similarity, then either the columns won't now contain all the info (for solution (a)) or will be different for different participants (b). It feels like both these outcomes are worse than the current issue of having to ignore the column that has duplicate info (for nReps=1). Especially for automated analyses, run in a script, it's inconvenient to have different columns for different participants.

We have, for the upcoming release, been working on a new option to alter the _order_ of the columns and I feel like this might answer the central issue you have (you can simply lower the priority of the column(s) you don't need) without actually losing any columns (so analysis scripts can expect a consistent input).

See:
https://github.com/psychopy/psychopy/pull/5679

PR #5544, opened 2023-04-22
ENH: Allow users to avoid duplicate trial info in data file

HadeidM:
Added arguments to trialHandler2 initialization so that the user can now specify whether the trialHandler adds data columns such as thisN, trialN, and thisRepN to the final output experiment data file


peircej:
Closing this, pending further discussion about the benefits. See #5386 for further explanation

PR #5546, opened 2023-04-23
ENH: #5386, adds option to ExperimentHandler to hide dup columns

nicolecuthbert:
Addresses issue #5386 very differently than pull request #5544—both solutions can likely coexist.

### DIFF
--- a/psychopy/data/experiment.py
+++ b/psychopy/data/experiment.py
@@ -38,7 +38,8 @@ class ExperimentHandler(_ComparisonMixin):
                  saveWideText=True,
                  dataFileName='',
                  autoLog=True,
-                 appendFiles=False):
+                 appendFiles=False,
+                 hideDuplicates=False):
         """
         :parameters:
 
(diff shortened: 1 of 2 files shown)


peircej:
Hi there, sorry for the delay handling this and thanks for your contribution

I've been struggling to decide if this is a good idea to have columns that might change. The 2 columns you point to store information about the trial number in this repeat versus the trial number in total. You're right that if participants only run 1 repeat of each condition then there is duplication but as soon as they run a second repeat the columns become different. If we remove one of the columns either (a) as a setting or (b) automatically based on the apparent similarity, then either the columns won't now contain all the info (for solution (a)) or will be different for different participants (b). It feels like both these outcomes are worse than the current issue of having to ignore the column that has duplicate info (for nReps=1). Especially for automated analyses, run in a script, it's inconvenient to have different columns for different participants.

We have, for the upcoming release, been working on a new option to alter the _order_ of the columns and I feel like this might answer the central issue you have (you can simply lower the priority of the column(s) you don't need) without actually losing any columns (so analysis scripts can expect a consistent input).

See:
https://github.com/psychopy/psychopy/pull/5679

Issue #5532, opened 2023-04-19
Updating Download Link on PsychoPy Website


jolgen:
Couldn't find anything on this:
On https://www.psychopy.org/download.html you still only can download version v2022.2.5 - although (as far as I understand) v2023.1.2 is stable.
Title of the download page is already changed to v2023.1.2 which suggests to me that you should be able to download the latest version from there.

PR #5547, opened 2023-04-24
DOC: Update standalone version on download.rst

EitanHemed:
Solving issue #5532 - updating the version on the "Download recent version" button to 2023.1.2.


peircej:
Thanks @EitanHemed 
We had actually been deliberately holding off making this release The Big Blue Button™️ but I think the time is right ;-)

PR #5551, opened 2023-04-26
NF: Improved streaming media support

mdcutone:
This PR contains changes which improve working with streaming media sources (movies, cameras, etc.) with most work refactoring camera support to make it pluggable to support multiple backends. 

- [x] Improved movie file writer with asynchronous disk writes
- [x] More reliable handling of camera streams
- [x] Cleaner shutdown for threads used for disk and device I/O
- [x] FFMPEG and OpenCV camera backends


mdcutone:
Fixed issues in builder, got the viewfinder feature also tested out and working. You'll need to have a valid `frameSize` entered at the beginning when using the OpenCV video streaming backend or the viewfinder texture will not initialize. We don't get a valid 'default' frameSize/frameRate until after `open()` is called and the interface is queried.

Issue #5561, opened 2023-05-03
Progress bar on 'Show Screen Numbers' is not seen


EitanHemed:
On Windows 11, using Builder view, version 2023.1.0.

Settings > Visual < Show Screen Numbers creates windows which identify the respective monitors. However, the progress bar which should appear during the presentation of each numbered window is not seen. 

![image](https://user-images.githubusercontent.com/37670372/235877791-2b987fa3-0fc0-4ba4-8913-d11d26bcd0f0.png)

EitanHemed:
Closed with #5562

PR #5562, opened 2023-05-03
BF: Colored progress bar on showScreenNumbers

EitanHemed:
The progress bar was not visible (#5561), as it was the same color as the background.


TEParsons:
Thanks, good catch! I think this is because the default has changed - it used to be white. So supplying the color explicitly is a good solution.

TEParsons:
Thanks, good catch! I think this is because the default has changed - it used to be white. So supplying the color explicitly is a good solution.

PR #5575, opened 2023-05-05
PKG: Pin version of urllib

mdcutone:
Fixes an issue preventing tests from running.

### DIFF
--- a/.github/workflows/pytests.yaml
+++ b/.github/workflows/pytests.yaml
@@ -38,6 +38,8 @@ jobs:
         pip install attrdict3  # used by wxpython setup
         python ./building/plat_custom_installs.py
 
+        pip install requests==2.30.0
+        pip install urllib3==1.26.15
         pip install pdm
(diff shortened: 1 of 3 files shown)


shun2wang (commented):
Hi, once this fixed by upstream at [here](https://github.com/ionrock/cachecontrol/pull/294)  and [here ](https://github.com/urllib3/urllib3/issues/3010) we will not need to pin the version of `uerlib `and `requests` right?

do we wait for it or just using static version?😀

peircej:
So it looks like this was slightly accidental by `urllib3` as their docs indicated it would be _deprecated_ now and _removed_ only in 2.1.0

But I have a feeling it's already fixed now in `requests` [version 2.30.0](https://github.com/psf/requests/releases/tag/v2.30.0) so maybe we don't need the pin ourselves

mdcutone:
Closing due to upstream fixes.

Issue #5600, opened 2023-05-28
Joystick component - missing closing paranthesis in generated code


EitanHemed:
When using a joystick code component in builder view and setting the 'Save joystick state' (Data tab) to 'each frame', and 'End Routine on press' (Basic tab) to 'Never', the generated script includes a list comprehension expression which is missing a closing parentheses.

This raises a SyntaxError when running the experiment:
```
 File "~joystick_issue.py", line 216
    [logging.data("joystick_{}_button: {}, pos=({:1.4f},{:1.4f})".format(joystick.device_number, i, joystick.getX(), joystick.getY()) for i in joystick.pressedButtons]
                                                                                                                                                                      ^
SyntaxError: closing parenthesis ']' does not match opening parenthesis '('
################ Experiment ended with exit code 1 [pid:22928] #################
```

Thanks to @ranibaron for tracking the bug.

PR #5601, opened 2023-05-28
BF: Joystick component - SyntaxError in generated code

EitanHemed:
Solving issue #5600. 
The list comprehension expression in the generated code was missing the closing parenthesis.


TEParsons (approved):
Good catch, thanks!

Issue #5141, opened 2022-09-20
column separator in output file


JensBoelte:
Hello,

while PsychoPy supports different output separators in the result csv-files, the csv recorded via Pavlovia only come with a comma as separator. It would be nice if Pavlovia uses the same separator as set in PsychoPy.

Best wishes and thanks for your work Jens

EitanHemed:
@TEParsons I've looked at what it requires here and on the PsychoJS repo, and would like to help with this one, if you're OK with that (i can't assign myself).

PR #5607, opened 2023-06-01
NF: Adding data file separator to expInfo

EitanHemed:
This comes to solve issue #5141, a feature-request to be able to pass the data file separator to psychoJS (currently uses only comma). 

The changes proposed here should be reviewed alongside [this modification](https://github.com/psychopy/psychojs/pull/566).

### DIFF
--- a/psychopy/experiment/components/settings/JS_setupExp.tmpl
+++ b/psychopy/experiment/components/settings/JS_setupExp.tmpl
@@ -21,6 +21,7 @@ async function updateInfo() {{
 
   
   psychoJS.experiment.dataFileName = (("." + "/") + {filename});
+  psychoJS.experiment.field_separator = {field_separator};
 
 
   return Scheduler.Event.NEXT;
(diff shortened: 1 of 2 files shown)


TEParsons (changes requested):
See my comments on the accompanying PsychoJS PR for feedback on using a field in `expInfo`

It's also worth noting that this would *only* appear in the compiled Python code - the functions to write JS code all end with JS (e.g. `writeInitCodeJS`)

TEParsons (approved):

TEParsons:
Looks great! The test suite is failing because you're using the Param object to index `delim_options` rather than its (string) value - easy mistake to make and very easy to fix. I'll submit a fix to your branch, once that's merged I'm happy to pull in both this and (pending approval from the JS team) the accompanying PsychoJS PR

TEParsons:
Oh, never mind, looks like I can push directly to your branch... So yes, once the accompanying PsychoJS PR is approved I'll pull this one in too :)

Issue #4604, opened 2022-02-08
Two small graphical glitches in the Builder


mmagnuski:
I don't know if there is something that can be done about this at the PsychoPy end, but I've experienced this over and over on multiple computers (all Windows) and decided to report just in case. See the attached gif, it shows:
1. a misaligned horizontal line when trying to rescale the flow part of the Builder
2. the "Insert ..." buttons in the flow part partially disappearing when scrolling the flow horizontally.

![psychopy_small_vis_glitches](https://user-images.githubusercontent.com/8452354/153060713-18e45dc2-edf9-42d6-ad9e-20b861178aef.gif)

mdcutone:
Thanks for pointing these out. First issue is likely a glitch in the GUI toolkit we're using. The second one is something we can fix.

Issue #2132, opened 2018-12-08
Builder UX issues


hoechenberger:
I've only recently started using the Builder again after taking a break of several years. I always found it difficult to force my ideas into the "Builder type of thinking", and my paradigms would have required to add quite some Code Components anyway, so I decided to write all code from scratch right from the beginning.

We now have some students in our lab that are still learning the basics of Python and I thought maybe I/we could give Builder another try to make it easier for them to implement their own experiments.

I came across a number of small things that I thought could be improved to help make for a better UX, which I will list in the following.

- [ ] `Insert Routine` and `Insert Loop` buttons are stuck to the left of the Flow panel. This makes it cumbersome to add additional Routines or Loops at the end of an experiment already consisting of numerous Routines, as one would first have to scroll all the way back to the left to reach the buttons, click them, then scroll back all the way to the right to insert the element.
![inserting routine](https://user-images.githubusercontent.com/2046265/49687795-97ce7400-fb08-11e8-9622-57ee28bfe7f2.gif)
(One can also see here that the scrolling position is not updated properly after the Routine has been inserted)

  **Possible Solution**: The Flow panel should consist of two sub-panels, one that contains the buttons, and one containing the Routines and Loops and a horizontal scroll bar. That way, the buttons would always remain on-screen, even when scrolling to the far-right Routines in a large experiment.
- [x] Inserting longer portions of text without line-breaks into the Text Component (and checking what's written there) is annoying as one would have to scroll around the window, since it does not support soft-wraps.
![text component](https://user-images.githubusercontent.com/2046265/49687887-563ec880-fb0a-11e8-89af-c0d4ae048a70.gif)

  **Possible Solution**: Remove the horizontal scroll bar from the input field, and add support for soft-wraps.
- [ ] The way one moves a Component inside a Routine (moving it up / down) is very atypical in the sense that I've never seen anything like that in any other app I've used for the past 3 years or so: one has to right-click and then click on "move up" or "move down", respectively. With a modern app I would expect to find some kind of clickable "anchor" left of each Component which I can simply click into and then drag it to the desired position.

  **Possible Solution**:  Insert a drag-and-droppable icon / anchor left of each Component present in a Routine, allowing the user to reorder Components via drag and drop.
- [x] The identity of `STOPPED` and `FINISHED` is confusing. In `psychopy.constants` we define:
  ```python
  STOPPED = -1
  FINISHED = STOPPED
  ```
[...]

PR #5620, opened 2023-06-08
ENH: Lock Flow panel buttons to the left when scrolling

TEParsons:
By separating the concept of the Flow *canvas* (window the flow is drawn to) and Flow *panel* (panel object within the UI), meaning the panel can stay in one place while just the canvas scrolls


TEParsons:
Ticks the penultimate box on #2132 and fixes #4604

PR #5637, opened 2023-06-14
ENH: Check in useVersion that requested PsychoPy version matches installed Python version

TEParsons:


peircej:
@TEParsons the tests are failing by (correctly) creating a failed scenario. I think you need the test to test for a raised error with [pytest.raises](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions)

PR #5646, opened 2023-06-20
ENH: Group and document status-related constants

TEParsons:
Doesn't change their function, but allows semantically similar constants to be accessed as a group and makes documentation easier as a docstring can be assigned to the SimpleNamespace object.


peircej:
does this not break a lot of existing code components?

TEParsons:
> does this not break a lot of existing code components?

Nope, because the original names and values are all there still, the namespace objects exist alongside them. It's a change from
`NOT_STARTED = 0`
to 
`status.NOT_STARTED = NOT_STARTED = 0`

PR #5696, opened 2023-07-06
ENH: If loading app windows with file causes error, try without

TEParsons:
Stops users getting stuck in the position where they can't open the app because their last open experiment got mangled somehow.

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -576,15 +576,42 @@ class PsychoPyApp(wx.App, handlers.ThemeMixin):
 
         # Create windows
         if view.runner:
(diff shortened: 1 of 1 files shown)


peircej (changes requested):
I feel like it would be good to get them the actual error message though or they won't be able to debug what has gone wrong (or provide us with good info). How about:
```
except Exception as err:
                # if Coder failed with file, try without
                logging.error(_translate(
                    "Failed to open Coder with requested scripts,opening with no scripts open. Requested: {}\n Err: {}"
                ).format(scripts, err))
                self.showCoder()
```

peircej (approved):

PR #5702, opened 2023-07-09
Fix VlcMovieStim

marcpabst:
As is, VlcMovieStim segfaults for me (tested on Win 10, Win 11, Ubuntu 22..10 and latest Linux Mint) when playing a movie. I don't really have time to investigate atm, but increasing the ref count seems to fix the problem.


peircej:
I'm comfortable pulling this in for now but what we're really wanting is for the ffpyplayer movies to be the single go-to movie player, rather than trying to support multiple solutions. Is that player not working for you?

marcpabst:
Last time I tried it seemed to have problems with larger movie files (audio/video not synced). Also, it still says "This movie class is very efficient and better suited for playing high-resolution videos (720p+) than the other movie classes." in the docs.

PR #5703, opened 2023-07-11
DOCS: shelf leaderboard documentation

suelynnmah:


RebeccaHirst (changes requested):
Before I pull this in @suelynnmah - please could we have more comments on the code components of the .pyexp file itself? ideally each line should have a comment, and we can probably remove any console.log statements that were used for debugging!

suelynnmah:
Changes to psyexp made and synced and project made public

PR #5705, opened 2023-07-11
FF: Fix bug where participant info dialog was never registering as "OK"

TEParsons:


peircej:
This was fixed marginally better in https://github.com/psychopy/psychopy/pull/5707

PR #5712, opened 2023-07-13
BF: Fixed audio not resetting during experiment

mdcutone:
Fixes an issue where successive calls to `play()` using the PTB sound backend does not play the sound again. This checks if audio is playing directly from the back end and does a reset.


peircej:
@mdcutone it looks like you have some debugging print statements in here

PR #5713, opened 2023-07-13
ENH: py2js support for additional functions

roelofsaj:
Refactoring to remove duplicate code and adding translation support for:
* math.ceil
* math.floor
* core.Clock
* 'delimiter'.join(array)
* array.pop(): correctly translate with input arguments
* array.split(): use the correct defaults when no arguments included
* array.insert()
* array.sort(reverse=True): correctly translate the 'reverse' keyword parameter

Also add tests for the above translations.

### DIFF
--- a/psychopy/experiment/py2js_transpiler.py
+++ b/psychopy/experiment/py2js_transpiler.py
@@ -337,8 +337,8 @@ class pythonTransformer(ast.NodeTransformer):
         # The default first argument for pop is -1 (remove the last item).
         elif func.attr == 'pop':
             func.attr = 'splice'
-            args = args if args else [ast.Constant(-1)]
-            args = [args, [ast.Constant(1)]]
+            args = args if args else [ast.Constant(value=-1, kind=None)]
+            args = [args, [ast.Constant(value=1, kind=None)]]
 
             return ast.Call(
                 func=func,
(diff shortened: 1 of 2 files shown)


peircej:
Thanks @roelofsaj 
Looks like the test suite is detecting an issue with this one:
https://github.com/psychopy/psychopy/actions/runs/5545193357/jobs/10123755138?pr=5713
```
self = <astunparse.unparser.Unparser object at 0x14d4def40>
t = <_ast.Constant object at 0x14d4d9490>

    def _Constant(self, t):
        value = t.value
        if isinstance(value, tuple):
            self.write("(")
            if len(value) == 1:
                self._write_constant(value[0])
                self.write(",")
            else:
                interleave(lambda: self.write(", "), self._write_constant, value)
            self.write(")")
        elif value is Ellipsis: # instead of `...` for Py2 compatibility
            self.write("...")
        else:
>           if t.kind == "u":
E           AttributeError: 'Constant' object has no attribute 'kind'
```

roelofsaj:
Sorry about that! I was using a newer version of python (3.9), and the tests were passing there. I've installed 3.8, and now I'm seeing the error. I'll try to track it down and fix it today.

roelofsaj:
I think this fixed it (unless the ubuntu test usually passes?).

TEParsons:
Don't worry about the Ubuntu tests, they're failing for unrelated reasons... This looks great, happy to pull in :)

PR #5723, opened 2023-07-19
Minor Textbox bugfixes

ollie-d:
Previously when `size` and `textgrid_shape` were both `None`, the program would crash. I fixed this by setting `textgrid_shape` to `(len(text),1)` when **both** are `None`.

I also fixed a `textgrid.py` crash caused by `size` being too small to render a single character. A `ValueError` is now raised which tells the user to increase the size or use the `textgrid_shape` option.

Finally, I added `AttributeError` to the exceptions of `__del__` in `textgrid.py`. Previously this would cause an additional error message whenever `textgrid.py` raised an error which makes actual problems difficult to detect.

_P.S. I got to TextBox via looking for rendering improvements for TextBox2. TextBox is much faster when drawing multiple instances. I plan on continuing my tinkering with TextBox2, but figured these quick bug fixes would be useful in the meantime._

### DIFF
--- a/psychopy/visual/textbox/__init__.py
+++ b/psychopy/visual/textbox/__init__.py
@@ -224,64 +224,59 @@ class TextBox:
     _gl_info = None
 
     def __init__(self,
-                 window=None,               # PsychoPy Window instance
-                 text='Default Test Text.',  # Initial text to be displayed.
-                 font_name=None,            # Family name of Font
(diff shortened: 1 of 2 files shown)


TEParsons:
The ultimate goal is to bring the speed of TextBox2 up to match TextBox so that we can remove TextBox altogether, but in the meantime these fixes seem a good stop gap :)

PR #5740, opened 2023-07-21
BF: Safer vertices default value for Shape

TEParsons:
Setting vertices to `()` causes a numpy error later on, so it shouldn't be used as the default.


peircej:
This has caused a load of errors about `windingRule`

PR #5790, opened 2023-08-07
ENH: fixing slow color __deepcopy__ resulting from unnecessary validation

cmacmillan:
Hi I noticed an extremely poorly performing part of the psychopy code. When you make duplicate of a color (which also happens if you perform math with a color e.g. mycolor + .1) the new color gets validated repeatedly (not only validated in the color.set function, but also validated in the individual setter functions as well) which really eats performance.

This validation is pointless since if we are copying a color we already know if the color is valid or not (see the valid flag)

So my solution is just to add a static feature flag to disable validation.

In this particular situation where we repeatedly draw textboxes, this improves performance by ~30%. Color manipulation is such a common operation I think this is worth fixing.

Attached is the code for the psychopy experiment which I used to do my profiling.
[badcolorcopyperfdemo.zip](https://github.com/psychopy/psychopy/files/12273471/badcolorcopyperfdemo.zip)

### DIFF
--- a/psychopy/colors.py
+++ b/psychopy/colors.py
@@ -270,11 +270,15 @@ class Color:
         self._requestedSpace = None
 
         self.set(color=color, space=space)
-
+    
+    perform_validation = True # Static flag
     def validate(self, color, space=None):
         """
         Check that a color value is valid in the given space, or all spaces if space==None.
         """
+        if (not Color.perform_validation):
+            return color, space
+
         # Treat None as a named color
         if color is None:
             color = "none"
@@ -487,10 +491,18 @@ class Color:
         return self.__deepcopy__()
 
     def __deepcopy__(self):
+        # optimization to disable validation while copying
+        # since it isn't necessary and is very slow
+        original_perform_validation = Color.perform_validation
+        Color.perform_validation = False
+
         dupe = self.__class__(
             self._requested, self._requestedSpace, self.contrast)
         dupe.rgba = self.rgba
         dupe.valid = self.valid
+
+        Color.perform_validation = original_perform_validation
+
         return dupe
 
     def getReadable(self, contrast=4.5/21):


cmacmillan:
Sorry forgot to tag the commit with enh

TEParsons:
Good catch! You make a good point that validation isn't needed when a Color is created by using deepcopy on an already valid Color - though I'm not sure an attribute of the class is the way to go with catching this as it would disable validation on all new Colors after one is copied. Instead, I think it's safer to have `set` accept a `validate=False` argument when calling it to skip validation. I'll put in a PR to your branch with the necessary changes; if you're happy with it, pull into your branch & I'll then pull your branch into dev :)

TEParsons:
Also, the file you shared would be a great addition to our test suite - check out `psychopy/tests/README.md` for guidance on writing a test. You could implement it as a function in `psychopy/tests/test_colors/test_Color.py` to make sure setting color values is fast enough.

cmacmillan:
>  it would disable validation on all new Colors after one is copied.

So actually I reset the flag after deep copying has done its business in order to avoid that problem.

>  I think it's safer to have set accept a validate=False argument when calling it to skip validation.

This is actually how I was doing it initially. But it's a little weird because there are setters all over the place.
For example if we step through colors.set on line 350.
If we are a valid color instance already, we hit the rgba setter, which performs a validation. Otherwise we hit the validation on line 366, and then once setattr is called, we will go into whatever function exists for the space we're in, which will call validate. Then it will go into the rgb setter where it will get validated again.
So adding validation via arguments requires modifying nearly every setter function in that class. But I do agree that the resulting code will be cleaner.

TEParsons:
Hm, I see what you mean about the setters - there are a lot of calls to validate which wouldn't be skipped just by changing the behaviour of `set`... But often those validate calls are needed. For example, if you had a stimulus whose colorSpace was `rgb` and you set its color to `"red"`; this doesn't need to throw an error as a named color is always detectable, but the function which detects it is `.validate`.

Maybe the better solution is to look deeper into why `.validate` is slow - in your testing, did you manage to find out which lines in `.validate` are taking the time?

cmacmillan:
Ah you're right since we are passing in the _requested into the constructor. I think it would make sense to just pass .color and .space into the constructor when copying, since then we can skip validation (instead of ._requested and ._requestedspace) Would potentially need other changes tho.

TEParsons:
I get what you mean, but we can do that at the Color level rather than the foreColor/fillColor/borderColor setter level: We could make a blank Color and then, rather than setting the color, we copy over the `_cache` attribute.

I've implemented this in the branch which I've put in to yours as a PR, running the test you supply gives a similar improvement in speed:

| Branch | Speed |
| -------- | ------- |
| psychopy/dev | 21.6649634 |
| cmacmillan:dev | 13.2716936 |
TEParsons:dev-cmacmillan-collab | 13.523499 |

cmacmillan:
Will work more on this fix when I have extra time in the near future.

TEParsons:
Closing in favour of #6412

PR #5796, opened 2023-08-07
EN: Speed up plugin dialog by caching resources locally

mdcutone:
This allows resources related to the plugin directory to be cached locally and reused during a session. This reduces the amount of requests to the server and speeds up successive instancing of the dialog box. Resources are kept locally until the remote file is updated or if the user manually deletes the cache folders and files locally.


TEParsons (approved):
I wonder if it's worth moving the cacheing code to `utils.ImageData` itself so that this applies to e.g. Pavlovia projects, but this is good still :)

PR #5809, opened 2023-08-14
BF: Missing call to activate plugins meant some eyetracker backends were missed

TEParsons:


peircej:
This is *probably* not needed in addition to https://github.com/psychopy/psychopy/pull/5808 but let's wait for opinions from @mdcutone

PR #5810, opened 2023-08-14
BF: Fixed Microphone component device selection

mdcutone:
Fixes an issue where the Microphone component was unable to select devices other than the default


mdcutone:
Closing because the fix is in release

PR #5828, opened 2023-08-18
NF: Functions to set/get experiment variables from Session

TEParsons:


peircej:
Instead we went with simply updating `expInfo` (in #5862 )

PR #5830, opened 2023-08-18
RF: Deprecate use of _localized

TEParsons:
Dict is no longer needed as _translate is cached, and attempting to index _localized without wx raises errors (while calling _translate returns the input unchanged, without error)

Also seems to have the happy side-effect of speeding up the app, presumably because we aren't defining & updating dicts so much:

| --- | Before | After |
| --- | --- | --- |
| Builder frame | 0.561022997 | 0.537538528 |
| Coder frame | 0.51135087 | 0.498919249 |
| Runner frame | 0.11471653 | 0.07901001 |
| Change theme | 0.334387779 | 0.30944705 |

(from psychopy.tests.test_app.test_speed.TestSpeed)

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -69,31 +69,6 @@ from psychopy.projects import pavlovia
 
 from psychopy.scripts.psyexpCompile import generateScript
 
-# _localized separates internal (functional) from displayed strings
-# long form here allows poedit string discovery
-_localized = {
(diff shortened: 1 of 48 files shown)


peircej:
I've pulled in the simple alternative at https://github.com/psychopy/psychopy/pull/5832 for now because the tests were passing there and we need something working right away for Wakefield testing

PR #5863, opened 2023-08-30
BF: Update path of README dialog on "save as"

TEParsons:
It was already updated on "open", but if you open Builder with a new psyexp and then do Save As..., the README dialog doesn't update its understanding of where the README file should be


TEParsons:
Task linked: [CU-8692mjfvm Error in saving readme when it's created within the app](https://app.clickup.com/t/8692mjfvm)

PR #5864, opened 2023-08-30
BF/RF: Handle when Component has no start time in JS

TEParsons:
Currently we still write the code to close the start check code's if statement regardless of whether the if statement was written or not. This applies the same fix we did to the Python end: Querying from the `writeStartTestCode` method whether it was written.


TEParsons:
Task linked: [CU-8692m3r9r No value in component start results in initializing screen online](https://app.clickup.com/t/8692m3r9r)

PR #5878, opened 2023-09-04
BF: Fix JS code when Polygon has custom vertices

TEParsons:


TEParsons:
Task linked: [CU-862k7h2an Syntax error in custom polygon code online (extra comma or missing parameter)](https://app.clickup.com/t/862k7h2an)

PR #5879, opened 2023-09-04
BF: Fix JS code when Routine has fixed duration in S

TEParsons:


TEParsons:
Task linked: [CU-862k6gnq4 tThisFlip is not defined](https://app.clickup.com/t/862k6gnq4)

PR #5897, opened 2023-09-13
Update messages.po

jingrun213:
Add translations to Simplified Chinese


shun2wang:
@patrickabolger @peircej  Should translation commit and PR have a message prefix? like `I18N` or something?

EDIT: well, in recode I see using `DOCS` so that's fine.

peircej:
> @patrickabolger @peircej Should translation commit and PR have a message prefix? like I18N or something?

Yes, I think this is OK for now. Maybe we'll find we need to distinguish translation from documentation but they're quite similar in my head

PR #5899, opened 2023-09-18
BF: Use Version couldn't compile when Python path contains a space

TEParsons:
I *think* the fix is as simple as just putting the Python path in quotes when calling the command, but I need to test this on Mac so keeping as draft until I can confirm.


TEParsons:
Task linked: [CU-862kafra4 Error using use version with multiple instances of PsychoPy where there is a space in the PsychoPy app name](https://app.clickup.com/t/862kafra4)

PR #5901, opened 2023-09-18
BF: Fixed eyetracker calibration routine freezing

mdcutone:
Convert deprecated `numpy.str` type to builtin `str`, which caused an error that locks up the routine.


TEParsons:
Task linked: [CU-8692rc3rg Eyetracker Calibration Routine Attribute Error](https://app.clickup.com/t/8692rc3rg)

Issue #5039, opened 2022-07-23
Confusing error when inheriting from lazily imported classes when importing them from visual, not from their specific files 


alexander-pastukhov:
This is not an issue to be fixed in the code but, perhaps, deserves extra information in docs. For classes that are imported non-lazily, you can import them from `psychopy.visual`  and both create objects and inherit from them
```
from psychopy.visual import ImageStim
class NewImageStimulus(ImageStim):
    pass
```

However, for lazily imported classes, importing from `psychopy.visual` allows to create objects but inheriting from them generates a very confusing looking error
```
from psychopy.visual import Circle
class NewCircleStimulus(Circle):
    pass
```
**TypeError: 'str' object does not support item assignment**

The issue is resolved when importing with the full path
```
from psychopy.visual.circle import Circle
class NewCircleStimulus(Circle):
    pass
```

Thus, the issue is not that it does not work but that it always works with full path but _sometimes_ when importing from `psychopy.visual`. Crowd wisdom: what would be a good place in documentation to warn "When inheriting from classes always import via the full path"? In each lazily imported class? I'll be happy to create a pull request once the consensus is reached.

TEParsons:
It looks like when lazy importing a child class, the scope passed to the lazy importer as it's building the module is just the name of the class rather than the output from a call to `globals()`... I'll try and track down where scope is actually coming from in this instance, but if not then yes I think you're right that we should advise users to import directly when extending classes

peircej:
Yes, for sure, subclasses should use the full location of the path not the top-level lazy-imported version. Not sure where to add that documentation but happy to take a pull request doing that

alexander-pastukhov:
@peircej How about adding this to the class description. E.g., for the Circle the text would be 
"Stimulus class for drawing circles. **This is a lazy-imported class, therefore import using full path `psychopy.visual.circle.Circle` when inheriting from it.**" If this sounds reasonable, I'll go through RSTs of all lazy-imported classes and create a pull request.

peircej:
Sure that sounds sensible 👍

PR #5913, opened 2023-09-25
DOCS: Guidance on subclassing lazy-imported classes

alexander-pastukhov:
This solves issue  #5039. All lazy imported class have the following information added to the documentation, e.g.
"This is a lazy-imported class, therefore import using full path `from psychopy.visual.aperture import Aperture` when inheriting from it."


peircej:
Thanks @alexander-pastukhov

PR #5914, opened 2023-09-25
EN: Improved movie playback support using ffpyplayer

mdcutone:
This PR adds the `MovieFileReader` class which will be used to replace the existing `ffpyplayer` backend used by `MovieStim` for playback. The new interface can also support other libraries in the future, providing a common interface for other backends, much like we did with `MovieFileWriter`.


TEParsons:
@mdcutone is this still relevant? I know you've made a lot of improvements since opening this, do any of those supersede this?

mdcutone:
Yes, this is still relevant, just been on the backburner since I focused mostly on plugins and other fixes this release. Holding off until the next release

TEParsons:
Okay, in the interest of keeping the PRs tab clean I'll close this and we can revisit after release

PR #5921, opened 2023-10-02
BF: Mute audio on movie control

mdcutone:
Make an additional mute call to silence audio during playback control. Helps deal with audio playing at inappropriate times.


peircej:
This looks like an interesting workaround for the buggy audio-not-stopping issue. Shall I pull in for .2.4 release or do we think it might cause new probs?

PR #5924, opened 2023-10-02
NF: Radial component

bartuatabek:
Added a new builder component for the `visual.RadialStim`.


TEParsons:
The component itself looks great, but I think this would make more sense in the [psychopy-visionscience](https://github.com/psychopy/psychopy-visionscience) plugin than in the main library. Could you submit this as a PR to that repo instead? I think this would be a great addition!

PR #5927, opened 2023-10-05
DOCS: update translation doc a little more & change Chinese translation-group email

patrickabolger:
@peircej Just a couple more fixes to the translation doc to make it conform better with what we've learned. And (@shun2wang) I added a change to the Chinese translation email that got combined with this pull request. I'll need to figure out how to separate those properly, LOL.


shun2wang (approved):
ah, yeah now we have email group and should change this. thanks:-)

PR #5928, opened 2023-10-06
BF: Static component was preventing Movie from starting

TEParsons:
Because it (the Static component) was setting the movie's status to STARTED rather than its own, so when the start code for the movie happened it wasn't being started because the code thinks it's already started.


TEParsons:
Task linked: [CU-1x1ktfv Movie loaded in Static Component only shows blank screen](https://app.clickup.com/t/1x1ktfv)

PR #5929, opened 2023-10-06
ENH: Give Components and Standalone Routines translated labels

TEParsons:
Adds a "label" attribute to BaseComponent and BaseStandaloneRoutine, which defaults to None. It can then be set by subclassing components to be a translated string, or left as None. If None, then Builder generates label as normal, if set, it uses the translated label.

Not a pressing issue but only took a mo and seemed like low hanging fruit for improving accessibility!

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -2661,7 +2661,10 @@ class ComponentsPanel(scrolledpanel.ScrolledPanel, handlers.ThemeMixin):
             self.component = comp
             self.category = cat
             # construct label
-            label = name
+            if comp.label is not None:
+                label = comp.label
+            else:
+                label = name
             # remove "Component" from the end
             for redundant in ['component', 'Component', "ButtonBox"]:
                 label = label.replace(redundant, "")
@@ -2794,7 +2797,10 @@ class ComponentsPanel(scrolledpanel.ScrolledPanel, handlers.ThemeMixin):
             self.routine = rt
             self.category = cat
             # construct label
-            label = name
+            if rt.label is not None:
+                label = rt.label
+            else:
+                label = name
             # remove "Routine" from the end
             for redundant in ['routine', 'Routine', "ButtonBox"]:
                 label = label.replace(redundant, "")
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -31,6 +31,7 @@ class BaseComponent:
     targets = []
     plugin = None
     iconFile = Path(__file__).parent / "unknown" / "unknown.png"
+    label = None
     tooltip = ""
     # what version was this Component added in?
     version = "0.0.0"
--- a/psychopy/experiment/components/aperture/__init__.py
+++ b/psychopy/experiment/components/aperture/__init__.py
@@ -23,6 +23,7 @@ class ApertureComponent(PolygonComponent):
     categories = ['Stimuli']
     targets = ['PsychoPy']
     iconFile = Path(__file__).parent / 'aperture.png'
+    label = _translate("Aperture")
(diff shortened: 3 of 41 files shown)


peircej:
Not sure about this. It's getting kind of close to being code, which I think is natural in English. The question is whether this adds more confusion than clarity but maybe we need a non-native speaker to help that decision. @hsogo @shun2wang what are your thoughts? Should components, like "Aperture" have their names translated?

TEParsons:
Things like "Brush" and "Button" I would have thought should be translated, but I see your point for things like "Eyetracker Calibration". We could always pick and choose which ones to translate I suppose. Or maybe we could have the labels show the derived code-y name and then translated in brackets? Like: "Brush (筆)"

shun2wang:
Yes, in fact, we did hesitate some words when translating, like "Aperture" yes, and we discussed in channel and found a appropriate translation.So I think that's OK.

hsogo:
As far as my experience, Japanese students don't have any trouble with English component labels with the aid of understandable icons and translated tooltips.  Translated component names would be nice, but the default name of the component (e.g. "button", "button\_2", ... when the Button component is placed in the routine) remains English even if the label of the component name is translated, doesn't it?  If so, users would still be required to understand what the English names mean.  I prefer to leave the component name labels in English, but it is good for us to have options to translate them.

hsogo:
I thought we could choose whether the component names are translated or not; however, I tried this PR and found that some component names are the same as labels in the property dialogs.  For example, the name of the Sound component appears in the property dialog.

In this case, I feel that "音" is strange as the component name (but not strange as the property name). "サウンド" would be fine both as the component and the property name.  But I'm not sure I can find an appropriate translation for every such case.  Actually, I found that the Progress component has the same issue and I wonder what word would be appropriate.

peircej:
So we were already translating Component names? Oh, I thought those were like protected terms in PsychoPy and not translated

I think the only way to distinguish the string in the Property from the string in the Component name would be to have something like `_translate("Sound Component")` as the actual name and in English we could "translate" that to just "Sound" for the sake of keeping it short, whereas the parameter would actually be `_translate("Sound")`. So in English these end up looking the same but have the flexibility in other langs

TEParsons:
So in this case the word "Component" acts as a syntax flag to say "do not translate the preceding word", right? What if instead we had a reserved symbol to denote "do not translate between these two points"? Maybe ` like in Markdown?

TEParsons:
So:
```
_translate("Sound") -> "音"
_translate("`Sound`") -> "Sound"
```

Would also be useful for referring to component names in tooltips, we could say "Set the sound of a `Sound` `Component`" for example

hsogo:
Unless we merge this pull request, component names are not translated.  The problem is that we can't give different translations to the component names and the property names when both are the same word in English ("Sound" and "Progress" are examples).  Simply Gettext searches for the original string from the .mo file and replaces it if a translation is found.  Gettext doesn't consider the context.

peircej:
The solution I was suggesting wasn't to enforce _non-translation_ it was to allow 2 different translations for the two forms into 2 different phrases in Japanese, while allowing the English version to converge on one final phrase to the user. 

Labelling non-translation is a different issue and your solution would be a good one but might be annoying to rework everything now to do that

peircej:
Did we reach a conclusion on this? Are we skipping it and not translating component names? How strongly do people feel ni favour?

TEParsons:
I don't think there's an urgent need for this, could push back to 2024.2?

hsogo:
As I said in https://github.com/psychopy/psychopy/pull/5929#issuecomment-1757084785, I prefer to leave the component name labels in English as far as the default name of the component is English (e.g. `button_1,` `button_2`,... for Button component).

peircej:
OK, I'm closing in favour of the current system. No need to overcomplicate things. No translated labels for Components for now

PR #5931, opened 2023-10-06
FF: Fix time printing in log and data file

TEParsons:
Because ISO format needs to be from epoch to be meaningful, if a Timestamp was created from an ISO string, it was always from epoch, even if `getTime(applyZero=True`. The Timestamp itself needs to know when last reset was so that `resolve` can account for it when outputting to float/str.


TEParsons:
Task linked: [CU-8692mjtnn Bug: Routine start/stop times are still in relative time from the OR app (should be ISO time)](https://app.clickup.com/t/8692mjtnn)

PR #5934, opened 2023-10-09
BF: Fix bug whereby Textbox's scrolling Aperture was applied to all stim

TEParsons:
Context: https://discourse.psychopy.org/t/text-input-component-causes-a-porthole-effect-on-other-routines-display/36671


TEParsons:
Task linked: [CU-8692v2ark Scrollable textbox creates portal in other routines](https://app.clickup.com/t/8692v2ark)

PR #5938, opened 2023-10-11
Pupil Labs Eyetracker Update

domstoppable:
This merge adds documentation for the Neon eyetracker and eliminates the need for a user to specify their camera calibration file


domstoppable:
@mdcutone - thanks for the merge!

Sorry to bother, but I noticed two minor errors in the documentation.

1. [The video I tried to embed](https://github.com/psychopy/psychopy/blob/0e8cbd4f351b1d5dbf5b721394ad8e7b42cf686b/docs/source/api/iohub/device/eyetracker_interface/PupilLabs_Neon_Implementation_Notes.rst?plain=1#L68) renders as raw text. I couldn't find any examples in other doc pages, so I went with (what I thought) would work, but I guess our RST renderer doesn't work like that.
2. There's [a bad link](https://github.com/psychopy/psychopy/blob/0e8cbd4f351b1d5dbf5b721394ad8e7b42cf686b/docs/source/api/iohub/device/eyetracker_interface/PupilLabs_Neon_Implementation_Notes.rst?plain=1#L96), but it's a simple fix.

I'm happy to fix both of these if you can provide some info or guidance on embedding the video.

PR #5939, opened 2023-10-11
BF: Let experiments run without git installed

TEParsons:
Git was previously needed to query available versions when making a SettingsComponent, instead we should only do this query when trying to open the UseVersion control.


TEParsons:
Task linked: [CU-8692v6u4b Python executable deleted mid-run](https://app.clickup.com/t/8692v6u4b)

PR #5962, opened 2023-10-18
Docs: fix some translation mistakes

shun2wang:
fixed some placholders missing error.

### DIFF
--- a/psychopy/app/locale/zh_CN/LC_MESSAGE/messages.po
+++ b/psychopy/app/locale/zh_CN/LC_MESSAGE/messages.po
@@ -116,7 +116,7 @@ msgid ""
 "{component.type} component \"{component.params[name].val}\" as a dollar sign "
 "has been used incorrectly."
 msgstr ""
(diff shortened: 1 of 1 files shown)


peircej:
This highlights an interesting case, actually. We've mostly been saying that translations won't actually break PsychoPy but in the case of the `{transcrbiver}` typo from `{transcriber}` it will actually cause an apparent error in the app. It will only error for people using this language that happen to trigger this translation, but still, it will appear to those people as an error "transcrbiver is not defined"

Maybe we should add something to the test suite to check if all translations where the target contains a {} variable also contain that same variable at least once in the translation

shun2wang:
>Maybe we should add something to the test suite to check if all translations where the target contains a {} variable also contain that same variable at least once in the translation

Yes, I actually tried a little bit to find a tool that could run this check from the CLI. in fact the Poedit we are using now can run this check in the GUI, but it cannot be executed in the CLI. I also tried [poutils](https://pypi.org/project/poutils/) but not apply for this case.Maybe we have to make it ourselves...

PR #5963, opened 2023-10-19
ENH/RF: Changes to allow individual items in a list param to be code ($)

TEParsons:
Bit nervous about this as there's a lot of change, but if we're to be able to process $ in individual items in a `list` type Param then I think it's necessary.

TLDR on what's changed:
- Functions to parse dollar syntax and stringify params are now modular, so they can be called on values rather than having to use the Param's attributes
- Param can now detect whether its value is an array and, if so, will parse dollar syntax and do stringifying on *each value* rather than on the whole thing
- This means that (and this is why this was requested) the FileListCtrl in Resource Manager Component can have an individual item within it be code, so you can have "image.png" alongside "$cond" and both will work


TEParsons:
Task linked: [CU-2k1gjy3 FileListCtrl needs open text entry option w/ dollar sign stuff](https://app.clickup.com/t/2k1gjy3)

PR #5971, opened 2023-10-25
BF: Fix shape setting on Polygon-derived components (e.g. ROI)

TEParsons:


TEParsons:
Task linked: [CU-8692x368k Eyetracker: ROI does not get custom vertices for Star](https://app.clickup.com/t/8692x368k)

PR #5973, opened 2023-10-25
RF: Allow audio playback interfaces to be configured by `DeviceManager`

mdcutone:
Initial work on getting sound backends configured from `DeviceManager`.


TEParsons:
This is now covered by `SpeakerDevice` so closing the PR

PR #5983, opened 2023-11-02
Add `infoMsg` parameter

### LINKED ISSUE #5937: [Feature Request]: more flexible settings for measuring frame rate
### PsychoPy Version

2023.2.3

Linux-based systems

### Bug Description

Creating a window with `psychopy.visual.Window()` now displays the text "Attempting to measure frame rate of screen, please wait…" as part of its initialization. Previously, PsychoPy measured frame rate without displaying any text.

### Expected Behaviour

I think the new text is highly distracting, and I'd prefer to have a blank screen instead (just like in previous versions). 

### Steps to Reproduce

1. Create a new window with `psychopy.visual.Window()`
2. You'll see the text while frame rate is being measured.

### Additional context

This question has also come up in the forum: https://discourse.psychopy.org/t/is-there-a-way-to-skip-frame-rate-measurement-on-each-initialisation/36232/1

cbrnr:
Add `infoMsg` parameter to `Window` as well as `Window.getActualFrameRate()`, which lets users control the displayed message. Defaults to `"Attempting to measure frame rate of screen, please wait ..."`, but I would rather have it default to an empty string. Fixes #5937.

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -189,7 +189,8 @@ class Window():
                  bpc=(8, 8, 8),
                  depthBits=8,
                  stencilBits=8,
-                 backendConf=None):
+                 backendConf=None,
+                 infoMsg=None):
         """
         These attributes can only be set at initialization. See further down
         for a list of attributes which can be changed after initialization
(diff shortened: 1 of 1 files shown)


TEParsons (approved):
This looks good! I'd actually recommend taking it further: We could add a pair of parameters to ExperimentSettings - one to control `checkTiming` (True/False) and another to control `infoMsg` when the call to create the window is written in boilerplate. Happy to pull this in and will then build upon it :)

Issue #6002, opened 2023-11-09
[Bug]: Inconsistent data for routines outside of trial loop


monique2208:
### PsychoPy Version

2023.2.3

### What OS are your PsychoPy running on?

Windows 10

### Bug Description

Data from routines outside of a trial loop is not added to the output csv in a consistent way. 

If the routine consists of only a visual presentation, information about the start and stop times of the routine will be written on the same row as the data from the following trial. However, if there is a button press or any type of input in the routine, the information will all be written on a separate row. This is because the keyboard component causes the `thisExp.nextEntry()` to be called, which is otherwise not there.

### Expected Behaviour

I would expect Routines outside of a loop to have their own row in the output csv file, instead of finding data about these routines combined with the first trial. I would also expect the data to end up in the same place regardless of which components are in the routine.

### Steps to Reproduce

When running the stroop demo as is, there will be a separate line on the csv for the routine.

[keypress_2023-11-09_17h54.56.121.csv](https://github.com/psychopy/psychopy/files/13310897/keypress_2023-11-09_17h54.56.121.csv)

If the keyboard component in the instruct routine is disabled and the text instruction has a duration instead, the information about the onset and offset of the instruction text will be on the same line as the first trial. 

[no_keypress_2023-11-09_17h57.10.269.csv](https://github.com/psychopy/psychopy/files/13310896/no_keypress_2023-11-09_17h57.10.269.csv)

### Additional context

I am currently working with the psychopy bids plugin to get a bids event file directly from the experiment. Because the `thisExp.nextEntry` call is written by the keyboard component, I struggle to get the right values in my bids event file. If I place the component before the keyboard component, the information about the keyboard response is not available. If I place the component after the keyboard component the nextEntry will be called and I have trouble accessing the information about the entire routine. It would be easier if nextEntry was called at the very end, like it is during a trial. All trial information remains accesible to custom code.

TEParsons:
Fixed by #6004

PR #6004, opened 2023-11-10
BF: fix inconsistent data outside trial loop

monique2208:
Possible solution to  #6002. Instead of having the input components write the call to  `nextEntry` if they do not occur within a trial loop, this is done by the routine. This way there is always a line of data for routines outside of a trial loop, regardless of which components are in there. Routines within a loop are not affected.


peircej:
I agree with this but, because it's a behaviour change (somebody may=have an analysis script expecting the current behaviour)  I'm going to push it to the dev branch for the 2024.1.0 release instead

thanks @monique2208 for your contribution! :-)

PR #6009, opened 2023-11-14
BF: Make `engine` param lower case

mdcutone:
Forces the engine param to be lowercase. This ensures that the interface can specify it in various ways.


TEParsons:
Should the output of `getAllTranscriberInterfaces` not then also change so that keys there are lowercase?

peircej:
It looks like the test fails because TestAllComponents is trying to load a mic and there isn't one on the ubuntu VM for testing

PR #6020, opened 2023-11-15
RF: Move Pavlovia sync output to a tab in Runner

TEParsons:


TEParsons:
Task linked: [CU-86934rryq Display output from Pavlovia sync in Runner tab rather than its own dialog](https://app.clickup.com/t/86934rryq)

PR #6021, opened 2023-11-16
ENH: Add Session method for getting frame rate from window

TEParsons:


TEParsons:
Task linked: [CU-86934xpcf Session needs to have a function to get the frame rate](https://app.clickup.com/t/86934xpcf)

PR #6033, opened 2023-11-22
RF: Move screen number show function to DeviceManager

TEParsons:


TEParsons:
Task linked: [CU-86936c8np Add Python function to show window numbers](https://app.clickup.com/t/86936c8np)

PR #6045, opened 2023-11-24
DOCS: fix fatal issue with es_ES locale

patrickabolger:
`tips.txt` was translated as `consejos.txt`, which would have caused a crash if someone had chosen this locale. I set it back to `tips.txt`. (if someone wants start-up tips in Spanish, they'll need a `tips_es_ES.txt` file for Spanish in Spain, or `tips_es_CO.txt` file for Spanish in Colombia) Also, the locale is `es_ES`, but someone changed the language within that locale to Spanish in Colombia (which should be a new locale: `es_CO`, instead). I changed the language in `es_ES` to Spanish in Spain (as it must be). I'm not sure if the translations in there are for Colombian or Iberian Spanish. I'll need to look closer. Or we can ask whoever made the pull request.


psychopy/app/locale/es_ES/LC_MESSAGE/messages.po
```diff
@@ -300,9 +300,11 @@ msgstr ""
 msgid "Compatibility information"
 msgstr "Información de compatibilidad"
 
+# I deleted this translation (consejos.txt) and replaced it with tips.txt again. This is because this is a pseudo-translation, and has to be an actual file labeled tips_es_CO.txt, which must be located in the directory: psychop/apps/Resources. It’s not there yet, so this has to stay as tips.txt. Putting consejos.txt here will cause the localization to crash immediately upon startup.
 #: psychopy/app/_psychopyApp.py:654
+#, fuzzy
 msgid "tips.txt"
-msgstr "consejos.txt"
+msgstr "tips.txt"
```
shun2wang, line 307: we should also remove the call of  `_translate()` from origin translation strings at https://github.com/psychopy/psychopy/blob/11a53f5420330d6c4080c3e1219b46a3a7258f21/psychopy/app/_psychopyApp.py#L654, then this `msgid "tips.txt"` will not get into the .po files. that will be safety for this issue.

shun2wang: and make file links for translation languages of tips.txt

PR #6056, opened 2023-11-29
BF: Fix bug whereby Brush wouldn't pick up on mouse clicks

TEParsons:
Release cherry pick of #6053


TEParsons:
Task linked: [CU-86937bxhv Brush (Python) doesn't record any responses if Press Button setting is used](https://app.clickup.com/t/86937bxhv)

PR #6064, opened 2023-12-01
DOCS: Update Simplified Chinese translations

ice-yu:


ice-yu:
There are only a few lines about "pump" left, which I am not sure how to translate. Plus, a few other lines scatter in previous part, and I think it's also hard for us to figure out a proper translation.

shun2wang:
@ice-yu it's means "syringe pump" 注射泵，some of the arguments you can refer to "Cetoni neMESYS" 微流注射泵。

shun2wang (approved):

PR #6066, opened 2023-12-04
NF: Piloting mode

TEParsons:
Allow experiments to run in "debug mode", which overrides some experiment settings to make debugging easier, leaving the settings intact for when they're ready to run for real


peircej:
@TEParsons given where we are in release cycle, are we going to pull this in and then work on fixing the test suite before the (major) release?

TEParsons:
Yeah I think that's best, I'd left it open as there's still discussion to be had around the appearance of the button which feels a bit feature-y, but any changes would be minor and easy to implement so I think we could stretch the definition of "bug fix" to include "making the button be a different direction" or "moving such and such label over here" ;)

Bec1993:
Hello! 
Unfortunately, I have accidentally run a couple of participants in pilot mode rather than experiment mode as I forgot to toggle the "running mode" on in builder. Are there any differences that would impact my ability to use this data, or should it be fine? 
Thanks!

peircej:
@Bec1993 you should use the forum for support queries. The psychopy issues list is for working out development issues (bugs etc.)
But in short - the key things that currently happen in pilotting mode:
1. the study runs in a window rather than full-screen
2. there's a big orange box and a message warning you that you're in pilot mode
3. there are more debugging messages

1 and 3 might have impacted the timing of your study (I can't tell you if that matters in your study) and 1 and 2 would have affected the experience of the participant (I can't tell you if that matters in your study)

PR #6068, opened 2023-12-06
FF: Correct param names for photodiode pos/size/units

TEParsons:


TEParsons:
Task linked: [CU-86938314y Photodiode Key error if Find diode unticked](https://app.clickup.com/t/86938314y)

PR #6070, opened 2023-12-06
FF: Update resp variable name from button box code

TEParsons:


TEParsons:
Task linked: [CU-869395qat Error when assigning a Correct answer for a buttonBox](https://app.clickup.com/t/869395qat)

PR #6079, opened 2023-12-08
ENH: Allow Sound to be initialised with a specific Speaker

TEParsons:
Rather than just always grabbing all computer audio


TEParsons:
Task linked: [CU-869363hez Streaming stops after hitting Start Recording button.](https://app.clickup.com/t/869363hez)

PR #6090, opened 2023-12-13
DOCS: Fixes bad link and adds video controls

domstoppable:


domstoppable:
Thanks, @peircej!

Did something go wrong in the conversion to HTML though? It seems no stylesheets are applied, and I'm seeing a bunch of 404s for CSS and JS resources in both Firefox and Chrome. I believe this was working ok before this merge.

Looks like a bad root URL? Resource URLs referenced by this page point to`/api/_static/`, whereas other pages simply point to `/_static/`.

peircej:
Where are you seeing the issue with style sheets? When I visit the [live page here](https://psychopy.org/api/iohub/device/eyetracker_interface/PupilLabs_Neon_Implementation_Notes.html) all seems to be working fine

The original problem with the video link, by the way, was about indentation. I guess it's partly confusing because of adding an embedded video within the text of a bullet point. That seems to work now, even though it looks a little weird IMHO

domstoppable:
Interesting... aha!

If you search [Google](https://www.google.com/search?q=psychopy+pupil+neon) or [DuckDuckGo](https://duckduckgo.com/?q=psychopy+pupil+neon) for "psychopy pupil neon", the first hit links to: `https://www.psychopy.org/api//iohub/device/eyetracker_interface/PupilLabs_Neon_Implementation_Notes.html`

The extra slash between `api` and `iohub` seems to cause the problem.

peircej:
Well, that's weird. I wonder why google thinks that's a valid location. Maybe somewhere on the web points to that URL and has been indexed? 🤔

PR #6097, opened 2023-12-14
Misc: Add translation workflows

shun2wang:
Now we use a github action to compile all .po to .mo files once the .po was updated. demo runner at https://github.com/shun2wang/psychopy/actions/runs/7207261091

How to test this action:
1. Cherry-pick this commit to your branch
2. Change works on branch to your test branch
3. Run this action or run by manually.

TODO:
add a step to get strings from .py and update them to .pot/.po

btw: I tried use pybabel but not success and I dont konw why...

### DIFF
--- a/.github/workflows/msgfmt.py
+++ b/.github/workflows/msgfmt.py
@@ -1,0 +1,246 @@
+#! /usr/bin/env python3
+# Written by Martin v. Löwis <loewis@informatik.hu-berlin.de>
+
+"""Generate binary message catalog from textual translation description.
+
+This program converts a textual Uniforum-style message catalog (.po file) into
+a binary GNU catalog (.mo file).  This is essentially the same function as the
+GNU msgfmt program, however, it is a simpler implementation.  Currently it
+does not handle plural forms but it does handle message contexts.
+
+Usage: msgfmt.py [OPTIONS] filename.po
+
+Options:
+    -o file
+    --output-file=file
+        Specify the output file to write to.  If omitted, output will go to a
+        file named filename.mo (based off the input file name).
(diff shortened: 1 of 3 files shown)


peircej:
I'm not sure if it's the right approach to compile the .po files to .mo during a commit. I feel like that will lead to

lots of merge conflicts (because translators have local .mo files that the repository mo file conflicts with)
binary files that frequently change in the repository eat up space (because the entire file has to be kept in history rather than just the changes)
I agree we should be auto-compiling all files but I think it might be best at build-time instead of commit time, no?

shun2wang:
>lots of merge conflicts (because translators have local .mo files that the repository mo file conflicts with)

Yeah, that's good a point. we can complie .mo when build a release.

Maybe we should make an action to auto update pot files and po files while source code changed. I saw that .pot/po files was out of date for a long time. and I tried use pybabel to do this but not successed, any suggestions?

All setps would be:
1. Extract origin translation strings from .py files to .POT file.
2. Update/generate locale .po files by .POT, then push to repo.
3. Make .mo files by .po files. (only release)

Should I move this PR steps to the build actions?

peircej:
For the .pot file, I believe Jeremy Gray used to use those but I believe @hsogo wasn't bothering with them and instead just importing from source code directly to the po file periodically. I guess there's some confusion because we currently have the old pot file (which went years without any updates at all). If there is some advantage to using the pot template, e.g. so that everyone in a team is working to the same set of strings, then I'm happy to switch and start using it and then automatically update it if useful

shun2wang:
OK I will close this for now.

PR #6098, opened 2023-12-14
Docs: Fix translation .po file error

shun2wang:
A typo error will failing compiling to .mo.


peircej:
I'm not sure if it's the right approach to compile the .po files to .mo during a commit. I feel like that will lead to
- lots of merge conflicts (because translators have local .mo files that the repository mo file conflicts with)
- binary files that frequently change in the repository eat up space (because the entire file has to be kept in history rather than just the changes)

I agree we should be auto-compiling all files but I think it might be best at build-time instead of commit time, no?

peircej:
Sorry, my comment was on the wrong pull request! ;-) 
Was meant to be referring to https://github.com/psychopy/psychopy/pull/6097

PR #6106, opened 2023-12-15
RF: Use standardised device param structure in CameraComponent

TEParsons:


TEParsons:
Task linked: [CU-8693aduuv Bring CameraComponent in line with new DeviceManager structure (as with Microphone, Sound, ButtonBox, etc.)](https://app.clickup.com/t/8693aduuv)

PR #6110, opened 2023-12-17
DOC: fix typos

afuetterer:
Hi, this is my first PR to the project. Please let me know, if I missed something or touched files, that should not be changed.

This PR fixes a few typos in the code base.


afuetterer:
Let me take a look at the failed tests.

peircej:
It's OK, the failed tests right now are due to a plugin. We'll fix that and hopefully all will be good again. Thanks for your help!

afuetterer:
Awesome, thanks. I was wondering why these tests failed. Never mind then.

Is the test suite suitable for parallelization with pytest-xdist? Maybe it would run faster in CI then?

peircej:
Some tests could be run in parallel but some not. Setting up the testing of visual stimuli is probably the longest part of the job and it's important that opengl calls are not made in more than one thread simultaneously. That would be the main concern, and that's why I haven't pursued the idea myself, but it might all be fine! :-)

PR #6113, opened 2023-12-18
FF: Only make one audio stream per physical microphone device

TEParsons:
Effectively makes `MicrophoneDevice._stream` a quasi-singleton, whereby if there's already one for the device's index then the same object is used behind the scenes. Gets around the "AudioStream already created" error.

Adding @mdcutone as a reviewer to make sure this is compatible with the rest of how Microphone works


TEParsons:
Task linked: [CU-8693bpjjr chooseMic fails in session-checks](https://app.clickup.com/t/8693bpjjr)

PR #6123, opened 2023-12-29
BF: indent setting device lines to be included in setDevice() function

RebeccaHirst:
fixes backend error on line 251:

"/Users/becca/Documents/GitHub/psychopy/psychopy/sound/__init__.py", line 251, in <module>
    if hasattr(backend, 'defaultOutput'):
NameError: name 'backend' is not defined

Note @mdcutone also submitted a pull request to fix this https://github.com/psychopy/psychopy/pull/6122 however I tested my side and still got the same error. I am not fully familiar with the code in this part of PsychoPy so would appreciate feedback if this fix is appropriate or not and why. Thanks!

### DIFF
--- a/psychopy/sound/__init__.py
+++ b/psychopy/sound/__init__.py
@@ -246,28 +246,28 @@ def setDevice(dev, kind=None):
                             "not {!r}".format(kind))
 
 
-# Set the device according to user prefs (if current lib allows it)
-deviceNames = []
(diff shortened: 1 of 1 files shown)


peircej:
@RebeccaHirst I'm guessing this one *is* actually solved elsewhere and I'm concerned the indentation change in this PR might change behaviour so I'm closing for now, but if you find it IS still an issue then reopen

PR #6128, opened 2024-01-03
FF: Explicitly import microphone into psychopy.sound

TEParsons:


TEParsons:
Task linked: [CU-8693bxxuv AttributeError: module 'psychopy.sound' has no attribute 'microphone'](https://app.clickup.com/t/8693bxxuv)

PR #6133, opened 2024-01-05
BF: Handle "None" for Microphone index

TEParsons:


TEParsons:
Task linked: [CU-8693f41t6 psychopy.sound.exceptions.AudioInvalidCaptureDeviceError if selected Microphone is not available](https://app.clickup.com/t/8693f41t6)

PR #6136, opened 2024-01-07
BF: Fix _set function to handle input length

simonfei123:
When I added a custom mask to visual.NoiseStim(), there was an error from val2array() function, saying "expected 2, but got (the length of the mask)". I traced the issue to the _set() function in basevisual.py, as it sends val to val2array() without specifying the length, so it's still using the default value length=2. I added length=len(val) and it fixed the issue for me.

### DIFF
--- a/psychopy/visual/basevisual.py
+++ b/psychopy/visual/basevisual.py
@@ -1942,7 +1942,7 @@ class BaseVisualStim(MinimalStim, WindowMixin, LegacyVisualMixin):
         """
         # format the input value as float vectors
         if type(val) in [tuple, list, numpy.ndarray]:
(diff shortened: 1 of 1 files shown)


peircej:
I've merged this but actually the original use case for the length attribute was to check that attributes were being set with the appropriate length (e.g a colour should be a triplet, a size should be a singleton or len=2 etc). It turns out that this test has not been applied in the codebase for some years now, anyway, leaving just this bug!

I wonder if it would be better to remove the conditional from the `_set` function itself, rather than calculating length here and then sending it to `_set`. That would have the same result as this commit but result in less code being executed

PR #6155, opened 2024-01-10
FF: PsychoPy was never registering as the active window on Mac

TEParsons:


TEParsons:
Task linked: [CU-8693drx47 Escape key not enabled for experiments ](https://app.clickup.com/t/8693drx47)

Issue #6157, opened 2024-01-10
[Bug]: Error when clicking Save in Coder when no file open: AttributeError: ‘NoneType’ object has no attribute ‘filename’


nvanhaute:
### PsychoPy Version

2023.2.3

### What OS are your PsychoPy running on?

macOS Silicon

### Bug Description

Clicking Save in the Coder window when there is no file open will yield the following unhandled internal error:

```
Traceback (most recent call last):
  File “/Applications/PsychoPy.app/Contents/Resources/lib/python3.8/psychopy/app/coder/coder.py”, line 2405, in fileSave
AttributeError: ‘NoneType’ object has no attribute ‘filename’
```

### Expected Behaviour

Probably should just disable the Save button when no file is open.

### Steps to Reproduce

1. Open Coder
2. Ensure there is no file open
3. Click Save button
...

PR #6158, opened 2024-01-10
BF: fixed error when using Save in Coder when there is no file open

nvanhaute:
Now, we will quit out early when trying to Save while there is no file open

fixes GH-6157


TEParsons:
I think we can actually go a step further and disable the save/saveas buttons when there's no file, but silently exiting on invalid save command is also a good thing to do. Thanks for contributing!

PR #6168, opened 2024-01-11
BF: disable Coder save buttons when currentDoc is None on open

nvanhaute:
Previously, it was possible to start a Coder window with no active document, click one of the save buttons, and then encounter an error.

A fix was added to quit early out of the fileSave method if this was the case, but this builds on that by disabling the save buttons if we're not in a valid state to save.

Related to GH-6157


peircej:
Thanks. I've confirmed the test failures are unrelated and we're working on those separately, so pulling in

PR #6180, opened 2024-01-17
ENH: Speed up experiment loading

TEParsons:
By only loading plugins on app init rather than each time an experiment loads


TEParsons:
Task linked: [CU-8693ftc64 Runner doesn't notice that a script has ended (should return to enabled Run buttons)](https://app.clickup.com/t/8693ftc64)

TEParsons:
Task linked: [CU-8693jkd40 Check the speed of loading of psyexp files](https://app.clickup.com/t/8693jkd40)

PR #6181, opened 2024-01-17
FF: Remove ref to pavlovia buttons when window is deleted

TEParsons:


TEParsons:
Task linked: [CU-8693jkgpv Error when switching user](https://app.clickup.com/t/8693jkgpv)

PR #6188, opened 2024-01-24
FF: ScreenBufferSampler wasn't appearing in list of devices

TEParsons:


TEParsons:
Task linked: [CU-869382xae Allow photodiode component to fail with warning not error](https://app.clickup.com/t/869382xae)

PR #6190, opened 2024-01-24
BF: Draw expected duration from Routine Settings in Builder

TEParsons:


TEParsons:
Task linked: [CU-8693gw134 Orange line to show routine timeout only appears for fixed duration](https://app.clickup.com/t/8693gw134)

PR #6192, opened 2024-01-25
BF: Error when AUI perspective loads if window size is negative

mdcutone:
This bug comes up when resizing the coder window. Since there is no `SetMinSize` call, the window can be resized to (-1. -1). The app will raise an unhandled exception, and the AUI perspective will be invalid. The coder window cannot be resized since it will always raise an error when the user attempts to resize it. If the app is closed, the invalid AUI layout will be saved and loaded next startup. This will crash the app since it is invalid, preventing the app from making it past the splash screen.

This fixes the issue by ensuring `SetMinSize()` is always called to prevent the window from being size too small in the first place. The second modification catches the error and manually sets up the window. This should generalize to other AUI perspective errors, allowing the app to get a fresh, default layout on any error.


TEParsons:
Task linked: [CU-2f9cve0 AUI frame manager error](https://app.clickup.com/t/2f9cve0)

PR #6193, opened 2024-01-26
ENH: Catch fails in Packages dialog

TEParsons:
Currently, if one version number from a package can't be parsed by packaging.version.Version, it raises an error. This tries a few workarounds which can take a decent guess at the version number, and fails without error otherwise


TEParsons:
Task linked: [CU-8693f5he4 Authors github button does not take you to github](https://app.clickup.com/t/8693f5he4)

PR #6208, opened 2024-02-02
BF: Fixed various unhandled UI exceptions and Runner window resizing

mdcutone:
Fixes various issues around the UI where some actions raise unhandled exceptions. Also, I added more reasonable minimum sizes for the runner frame, it's too large for 720p displays which is a size many classroom projectors use.


TEParsons:
I fixed the same bugs in Builder [here](https://github.com/psychopy/psychopy/pull/6204) and I did so slightly differently (prompting a saveas rather than returning), but you also caught error cases in Coder and Runner which I missed so will do a cherry pick to get those changes in :)

TEParsons:
Closing in favour of #6213

PR #6215, opened 2024-02-07
FF: Bugs with DeviceManager not detecting/setting up certain devices

TEParsons:


TEParsons:
Task linked: [CU-8693rprb9 Button Box and Photodiode should be available to add before the session is started](https://app.clickup.com/t/8693rprb9)

PR #6221, opened 2024-02-08
RF/BF: Experiments weren't working online if exp name differed from filename

TEParsons:


TEParsons:
Task linked: [CU-8693jy8fc Experiments fail online if Experiment Name in the settings doesn't match the experiment file name.](https://app.clickup.com/t/8693jy8fc)

PR #6224, opened 2024-02-09
FF: If Device fails to initialise, raise error with name info

TEParsons:


TEParsons:
Task linked: [CU-8693ttyzf Better error handling when device fails to initialise](https://app.clickup.com/t/8693ttyzf)

PR #6233, opened 2024-02-14
FF: Remove circular reference when converting experiment data to JSON

TEParsons:


TEParsons:
Task linked: [CU-8693twetu Circular reference at end of session-checks](https://app.clickup.com/t/8693twetu)

PR #6237, opened 2024-02-16
FF: Avoid setting text control values to None

TEParsons:


TEParsons:
Task linked: [CU-86939wjqj StyledTextCtrl.SetValue(): error when selecting task from "Search projects online" dlg](https://app.clickup.com/t/86939wjqj)

PR #6247, opened 2024-02-19
Fix numpy deprecation

mscheltienne:
Found through a strict pytest run:

```
    @alpha.setter
    def alpha(self, value):
        # Treat 1x1 arrays as a float
        if isinstance(value, np.ndarray):
            if value.size == 1:
>               value = float(value)
E               DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
```

I also slightly simplified the validation logic by clipping to the acceptable range *after* the conversion to a `float`.

### DIFF
--- a/psychopy/colors.py
+++ b/psychopy/colors.py
@@ -541,20 +541,17 @@ class Color:
         # Treat 1x1 arrays as a float
         if isinstance(value, np.ndarray):
             if value.size == 1:
-                value = float(value)
-        # Clip value(s) to within range
-        if isinstance(value, np.ndarray):
-            value = np.clip(value, 0, 1)
+                value = float(value[0])
         else:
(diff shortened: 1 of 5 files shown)


peircej:
Thanks @mscheltienne that looks good. I'll pull once the tests pass, but I think they will do

mscheltienne:
Wait a minute.. finding other similar deprecation..

mscheltienne:
That was the second one caught:

```
DeprecationWarning: `alltrue` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `all` instead.
```

I grep through the codebase to replace all occurences. Good to go!

PR #6248, opened 2024-02-19
BF: Handle when Static component has no start time

TEParsons:
Write start/stop code inside of an `if` statement so it only gets written (including the dedent and close `}`) if the statement was opened in the first place


TEParsons:
Task linked: [CU-862hzvt94 Error when using static period to set movie: TypeError: Cannot set properties of undefined (setting 'status')](https://app.clickup.com/t/862hzvt94)

PR #6249, opened 2024-02-19
FF: Fix bug where plugins dlg would crash due to `write` error

TEParsons:


TEParsons:
Task linked: [CU-8693vn7ke Unhandled error when selecting "Get More" in Builder](https://app.clickup.com/t/8693vn7ke)

Issue #6256, opened 2024-02-21
[Bug]: Conda installation via `psychopy-env.yml` python dependency fails


tomdstone:
### PsychoPy Version

2022.2.5

### What OS are your PsychoPy running on?

macOS Silicon

### Bug Description

I installed psychopy following the conda instructions on the website ([link](https://www.psychopy.org/download.html#conda)) and got a dependency conflict error. Here is the one produced by mamba as it is easier to read:
<img width="683" alt="Screenshot 2024-02-20 at 6 03 04 PM" src="https://github.com/psychopy/psychopy/assets/77251489/5cce7600-1db6-41d0-995d-cca791b168db">

Inside the `psychopy-env.yml` file I changed `python=3.8` to `python>=3.8` and it installed without an error. I got some wxpython errors (see below) when I tried to run `psychopy` straight from the terminal, but I was able to import the package in python without issue and open a basic gui.

<img width="1167" alt="Screenshot 2024-02-20 at 6 08 43 PM" src="https://github.com/psychopy/psychopy/assets/77251489/7a8e89d4-d5de-499f-aec7-2c9ce4240c5f">

### Expected Behaviour

No error message

### Steps to Reproduce

1. Download the `psychopy-env.yml` file and put it in your home directory
2. run `conda env create -n psychopy -f ~/psychopy-env.yml`, or `mamba` over `conda`
3. Edit the file to have `python>=3.8` instead of `python=3.8`
4. run `conda env create -n psychopy -f ~/psychopy-env.yml` and it should install

### Additional context

`Conda` version is `23.3.1`
The `python` version that was selected by mamba is `3.10.13`
The `psychopy` version selected by mamba is `2022.2.5`

PR #6261, opened 2024-02-21
PKG: Changed python dependency in conda env to >=3.8

tomdstone:
The conda environment file appears to create conflicting dependency requirements, was able to fix by changing `python=3.8` to `python>=3.8`. See issue #6256. Not entirely sure if this is a bug on my end, but I was able to reproduce it on two separate computers. Opened the pull request just in case folks decide its best to make the change.

### DIFF
--- a/conda/psychopy-env.yml
+++ b/conda/psychopy-env.yml
@@ -2,7 +2,7 @@ name: psychopy
 channels:
 - conda-forge
 dependencies:
-- python=3.8
+- python>=3.8
 - psychopy
 - pip
 - pip:


peircej:
The long and short of it is that the conda env file is very outdated. None of the core devs on the project actually use conda (we're all python.org users) so support for conda comes purely from user contributions and these have been sparse. If you're able to improve things with fixes to the conda env then I think this is going to be useufl to many others. Note that, in terms of the ideal version of Python, I believe psychopy itself is fully compatible with 3.10 (and possibly 3.11?) but there is also a question of what's best in terms of hardware compatibility

tomdstone:
> Note that, in terms of the ideal version of Python, I believe psychopy itself is fully compatible with 3.10 (and possibly 3.11?) but there is also a question of what's best in terms of hardware compatibility

For what it's worth, I was able to get it mostly working with 3.10 doing plain old `conda install` but 3.11 failed to work.

PR #6270, opened 2024-02-25
Update staircase.py

alexholcombe:
Please check if I'm misunderstanding something, but looks like the staircase is estimating the 63% correct point, not "50 / 50"


peircej:
I believe you're right @alexholcombe 
I expect we originally set up the data for measuring a PSE and then changed our mind ;-)

PR #6299, opened 2024-03-02
BF: Selecting audio device in builder and coder (fixes #6142)

### LINKED ISSUE #6142: [Bug]: Selecting audio device in builder and coder.
### PsychoPy Version

2023.2.3

Windows 10

### Bug Description

As described in this forum [post](https://discourse.psychopy.org/t/how-to-change-audiodevice-in-hardware-preferences/30982), switching audio device in builder preferences just reverts back to the default after hitting apply. 

I've tried to work around this by adding a code component to manually set the audio device as suggested by another user in that thread, but I then see the issue shown in this [thread](https://discourse.psychopy.org/t/error-with-psychportaudio-although-it-works-in-psychtoolbox/27764). I've tried both the ptb and pyo libraries. 

### Expected Behaviour

Sound device should change in preferences and be used in the experiment. 

### Steps to Reproduce

1. Install low latency sound card linked below with driver. 
2. Open psychopy builder and add a sound component. Specifying 250 (pure tone) as the Sound.
3. Open preferences > Hardware > audio device drop down > select your audio device (in my case Speakers (Lynx E44)
4. Run experiment. 

### Additional context

I'm using a low latency sound card, [Lynx e44](https://www.lynxstudio.com/downloads/e44/windows-driver-v2-build-26/) as my applications require audio output at frequencies not possible with the default windows card. This comes with its own driver that I've downloaded and added the .sys file name from the drivers folder (without the .sys extension), I'm not sure if this is correct. 

I'm able to produce sound from the sound card outside of psychopy, so the driver is properly installed.

zuazo:
## PsychoPy Versions

2023.2.3, release, dev

Apparently in 2022.2.5 version it worked. The bug was introduced afterwards.

## What OSes is your PsychoPy running on?

Ubuntu-20.4, Windows 10

## Bug Description

This patch fixes an issue where selecting an audio device in PsychoPy Builder preferences is not being saved correctly. There is no way to change the audio device but to manually edit the user configuration.

This same issue has been described here before:

- https://github.com/psychopy/psychopy/issues/6142
- https://discourse.psychopy.org/t/how-to-change-audiodevice-in-hardware-preferences/30982

This was closed believing it is fixed in 2024.1.0/dev, but it is not.

## Solution

- The `audioDevice` preference must be a `list` for the PsychoPy's internal settings validation to pass. This fixes the Apply and OK buttons.
- Additionally, it ensures `self.audioDevDefault` is updated to reflect the newly selected audio device, allowing the Apply button to function correctly.

This change allows the audio device to be correctly selected, saved, and utilized in experiments.

### DIFF
--- a/psychopy/app/preferencesDlg.py
+++ b/psychopy/app/preferencesDlg.py
@@ -588,7 +588,7 @@ class PreferencesDlg(wx.Dialog):
                         options = self.audioDevNames
                         try:
                             default = self.audioDevNames.index(
-                                self.audioDevDefault)
+                                self.audioDevDefault[0])
                         except ValueError:
                             default = 0
                     else:
@@ -663,8 +663,8 @@ class PreferencesDlg(wx.Dialog):
                     self.app.theme = self.prefsCfg[sectionName][prefName] = self.themeList[thisPref]
                     continue
                 elif prefName == 'audioDevice':
-                    self.prefsCfg[sectionName][prefName] = \
-                        self.audioDevNames[thisPref]
+                    self.audioDevDefault = [self.audioDevNames[thisPref]]
+                    self.prefsCfg[sectionName][prefName] = self.audioDevDefault
                     continue
                 elif prefName == 'locale':
                     # '' corresponds to system locale


TEParsons:
We marked as fixed because the audio device can now be specified in the Sound component itself, the pref is only used as a backup now, but didn't realise the pref was breaking the dialog! Thanks for finding & fixing this - I pulled down your branch to test and all seems fine.

PR #6300, opened 2024-03-03
BF: Fix handling of code expressions in sound component stop values

zuazo:
## PsychoPy Versions

2023.2.3, dev

## What OSes is your PsychoPy running on?

Ubuntu-20.4

## Bug Description

There is an issue in the sound component where code expressions for the stop value without preceding `$` were causing errors. This does not happen with start value, only with the stop property, but both are marked with `$`.

sound-component-stopVal-bug

Traceback error example:
```
  File "~/src/psychopy/psychopy/app/builder/builder.py", line 1392, in pilotFile
    self.app.runner.panel.pilotLocal(event)
  File "~/src/psychopy/psychopy/app/runner/runner.py", line 653, in pilotLocal
    self.runLocal(evt, args=["--pilot"], focusOnExit=focusOnExit)
  File "~/src/psychopy/psychopy/app/runner/runner.py", line 635, in runLocal
    generateScript(
  File "~/src/psychopy/psychopy/scripts/psyexpCompile.py", line 87, in generateScript
    compileScript(infile=exp, version=None, outfile=outfile)
  File "~/src/psychopy/psychopy/scripts/psyexpCompile.py", line 225, in compileScript
    _makeTarget(thisExp, outfile, targetOutput)
  File "~/src/psychopy/psychopy/scripts/psyexpCompile.py", line 205, in _makeTarget
    script = thisExp.writeScript(outfile, target=targetOutput)
  File "~/src/psychopy/psychopy/experiment/_experiment.py", line 327, in writeScript
    self_copy.flow.writeBody(script)
  File "~/src/psychopy/psychopy/experiment/flow.py", line 301, in writeBody
    entry.writeInitCode(script)
  File "~/src/psychopy/psychopy/experiment/routines/_base.py", line 482, in writeInitCode
    thisCompon.writeInitCode(buff)
  File "~/src/psychopy/psychopy/experiment/components/sound/__init__.py", line 147, in writeInitCode
    elif float(inits['stopVal'].val) > 2:
ValueError: could not convert string to float: 'sound_time'
```

## Solution

- Fix handling of code expressions in sound component stop values by using `canBeNumeric()` to check the value correctly.
- This change also moves `canBeNumeric()` to a more central location: `experiment.utils`.

### DIFF
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -13,6 +13,7 @@ from xml.etree.ElementTree import Element
 from psychopy import prefs
 from psychopy.constants import FOREVER
 from ..params import Param
+from psychopy.experiment.utils import canBeNumeric
 from psychopy.experiment.utils import CodeGenerationException
 from psychopy.experiment.utils import unescapedDollarSign_re
 from psychopy.experiment.params import getCodeFromParamStr
@@ -1349,14 +1350,3 @@ class BaseVisualComponent(BaseComponent):
                     "\n"
                 )
                 indented -= 1
-
-
-def canBeNumeric(inStr):
-    """Determines whether the input can be converted to a float
-    (using a try: float(instr))
-    """
-    try:
-        float(inStr)
-        return True
-    except Exception:
-        return False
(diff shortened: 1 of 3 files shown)


TEParsons:
Good catch, thanks! This all looks good, though I wonder whether we need that whole section of code at all - if the point is to choose a sensible default when a value is set each frame/repeat, then I would have thought just substituting -1 when updates != "constant" would be a better approach... Anyway, happy to pull this in once tests pass.

PR #6332, opened 2024-03-18
EN: Coder search panel

mdcutone:
This PR adds a new 'Search' panel to Coder for text search in all open documents. This is an upgrade from the old Notepad style find/replace dialog. Displays the document name, line contents and column/line numbers for each match. You can specify options for matching case/word and even use regular expressions. The replace function will be added at a later time.


TEParsons:

PR #6333, opened 2024-03-20
RF: Move assets needed by psychopy-lib out of the app folder

TEParsons:
Essentially creates a distinction between:
- `psychopy.app.Resources`: Files needed by the app (i.e. button icons)
- `psychopy.assets`: Files needed by everything (i.e. stim defaults, assets for templates)


TEParsons:
NOTE: Not to be pulled in until after releasing 2024.1.0 - I can see this being crunchy and would like a full dev cycle to make sure we run into any errors before it hits users

PR #6335, opened 2024-03-21
RF: Standardise functions between QT and WX dialog backends by getting both from a base class

TEParsons:
This has the following advantages:
- Changing the behaviour of the Dlg class is now much easier for things like parsing syntax, where the backend doesn't matter, as you only have to change it in one place
- Reduces the danger of multiple backends slipping out of parity (as they already had a bit)
- If we wanted to add another backend (e.g. tkinter or any new lightweight GUI packages we come across) it's easy to get it up to the level qt and wx are at as most of the API is already written


TEParsons:

PR #6342, opened 2024-03-27
ENH: Device selection for multiple sounds (ptb library)

janna-stb:
I implemented the option to select different output devices for multiple streams using the ptb library. The ptb library already had this functionality, but until now the psychopy code did not pass the device ID correctly. As a result, the sound was not output on the selected device. This will now work correctly and the sound will be played through the corresponding device/output for multiple simultaneous sounds on different devices.

### DIFF
--- a/psychopy/sound/_base.py
+++ b/psychopy/sound/_base.py
@@ -137,7 +137,7 @@ class _SoundBase(AttributeGetSetMixin):
 
         return device
 
-    def setSound(self, value, secs=0.5, octave=4, hamming=True, log=True):
+    def setSound(self, value, device, secs=0.5, octave=4, hamming=True, log=True):
         """Set the sound to be played.
 
         Often this is not needed by the user - it is called implicitly during
--- a/psychopy/sound/backend_ptb.py
+++ b/psychopy/sound/backend_ptb.py
@@ -115,10 +115,10 @@ def getDevices(kind=None):
     return devs
 
 
-def getStreamLabel(sampleRate, channels, blockSize):
+def getStreamLabel(sampleRate, channels, blockSize, deviceID):
     """Returns the string repr of the stream label
     """
-    return "{}_{}_{}".format(sampleRate, channels, blockSize)
+    return "{}_{}_{}_{}".format(sampleRate, channels, blockSize, deviceID)
 
 
 class _StreamsDict(dict):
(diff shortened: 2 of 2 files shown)


psychopy/sound/_base.py
```diff
@@ -137,7 +137,7 @@ def _parseSpeaker(self, speaker):
 
         return device
 
-    def setSound(self, value, secs=0.5, octave=4, hamming=True, log=True):
+    def setSound(self, value, device, secs=0.5, octave=4, hamming=True, log=True):
```
mdcutone, line 140: I'm concerned adding `device` here may break existing code that doesn't set that value. Possible solution would be to append `device` to the end of the keyword arguments and allow it to take None.

psychopy/sound/backend_ptb.py
```diff
@@ -165,28 +166,28 @@ def _getSimilar(self, sampleRate, channels=-1, blockSize=-1):
...
         """Strict check for this format or create new
         """
-        label = getStreamLabel(sampleRate, channels, blockSize)
+        label = getStreamLabel(sampleRate, channels, blockSize, device_id)
         # try to retrieve existing stream of that name
         if label in self:
             pass
         # todo: check if this is still needed on win32
         # on some systems more than one stream isn't supported so check
-        elif sys.platform == 'win32' and len(self):
```
peircej, line 179: @janna-stb is this really not needed any longer, or did you just remove it while making changes to `getStreamLabel` ?

peircej (changes requested):

zeyus:
I've just had to implement dual audio output (one for stimuli and another for audio triggers), but I have managed to do this with the current implementation (with ptb on windows 11) using:
```python
sound1 = Sound("./test.wav", speaker="Headphhones...")
sound2 = Sound("./test2.wav", speaker="Speaker...")

# ...
sound1.play()
sound2.play()
```

and you could update the speaker by using `sound1.speaker = DeviceManager.getDevice("Another device")`

Just curious about the difference.

TEParsons:
We're currently making changes with multi-speaker setups in mind (see #6953) so this will no longer be needed - as the ptb stream will be an attribute of the speaker rather than of the sound, the device index will be set by pointing the sound object to the correct speaker object. Thanks for contributing still though! I'll make sure the core changes from here are included.

PR #6346, opened 2024-04-03
WIP: first ideas of how to make TrialHandler2 that can create future …

peircej:
…trials

The basic concept is to store a copy of the expected `state` at each point in time as if that was our current position.

We would then generate the `trialHandler.sequence` and use that to extract thisTrial info


TEParsons:
Closing in favour of #6408 (which is based off of this)

PR #6348, opened 2024-04-04
FIX: fix translation issues

hsogo:
## Fix translation of dropdown menu in 2024.1.0
As shown below, translation of dropdown menu is broken.  

image

Labels must be translated in `ChoiceCtrls`.  `_translate(labels[i])` seems fine here, but an empty string is in "Use PsychoPy version" in the Experiment Settings dialog.  Because an empty string is reserved by GNU gettext, providing an empty string to `_translate()` causes an unexpected result.  So, we have to avoid applying `_translate()` when the label is an empty string.

```
--- a/psychopy/app/builder/dialogs/paramCtrls.py
+++ b/psychopy/app/builder/dialogs/paramCtrls.py
@@ -362,9 +362,9 @@ class ChoiceCtrl(wx.Choice, _ValidatorMixin, _HideMixin):
         _labels = {}
         for i, value in enumerate(choices):
             if i < len(labels):
-                _labels[value] = labels[i]
+                _labels[value] = _translate(labels[i]) if labels[i] != '' else ''
             else:
-                _labels[value] = value
+                _labels[value] = _translate(value) if value != '' else ''
         labels = _labels
         # store labels and choices
         self.labels = label
```

# Fix corrupted locale list in 2024.1.0
As shown below, the labels of locales in the Preference Dialog looks strange.  This was fixed by editing `preferencesDlg` in app/preferencesDlg.py

image

## Fix strings in `_translate()`

Formatted string literals should not be used in `_translate()` (in hardware/button.py).  The case in alarts/\_alarts.py is a bit complidated.
```
self.msg = _translate(catalog.alert[code]['msg']).format(**strFields)
```
This works as expected, but pybabel incorrectly detect `'msg'` as the string to be translated.  So it would be safer to avoid using a string in an expression in `_translate()`.

## Move entries in `_localized` in builder.py to localizedStrings.py
In the previous version, translation of some Builder labels were defined in `_localized` in app/builder/builder.py.  But `_localized` was deleted in the current builder.py.  I merged deleted labels to app/builder/localizedStrings.py instead reverting builder.py.

## generateHints.py supports [piloting] category 
The newly added [piloting] category was not recognized by preferences/generateHints.py, resulting that help messages in the Piloting section of the Preference dialog was not detectable by pybabel and Poedit.  I fixed this and updated localization/generateTranslationTemplate.py to run generateHints.py and generateAlertmsg.py.

## known but unfixed issues
Now we can edit the "Frame rate message" in the Experiment Settings dialog, but the message may be corrupted if non-ASCII characters are included in it.  I think that the font should be selectable by the user to fix this issue (also, I think it would be convenient if font size could be changed).  The message shown with the orange border around the window in the piloting mode is also corrupted if non-ASCII characters are included.  This message cannot be edited by a user, but translators may use non-ASCII characters for translation of this message.


peircej:
Thanks @hsogo that will be in 24.1.1 (imminent) and sorry we didn't alert you in advance of the new version coming. We really need to sort out a way for you to do your tests earlier :-) 

Regarding the `_localized` list we had removed this because it was causing more problems than solving. It had bitten us a couple of times by creating confusion with contributors (and plugin developers) around whether to use that or `_translate` and I think the speed advantage over using _translate() more than once is negligible. If _`localized` is still needed/useful in ways we hadn't seen then let's keep it but we were sort of hoping to phase it out in favour of having just one way to do translation

hsogo:
It was about 10 years ago that `_localized` was introduced, and I don't clearly remember the discussion on it.  But, the most important purpose was to separate the logic of the program and the appearance strictly.  In the early days of localization, there were troubles such as translated strings being output as parameter values to the psyexp file.  On the other hand, I understand that `_localized` is confusing especially for new contributors.  As far as it is agreed by contributors that translated strings are used for visual purpose only, I'm in favor of removing `_localized`.

Issue #1226, opened 2016-08-23
Incorrect error message about closing an iohub session


mwaskom:
I'm trying to debug a script that is using iohub and ran into an error message that I do not believe is correct. When restarting the script in the same IPython session it fails with

```
AttributeError: An existing ioHubConnection is already open. Use ioHubConnection.getActiveConnection() to access it; or use ioHubConnection.quit() to close it.
```

but this fails:

```
iohub.ioHubConnection.quit()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-10-23d56e86886c> in <module>()
----> 1 io = iohub.ioHubConnection.quit()

TypeError: unbound method quit() must be called with ioHubConnection instance as first argument (got nothing instead)
```

as does this

```
iohub.ioHubConnection().quit()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-123239c79580> in <module>()
----> 1 iohub.ioHubConnection().quit()

/Users/mwaskom/anaconda/envs/psychopy/lib/python2.7/site-packages/psychopy/iohub/client/__init__.pyc in __init__(self, ioHubConfig, ioHubConfigAbsPath)
    245
    246         if ioHubConnection.ACTIVE_CONNECTION is not None:
--> 247             raise AttributeError("An existing ioHubConnection is already open. Use ioHubConnection.getActiveConnection() to access it; or use ioHubConnection.quit() to close it.")
    248
    249         Computer.psychopy_process = psutil.Process()

AttributeError: An existing ioHubConnection is already open. Use ioHubConnection.getActiveConnection() to access it; or use ioHubConnection.quit() to close it.
```

`psychopy.__version__` shows '1.83.04'

mwaskom:
Oh for the record I closed it with `iohub.ioHubConnection.getActiveConnection().quit()`.

hoechenberger:
You will want to invoke the `quit()` method on your `ioHubConnection` instance. The following code works:

```python
from psychopy import iohub
io = iohub.launchHubServer()

# Lose "track" of the ioHubConnection.
del io

# Regain access to that connection.
io = iohub.client.ioHubConnection.ACTIVE_CONNECTION

# Quit ioHub.
io.quit()
```

mwaskom:
OK, thanks. The issue wasn't really that I couldn't eventually figure out how to do it, but that the error message I got was confusing because it dictated a nonfunctional solution.

[...]

PR #6354, opened 2024-04-08
ENH: Improved error message for issue #1226

MarkellaEgglezou:
In issue #1226 it is referred that the error messages are not clear. I searched a bit about the proper messages and changed them in order to be more helpful and understandable. Hope it is useful!


peircej:
That sounds fair enough @MarkellaEgglezou thanks

Issue #6355, opened 2024-04-08
[Bug]: Editable TextBox2 line spacing contamination


Catherine9811:
### PsychoPy Version

2023.2.3

### What OS are your PsychoPy running on?

Windows 11

### Bug Description

If multiple editable TextBox2 instance exists within the same experiment, the line spacing parameter seems to be contaminated by the last defined instance of TextBox2. I quickly reproduced it by creating two routine, both containing TextBox2 component but with different line spacing settings.

![image](https://github.com/psychopy/psychopy/assets/51292872/98c566c3-559c-413f-8dd3-fd94f1182ec6)
![image](https://github.com/psychopy/psychopy/assets/51292872/fb7df24b-5ba9-4fd9-91dd-2165f8cd54b7)
![image](https://github.com/psychopy/psychopy/assets/51292872/dd707b6f-2805-4437-a999-09677825255c)

The line spacing for components in the first routine all shares the same line spacing parameters as 2.0, even though they have been set to 1.0 and 0.1 separately.

![image](https://github.com/psychopy/psychopy/assets/51292872/3ba008ba-1241-46bb-9380-681fd6ab53b8)

[bug_reproduce.zip](https://github.com/psychopy/psychopy/files/14904844/bug_reproduce.zip)

### Expected Behaviour

Each editable TextBox2 instance should have individual settings.

### Steps to Reproduce

1. Download the zip file
2. Run and test

Catherine9811:
Created #6356 to fix the bug

Catherine9811:
Close as the fix has been merged

PR #6356, opened 2024-04-09
BF: Fixed line spacing contamination when multiple editable TextBox2 are present

Catherine9811:
Fixed line spacing contamination when multiple editable TextBox2 instance exists in the same experiment by adding extra identifier for lineSpacing parameter in fontManager.

- Modified the way that TextBox2 initialize its font property with the lineSpacing prop
- Modified fontManager getFont function to include lineSpacing as part of the identifier

Example and reproduction codes can be found in this issue #6355 

With this modification, I have the assumption that the bold, italic and lineSpacing property will not be updated until the font property is being re-assigned, but not sure if this is valid or not.

### DIFF
--- a/psychopy/tools/fontmanager.py
+++ b/psychopy/tools/fontmanager.py
@@ -947,7 +947,7 @@ class FontManager():
                 return False
         # If font is found, make glfont
         fontInfo = fontInfos[0]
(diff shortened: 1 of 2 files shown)


peircej:
Thanks @BXYMartin and @Catherine9811 that's now pulled in

PR #6374, opened 2024-04-15
ENH: Fixed #TODO: replace exec() with eval()

MarkellaEgglezou:
In psychopy\app\builder\dialogs\dlgsConditions.py in lines 385 to 426 there was "# todo: replace exec() with eval()". I changed the exec() with eval() as the comment said. Hope this helps!


peircej:
he he, I don't actually know what the aim was of using eval here instead of exec. My guess is that it was aiming to be more secure by not executing code, but I'm not sure if that's actually true! 😆

peircej:
thanks for your contribution(s) @MarkellaEgglezou

PR #6396, opened 2024-04-18
ENH: Improvements to attributeSetter decorator

TEParsons:
Intention being to make it easier for plugin developers to use.

- Adding an attributeSetter attribute now automatically creates `setWhatever`/`getWhatever` functions, reducing `setWhatever is not defined` errors coming out of Builder because xyz developer forgot to create that function
- Querying the value of an attributeSetter attribute before it's been set will now return the decorated function's default value rather than the __getattribute__ method handle, making error messages much clearer and reducing the chance of getting `attributeSetter has no attribute xyz` because you forgot to set something in __init__
- While I was in the neighbourhood I expanded the documentation to help see how to use it

Tested the speed and it doesn't seem any slower, which makes sense as there's not really any extra overhead once the attribute is set once - the only things which take any time are setting the get/set methods (which only happens once, on class import) and generating the method sig to get default (which only happens if the value hasn't been set)

### DIFF
--- a/psychopy/tests/test_misc/test_event.py
+++ b/psychopy/tests/test_misc/test_event.py
@@ -259,7 +259,7 @@ class _baseTest():
         m.getPressed(getTime=True)
 
     def test_isPressedIn(self):
-        m = event.Mouse(self.win, newPos=(0,0))
+        m = event.Mouse(win=self.win, newPos=(0,0))
         s = ShapeStim(self.win, vertices=[[10,10],[10,-10],[-10,-10],[-10,10]], autoLog=False)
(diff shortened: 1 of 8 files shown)


peircej (commented):
This has many (working) lines changed and may be producing errors. The changes don't seem useful to users. So to the user, it feels like this PR is a net loss - more potential bugs and no benefits

peircej:
Closing, This is a substantial refactor. I'm not clear that it's necessary and has a high likelihood of introducing bugs and behaviour changes that we aren't yet aware of. It certainly isn't a priority issue for users to have changed. Let's focus on what the users need fixing

PR #6403, opened 2024-04-22
BF: Always use Arial as default font and package it with PsychoPy

TEParsons:


TEParsons:
Closing this as it doesn't seem worth the effort - it would be nice for PsychoPy library to have the same default font as PsychoPy Builder, but ultimately the reason we want Arial is because it's web standard, which is irrelevant to the Python library. Open Sans seems to work better for the test suite as it's included in app.Resources, so let's just leave it for now

PR #6412, opened 2024-04-26
BF: Speed up color rendering

TEParsons:
See #5790 for more details


TEParsons:
Going back to the drawing board with this one - it's tricky figuring out when validation calls are and aren't needed

PR #6413, opened 2024-04-26
TESTS: get tests working again on macos

peircej:
failing during install of HDF5 but that's because the github runner has switched to using ARM architecture instead of intel


peircej:
This was superceded by 
https://github.com/psychopy/psychopy/pull/6419

PR #6423, opened 2024-04-30
PsychoPy App not launching on Anaconda python >3.9 due to removal of pythonw

peircej:


mh105:
@peircej Wouldn't we want to do `< (3,9)` if the idea is to only enforce `pythonw` calls in older python versions? Also the current conda-forge built for 3.8 is 3.8.19 and also does not come with pythonw. As far as I could tell, python3.6 and after is GUI-enabled python executable already so pythonw might be obsolete.

The latest conda itself only supports 3.8 onwards, and installing python3.8 from either defaults or conda-forge channels installs 3.8.19 and no longer has pythonw (coincidentally would not throw an error here because of the `>= (3,9)`). Perhaps this whole if statement could be removed and just call `start_app()`?

peircej:
For the direction of the comparison, sorry, yes you're right. That's what happens when we fail to test (none of us use conda)!

For the version, I don't know the answer. According to the docs even [if you track back to docs of v2.7](https://docs.python.org/2.7/using/mac.html#running-scripts-with-a-gui) it claims you can use either python or pythonw but I'm certain that wasn't true for wxpython apps or this code would not have been added. So, basically, I don't trust that line of the documentation. But if current conda users are finding they can launch the app without using pythonw then I'm happy for it to be removed. The vast majority of the users are on standalone where we control the python executable ourselves and test constantly, so the affected users will be few if this breaks it. 

Do you want to put in a new PR or shall I?

mh105:
Happy to PR. Do you want this merged to release or dev branch?

peircej:
I guess the release branch makes sense if this is actually a BugFix for current conda users. Then it will get in the 2024.1.6 release

PR #6432, opened 2024-05-02
ENH: Allow findPhotodiode to detect channel automatically

TEParsons:


TEParsons:
Converting this to a draft until I can test with the physical box

PR #6436, opened 2024-05-03
Fix the launch of PsychoPy App on conda envs without pythonw

mh105:
After python3.8, conda installs of python no longer distribute the framework build by default, so `pythonw` is not created in /bin. This is because `python` now natively supports GUI on macOS without the need for `pythonw`. Executing `psychopyApp.py` can successfully open the GUI window in most modern macOS/python/conda platforms. Therefore we check for the existence of a `pythonw` executable in the script on macOS to avoid `FileNotFoundError` in settings without `pythonw`.


mh105:
I'm sorry I forgot that I had this PR open - let me know if you would like me to open a new branch + PR with just the first two commits. The later commits are probably helpful but unrelated to the pythonw fix.

peircej:
Thanks @mh105 
Yes, given the additional non-BF commits I'll move this to target `dev`, then cherry-pick the actual BFs to `release`

PR #6437, opened 2024-05-03
EN: PsychoPy launcher improvements

mdcutone:
This PR makes several enhancements to the PsychoPy launcher: 

* The main PsychoPy app is now started in a sub-process with the correct environment for plugins to work without hacks to `pkg_resources` and `site` modules at runtime. Should be more compatible with py2app on Mac
* Packages can be installed/upgraded/removed prior to starting the main application so they aren't installed into a running environment. Likely possible to do in-place upgrades of PsychoPy in the future.
* PsychoPy is capable of self-restart after crashes or if initiated by the user: no more unrecoverable splash screen crashes
* Session lock file for determining if the app crashed in a previous session to allow for recovery tasks on the next startup

WIP or Planned features:
* Safe mode for recovering from startup crashes which disables plugins
* Hardware enumeration on startup in subprocess to avoid loading PTB and others with the GUI. Save hardware mapping to JSON in user preferences directory for populating dialog fields
* Command line options for clearing plugin cache, etc.

### DIFF
--- a/psychopy/app/__init__.py
+++ b/psychopy/app/__init__.py
@@ -11,6 +11,7 @@
 __all__ = [
     'startApp',
     'quitApp',
+    'restartApp',
     'getAppInstance',
     'getAppFrame',
     'isAppStarted']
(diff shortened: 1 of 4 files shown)


TEParsons:

PR #6439, opened 2024-05-06
Speed-up of MovieStim (noticeable for larger videos)

aforren1:
These were some lower-hanging fruit I noticed when others were trying to play 4k videos via MovieStim.

1. MovieStim's `.play` was being called every `.draw()`, since the MovieStim's state never changed (but the `_player` did)
2. Copying the frame's bytearray was expensive, so use the memoryview and keep the frame alive until the frame is copied to the PBO

(1) saves about 8ms per call on my Windows laptop, and (2) saves about 10ms per call.

The switch to `memmove(...)` was more of a clarity thing, happy to keep/drop...

I also noticed the glTexSubImage2D blocking on my AMD linux laptop, but that might be something more driver-related.

### DIFF
--- a/psychopy/visual/movies/__init__.py
+++ b/psychopy/visual/movies/__init__.py
@@ -20,7 +20,7 @@ from psychopy.tools.filetools import pathToString, defaultStim
 from psychopy.visual.basevisual import (
     BaseVisualStim, DraggingMixin, ContainerMixin, ColorMixin
 )
-from psychopy.constants import FINISHED, NOT_STARTED, PAUSED, PLAYING, STOPPED
+from psychopy.constants import FINISHED, NOT_STARTED, PAUSED, PLAYING, STOPPED, INVALID
 
 from .players import getMoviePlayer
 from .metadata import MovieMetadata, NULL_MOVIE_METADATA
@@ -320,6 +320,17 @@ class MovieStim(BaseVisualStim, DraggingMixin, ColorMixin, ContainerMixin):
     # --------------------------------------------------------------------------
     # Video playback controls and status
     #
+
+    @property
+    def status(self):
+        if self._player is not None:
+            return self._player.status
+        
+        return INVALID
+    
+    @status.setter
+    def status(self, value):
+        pass
 
     @property
     def isPlaying(self):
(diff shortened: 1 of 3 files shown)


peircej:
Thanks @aforren1 (long time no see! 👋 😃  )
@mdcutone this looks good to me but would you look over it too? 
Two key questions:
- is this important/safe enough to warrant release branch or should we move to dev branch ready for summer release after more testing on a range of machines/drivers?
- will it clash with the work you've been doing on updating movie rendering?

TEParsons:
I'm not sure about making `.status` be a property/setter pair - status not being set internally is the correct behaviour, the problem is that status appears to be queried internally (in `MovieStim.draw`). Status is really a Builder thing - it's only set by the compiled Builder code and should only be queried by the compiled Builder code. So I think we need to remove that `self.status == NOT_STARTED` from MovieStim.draw and replace it with `self.isNotStarted`.

Everything else looks good to me :)

aforren1:
Thanks all for looking! Maybe I'll just split the `MovieStim.play()` part into a separate BF & base the rest off the dev branch as an ENH? I get some other substantial wins by switching the texture format to BGRA, but that seems like the sort of thing that needs more testing...

aforren1:
And here's a [link](https://yaleedu-my.sharepoint.com/:v:/g/personal/alexander_forrence_yale_edu/EWxuzXDJmRJBrVBwqcTT0z0BqojhM00TV2fe7Rshx9txdQ?e=MlPeuM) to the video I'm testing with (just a 6sec snippet of https://commons.wikimedia.org/wiki/File:(4k)_Wild_Animal_-_Ultra_HD_Video_TV_60fps_(2160p).webm).

TEParsons:
Splitting into two PRs sounds good :)

aforren1:
@TEParsons re: the `.status`, does any of the Builder machinery rely on the status of the MovieStim? Before this change, I think it would've just been stuck at `NOT_STARTED` even while the video was playing. I'll definitely change the check in `.draw()` to `.isNotStarted` though.

TEParsons:
Builder does use `.status` but it also sets it - so the MovieStim doesn't need to update its status when it starts playing, Builder updates it when it tells it to start playing.

aforren1:
Thanks, opened a new PR with the updated fix and will follow up with the other changes.

PR #6441, opened 2024-05-07
Installed plugins were not being found on Linux/MacOS

mdcutone:
This PR fixes various bugs around plugin installation and discovery. Mainly, we ensure that the directory structure of the package folder is compliant with the scheme for user site packages on a given platform. We also configure environment variables to allow sub-processes to access the user package directory. 

WIP:
* ~Enable PIP installations on MacOS again and see if it works now~
* ~Update build script with new `psychopy.__init__.py` contents~

### DIFF
--- a/building/createInitFile.py
+++ b/building/createInitFile.py
@@ -71,15 +71,15 @@ template = """#!/usr/bin/env python
 import os
 import sys
 
-__version__ = '2024.1.4'
-__license__ = 'GPL v3'
-__author__ = 'Open Science Tools Ltd'
-__author_email__ = 'support@opensciencetools.org'
-__maintainer_email__ = 'support@opensciencetools.org'
-__url__ = 'https://www.psychopy.org/'
-__download_url__ = 'https://github.com/psychopy/psychopy/releases/'
-__git_sha__ = 'n/a'
-__build_platform__ = 'n/a'
+__version__ = '{version}'
+__license__ = '{license}'
+__author__ = '{author}'
+__author_email__ = '{author_email}'
+__maintainer_email__ = '{maintainer_email}'
+__url__ = '{url}'
+__download_url__ = '{download_url}'
+__git_sha__ = '{shaStr}'
+__build_platform__ = '{platform}'
 
 __all__ = ["gui", "misc", "visual", "core",
            "event", "data", "sound", "microphone"]
--- a/building/createInitFile.py
+++ b/building/createInitFile.py
@@ -71,15 +71,15 @@ template = """#!/usr/bin/env python
 import os
 import sys
 
-__version__ = '{version}'
-__license__ = '{license}'
-__author__ = '{author}'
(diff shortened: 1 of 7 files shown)


building/createInitFile.py
```diff
@@ -71,15 +71,15 @@ def createInitFile(dist=None, version=None, sha=None):
...
-__version__ = '{version}'
-__license__ = '{license}'
-__author__ = '{author}'
-__author_email__ = '{author_email}'
-__maintainer_email__ = '{maintainer_email}'
-__url__ = '{url}'
-__download_url__ = '{download_url}'
-__git_sha__ = '{shaStr}'
-__build_platform__ = '{platform}'
+__version__ = '2024.1.4'
```
peircej: These shouldn't be getting hard-coded. Keep them as variables

psychopy/preferences/preferences.py
```diff
@@ -164,29 +164,42 @@ def getPaths(self):
...
+        userBasePath = Path(self.paths['userPrefsDir']) / 'packages'
 
-        # populate directory structure for user-installed packages
-        if not prefixRootDir.is_dir():
-            prefixRootDir.mkdir()
-        
-        userSiteDir = prefixRootDir / 'site-packages'
-        if not userSiteDir.is_dir():
-            userSiteDir.mkdir()
+        # Package paths for custom user site-packages, these should be compiant
```
peircej: typo: compliant

psychopy/preferences/preferences.py
```diff
@@ -164,29 +164,42 @@ def getPaths(self):
...
+        # with platform specify conventions.
```
peircej: typo: specific

psychopy/preferences/preferences.py
```diff
@@ -164,29 +164,42 @@ def getPaths(self):
...
+            userPackagePath.mkdir(parents=True)
+        if not userIncludePath.is_dir():
+            userIncludePath.mkdir(parents=True)
+        if not userScriptsPath.is_dir():
+            userScriptsPath.mkdir(parents=True)
 
         # add paths from plugins/packages (installed by plugins manager)
-        self.paths['userPackages'] = userSiteDir
-        self.paths['userScripts'] = userScriptsDir
+        self.paths['userPackages'] = userBasePath
```
peircej: is that true? It feels like userBasePath is going to be `~/psychopy3` whereas you need `~/psychopy3/packages`
Or, maybe the question is, why do we need `prefs.paths['packages']` and `prefs.paths['userPackages']` ?

peircej: ah, OK, I see the changes now. I guess I find the name `userBasePath` unintuitive, but it looks correct in its operation

mdcutone: Loosely following conventions in `site` for naming since I get them mixed up otherwise, could change them of course.

PR #6443, opened 2024-05-08
Fixed units for mouse gaze calibration targets (pix -> height)

KirstenWilliams:
Units for mousegaze calibration targets were set to pixels instead of height units so they appeared very small. This has now been changed, calibration targets should now be of appropriate size.


TEParsons:
Just realised this was pointing at dev but is probably needed in the next bug fix release, could you reopen another PR into `psychopy:release`?

PR #6450, opened 2024-05-08
Suggest the non-symlink path to the interpreter for `core.rush()` on Debian-derived systems

aforren1:
Really minor thing, `setcap` only operates on non-symlink files. It looks like Python 3 has been the default python since Ubuntu 20.04, so also updated that text to reflect the likely system Python.

### DIFF
--- a/psychopy/platform_specific/linux.py
+++ b/psychopy/platform_specific/linux.py
@@ -9,6 +9,7 @@
 """
 from psychopy import logging
 import sys
+import os.path
 try:
     import ctypes
     import ctypes.util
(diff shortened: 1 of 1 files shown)


peircej:
I'm actually now not certain how this interacts with the suggested config setting at https://github.com/psychopy/psychopy/blob/dev/psychopy/app/linuxconfig/__init__.py#L39
My memory is that allows anybody in the psychopy group (to which the user is added) to re-nice to -20 without needing sudo (but I think that's not taking effect unless you then call `rush`)

aforren1:
Thinking about it more, I think the config should be the way to go rather than setting the capabilities on the executable (i.e. replacing the setcap recommendation).

However, a member of the psychopy group might not be able to set the priority currently requested in `rush` (50 < `sched_get_priority_max()`). Either the `sched_priority` needs to be reduced to fall within the upper bound set by `rtprio` in the config, or the config needs to bump the allowed `rtprio` up to the 99 (see the [sched_setscheduler](https://man7.org/linux/man-pages/man2/sched_setscheduler.2.html) man page). Maybe the former? FWIW, PTB recommends using the [minimum realtime priority](https://github.com/kleinerm/Psychtoolbox-3/blob/c20afa6339e88ef1b96625b3aaab110c4afeb5e9/Psychtoolbox/PsychPriority/Priority.m#L100).

peircej:
Right, yes, I agree. We certainly want `99-psychopylimits.conf` to be compatible with the actual call to rush, and If PTB now recommends a lower value to achieve good timing while still playing nice that sounds ideal to me (and means the code can change without changing the existing config files)

aforren1:
I probably want to hit a few related things at the same time, will open issues to discuss.

PR #6456, opened 2024-05-10
BF: show start parameter for stop only eyetracker components

RebeccaHirst:
Currently the Stop Only action type causes experiments to crash. This is because the stop recording component never has its status set to START in the compiled python code. It does make sense to have a Start time because in theory you might want to Stop following a stimulus or something like that. This edit means that the stop recording component is set to start and therefore the recording is stopped.

### DIFF
--- a/psychopy/experiment/components/eyetracker_record/__init__.py
+++ b/psychopy/experiment/components/eyetracker_record/__init__.py
@@ -54,14 +54,6 @@ class EyetrackerRecordComponent(BaseComponent):
               }
          )
 
-        self.depends.append(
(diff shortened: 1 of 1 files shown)


TEParsons:
I'm not sure about "Stop only" having a start time exposed to the user - it would be better to change the boilerplate so that its status is set regardless of start time if type is "Stop only". What's the use case in which listening for a stop needs a custom start rather than just the beginning of the Routine? If you wanted to stop it when a stimulus finishes, surely you'd just use stimName.status == FINISHED  in the value for "Stop" and "conditional" for "Stop type"?

PR #6457, opened 2024-05-13
ENH: Improvements to start/stop behaviour of Eyetracker Record Components

TEParsons:
Attempt to fix the problem identified in #6456 

The problem:
- An EyetrackerRecordComponent set to "Stop only" doesn't have a start time, so was never marked with `.status = STARTED`
- Marking it with `.status = STARTED` calls a property/setter pair which actually starts the eyetracker, this is generally inadvisable as sometimes we want to just tell Builder it's started (i.e. when there's a new *Component* but not a new *device*, we want to mark the Component as `STARTED` but don't want to start the device again)
- This meant the eyetracker was never stopped so crashed when the experiment finished

The solution(s):
- Instead of `.status = STARTED` starting the eyetracker, have the start code in a function called `startRecording` and call this function explicitly in the Builder code (ditto for stopping the recording)
- Remove the property/setter so that `status` returns to being just an attribute for tracking progress in Builder
- If the component is "Stop only", have the Component set `.status = STARTED` at the very beginning (which we can now do safely)

Bonus:
- I noticed that the dollar signs by the start/stop aren't being hidden/shown according to state, so I fixed that while I was in the neighbourhood

### DIFF
--- a/psychopy/app/builder/dialogs/__init__.py
+++ b/psychopy/app/builder/dialogs/__init__.py
@@ -522,13 +522,16 @@ class StartStopCtrls(wx.GridBagSizer):
             self.estimLabel.Show(visible)
         if hasattr(self, "label"):
             self.label.Show(visible)
+        # show/hide dollars
+        if hasattr(self, "dollar"):
+            self.dollar.Show(visible)
         # Set value to None if hidden (specific to start/stop)
         if not visible:
             if "startVal" in self.ctrls:
                 self.ctrls["startVal"].Value = ""
             if "stopVal" in self.ctrls:
                 self.ctrls["stopVal"].Value = ""
-        # Layout
+        # layout
         self.parent.Layout()
 
     def updateCodeFont(self, evt=None):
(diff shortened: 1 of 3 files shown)


RebeccaHirst:
We have tested this and unfortunately it throws the error message 

  File "C:\Users\lpykgw\Downloads\hello (1)_lastrun.py", line 935, in <module>
    run(
  File "C:\Users\lpykgw\Downloads\hello (1)_lastrun.py", line 807, in run
    if tThisFlipGlobal > etRecord_3.tStartRefresh + 1.0-frameTolerance:
TypeError: unsupported operand type(s) for +: 'NoneType' and 'float'
################ Experiment ended with exit code 1 [pid:12848] #################


this is because tStartRefresh is never created in the compiled python code 


```
if etRecord_3.status == NOT_STARTED:
            etRecord_3.status = STARTED
        # if etRecord_3 is stopping this frame...
        if etRecord_3.status == STARTED:
            # is it time to stop? (based on global clock, using actual start)
            if tThisFlipGlobal > etRecord_3.tStartRefresh + 1.0-frameTolerance:
                # keep track of stop time/frame for later
                etRecord_3.tStop = t  # not accounting for scr refresh
                etRecord_3.tStopRefresh = tThisFlipGlobal  # on global time
                etRecord_3.frameNStop = frameN  # exact frame index
                # add timestamp to datafile
                thisExp.addData('etRecord_3.stopped', t)
                # update status
                etRecord_3.status = FINISHED
                etRecord_3.stopRecording()
```

TEParsons:
Give it a go now with the new commits :)

PR #6458, opened 2024-05-13
BF: KeyboardComponent allowed keys from spreadsheet caused namespace errors

TEParsons:
It's because of local vs global namespaces - the variables from the spreadsheet are defined globally, but because the allowedKeys code in a Keyboard Component can set the variable (when it converts it to a tuple or does eval or etc.) and because the run code is now in a function, Python reserves the local name and so if you reference it you get "Variable not defined" (even though it's in globals).

This fixes it by saying that, if the same conditions under which the allowedKeys code is written is True, add some code to the start of the Routine to use the global variable's value for the undefined local variable.

### DIFF
--- a/psychopy/experiment/components/keyboard/__init__.py
+++ b/psychopy/experiment/components/keyboard/__init__.py
@@ -171,7 +171,7 @@ class KeyboardComponent(BaseDeviceComponent):
 
     def writeRoutineStartCode(self, buff):
         code = (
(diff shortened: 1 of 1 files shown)


TEParsons:
Spotted by [this forum user](https://discourse.psychopy.org/t/how-to-make-changes-allowed-keys-in-routines/39977)

PR #6460, opened 2024-05-14
BF: fix an issue with etRecord component for Stop Only

mh105:
There is a bug with the etRecord component in builder because the constructor for `psychopy.hardware.eyetracker.EyetrackerControl` always initializes the `_status` attribute as `NOT_STARTED`. This causes an issue when the component is used under the "Stop Only" Record actions. In this scenario, two independent etRecord components are initialized, causing the bug, because the experiment can only change the status of the second component (supposed to stop recording) to `FINISHED` if the status is already `STARTED`.

Instead, the status of any `EyetrackerControl` instance should reflect whether recording is currently enabled for the tracker instance passed into the constructor. Looking through `psychopy.iohub.devices.eyetracker.hw`, `isRecordingEnabled()` is appropriately defined for everyone so this fix should work seamlessly.

### DIFF
--- a/psychopy/hardware/eyetracker.py
+++ b/psychopy/hardware/eyetracker.py
@@ -12,7 +12,7 @@ class EyetrackerControl(AttributeGetSetMixin):
     def __init__(self, tracker, actionType="Start and Stop"):
         self.tracker = tracker
         self.actionType = actionType
-        self._status = NOT_STARTED
+        self._status = tracker.isRecordingEnabled()
 
     @property
     def status(self):


peircej:
@RebeccaHirst @TEParsons this is an even shorter changed compared to your solutions in #6457 and #6456

Is this going to be sufficient for the short-term BF (postponing the other discussion about whether "start/stop" times refer to the start _of the component_ versus the start _of the eyetracker_?

TEParsons:
Yep sounds like a plan :) @mh105 would you mind creating a new branch from `psychopy/release` and cherry picking this change across? Then we can pull this in for the next bug fix release and continue working on the other stuff @RebeccaHirst and I are figuring out in the `dev` branch

mh105:
Sure. Just created a new PR to `psychopy/release`. I'll close this now.

PR #6461, opened 2024-05-14
BF: should not fix units="pix" when drawing TargetStim during calibration

mh105:
_This was hell of a bug to hunt down..._

The problem is as the following: when using the EyeTracker Calibration/Validation components in Builder, the GUI allows setting of Outer radius and Inner radius, and the Spatial units is grayed out as following the screen unit in the experiment settings. This is great and parsimonious design. But after some convoluted layers of overloaded `dict()` calls, we get to this part of `psychopy.iohub.devices.eyetracker.calibration.procedure.BaseCalibrationProcedure`. And it assumes the `target_attributes` to contain `outer_diameter` and `inner_diameter` in pixels.

This causes the strange behavior that the validation component works (because it correctly uses unit_type) while the calibration component fails to display any targets, giving the appearance of screen freezing. It is actually because the default value for these components under `height` unit setting for screen is `0.01`. Since it's less than one pixel, nothing got displayed.

Fortunately the fix is simple because the `unit_type` got carried all the way into this function call, and we just need to set `units=unit_type` when using `layout.Size()`. Then everything works nicely and the calibration screen gets displayed.


mh105:
Ah, looks like this was fixed by d62efd5 already, and we had the exact same fix :) Nevertheless, I think the ENH commit above is worth merging, since there's no need to keep the confusing `Pix` variables around when everything can be organized cleanly as arguments for the `visual.TargetStim()` constructor.

TEParsons:
I think that code was added initially as a fix for Tobii eyetrackers, they use their own proprietary calibration code which needs to be in pixels. I'll test this branch with a Tobii next time I've got one to hand (we have one in the office), it may well be unnecessary following yours and @KirstenWilliams's other fixes but want to confirm before pulling in

mh105:
It is possible that the `layout.Size` and `getattr()` call to get the `radius` and `innerRadius` variables are still necessary. Closing this PR since the core fix of using `unit_type` is incorporated in release 2024.1.5.

PR #6474, opened 2024-05-17
BF: Allow meaningful comparisons of KeyPress object instances with the same name

mh105:
This is a subtle bug I noticed when I was trying to understand how `KeyboardDevice.getKeys()` works. There is [this line that checks if a key is still wanted and not present before adding to output](https://github.com/psychopy/psychopy/blob/8558cace4cd07f1d81b68f6d2ee6bcf397fb49d1/psychopy/hardware/keyboard.py#L496), but the `in` operator relies on a robust implementation of `__eq__()` and `__ne__()` dunder methods. Looking at the implementations of them for `KeyPress` objects, they are just comparing the `name` attribute.

This causes a problem because if a subject presses the same key twice in-between `.getKeys()` calls, only the first press/release event of that key would be registered and outputted.

This PR improves how KeyPress object instances are compared by leveraging the `.tDown` attribute, with the design assumption that the same key can't be pressed twice at the exact the same, or at least in this universe :)

### DIFF
--- a/psychopy/hardware/keyboard.py
+++ b/psychopy/hardware/keyboard.py
@@ -584,7 +584,8 @@ class KeyboardDevice(BaseResponseDevice, aliases=["keyboard"]):
             if message.type == "KEYBOARD_PRESS":
                 # if message is from a key down event, make a new response
                 response = KeyPress(code=message.char, tDown=message.time, name=message.key)
-                response.rt = response.tDown - (self.clock.getLastResetTime() - self._iohubKeyboard.clock.getLastResetTime())
+                response.rt = response.tDown - (
+                    self.clock.getLastResetTime() - self._iohubKeyboard.clock.getLastResetTime())
                 self._keysStillDown.append(response)
             else:
                 # if message is from a key up event, alter existing response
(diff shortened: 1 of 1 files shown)


TEParsons:
I think comparing only the value is the correct (and intuitive, for users) thing to do in general - if you're asking "is this keypress the same as this other keypress?" usually the user means is "is this the same key?". Most of the time these kinds of comparisons will be against a string (e.g. `"a" in key_resp.getKeys()`) as users generally won't make a KeyPress object themselves, but they may want to compare a previous response with the current one (e.g. `key_resp_1.keys[-1] in key_resp_2.getKeys()`) and this PR would change how that evaluates.

Instead I think we should change how `getKeys` checks for duplicates to use something other than `in`, [this line](https://github.com/psychopy/psychopy/blob/8558cace4cd07f1d81b68f6d2ee6bcf397fb49d1/psychopy/hardware/keyboard.py#L496) could maybe be `if wanted and not any(k is resp for k in keys):`?

mh105:
Makes sense! I'll open another PR to use the line you suggested. I do need this fixed because otherwise I can't retrieve multiple presses of the same key with `.getKeys()` at the moment.

PR #6476, opened 2024-05-19
DOC: Arabic translation

Ahmad-Abudoush:
one string of Arabic translation


peircej:
Thanks @Ahmad-Abudoush :-)

PR #6478, opened 2024-05-20
BF: allow getKeys() to return the same key pressed multiple times

mh105:
Fixes the bug of `.getKeys()` not returning the same key pressed multiple times using the change suggested by @TEParsons  [here](https://github.com/psychopy/psychopy/pull/6474#issuecomment-2120394130).

This BF maintains the KeyPress() == KeyPress() comparison using only the `name` attribute.


mh105:
Sorry I messed up my detached head of `psychopy/release` when creating this PR. Let me do it again so the PR history is cleaner.

PR #6479, opened 2024-05-20
BF: allow getKeys() to return the same key pressed multiple times

mh105:
_(Replaces #6478 I accidentally messed up detached HEAD with `psychopy/release` using rebase in that PR. This one should be correct)._

Fixes the bug of `.getKeys()` not returning the same key pressed multiple times using the change suggested by @TEParsons  [here](https://github.com/psychopy/psychopy/pull/6474#issuecomment-2120394130).

This BF maintains the `KeyPress() == KeyPress()` comparison using only the `name` attribute.


TEParsons:
This looks good :) Just added a test so we don't backslide on this in future, once the test suite passes (it passed for me locally so I expect it to) I'm happy to pull in

PR #6485, opened 2024-05-22
ENH: Materialized setPavloviaUser method in builder.py

MarkellaEgglezou:
In psychopy\app\builder\builder.py in lines 1635 to 1637 it said: 
def setPavloviaUser(self, user):
    # TODO: update user icon on button to user avatar
    pass
I materialized this method and added it to the code. Hope it is useful!

### DIFF
--- a/psychopy/app/builder/builder.py
+++ b/psychopy/app/builder/builder.py
@@ -1633,8 +1633,28 @@ class BuilderFrame(BaseAuiFrame, handlers.ThemeMixin):
         self.app.runner.panel.runOnlineDebug(evt=evt)
 
     def setPavloviaUser(self, user):
-        # TODO: update user icon on button to user avatar
-        pass
+        if user is None:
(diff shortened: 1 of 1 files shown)


TEParsons:
Thanks for catching this, but I don't think this code will work - BuilderFrame doesn't have an attribute `button` so this function would raise an error. Updating the avatar is already handled by `psychopy.projects.pavlovia:login`:
```
    # update Pavlovia button(s)
    appInstance = app.getAppInstance()
    if appInstance:
        for btn in appInstance.pavloviaButtons['user'] + appInstance.pavloviaButtons['project']:
            btn.updateInfo()
```
and I don't think the `setPavloviaUser` function is actually used (same as with the equivalent function in Coder) so the best course of action may be to just delete the function. Closing this PR but please do put in another removing the function (also happy to do it myself but don't want to deny you credit for finding this)

PR #6486, opened 2024-05-22
ENH: Materialized setPavloviaUser method in coder.py

MarkellaEgglezou:
In psychopy\app\coder\coder.py in lines 2876 to 2878 it said:
def setPavloviaUser(self, user):
# TODO: update user icon on button to user avatar
pass
I materialized this method and added it to the code along with some imports. The code in this method is the same as the code in the namesake method in builder.py. Hope it is useful!


TEParsons:
Closing for the same reason as #6485

PR #6490, opened 2024-05-22
Add hyperlink to plugins page

KirstenWilliams:
take 2: fix hyperlink


RebeccaHirst:
this includes more files than expected for this change

PR #6493, opened 2024-05-23
PKG: uninstall legacy installation from `%appdata%`

shun2wang:
I had a issue while remove old installation and install new version on Windows. new version was detected old packages from these appdata to block me to open the Psychopy setting GUI unless I removed the dir.

I'm not sure of the impact of this, but it seems like the uninstall operation should keep clean?

### DIFF
--- a/buildCompleteInstaller.nsi
+++ b/buildCompleteInstaller.nsi
@@ -198,6 +198,7 @@ Section Uninstall
   Delete "$INSTDIR\${PRODUCT_NAME}.url"
   Delete "$INSTDIR\uninst.exe"
   RMDir /r "$INSTDIR"
+  RMDir /r "$APPDATA\psychopy3"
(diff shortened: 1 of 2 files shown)


peircej:
Hi there, no I think in general this isn't ideal. It would mean that all preferences and plugins get lost when the user updates the software which isn't expected behaviour. What we want, instead, is that the app successfully launches (always) but then detects that something went wrong (e.g. loading incompatible plugins, or incompatible preferences) and then offers to fix the issue (reset preferences, delete existing plugins etc)

shun2wang:
>What we want, instead, is that the app successfully launches (always) but then detects that something went wrong (e.g. loading incompatible plugins, or incompatible preferences) and then offers to fix the issue (reset preferences, delete existing plugins etc)

Good point, thanks!

PR #6500, opened 2024-05-24
ENH: Save dwell times from roi component

RebeccaHirst:
Enhancement to ROI component to save dwell times in addition to timesOn and timesOff 

dwell time is the summed difference between the times On and times Off. 

Requested on forum https://discourse.psychopy.org/t/calculating-looking-time-duration-output-file/25634


TEParsons (approved):
Seems reasonable - usually I'd say "give it to one of the science team to test" but I presume you've already done that ;)

PR #6501, opened 2024-05-24
BF: update param controls to fetch loop templates

RebeccaHirst:
Template xltx files were not being fetched for each loop type because the frame was not returning with a type. Updated so that files are fetched for loops.


TEParsons (changes requested):
Sorry to kill your bugfixing buzz, but this now *only* works for loops - if you try making a Form or a Counterbalance, the excel buttons on these now don't get the correct templates.

Honestly I think it might be better to supply this information from the Param rather than trying to trace parentage in the interface - I'll see what I can come up with without changing too much

PR #6515, opened 2024-05-31
BF: Fixed .js file not being found when using "Use version" (removed _legacy from suffix)

RebeccaHirst:
When a psychopy builder file has been synced to pavlovia, the pavlovia project creates the url run.pavlovia.org/username/projectname where projectname is the name of the js file it is looking for. If you use Use Version then at the moment it compiles projectname_legacy.js - which means that file is not used by the pavlovia project. We need the js file that is compiled to always be the same name as the .psyexp file we are working on and syncing from, even if a different version is used.

Here we prevent the "_legacy.js" ONLY on the js files so it only affects this part of the compiling process.

### DIFF
--- a/psychopy/scripts/psyexpCompile.py
+++ b/psychopy/scripts/psyexpCompile.py
@@ -61,7 +61,7 @@ def generateScript(exp, outfile, target="PsychoPy"):
         # if compiling to JS, js file needs to have legacy filename
         _stem, _ext = os.path.splitext(outfile)
         if _ext == ".js":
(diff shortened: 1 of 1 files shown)


TEParsons:
Okay yep, I've dug a bit deeper and this is the correct solution :) Basically, that code was originally there because the index file was looking for a file with `_legacy`, but now it's (correctly) looking for just the experiment name. I'll just add a commit to delete the lines around it before committing, as they're now doing nothing (just splitting outfile apart then putting it back together unchanged)

Issue #6509, opened 2024-05-29
[Bug]: Builder: Line drawing anchoring behavior is inconsistent with other polygons


tomdstone:
### PsychoPy Version

2024.1.4

### What OS are your PsychoPy running on?

macOS Silicon

### Bug Description

Words with 'apostraphes' surrounding are specific field names in the Polygon Layout view pane. This experimentation was taking place with 'height' set as the units, but I found similar behavior with 'norm'. Popup window shown below to help with clarity.

<img width="622" alt="Screenshot 2024-05-29 at 12 14 37 PM" src="https://github.com/psychopy/psychopy/assets/77251489/a6f215bd-dc53-4059-9b43-787915565a52">

When drawing a non-line polygon, the behavior is that the object is drawn with width and height, translated so that the 'Anchor' point of the object is located at 'Position', and then the object is rotated 'Orientation' degrees clockwise.

With lines, it appears that the line is first drawn centered at `(0,0)` with a width of 'w'. A pivot point is then placed at the 'position' coordinates. The line is then rotated relative to that pivot point 'Orientation' degrees clockwise. The below has a line with width `0.2`, position `(-0.1, 0.1)`, with angles `0, 22.5, 45, 67.5, 90`. You can see below how they pivot clockwise around the `(-0.1, 0.1)` circled position. The 'Anchor' option appears to have no effect. The `(0,0)` origin is in the middle of the rightmost flat line.

<img width="343" alt="Screenshot 2024-05-29 at 12 18 59 PM" src="https://github.com/psychopy/psychopy/assets/77251489/30a69106-35b8-4e10-9235-063732057e81">

### Expected Behaviour

A horizontal line would be placed at 'Position' using the 'Anchor' with line length 'w', and then rotated 'Orientation' around that anchor point.

### Steps to Reproduce

1. Set the 'position' to something not `(0,0)`
2. Vary the 'Orientation' angle among several duplicate lines

PR #6525, opened 2024-06-04
BF: fix double position offset for Line polygon shape

mh105:
As reported by #6509, polygon `Line` shapes in builder do not have correct position setting. After looking into this a bit, I believe the problem comes from the fact that `WindowMixin` methods already take the `.pos` attribute into account when drawing shapes that carry the `Vertices` class instances in the `_vertices` attribute. So subtracting `.pos` again will offset the `Line` shape twice, resulting in incorrect positioning and orientation.

@TEParsons I noticed that the `Vertices.setas()` method is really only used by the `visual.Line` class alone and not by any other polygon shapes. This makes me wonder whether [these lines](https://github.com/psychopy/psychopy/blob/aebf519ede786c4a16749c56f44fd0d89bd519d8/psychopy/layout.py#L831) accounting for flip and anchor are also erroneously repeating what shape drawing functions already take care of, when the changes are non-zero. I didn't have time to look into this too closely but I want to flag it for you.

### DIFF
--- a/psychopy/experiment/components/polygon/__init__.py
+++ b/psychopy/experiment/components/polygon/__init__.py
@@ -187,8 +187,8 @@ class PolygonComponent(BaseVisualComponent):
                     "    size=%(size)s,\n" % inits)
 
         code += ("    ori=%(ori)s, pos=%(pos)s, anchor=%(anchor)s,\n"
-                 "    lineWidth=%(lineWidth)s, "
-                 "    colorSpace=%(colorSpace)s,  lineColor=%(lineColor)s, fillColor=%(fillColor)s,\n"
+                 "    lineWidth=%(lineWidth)s,\n"
+                 "    colorSpace=%(colorSpace)s, lineColor=%(lineColor)s, fillColor=%(fillColor)s,\n"
                  "    opacity=%(opacity)s, " % inits)
 
         depth = -self.getPosInRoutine()
--- a/psychopy/layout.py
+++ b/psychopy/layout.py
@@ -171,16 +171,16 @@ class Vector:
             self.valid = False
 
         # Replace None with the matching window dimension
-        if (value == None).any() or np.isnan(value).any():
+        if (value == None).any() or np.isnan(value).any():  # noqa: E711
             win = Vector((1, 1), units="norm", win=self.win)
             if len(value.shape) == 1:
-                value[value == None] = getattr(win, units)[value == None]
+                value[value == None] = getattr(win, units)[value == None]  # noqa: E711
                 value[np.isnan(value)] = getattr(win, units)[np.isnan(value)]
             else:
                 value[np.isnan(value[:, 0]), 0] = getattr(win, units)[0]
                 value[np.isnan(value[:, 1]), 1] = getattr(win, units)[1]
-                value[value[:, 0] == None, 0] = getattr(win, units)[0]
-                value[value[:, 1] == None, 1] = getattr(win, units)[1]
+                value[value[:, 0] == None, 0] = getattr(win, units)[0]  # noqa: E711
+                value[value[:, 1] == None, 1] = getattr(win, units)[1]  # noqa: E711
 
         assert self.valid, (f"Array of position/size values must be either "
                             f"Nx1, Nx2 or Nx3, not {value.shape}")
--- a/psychopy/layout.py
+++ b/psychopy/layout.py
@@ -832,19 +832,13 @@ class Vertices:
         value *= self._flip
         # Account for anchor
         value -= self.anchorAdjust * getattr(self.size, units)
+        # Account for pos
+        if self.pos is None:
+            raise ValueError(
+                u"Cannot not calculate absolute positions of vertices without "
+                u"a pos attribute")
 
-        # NOTE: This is commented out because the pos attribute is already used
-        # to calculate the absolute position of the vertices during drawing.
-        # If we subtract the pos attribute again, the vertices will be offset
-        # by the pos attribute twice, and the orientation calculation will be off.
-
-        # # Account for pos
-        # if self.pos is None:
-        #     raise ValueError(
-        #         u"Cannot not calculate absolute positions of vertices without "
-        #         u"a pos attribute")
-        # value -= getattr(self.pos, units)
-
+        value -= getattr(self.pos, units)
         self.base = value  # apply
 
     @property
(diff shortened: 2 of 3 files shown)


TEParsons:
I'm not sure about this one - `Vertices.setas` is used extensively, it's just used via the attribute setters (`myStim._vertices.pix = ...` for example will call `myStim._vertices.setas(..., units="pix")`). Accounting for the stimulus position is quite important as it's converting from absolute positions to relative positions (i.e. from positions on screen to vertices). Setting `flip`, `anchor`, etc. in a WindowMixin stimulus doesn't do the calculation there and then, it tells the `Vertices` object what the setting is and the object then does the calculation when next converting the vertices to absolute points for drawing, so there isn't a duplication there; it's subtracting on set and adding on get.

The problem with Line is that it's created by setting a start and stop as vertices (i.e. relative to the size, pos, anchor, flip, etc. of the stimulus), meaning when you change any of these attributes it transforms the shape which the vertices are relative to, rather than transforming the vertices. This gets especially tricky when anchor isn't center, as the point of rotation then becomes the imaginary corner of a box whose size and shape has little to do with the actual position of the line. Maybe the best solution is to just lock Line vertices as `(-0.5, 0.5)` and then transform its size, pos and ori to get the requested start and stop points?

mh105:
@TEParsons Good point. Please see my new BF commit - I believe using the `ShapeStim` constructor with the absolute vertex coordinates for the line is the easiest way to get around the issues created by making an explicit call to `_vertices.setas()`.

However, `anchor` still doesn't work properly for `Line` shapes. I don't know if the concept of `anchor` actually makes sense for Line shapes in the first place. Perhaps when the shape is Line, the anchor option could be grayed out and forced to be `"center"`?

TEParsons:
I agree on anchor not making much sense for Line, I suppose having "start", "center" and "end" as anchor options does make sense but only when the line is set by start/end points rather then size/pos. You could add a new Param to PolygonComponent called "lineAnchor" or something, which is shown/hidden by an entry in `depends` according to whether shape is Line, and give it "center" as the only option. Then when shape=line we can use that param instead of anchor. 

Maybe in future this Param could be extended to have "start" and "end" as options, which I can see being useful if someone wanted to rotate their line from one end (like the hand on a clock), but this is beyond the scope of a bug fix.

I'm not altogether clear what the effect of setting vertices via `BaseShapeStim.vertices` vs setting via `self._vertices.setas()` is - from some preliminary testing of your branch it seems to have the same effect - what's the use case in which `setas` is breaking for you? Do you have a test script which showcases the difference?

mh105:
> I'm not altogether clear what the effect of setting vertices via `BaseShapeStim.vertices` vs setting via `self._vertices.setas()` is - from some preliminary testing of your branch it seems to have the same effect - what's the use case in which `setas` is breaking for you? Do you have a test script which showcases the difference?

Ah, so the bug I encountered is that in release 2024.1.5, if we draw a line **in Builder** with non-zero position, say (0.2, 0.2), then using `self._vertices.setas()` doesn't actually place the line in that position. The line still gets drawn at the center of the screen. I looked at how `._vertices` gets printed after `.setas()`, and it looks like this bug is because the shifted `value` accounting for pos (which I commented out in BF [d81e406](https://github.com/psychopy/psychopy/pull/6525/commits/d81e406eaa5168d7a47924cf60559b39b28e5f75)) is cancelling out whatever position shift stored in the `.pos` attribute that gets used when the vertices are drawn. This is why I believe there is double position offset going on for the Line polygon shape. If you add non-zero `Orientation` on top of this bug, we get what  #6509 is reporting.

I have no idea why passing the same `(start, end)` into `super(Line, self).__init__()` as proposed in BF [69226ba](https://github.com/psychopy/psychopy/pull/6525/commits/69226ba8983518c8cc90a914b9fcd0eb1a58e36d) doesn't have the same buggy behavior. I suspect it's something you said about the paired use of `.setas()` and `.getas()`. Maybe calling `.setas()` directly causes a unilateral shift of `_vertices.base` that isn't matched by another `.getas()`.

Could you see if you can reproduce the bug above? If so and if you have some time, I would greatly appreciate if you could help add the modification to PolygonComponent to handle the special case of `anchor` for line shapes. I just haven't touched that part of PsychoPy yet, and I don't want to inadvertently introduce new bugs because I don't understand how the dependencies around GUI are handled under the hood 😅

TEParsons:
Ah I see what you mean - I was just testing direct from code so wasn't seeing any difference, but Builder sets both "pos" and "start/end", hence what you're seeing. So Builder is saying "move the shape (line) to here, then ignore that and set its start and stop points to this (relative to center)"... Whereas setting vertices as you've done here means that pos and size are preserved.

I'm not sure why the Builder code is like that in the first place - it's setting start/end as if size is absolute but position is relative, which I don't think has ever been the case. I'll pull this in as it does fix the position, though size is still broken so I'll make a subsequent PR so fix that too, and I'll add the Builder code to grey out anchor there too.

mh105:
>  ...though size is still broken so I'll make a subsequent PR so fix that too, and I'll add the Builder code to grey out anchor there too.

Sorry would you mind explaining a little in what way is size still broken? In my testing within this branch, it seems like the size of the line is set correctly, even with non-zero `pos` and Orientation. Perhaps I missed a subtle difference here.

mh105:
@TEParsons I'll pull the `dev` and test out whether the Line shapes we are trying to draw are showing up correctly. But I wanted to follow up on your previous comment about how the size was still broken. What did you mean by that? I just want to make sure that there isn't an incorrect size that I'm not aware of and therefore not paying attention to when testing the `dev` behavior. Thanks a lot!

TEParsons:
I'll try to create a script/experiment to recreate the behaviour, but for me at least when I set the size and put the line in front of a rectangle with the same size, the line always seemed to be bigger.

mh105:
> I'll try to create a script/experiment to recreate the behaviour, but for me at least when I set the size and put the line in front of a rectangle with the same size, the line always seemed to be bigger.

@TEParsons I just restored this branch and tested it out. Commit 69226ba seems to get the job done correctly - the Line I get matches the rectangle specified with the same size and unit, and this is true regardless of `pos` and `Orientation`. Could you try recreating the behavior you encountered?

Also, I just tested out #6543, which is a PR into `dev`. It unfortunately doesn't draw correctly at the moment. I've added a comment in that merged PR. Maybe we should revert.

PR #6535, opened 2024-06-07
RF: Rush cleanup, Linux+Builder permissions check

### LINKED ISSUE #6530: [Feature Request]: Builder: detect whether a Linux user can already change the priority, rather than checking for whether the .conf file exists?
### Description

Detect whether the current user has the power to change the priority of the process, rather than looking for an existing `99-psychopylimits.conf` file.

### Purpose

We pre-install PTB on our machines, and through that process users are already members of a group that can change nice/rtprio. When users run the Builder, they're prompted to add a config file that is redundant (in our case) with the existing PTB config/group.

Another potential issue (that I haven't run into & I don't know how likely it is) is that if a user didn't completely follow the instructions, the .conf file _could_ exist, but the user might not be a member of the proper group yet.

### Is your feature request related to a problem?

_No response_

### Describe the solution you would like

I would return `False` in the `rush` implementations when the user fails to change the priority, such that PsychoPy can detect whether the priority change actually succeeded or not (currently only "fails" when the ctypes import fails). The Builder can then check whether `Priority(1)` succeeds, and display the config file info if it fails (or be more nuanced about "file exists, but user can't change priority"). 

This does change what `False` means in this context. Another potential drawback is that it would fail to detect if the user has already applied the current advice of setting `cap_sys_nice`, and I'm not sure if that's entirely equivalent to everything the config file does.

For uniformity, I would apply this change to all the `rush` implementations.

### Describe alternatives that you have considered

_No response_

### Additional context

_No response_

aforren1:
- All the `rush` implementations now return False if the function fails to change the priority, rather than only if ctypes-related things failed.
 - Removed the 'posix' platform, which didn't have an implementation and didn't seem to be a possible value? https://docs.python.org/3/library/sys.html#sys.platform
 - Windows `rush` now handles truthy/falsy values, rather than strict true/false
 - Simplified the Linux `rush` impl, and updated the docstring + warning text
   - Is it too verbose? Also just copied from the app help text...
 - For builder, replaced the initial config file check with a test of whether the user could already change the priority

Only tested on Ubuntu 20.04 so far, and don't have access to a Mac to sanity check those changes.

Closes https://github.com/psychopy/psychopy/issues/6530


peircej:
Thanks Alex!

peircej:
FYI we're now working towards a mid-July release for 2024.2.0 and this will be in there

aforren1:
Thanks, and sounds good!

PR #6543, opened 2024-06-10
BF: Fix size and pos of Line

### LINKED ISSUE #6509: [Bug]: Builder: Line drawing anchoring behavior is inconsistent with other polygons

TEParsons:
Previously, start/end of Line were set by figuring out the relative vertices needed for the start/end points of the line to appear in the absolute positions specified. This lead to some strange behaviour (described in #6509 and #6525) around setting position/size/anchor/orientation.

This PR instead keeps Line vertices fixed as going from the bottom left to top right corner, and sets start/end by adjusting the Line's size and pos to put the relevant corner (bottom left = start, top right = end) in the desired position while keeping the other in the same place as before. This means orientation, anchor, etc. all work as with any other ShapeStim.

Fixes #6509


TEParsons:
@mh105 could you pull this down and try it out to make sure Line behaves as you'd expect?

mh105:
@TEParsons @peircej Sorry I was a bit delayed in testing this. Unfortunately the implementation here in PR to `dev` doesn't draw rotated lines correctly. It shows up like this:

Screenshot 2024-06-15 at 10 47 32

While the fix in #6525 to `release` does draw correctly like this:

Screenshot 2024-06-15 at 10 49 35

I'm attaching the `psyexp` and `py` files here as well so you can test it.
[line_drawing.zip](https://github.com/user-attachments/files/15849253/line_drawing.zip)

Should we revert the PR merge and restore the branch to continue working on it?

PR #6559, opened 2024-06-12
BF: take into account that speaker index could jump over integers

mh105:
I added a logging error warning in #6538 if the specified speaker index is out of bounds, but I forgot to consider that the indices may skip numbers so they could exceed the length of `profiles`. This BF uses `not in` a list comprehension to get around that. Sorry about that!

### DIFF
--- a/psychopy/hardware/speaker.py
+++ b/psychopy/hardware/speaker.py
@@ -7,11 +7,11 @@ class SpeakerDevice(BaseDevice):
     def __init__(self, index):
         profiles = self.getAvailableDevices()
 
-        # use first device if index is default
+        # if index is default (-1), setup a default device index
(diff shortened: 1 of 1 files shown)


mh105:
Although when I accidentally triggered the error logging, I noticed that it shows up in stdout as DEPRECATED instead of ERROR. Perhaps I shouldn't be using `logging.error()` and should use something else? @TEParsons

TEParsons:
I think DEPRECATED rather than ERROR is a bug in how Runner/Coder display log messages, probably because the constants DEPRECATED and ERROR have the same value. `logging.error()` is the correct function, we need to fix it further up.

mh105:
Sounds good. Could you merge this when you get a chance? It'll fix the indexing issue I introduced in the previous PR. Thanks!

Issue #6557, opened 2024-06-12
[Bug]: Responses interrupting sound component prevent said component running in the next trial in a loop


tomdstone:
### PsychoPy Version

2024.1.5

### What OS are your PsychoPy running on?

macOS Silicon

### Bug Description

When an audio component in Builder starts at a non-zero time and lasts a finite duration, pressing a response that forces the end of the routine while the audio is playing prevents it from running in the next loop of the trial. In the example below, the audio stimuli is an 'A' note, starts at time 1, and lasts 2 seconds. The response options all last forever. If a response is pressed at, say, 2 seconds, the next run of the routine begins but the audio doesn't play. If pressed before or after the audio stimuli, it does play in the next loop.

<img width="948" alt="Screenshot 2024-06-12 at 8 42 50 AM" src="https://github.com/psychopy/psychopy/assets/77251489/7b30adf9-fdfe-44e3-80a9-ce165337625e">

Some observations:
- I can only reproduce the bug when the audio starts at a time > 0 and has a finite duration
- If there is another sound component that isn't interrupted, say for example one that went from time 0.25 to time 0.75 and the button is pressed at time 2, that component plays just fine in the next loop, but the interrupted one doesn't.
- Button, mouse, and key responses all have this bug
- Adding a code component that at the end of the routine says `audio.status = FINISHED` fixes this behavior
- So I suspect that it has something to do with how audio components are handled when a force quit routine happens before `audio.isFinished = True` occurs, which would usually trigger:
```
if audio.isPlaying:
    audio.status = STARTED
elif audio.isFinished:
    audio.status = FINISHED
```

### Expected Behaviour

In the above builder setup, pressing the button while the audio component is playing stops the sound for one second and then the sound plays again. Instead, the sound doesn't play again.

### Steps to Reproduce

1. Create an audio component starting at time 1 with duration 2
2. Create a response component that lasts forever and starts at or before the time the audio component starts
3. Create a loop around the single routine with many repeats
4. When running the experiment, activate the response component while the audio component is playing, the sound will cut out and not come back on.
5. a psychopy experiment file is located below with the extension changed to `.txt` so that it could be uploaded to github.

tomdstone:
This is the `.psyexp` file in the example with the extention changed to a `.txt` so I could upload it to github.
[audio_bug_testing.txt](https://github.com/user-attachments/files/15807855/audio_bug_testing.txt)

PR #6560, opened 2024-06-12
BF: correct how sound component status is updated in loops of routines

mh105:
There are three inter-related bugs in release 2024.1.5 with the PTB audio backend (likely with other backends as well) in Builder generated scripts under a specific setting when:
1) a routine containing a sound component is repeated in a loop, and 
2) there is a concurrent response component that can force end the routine.

The most obvious behavior is reported in #6557, but they have other more subtle unwanted outcomes. All three problems came from logics around [this section](https://github.com/psychopy/psychopy/blob/f53539e50633afb044d212385b99b08f89925e37/psychopy/sound/backend_ptb.py#L370) of `@property def status(self)` in the PTB backend.

1. Using `stop()` does not properly end a sound component.
**Explanation:** This bug is because in the `writeFrameCode()`, `.stop()` is called after `.status = FINISHED`. This means when `.status` was called, `._isPlaying` is still `True`, hence no `_EOS()` call is triggered as it should. Then the `._isFinished` attribute is never turned to `True` for the sound component. Also, after `.status` gets updated to `FINISHED`, even if we make another call to trigger the `@property status` function, we can't enter the `_EOS` if statement since `status` is no longer `STARTED`. So the ordering of setting status and `stop()` is an issue. Calling `stop()` first and then `.status = FINISHED` would have been perfectly fine, but it's tricky to do with re-using the `BaseDeviceComponent.writeStopTestCode()` to write this part of the code in generating scripts with Builder.
**Bug Fix:** I played around with a few approaches to handle this, including making additional `.status` calls or exposing a new method. But really the cleanest way is to call `_EOS()` directly, which will in turn call `stop()`. I do realize calling a private function isn't good practice, so feel free to push to this branch to make a wrapper to `_EOS()` and call that instead. But under the hood, calling `_EOS()` instead of `stop()` is what we need to execute at the stopping frame, which will properly set `._isFinished` to `False`.

2. Using `pause()` at the end of a routine does not set the status correctly.
**Explanation:** At the end of a routine, `pause()` sets `._isPlaying` to `False` but does not update the `.status`. At the next loop of the routine, `.seek(0)` doesn't change `_isFinished` because of the first bug, and the `.status = NOT_STARTED` immediately below will trigger the `@property status` function, since `.status` is `STARTED` before the setter, this calls the `_EOS()` routine and sets `._isFinished` to `True`. Because of the bug below that repeatedly updates `status` in every frame, if the sound component is not scheduled to start at time 0, the first frame will toggle the `.status` to `FINISHED` since `._isFinished` is `True`, and the sound on this next loop never gets played.
**Bug Fix:** I added `.status = PAUSED` after `pause()` at the end of a routine, which promptly triggers the `@property status` function and updates the status.

3. Setting sound component status every frame can end / start sound playing at incorrect times.
**Explanation:** I'm not sure what led to adding this extra section in the first place, but it looks redundant from the codes already there checking for sound component start/end in each frame. In fact, invoking the `@property status` function on every frame could have unexpected buggy behaviors. For example, when a sound finishes playing in-between two frames, this status update section will call `.status = STARTED` because `._isPlaying` is still `True`. Then since the sound stream has reached the end of the sound file, the `._EOS()` function gets called, and then `.stop()` gets called before the next frame that is supposed to log when the sound stopped, introducing an offset in the logged timing.
**Bug Fix:** I simply deleted this section and in all my testing, sound components seem to play just fine. It seems dangerous to me to use `isPlaying` and `isFinished` to update status, since these two private attributes are not directly accessed by users. Having this status update every frame will often skip important starting and stopping frame code sections gated by if statements like `.status == NOT_STARTED and tThisFlip >= 0.0-frameTolerance` and `.status == STARTED`.

### DIFF
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -223,7 +223,7 @@ class SoundComponent(BaseDeviceComponent):
         # Write stop code
         indented = self.writeStopTestCode(buff)
         if indented:
-            code = ("%(name)s.stop()\n")
+            code = ("%(name)s._EOS()\n")
             buff.writeIndentedLines(code % self.params)
         # because of the 'if' statement of the time test
         buff.setIndentLevel(-indented, relative=True)
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -228,16 +228,6 @@ class SoundComponent(BaseDeviceComponent):
         # because of the 'if' statement of the time test
         buff.setIndentLevel(-indented, relative=True)
 
-        # Update status
-        code = (
-            "# update %(name)s status according to whether it's playing\n"
-            "if %(name)s.isPlaying:\n"
-            "    %(name)s.status = STARTED\n"
-            "elif %(name)s.isFinished:\n"
-            "    %(name)s.status = FINISHED\n"
-        )
-        buff.writeIndentedLines(code % self.params)
-
     def writeFrameCodeJS(self, buff):
         """Write the code that will be called every frame
         """
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -293,6 +293,7 @@ class SoundComponent(BaseDeviceComponent):
             # stop at the end of the Routine, if requested
             code = (
                 "%(name)s.pause()  # ensure sound has stopped at end of Routine\n"
+                "%(name)s.status = PAUSED\n"
             )
             buff.writeIndentedLines(code % self.params)
         # get parent to write code too (e.g. store onset/offset times)


TEParsons:
The code to continuously check `isPlaying` and `isFinished` was added because users were using `.status` to query whether the sound had finished playing so it made sense for status to be FINISHED when the sound had finished (even if the Component still had some time left). 

I agree with you that it doesn't make sense when status can also trigger calls affecting playback, it has the potential to get worryingly circular, but rather than not setting status I think it's better for status to not affect playback (other than by e.g. ending the Routine). That way the use case you describe (where a Sound is ended before its Component's stop condition is hit) is still handled, as `.pause()` is called on Routine end and `.seek(0)` at Routine start, so without `.status` setters complicating things this should return `.isFinished` to `False` as intended.

I'll close this PR and put in one to dev with broader changes to `.status` to bring this about, if that's okay (don't wanna steal your thunder!). If you have time could you pull it down and test the use cases you describe here? If you could construct a .psyexp file which raises an error when this behaviour isn't correct, I think it would be a great addition to the test suite - you could maybe add a `test_SoundComponent` file to `psychopy.tests.test_experiment.test_components` with some psyexp files representing common use cases and some minimal code to compile and run them via `psychopy.session.Session`? Happy to walk you through it if any of that is unfamiliar :)

mh105:
Thanks for the detailed thinking here @TEParsons! What you said makes sense, although one would have to be very explicit in handling status and `._isPlaying` and `._isFinished` separately then. I'll take a look at your `dev` PR when I get a chance. 

I don't know I'm in a good position to add the test suites, as I only know how my groups need to use the sound component rather than comprehensive use cases (such as other users are using `status` to query whether the sound has finished).

PR #6562, opened 2024-06-13
RF: Decouple setting status from triggering playback/clock resets/etc.

### LINKED ISSUE #6557: [Bug]: Responses interrupting sound component prevent said component running in the next trial in a loop

TEParsons:
More detail:
- Camera and SoundPysound had setters for status which didn't do anything
- Rather than ButtonStim assuming it needs to reset its clock when set to STARTED from NOT_STARTED, explicitly reset its clock in the Builder code
- As identified by @mh105 in #6560, SoundPTB was calling `_EOS()` when its status was set to STARTED, which could stop playback if `isFinished` and `isPlaying` aren't set correctly (which, if the previous Routine was force ended, they weren't) - this PR addresses this by removing the `.status` attribute setter and instead updating `isFinished`/`isPlaying` when they are queried

Fixes #6557

### DIFF
--- a/psychopy/experiment/components/button/__init__.py
+++ b/psychopy/experiment/components/button/__init__.py
@@ -275,6 +275,7 @@ class ButtonComponent(BaseVisualComponent):
         indented = self.writeStartTestCode(buff)
         if indented:
             code = (
+                "%(name)s.buttonClock.reset()"
                 "%(name)s.setAutoDraw(True)\n"
             )
             buff.writeIndentedLines(code % self.params)
--- a/psychopy/experiment/components/button/__init__.py
+++ b/psychopy/experiment/components/button/__init__.py
@@ -275,7 +275,7 @@ class ButtonComponent(BaseVisualComponent):
         indented = self.writeStartTestCode(buff)
         if indented:
             code = (
-                "%(name)s.buttonClock.reset()"
+                "win.callOnFlip(%(name)s.buttonClock.reset)\n"
                 "%(name)s.setAutoDraw(True)\n"
             )
             buff.writeIndentedLines(code % self.params)
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -223,7 +223,7 @@ class SoundComponent(BaseDeviceComponent):
         buff.setIndentLevel(-indented, relative=True)
 
         # write code for stopping
-        indented = self.writeStopTestCode(buff)
+        indented = self.writeStopTestCode(buff, extra=" or %(name)s.isFinished")
         if indented:
             code = ("%(name)s.stop()\n")
             buff.writeIndentedLines(code % self.params)
@@ -231,15 +231,8 @@ class SoundComponent(BaseDeviceComponent):
         buff.setIndentLevel(-indented, relative=True)
 
         # write code to run while active
-        indented = self.writeActiveTestCode(buff)
+        indented = self.writeActiveTestCode(buff, extra=" or %(name)s.isPlaying")
         if indented:
-            # update status according to playback
-            code = (
-                "# if sound is finished but %(name)s has time left, finish it now\n"
-                "if %(name)s.isFinished:\n"
-                "    %(name)s.status = FINISHED\n" 
-            )
-            buff.writeIndentedLines(code % self.params)
             # dedent
             buff.setIndentLevel(-indented, relative=True)
 
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -230,13 +230,18 @@ class SoundComponent(BaseDeviceComponent):
         # because of the 'if' statement of the time test
         buff.setIndentLevel(-indented, relative=True)
 
-        # update status according to playback
-        code = (
-            "# if sound is finished but %(name)s has time left, finish it now\n"
-            "if %(name)s.isFinished:\n"
-            "    %(name)s.status = FINISHED\n" 
-        )
-        buff.writeIndentedLines(code % self.params)
+        # write code to run while active
+        indented = self.writeActiveTestCode(buff)
+        if indented:
+            # update status according to playback
+            code = (
+                "# if sound is finished but %(name)s has time left, finish it now\n"
+                "if %(name)s.isFinished:\n"
+                "    %(name)s.status = FINISHED\n" 
+            )
+            buff.writeIndentedLines(code % self.params)
+            # dedent
+            buff.setIndentLevel(-indented, relative=True)
 
     def writeFrameCodeJS(self, buff):
         """Write the code that will be called every frame
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -211,32 +211,37 @@ class SoundComponent(BaseDeviceComponent):
         """
         # Write start code
         self.writeParamUpdates(buff, 'set every frame')
+
+        # write code for starting
         indented = self.writeStartTestCode(buff)
         if indented:
             if self.params['syncScreenRefresh'].val:
                 code = ("%(name)s.play(when=win)  # sync with win flip\n") % self.params
             else:
                 code = "%(name)s.play()  # start the sound (it finishes automatically)\n" % self.params
-            buff.writeIndented(code)
+            buff.writeIndentedLines(code % self.params)
         buff.setIndentLevel(-indented, relative=True)
 
-        # Write stop code
+        # write code while active
+        indented = self.writeActiveTestCode(buff)
+        if indented:
+            code = (
+                "# update %(name)s status according to whether it's playing\n"
+                "if %(name)s.isPlaying:\n"
+                "    %(name)s.status = STARTED\n"
+                "elif %(name)s.isFinished:\n"
+                "    %(name)s.status = FINISHED\n" 
+            )
+            buff.writeIndentedLines(code % self.params)
+        buff.setIndentLevel(-indented, relative=True)
+
+        # write code for stopping
         indented = self.writeStopTestCode(buff)
         if indented:
             code = ("%(name)s.stop()\n")
             buff.writeIndentedLines(code % self.params)
         # because of the 'if' statement of the time test
         buff.setIndentLevel(-indented, relative=True)
-
-        # Update status
-        code = (
-            "# update %(name)s status according to whether it's playing\n"
-            "if %(name)s.isPlaying:\n"
-            "    %(name)s.status = STARTED\n"
-            "elif %(name)s.isFinished:\n"
-            "    %(name)s.status = FINISHED\n"
-        )
-        buff.writeIndentedLines(code % self.params)
 
     def writeFrameCodeJS(self, buff):
         """Write the code that will be called every frame
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -232,10 +232,8 @@ class SoundComponent(BaseDeviceComponent):
 
         # update status according to playback
         code = (
-            "# update %(name)s status according to whether it's playing\n"
-            "if %(name)s.isPlaying:\n"
-            "    %(name)s.status = STARTED\n"
-            "elif %(name)s.isFinished:\n"
+            "# if sound is finished but %(name)s has time left, finish it now\n"
+            "if %(name)s.isFinished:\n"
             "    %(name)s.status = FINISHED\n" 
         )
         buff.writeIndentedLines(code % self.params)
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -222,18 +222,15 @@ class SoundComponent(BaseDeviceComponent):
             buff.writeIndentedLines(code % self.params)
         buff.setIndentLevel(-indented, relative=True)
 
-        # write code while active
-        indented = self.writeActiveTestCode(buff)
-        if indented:
-            code = (
-                "# update %(name)s status according to whether it's playing\n"
-                "if %(name)s.isPlaying:\n"
-                "    %(name)s.status = STARTED\n"
-                "elif %(name)s.isFinished:\n"
-                "    %(name)s.status = FINISHED\n" 
-            )
-            buff.writeIndentedLines(code % self.params)
-        buff.setIndentLevel(-indented, relative=True)
+        # update status according to playback
+        code = (
+            "# update %(name)s status according to whether it's playing\n"
+            "if %(name)s.isPlaying:\n"
+            "    %(name)s.status = STARTED\n"
+            "elif %(name)s.isFinished:\n"
+            "    %(name)s.status = FINISHED\n" 
+        )
+        buff.writeIndentedLines(code % self.params)
 
         # write code for stopping
         indented = self.writeStopTestCode(buff)
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -222,6 +222,14 @@ class SoundComponent(BaseDeviceComponent):
             buff.writeIndentedLines(code % self.params)
         buff.setIndentLevel(-indented, relative=True)
 
+        # write code for stopping
+        indented = self.writeStopTestCode(buff)
+        if indented:
+            code = ("%(name)s.stop()\n")
+            buff.writeIndentedLines(code % self.params)
+        # because of the 'if' statement of the time test
+        buff.setIndentLevel(-indented, relative=True)
+
         # update status according to playback
         code = (
             "# update %(name)s status according to whether it's playing\n"
@@ -231,14 +239,6 @@ class SoundComponent(BaseDeviceComponent):
             "    %(name)s.status = FINISHED\n" 
         )
         buff.writeIndentedLines(code % self.params)
-
-        # write code for stopping
-        indented = self.writeStopTestCode(buff)
-        if indented:
-            code = ("%(name)s.stop()\n")
-            buff.writeIndentedLines(code % self.params)
-        # because of the 'if' statement of the time test
-        buff.setIndentLevel(-indented, relative=True)
 
     def writeFrameCodeJS(self, buff):
         """Write the code that will be called every frame
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -216,9 +216,9 @@ class SoundComponent(BaseDeviceComponent):
         indented = self.writeStartTestCode(buff)
         if indented:
             if self.params['syncScreenRefresh'].val:
-                code = ("%(name)s.play(when=win)  # sync with win flip\n") % self.params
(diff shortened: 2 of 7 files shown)


psychopy/experiment/components/button/__init__.py
```diff
@@ -275,6 +275,7 @@ def writeFrameCode(self, buff):
         indented = self.writeStartTestCode(buff)
         if indented:
             code = (
+                "%(name)s.buttonClock.reset()"
```
peircej: the clock for the button should surely be set during the win.flip, shouldn't it? When you set autodraw to be true the button is not visible. I think this should be a callOnFlip call

peircej (changes requested):

psychopy/experiment/components/button/__init__.py
TEParsons: Good call, just made the change

peircej (approved):

psychopy/experiment/components/sound/__init__.py
```diff
@@ -211,33 +211,35 @@ def writeFrameCode(self, buff):
         """
         # Write start code
         self.writeParamUpdates(buff, 'set every frame')
+
+        # write code for starting
         indented = self.writeStartTestCode(buff)
         if indented:
             if self.params['syncScreenRefresh'].val:
                 code = ("%(name)s.play(when=win)  # sync with win flip\n") % self.params
```
mh105: I think these two ` % self.params` should be removed since now you use `buff.writeIndentedLines(code % self.params)`. Having the extra specifier is harmless but still redundant coding.

psychopy/experiment/components/sound/__init__.py
```diff
@@ -211,33 +211,35 @@ def writeFrameCode(self, buff):
...
-        # Write stop code
-        indented = self.writeStopTestCode(buff)
-        if indented:
-            code = ("%(name)s.stop()\n")
             buff.writeIndentedLines(code % self.params)
-        # because of the 'if' statement of the time test
         buff.setIndentLevel(-indented, relative=True)
 
-        # Update status
+        # update status according to playback
```
mh105: This status update needs to occur after the `# write code for stopping` section. Otherwise imagine we are on the frame that the sound has finished playing. Then calling the new `.isFinished()` triggers `_checkPlaybackFinished()` and then `._isFinished` gets set to `True`, then `.status` gets set to `FINISHED`. And the routine frame will never enter into the stopping frame section, and `.stop()` never gets called.

Moving it to after the stopping section will work fine.

psychopy/experiment/components/sound/__init__.py
```diff
@@ -211,33 +211,35 @@ def writeFrameCode(self, buff):
...
         code = (
             "# update %(name)s status according to whether it's playing\n"
             "if %(name)s.isPlaying:\n"
```
mh105: I suggest that we remove the if statement checking on `.isPlaying`/updating status to `STARTED`. This is already taken care of by the starting frame section, and having this here will only repeatedly set `status` to `STARTED` on every frame, which might make future code maintenance harder if things are not working correctly.

The comment could also instead be `"# update %(name)s status according to whether finished\n"`

psychopy/sound/backend_ptb.py
```diff
@@ -340,11 +340,13 @@ def __init__(self, value="C", secs=0.5, octave=4, stereo=-1,
     @property
     def isPlaying(self):
         """`True` if the audio playback is ongoing."""
+        self._checkPlaybackFinished()
```
mh105: Since `._checkPlaybackFinished()` doesn't affect `._isPlaying`, maybe we shouldn't add the extra method call here.

psychopy/sound/backend_ptb.py
```diff
@@ -340,11 +340,13 @@ def __init__(self, value="C", secs=0.5, octave=4, stereo=-1,
...
         return self._isPlaying
 
     @property
     def isFinished(self):
         """`True` if the audio playback has completed."""
+        self._checkPlaybackFinished()
         return self._isFinished
```
mh105: We can have this method simply `return self._checkPlaybackFinished()`

psychopy/sound/backend_ptb.py
```diff
@@ -360,27 +362,6 @@ def statusDetailed(self):
...
-        NOT_STARTED, STARTED, FINISHED, PAUSED
-
-        Psychtoolbox sounds also have a statusDetailed property with further info"""
-
-        if self.__dict__['status']==STARTED:
-            # check portaudio to see if still playing
-            pa_status = self.statusDetailed
-            if not pa_status['Active'] and pa_status['State']==0:
-                # we were playing and now not so presumably FINISHED
-                self._EOS()
```
mh105, line 375: With the `@status.setter` and `@property def status()` removed, the method `_EOS()` is no longer reachable unless explicitly called. Previously by accessing `status` every frame, the status property function
```
        if self.__dict__['status']==STARTED:
            # check portaudio to see if still playing
            pa_status = self.statusDetailed
            if not pa_status['Active'] and pa_status['State']==0:
                # we were playing and now not so presumably FINISHED
                self._EOS()
```
acts as a reporter to triple serve three different purposes through `_EOS()`  when `not pa_status['Active'] and pa_status['State']==0`:
1. accumulate `._loopsFinished` to make looped sound playing work.
2. call `.stop()`
3. flip `_isFinished` to `True` only when no looping.

To continue with these operations, what if we change `_checkPlaybackFinished()` to the following:
```
    def _checkPlaybackFinished(self):
        """Checks whether playback has finished by looking up the status.
        """
        pa_status = self.statusDetailed
        isFinished = not pa_status['Active'] and pa_status['State'] == 0
        if isFinished and not self._isFinished:
            self._EOS()
        else:
            self._isFinished = isFinished
        return self._isFinished
```

mh105: One small caveat is that if `self.loops > 0`, then there will be a period of time if somehow `not pa_status['Active'] and pa_status['State'] == 0` evaluates to `True`, but `._loopsFinished` hasn't reached the target `.loops` number, then the above proposal will accumulate `._loopsFinished` by 1 every time `isFinished()` is called. But I guess this may be a ok behavior, since if a sound should be looping while `not pa_status['Active'] and pa_status['State'] == 0` is `True`, it likely indicates that a loop has finished. The existing solution with `@status.setter` isn't great either since it is gated by `self.__dict__['status']==STARTED:`.

mh105 (changes requested):
There are two more things that I noticed that I want to flag for you. They aren't changed in this PR so I couldn't do inline comments.

1. Looking at `_EOS()` in the backend_ptb.py, I don't know if it is actually correct.
In the else if statement
```
        elif 0 < self.loops <= self._loopsFinished:
            self.stop(reset=reset, log=False)
```
shouldn't it also set `self._isFinished = True`? Since the accumulating counter `._loopsFinished` has exceeded `.loops`, it should indicate that the loops of sound have finished, right?

2. in the definition of `pause()`, the call of `stop()` should be `self.stop(reset=False, log=False)`, otherwise there will be double logging.

mh105:
Sorry about the number of comments. I didn't realize implementing a sound backend has these intricacies... which makes me appreciate more of the amazing work you guys are doing to make PsychoPy open source and available to everyone in the scientific community!

psychopy/experiment/components/sound/__init__.py
TEParsons: I tend to leave redundant `% self.params` in, it just means if other contributors (or me in the future) add a reference to a param in that string later then there's not a danger of `%(whatever)s` ending up in the compiled code. Learned the hard way after forgetting to put it back in enough times :')

TEParsons: Oh wait, my mistake, I see what you mean... Sorry, thought you meant adding `buff.writeIndentedLines(code % params)` when there's no `%(...)s` in the value of `code`. Yep, this redundancy is genuinely not needed!

psychopy/experiment/components/sound/__init__.py
TEParsons: Good point yeah, FINISHED is there because the Sound can finish playing without the Component finishing, but the same isn't true of starting so that line isn't needed.

I guess the concern is when people have Code components set up to call `.start()` and are using `.status == STARTED` to check whether it's still playing before their logic to stop it, as previously calling `.start()` would set `.status`, but in these instances I think we should be advising users to set `.status = STARTED` themselves, or query `.isPlaying` instead of `.status`.

psychopy/sound/backend_ptb.py
TEParsons: Good point, I was thinking we need to check in case it's finished but actually _checkPlaybackFinished doesn't affect that, it's just set to False on `.stop()`

psychopy/sound/backend_ptb.py
TEParsons: If nothing in _EOS is gated by `._isFinished`, what if we set `self._isFinished` before calling `_EOS()` rather than as an alternative to calling `_EOS()`? That way subsequent calls wouldn't duplicate.

So like this:
```
        # was the sound already finished?
        wasFinished = self._isFinished
        # is it finished now? (also sets _isFinished)
        isFinished = self._isFinished = not pa_status['Active'] and pa_status['State'] == 0
        # if it wasn't finished but now is, do end of stream behaviour
        if isFinished and not wasFinished:
            self._EOS()
```

psychopy/experiment/components/sound/__init__.py
mh105: @TEParsons I see your new implementation using `self.writeActiveTestCode(buff)`, but looking at this more carefully, I still struggle to see proper functioning of this activate status setting.

Let's look at one concrete example of Builder generated script:
```
            # if sound_1 is stopping this frame...
            if sound_1.status == STARTED:
                # is it time to stop? (based on global clock, using actual start)
                if tThisFlipGlobal > sound_1.tStartRefresh + 2-frameTolerance:
                    # keep track of stop time/frame for later
                    sound_1.tStop = t  # not accounting for scr refresh
                    sound_1.tStopRefresh = tThisFlipGlobal  # on global time
                    sound_1.frameNStop = frameN  # exact frame index
                    # add timestamp to datafile
                    thisExp.timestampOnFlip(win, 'sound_1.stopped')
                    # update status
                    sound_1.status = FINISHED
                    sound_1.stop()
            
            # if sound_1 is active this frame...
            if sound_1.status == STARTED:
                # update params
                pass
                # if sound is finished but sound_1 has time left, finish it now
                if sound_1.isFinished:
                    sound_1.status = FINISHED
```

If the sound finishes mid-frame, then the `if sound_1.isFinished:` will set `.status` to `FINISHED`. This means the actual stopping frame code on this next frame loop will get skipped. While users who are queuing `sound_1.status` will promptly know when the sound finished, but they have now lost access to updated attributes like `.tStop`, `.tStopRefresh`, and timestamp in data file on when the **sound component**. 

So my suggestion is not to use `writeActiveTestCode()`. Instead, we could overwrite `writeStopTestCode()` to add an `or` statement into the stopping frame code so it reads something like
```
            # if sound_1 is stopping this frame...
            if sound_1.status == STARTED:
                # is it time to stop? (based on global clock, using actual start)
                if tThisFlipGlobal > sound_1.tStartRefresh + 2-frameTolerance **or sound_1.isFinished**:
                    # keep track of stop time/frame for later
                    sound_1.tStop = t  # not accounting for scr refresh
                    sound_1.tStopRefresh = tThisFlipGlobal  # on global time
                    sound_1.frameNStop = frameN  # exact frame index
                    # add timestamp to datafile
                    thisExp.timestampOnFlip(win, 'sound_1.stopped')
                    # update status
                    sound_1.status = FINISHED
                    sound_1.stop()
```

I personally find this to be the most intuitive and elegant way of doing this. More generally, I think it could be useful to expand `writeStopTestCode()` to handle the use of additional stopping criteria by iterating over `self.params['stopType']` if it is a list, and then cascade multiple stopping criteria with `or`. Alternatively, we could add an internal `_extraStopType` parameter to `self.params`. Then `writeStopTestCode()` and `writeStopTestCodeJS()` just needs to be changed a little to accommodate this internal parameter that will be set by special components like a `SoundComponent`.

I'll make a PR to your branch to show you what I was thinking - can't promise it has handled all the details of `self.params` correctly...



This might be beyond the scope of a BF, so maybe overwriting `writeStopTestCode()` and `writeFrameCodeJS()` is the simplest.

psychopy/sound/backend_ptb.py
mh105: @TEParsons I guess one thing that is still unclear to me is how should `._isFinished` be set when there are **loops** specified? From reading `_EOS()`, I get the sense that `_isFinished` should only get turned to true when `._loopsFinished` exceeds `.loops`. This is a separate criteria from `not pa_status['Active'] and pa_status['State'] == 0` right?

So we essentially need to decide on does `not pa_status['Active'] and pa_status['State'] == 0` supersede the relevance of loops. Or asked in a different way, when the sound file is set to loop, would `not pa_status['Active'] and pa_status['State'] == 0` evaluate to `True` before all loops have finished?

**If the answer is yes:** right now whenever `not pa_status['Active'] and pa_status['State'] == 0` is `True`, we set `self._isFinished` as `True` without accumulating or checking `self._loopsFinished`. So attributes are not updated correctly.

**If the answer is no:** then there is no need to accumulate or keep track of `self._loopsFinished` at all. We just need to check `not pa_status['Active'] and pa_status['State'] == 0` to call `stop()` and set `self._isFinished` as `True`.


Regardless of what you decide how to handle this problem, right now we only enter `_EOS()` when `self._isFinished` is `True`. Then there is no longer the need to call `self._isFinished = True` during `_EOS()` right?

Idk if `_EOS()` itself is a legacy method and can be removed completely. But I have a hunch that without it, looped audio playing won't work properly, and the answer to my question 
> would `not pa_status['Active'] and pa_status['State'] == 0` evaluate to `True` before all loops have finished?

is yes, and we need to handle it carefully in `_checkPlaybackFinished()`.

mh105:
Added two more follow-up comments in our previous threads!

mh105 (changes requested):
Added a few more followup comments in our threads above!

psychopy/experiment/components/sound/__init__.py
mh105: Agreed. If they are manually calling `.start()` but using `.status` to check, they should set `.status` manually also, as demonstrated by Builder generated scripts.

psychopy/sound/backend_ptb.py
mh105: Interestingly, because of the change to `isFinished()`, we now have the chain `isFinished()` >> `_checkPlaybackFinished()` >> `_EOS()` >> `.stop()` >> `self._isPlaying = False`!

So we should actually now call `self._checkPlaybackFinished()` before returning `self._isPlaying` like how you originally had it 😅

One more thing that needs to go along with resurrecting this line, change the 
```
        if not self.isPlaying:
            return
```
in `def stop()` to
```
        if not self._isPlaying:
            return
```
Otherwise we will have circular calls that result in infinite loops.

psychopy/experiment/components/sound/__init__.py
TEParsons: I like the idea of `writeStart/Stop/ActiveTestCode` functions accepting an additional condition - I think that's an elegant way around special cases like Sound where we need to query more than just `.status`. I don't think it needs to be anything as complicated as an internal parameter or differing behaviour for `stopType` - as the Component knows whether or not it needs it, we can just have it as an optional argument to those functions, called by the Components which need it.

psychopy/sound/backend_ptb.py
TEParsons: I believe the answer is yes (@mdcutone please correct me if I'm wrong), and I think I follow: If we check finished and it looks like it is finished, we call `_EOS`, which sets `_isFinished` to True if all loops are finished, but if all loops *aren't* finished then `_isFinished` still remains `True`. So the solution should be as simple as adding `else: self._isFinished = False` to `_EOS`? I'll do that now and test it out

psychopy/sound/backend_ptb.py
TEParsons: I'm not sure it is infinite... You'd get `_checkPlaybackFinished` >> `_EOS` >> `stop` >> `isPlaying` >> `_checkPlaybackFinished`, but on that second `_checkPlaybackFinished` you wouldn't get `_EOS` as `wasFinished` would now be `True`. You also won't get `self._isPlaying = False` unless `isFinished` is `True` and `self._loopsFinished <= self.loops`, which seems reasonable as playing has genuinely stopped. So I think this is fine to leave as is, and certainly testing it out it seems to work.

TEParsons:
@mh105 I think I've fixed everything you raised so I'm going to pull in, then any bugs arising we can fix in separate PRs (just to stop this one getting monstrously large!)

psychopy/sound/backend_ptb.py
mh105: Perfect! That's it. Somehow I was overcomplicating it in my mind.

psychopy/sound/backend_ptb.py
mh105: I agree that `wasFinished` stops the infinite loop from happening. But we should call `self._checkPlaybackFinished()` in `isPlaying()` because it now has a possible route to toggle the `_isPlaying` attribute. Imagine I only call `isPlaying()` in a loop to check if a sound has finished playing. Without this call it can't possibly update `_isPlaying`. I'll make a separate PR.

PR #6565, opened 2024-06-13
EyeLogic integration

EyeLogicSolutions:
4af446f1a6b2547b9f07a6fc15ef48837f6af1be
Due to an issue, the origin of which I do not know, eyetracker iOHubDeviceView class names do not point to iohub.devices.eyetracker... but to the plugin packages. Prepending "psychopy.iohub.devices" to these will lead to a failed call to import_module.
I am currently in communication with @TEParsons on this issue. We may yet be resolve the issue in a different manner.

636edf612d158cdcec115ac03490fc5230ff1eac
Adds basic support. Tracker will be found and work (mostly): 
- calibration procedure settings tab contains a large number of unnecessary parameters
- the possible values of the only necessary parameter "Target Layout" are only partially applicable (FIVE & NINE)

59e903ff1b8226e22e0820cbb9cb6d34b8b5c755
- removes unncessary calibration parameters
- updates calibration process settings tab on change of eyetracker
- fixes the above issue concerning the applicability of "Target Layout"'s values

### DIFF
--- a/psychopy/app/builder/dialogs/__init__.py
+++ b/psychopy/app/builder/dialogs/__init__.py
@@ -41,6 +41,9 @@ white = wx.Colour(255, 255, 255, 255)
 codeSyntaxOkay = wx.Colour(220, 250, 220, 255)  # light green
 
 
+class EyeTrackerChoiceEvent(wx.CommandEvent):
+    pass
+
 class ParamCtrls():
 
     def __init__(self, dlg, label, param, parent, fieldName,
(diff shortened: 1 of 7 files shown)


TEParsons:
Closing this as we have a different solution for finding the eyetracker which doesn't require adding stubs to PsychoPy, and making the eyetracker calibration params show/hide according to device is a bigger question we can revisit after the upcoming release

PR #6571, opened 2024-06-15
DOC: make a note that writeRoutineEndCode does not save onset/offset times

mh105:
I noticed this while working on some other things for `SoundComponent`, but it looks like the `writeRoutineEndCode()` method of `BaseComponent` doesn't actually write code when generating scripts from Builder to save onset/offset times at the end of a routine. Is this expected...? It's just that almost all components have this parameter, and it was surprising there isn't corresponding code gets written for it. I apologize in advance if this is mistaken, and feel free to close it.

### DIFF
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -300,6 +300,8 @@ class BaseComponent:
             loop = currLoop.params['name']
             name = self.params['name']
 
+            # NOTE: this function does not write any code right now!
+
     def writeRoutineEndCodeJS(self, buff):
         """Write the code that will be called at the end of
         a routine (e.g. to save data)


mh105:
@TEParsons Sorry is it correct that this part doesn't write any code...? Just want to make sure toggling the option of Save Onset/Offset Time under the Data tab does have an effect as intended. Perhaps it is now handled using a different mechanism using `self.params['saveStartStop']` without adding a code section into the generated script? An example use is found [here](https://github.com/psychopy/psychopy/blob/1dab24f1959fa9e9e7ff57d868bdbe5cbe42be53/psychopy/experiment/components/sound/__init__.py#L299): I couldn't confirm whether that line has no effect whatsoever at the moment, so I just want to double check with you.

PR #6575, opened 2024-06-17
Merging release into dev branch

peircej:
@mh105 and @TEParsons I think you'll both need to check the resulting line.py to make sure it's OK after the merge conflict resolution around line 176 and line 195


mh105:
> @mh105 and @TEParsons I think you'll both need to check the resulting line.py to make sure it's OK after the merge conflict resolution around line 176 and line 195

Yeah as the comment I added to #6543, it looks like the fix @TEParsons merged into `dev` isn't drawing lines correctly yet. Maybe we could revert and stash those changes and test this out separately?

peircej:
In this merge the tests are failing because the vertices object is receiving a single vertex (shape=(2,) not (2,N)) and an assertion test expects at least 2 vertices:
https://github.com/psychopy/psychopy/blob/00ab1dd7d41e7711dea4295ba59b009955bcac89/psychopy/layout.py#L640

peircej:
I'll try reverting Todd's changes and redo the merge

mh105:
> In this merge the tests are failing because the vertices object is receiving a single vertex (shape=(2,) not (2,N)) and an assertion test expects at least 2 vertices:
> 
> https://github.com/psychopy/psychopy/blob/00ab1dd7d41e7711dea4295ba59b009955bcac89/psychopy/layout.py#L640

Right, this assertion is failing because of this change 945b652 to `writeInitCode()`. It might be safe to revert the 4 commits in #6543 and resolve it on a separate branch before merges.

PR #6576, opened 2024-06-17
SYS: Add GitHub action to check PRs to release can be merged into dev

TEParsons:
Meaning we can pre-emp merge conflicts with the dev branch before we choose to pull in something to release.

I expect when this fails we'll just pull in anyway then handle the conflict, but it means we know when one is about to crop up so can handle it while it's just a few lines rather than having to trawl through loads every time.

Worth noting: This doesn't actually *commit* the merge, it just does it locally and it's abandoned when the runner shuts down.


peircej:
I think this won't have any effect for a while (the action only occurs on release branch, which doesn't have the code in it because this PR is to dev branch) but that's fine

peircej:
I would actually run it on linux-latest, not macos as there are more linux runners and these are cheaper (on plans that pay)

Issue #5906, opened 2023-09-21
[Bug]: event.waitKeys and event.getKeys hanging on Ubuntu 22.04 LTS


mscheltienne:
### PsychoPy Version

2023.2.2 (main)

### What OS are your PsychoPy running on?

Linux-based systems

### Bug Description

`psychopy.event.waitKeys` and `psychopy.event.getKeys` are both hanging the interpreter, while `psychopy.hardware.keyboard.Keyboard.waitKeys` and `psychopy.hardware.keyboard.Keyboard.getKeys` work as intended. The issue was observed through the `EyeLinkCoreGraphicsPsychoPy.py` file provided by SR-research to control the EyeLink eye-tracker from a remote PC, and then through simple `psychopy` scripts (see below), on 2 computers with:
- POP_OS 22.04 LTS and Python 3.10
- Ubuntu 22.04 LTS and Python 3.10

PsychoPy is installed from the dev branch `pip install git+https://github.com/psychopy/psychopy` to include the monitor selection issue fixed in #5889 

### Reproduce

1. `event.waitKeys` is hanging:

```
from psychopy.event import waitKeys
from psychopy.visual import Window, TextStim


win = Window(size=(1920, 1080), fullscr=True, allowGUI=False, screen=1)
win.color = (0, 0, 0)
win.flip()
waitKeys(keyList=["return"])
text = TextStim(win)
text.draw()
win.flip()
```

2. `event.getKeys` is hanging:

```
from psychopy.event import getKeys
from psychopy.visual import Window, TextStim


win = Window(size=(1920, 1080), fullscr=True, allowGUI=False, screen=1)
win.color = (0, 0, 0)
win.flip()
while True:
    events = getKeys(modifiers=True)
    if len(events) != 1:
        continue
    keycode, modifier = events[0]
    if keycode == "escape":
        break
    elif keycode == "return":
        text = TextStim(win)
        text.draw()
        win.flip()
```

3. `psychopy.hardware.keyboard.Keyboard.waitKeys`  is working:

```
from psychopy.hardware.keyboard import Keyboard
from psychopy.visual import Window, TextStim


win = Window(size=(1920, 1080), fullscr=True, allowGUI=False, screen=1)
win.color = (0, 0, 0)
win.flip()
keyboard = Keyboard()
keyboard.start()
keyboard.waitKeys(keyList=["return"])
text = TextStim(win)
text.draw()
win.flip()
keyboard.stop()
```

[...]

PR #5889, opened 2023-09-08
Fix screen selection with pyglet on Ubuntu

mscheltienne:
Closes #5888
I'm not very familiar with pyglet/screen management. Let me know if you believe this change makes sense.

Issue #5888, opened 2023-09-08
[Bug]: Screen selection fails with pyglet on Ubuntu 22.04 LTS


mscheltienne:
### PsychoPy Version

2023.2.2

### What OS are your PsychoPy running on?

Linux-based systems

### Bug Description

Creating a window with `screen != 0` fails:

```
win = Window(
    size=_SCREEN_SIZE,
    color=_BACKGROUND_COLOR,
    units="norm",
    winType="pyglet",
    fullscr=True,
    allowGUI=False,
    monitor=monitor,
    screen=0,
)
```

The error raised is:

```
Traceback (most recent call last):
  File "/home/meguser/Projects/meg-sophie-schwartz/psychopy-faces.py", line 33, in <module>
    win = Window(
  File "/home/meguser/pyvenv/meg-sophie-schwartz/lib/python3.10/site-packages/psychopy/visual/window.py", line 469, in __init__
    self.backend = backends.getBackend(win=self, backendConf=backendConf)
  File "/home/meguser/pyvenv/meg-sophie-schwartz/lib/python3.10/site-packages/psychopy/visual/backends/__init__.py", line 61, in getBackend
    return Backend(win, *args, **kwargs)
  File "/home/meguser/pyvenv/meg-sophie-schwartz/lib/python3.10/site-packages/psychopy/visual/backends/pygletbackend.py", line 183, in __init__
    display = pyglet.canvas.Display(x_screen=win.screen)
  File "/home/meguser/pyvenv/meg-sophie-schwartz/lib/python3.10/site-packages/pyglet/canvas/xlib.py", line 127, in __init__
    raise NoSuchDisplayException(
pyglet.canvas.xlib.NoSuchDisplayException: Display "None" has no screen 1
Exception ignored in: <function Window.__del__ at 0x7ff44f9aa290>
Traceback (most recent call last):
  File "/home/meguser/pyvenv/meg-sophie-schwartz/lib/python3.10/site-packages/psychopy/visual/window.py", line 647, in __del__
    self.close()
  File "/home/meguser/pyvenv/meg-sophie-schwartz/lib/python3.10/site-packages/psychopy/visual/window.py", line 2536, in close
    self.backend.close()  # moved here, dereferencing the window prevents
AttributeError: 'NoneType' object has no attribute 'close'
```

### Expected Behaviour

With screen=1, the window should open in full screen on our second monitor.

### Additional context

[...]

PR #6578, opened 2024-06-17
Window & event fixes for Linux

aforren1:
- On linux, the `Display` is not the same as the `_default_display_`, meaning that events on the opened window were not being dispatched. This switches to pumping all the windows listed in `pyglet.app.windows`. Only tested on Ubuntu 20.04 so far.
 - GLFW events were not pumped in the event module, and so e.g. `waitKeys` would hang. This pumps GLFW at the same points as pyglet, and makes sure glfw is properly initialized before assuming it's available.
 - Unfortunately, https://github.com/psychopy/psychopy/pull/5889 broke the separate-X-screen case (e.g. as seen in some PTB setups). This hopefully handles both cases, though for my specific setup the pyglet shadow window also needed to be disabled (something about sharing GLX contexts across separate X screens?). Tested on Ubuntu 20.04, both with xorg.config enabled and disabled.
    - The `tesselate` import was moved to later in the process to allow the shadow window to be disabled.

Related issues:
https://github.com/psychopy/psychopy/issues/5407
https://github.com/psychopy/psychopy/issues/5906
https://github.com/psychopy/psychopy/issues/5888

### DIFF
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -52,15 +52,6 @@ if haveGLFW:
 else:
     useGLFW = False
 
-
-if havePyglet:
-    # get the default display
-    if pyglet.version < '1.4':
-        _default_display_ = pyglet.window.get_platform().get_default_display()
-    else:
-        _default_display_ = pyglet.canvas.get_display()
-
-
 import psychopy.core
 from psychopy.tools.monitorunittools import cm2pix, deg2pix, pix2cm, pix2deg
 from psychopy import logging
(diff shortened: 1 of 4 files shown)


peircej:
The tests are failing here and elsewhere (ie. not related to this PR) with message:
```
Fatal Python error: Segmentation fault

Current thread 0x0000000117392e00 (most recent call first):
  File "/Users/runner/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/psychtoolbox/audio.py", line 48 in get_version_info
  File "/Users/runner/work/psychopy/psychopy/psychopy/sound/backend_ptb.py", line 64 in <module>
  File "<frozen importlib._bootstrap>", line [24](https://github.com/psychopy/psychopy/actions/runs/9552892165/job/26330466388?pr=6578#step:11:25)1 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 883 in exec_module
collecting ...
```

Interestingly they fail the same way on Linux and MacOS but only on python3.10 (3.8 is fine)

@aforren1 I wonder if this is an issue with the PTB builds having some new incompatibility with py3.10 🤔

PR #6599, opened 2024-06-21
BF: Add default duration to eyetracker record component

RebeccaHirst:
Eyetracker Record component needs duration in order to stop - so adding a duration value by default so users do not get confused why the experiment halts at the stop record.

### DIFF
--- a/psychopy/experiment/components/eyetracker_record/__init__.py
+++ b/psychopy/experiment/components/eyetracker_record/__init__.py
@@ -26,7 +26,7 @@ class EyetrackerRecordComponent(BaseComponent):
     def __init__(self, exp, parentName, name='etRecord',
                  startType='time (s)', startVal=0.0,
                  stopType='duration (s)', stopVal=1.0,
-                 startEstim='', durationEstim='',
+                 startEstim='', durationEstim=1.0,
                  actionType="Start and Stop",
                  #legacy
                  save='final', configFile='myTracker.yaml'):


TEParsons:
I think this is also fixed by #6567 - which adds a "stop with Routine" parameter. But having a default stop time doesn't conflict so happy for both to happen (just trying to get tests passing on the other one)

mh105:
Sorry @RebeccaHirst @TEParsons: This PR is already merged so my comment might be too late. But I don't know if the change here actually has the intended effect. Isn't the default duration of the `EyetrackerRecordComponent` already set to be 1.0s by the `stopType='duration (s)', stopVal=1.0` arguments? I thought the `durationEstim` parameter is only used for visual display on the timeline as commented [here](https://github.com/psychopy/psychopy/blob/b41197a5b528f8cc8f1c7c8b571e2559c949ee82/psychopy/experiment/components/_base.py#L105). So does the experiment actually halt when `durationEstim` is not set? If so I'm worried that there's another bug somewhere that isn't due to `durationEstim`.

TEParsons:
Ah, I didn't realise - I thought you'd added a default for duration. Yes, duration estim is (or at least, should) only be used in the UI, the experiment shouldn't be affected

mh105:
Might make sense to revert so that users are not confused by why estimated duration defaults to 1s. But more importantly does the experiment actually halt when `EyetrackerRecordComponent` stops eyetracking recording? That's what I'm worrying about...

TEParsons:
The only thing I can think is maybe it's leading to the routine being marked as non-slip falsely? Will investigate on Monday

PR #6603, opened 2024-06-21
NF: Add "force nonslip" param to Routine Settings

TEParsons:


TEParsons:
@peircej could you check this over? I think this works as expected (from testing just now) but the concept of non-slip is kind of a headache at the best of times so would really appreciate a second pair of eyes on this

PR #6610, opened 2024-06-23
BF: enable correct behavior of entry-points to load eyetrackers

mh105:
I looked through the recent `dev` commits about using entry-points to manage eyetrackers as plugins to PsychoPy. There are some subtleties in `importDeviceModule()` and `getDevicePaths()` that weren't handled correctly, which prevented the desired uses of entry-points. These are all fixed now. At least EyeLink eyetrackers can be imported as plugins as long as the plugin repository (`psychopy-eyetracker-sr-research`) also has all the correct fixes. I've made a [PR](https://github.com/psychopy/psychopy-eyetracker-sr-research/pull/10) to `psychopy-eyetracker-sr-research` to document changes that may need to be mirrored to other device plugins.

----------------

There are a few points that I would like to raise here for future references, in case the development team has a different design principle in mind.

1) The motivation of using entry-points as plugins is usually to remove any explicit binding of plugin modules in the main application, and to resolve them dynamically by checking entry-points. This means that all the device-specific directories under [`psychopy/iohub/devices/eyetracker/hw`](https://github.com/psychopy/psychopy/tree/d38eab61d954511eba1c65a9e83f418d4cd4cd97/psychopy/iohub/devices/eyetracker/hw) except for `mouse` can be safely removed when things work properly. I would recommend to do that and **then test** how all the eyetrackers work with plugins installed. If there's a hidden dependency on the `psychopy/iohub/devices/eyetracker/hw` path somewhere in PsychoPy, this is an efficient way to catch them all.

2) Standalone apps based on python libraries have a peculiar behavior: when you first open the Builder App, it first looks into packaged executables in `/Applications/PsychoPy.app/Contents/Resources/lib/python38.zip`. This means that even though an updated user-specific package got installed to `~/.psychopy3/packages/lib/python/site-packages`, the `entry-point.txt` specification will point incorrectly into subdirectories of `python38.zip`! `os.walk()` doesn't work over zip files, and therefore some of the logics around getting config files will fail. This is a general problem with conflicting package namespaces in Python, so when packaging the python library for the next standalone release, all executables for plugins that can now be installed to `~/.psychopy3/packages/lib/python/site-packages` and managed with the GUI should be removed. This way there is no ambiguity about which plugin module is actually getting used. Interestingly, if you open a new Python console session in the Coder app, it does use the later namespace installed in the user specific package directory instead of from `python38.zip`, which only seems to get used when first starting the PsychoPy App. This matters because with the new entry-point logic, I couldn't get my `SR Research Ltd` option to show up in the experiment setting panel, and it took me a while to figure out why. I had to rezip `python38.zip` after deleting the version 0.0.2 `psychopy-eyetracker-sr-research` packaged inside in order for things to work.

3) One potential misconception of entry-points is that they are distinct from module trees specific to a process. This means that `setattr(targObj, attr, ep)` allows one to do `from targObj import attr` but it doesn't allow one to do `import targObj.attr` because the `ep` module is added with the namespace `attr` as an **attribute**, not **module**, to the existing entry point module `targObj`. So one shouldn't expect absolute or relative imports to work automatically. This issue shows up in how the added `importDeviceModule()` method was hoping to import plugin modules, which was failing.

4) Because ioHub manages devices through a greenlet server hosted on a separate subprocess, this complexity of entry-points and module namespace gets doubled, and it requires special attention. Even though [`psychopy.plugins` is adding entry point targets to existing class/module objects](https://github.com/psychopy/psychopy/blob/d38eab61d954511eba1c65a9e83f418d4cd4cd97/psychopy/plugins/__init__.py#L845C13-L845C39), this action has no effect whatsoever on the ioHub subprocess, which doesn't inherit these Python objects in the first place. So in order for things to work, we need to modify the `sys.modules` dictionary (again even already done on the main process) based on entry-points when first attempting to import devices from plugins on the `ioHub` subprocess, such that future imports can work intuitively like `import targObj.attr` or even `from targObj.attr.subpkg import xxx`.

5) I reduced the `os.walk()` from going through both the provided `iohub_device_path` and the entire user-level packages folder. This is because if a user installed multiple plugins with files called `supported_config_settings.yaml`, this may fail in confusing ways. The issue is that the for loop breaks at the first `supported_config_settings.yaml` found without checking whether it is the correct `supported_config_settings.yaml` for the requested device. It is probably ok when used with `device_name=""`, but when requesting configs for specific devices this could be very problematic. Since entry-points are working correctly now, the specified `iohub_device_path` should be accurate based on the entry-point `__file__`.

### DIFF
--- a/psychopy/iohub/util/__init__.py
+++ b/psychopy/iohub/util/__init__.py
@@ -233,8 +233,8 @@ def getDevicePaths(device_name=""):
 
         """
         yaml_paths = []
-        # try to walk both the internal iohub_device_path and user-level packages folder
-        for route in (os.walk(iohub_device_path), os.walk(prefs.paths['packages'])):
+        # search the internal iohub_device_path for device config files
+        for route in os.walk(iohub_device_path):
             for root, _, files in route:
                 # check each file in the route to see if it's a config yaml
                 device_folder = None
--- a/psychopy/iohub/util/__init__.py
+++ b/psychopy/iohub/util/__init__.py
@@ -18,6 +18,7 @@ import psychopy.logging as logging
 import psychopy.plugins as plugins
 from importlib.metadata import entry_points
 from pathlib import Path
+from psychopy.preferences import prefs
 
 ########################
 #
@@ -34,8 +35,6 @@ try:
     from collections.abc import Iterable
 except ImportError:
     from collections import Iterable
-
-from psychopy.preferences import prefs
 
 
 def saveConfig(config, dst_path):
@@ -143,7 +142,7 @@ def module_directory(local_function):
 def getSupportedConfigSettings(moduleName, deviceClassName=None):
     """Get the supported configuration settings for a device.
 
-    These are usually stored as YAML files within the module directory that 
+    These are usually stored as YAML files within the module directory that
     defines the device class.
 
     Parameters
@@ -152,10 +151,10 @@ def getSupportedConfigSettings(moduleName, deviceClassName=None):
         The name of the module to get the path for. Must be a package that defines
         `__init__.py`.
     deviceClassName : str, optional
-        The name of the specific device class to get the path for. If not provided, 
-        the default configuration file will be searched for in the module 
+        The name of the specific device class to get the path for. If not provided,
+        the default configuration file will be searched for in the module
         directory.
-    
+
     Returns
     -------
     str
@@ -176,13 +175,13 @@ def getSupportedConfigSettings(moduleName, deviceClassName=None):
             "Found ioHub device configuration file: {0}".format(yamlFile))
 
         return str(yamlFile)
-        
+
     # file name for yaml file name convention for single file
     yamlFile = yamlRoot / pathlib.Path('supported_config_settings.yaml')
     if not yamlFile.exists():  # nothing is found
         raise FileNotFoundError(
             "No config file found in module dir {0}".format(moduleName))
-    
+
     logging.debug(
         "Found ioHub device configuration file: {0}".format(yamlFile))
 
@@ -385,6 +384,7 @@ def getDeviceNames(device_name="eyetracker.hw", get_paths=True):
             names.append((d_config.get('manufacturer_name'), d_path))
     return names
 
+
 def getDeviceFile(device_name, file_name):
     """
     Returns the contents of file_name for the specified device. If file_name does not exist, None is returned.
@@ -396,6 +396,7 @@ def getDeviceFile(device_name, file_name):
     if device_name.endswith(".EyeTracker"):
         device_name = device_name[:-11]
     device_paths = getDevicePaths(device_name)
+
     device_sconfigs = []
     for dpath, _ in device_paths:
         device_sconfigs.append(readConfig(os.path.join(dpath, file_name)))
@@ -403,6 +404,7 @@ def getDeviceFile(device_name, file_name):
         # simplify return value when only one device was requested
         return list(device_sconfigs[0].values())[0]
     return device_sconfigs
+
 
 def getDeviceSupportedConfig(device_name):
     """
@@ -412,6 +414,7 @@ def getDeviceSupportedConfig(device_name):
     :return: dict
     """
     return getDeviceFile(device_name, 'supported_config_settings.yaml')
+
 
 if sys.platform == 'win32':
     import pythoncom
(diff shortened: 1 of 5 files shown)


mdcutone:
Hello, thanks for putting time towards fixing this issue. We can't pull these changes in since tests are failing with the following traceback:

```
psychopy/iohub/util/__init__.py:374: in getDeviceNames
    dconfigs = getDeviceDefaultConfig(device_name)
psychopy/iohub/util/__init__.py:327: in getDeviceDefaultConfig
    device_paths = getDevicePaths(device_name)
psychopy/iohub/util/__init__.py:272: in getDevicePaths
    for ep in entry_points()['psychopy.iohub.devices.eyetracker']:
../../../hostedtoolcache/Python/3.10.13/x64/lib/python3.10/importlib/metadata/__init__.py:427: in __getitem__
    return super().__getitem__(name)
E   KeyError: 'psychopy.iohub.devices.eyetracker'
```

mh105:
Thanks for flagging. Should be all good now.

psychopy/iohub/util/__init__.py
```diff
@@ -255,45 +263,40 @@ def _getDevicePaths(iohub_device_path):
...
-                )
-            else:
-                # otherwise, check its local folder
-                import inspect
-                deviceConfig = _getDevicePaths(
-                    os.path.dirname(inspect.getfile(cls))
+                    (ep_target.configFile.parent, ep_target.configFile.name)
                 )
+            else:  # otherwise, check the local folder of the entry point target
+                deviceConfig = _getDevicePaths(os.path.dirname(ep_target.__file__))
```
TEParsons, line 287: Will ep_target always have an attribute `__file__`? The reason I used `inspect.getfile` here was for when the target was a class (which didn't have `__file__`), is the assumption here that only classes will ever have an attribute `configFile`?

psychopy/iohub/util/__init__.py
```diff
@@ -255,45 +263,40 @@ def _getDevicePaths(iohub_device_path):
...
-            continue
-        # process each point
-        for ep in points:
-            # load class the entry point points to
+    # NOTE: The “selectable” entry points were introduced in importlib_metadata 3.6 and Python 3.10.
+    # Prior to those changes, entry_points accepted no parameters and always returned a dictionary
+    # of entry points, keyed by group. With importlib_metadata 5.0 and Python 3.12, entry_points
+    # always returns an EntryPoints object.
+
+    if 'eyetracker' in device_name.lower():
```
TEParsons, line 271: Is this because of something later enforces that the device name contain the word "eyetracker"?

mh105: This is because `device_name` got passed down all the way from the ioHub util function [getDeviceNames()](https://github.com/psychopy/psychopy/blob/8a228c2f14c134ee81d74e92e0a8ad72e50eeded/psychopy/iohub/util/__init__.py#L359). Previously PsychoPy setting component just calls `getDeviceNames()` without argument, therefore setting `device_name` to be `eyetracker.hw`. I've [updated this line](https://github.com/psychopy/psychopy/blob/8a228c2f14c134ee81d74e92e0a8ad72e50eeded/psychopy/experiment/components/settings/__init__.py#L46) to explicitly pass in the `device_name` argument to make it clear that this is used to load eyetrackers. This is how the mouse eyetracker implementation gets imported right now.

The reason for this if statement is because the `getDevicePaths()` function should be quite general, but we have this block of code that is hard-coded to look for entry-points related to eyetrackers. So I thought it's fair to check if `device_name` contains `eyetracker`. I checked everywhere where this ioHub function is used in PsychoPy, and this should produce a reasonable behavior.

My concern of not adding this `if 'eyetracker' in device_name.lower():` check is that if someone uses the `getDevicePaths()` function with a specific `device_name` in the future. For example, if EEG systems got incorporated into PsychoPy, then no matter what specific `device_name` I request, this function will also return all of the eyetracker config files, which feels like a buggy behavior to me. Let me know if you think otherwise.

If you would like `getDevicePaths()` without argument to return ALL device configs, more changes are needed for the built-in devices as well, which use `import_device()`. But for the eyetrackers using entry points, you can just change this line into
```
if 'eyetracker' in device_name.lower() or not device_name:
```

psychopy/iohub/util/__init__.py
mh105: Ah! This is why Jon had the question here: https://github.com/psychopy/psychopy/commit/fd5f50750ce1515da99b4721b4e7a30902684f98#commitcomment-143749931

I didn't realize that you are also using entry points for classes with the `module:class` syntax (in addition to modules). My bad! Using `inspect.getfile()` is indeed necessary. I'll open a PR to fix this right away. Thanks for flagging!

PR #6618, opened 2024-06-25
FF: suppress GLFW-related warnings during init in events.py

aforren1:
This should catch warnings seemingly related to M1+ Macs + older GLFW versions? I don't have access to a Mac, so haven't tested yet.

### DIFF
--- a/psychopy/event.py
+++ b/psychopy/event.py
@@ -32,9 +32,12 @@ try:
 except ImportError:
     havePyglet = False
 try:
+    import warnings
     import glfw
-    if not glfw.init():
-        raise ImportError
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore')
+        if not glfw.init():
+            raise ImportError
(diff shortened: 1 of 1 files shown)


mh105:
It does suppress the warning, although I usually don't like to suppress warnings. While we know the message is harmless for MacOS and older GLFW versions at the moment, I worry that a different warning message under a different setting could be crucial for debugging. With warnings filtered out, it may create unnecessary headache down the road for others. Maybe we just close this PR and let it warn for now, since Jon will upgrade `glfw` version that should fix this issue.

aforren1:
Yeah, I thought the warning might be shown again when someone tries to init a GLFW window, but it looks like it only hits the error callback function once (i.e. this change would swallow the one occurrence).

mh105:
Exactly, the py binding of GLFW3 seems to manually yield warnings so it'll be the only chance to see the error callback :) Thanks for your help still!!

PR #6619, opened 2024-06-25
BF: Builder: Ensure frame rate is stored.

aforren1:
The frame rate wasn't being saved into win._monitorFrameRate, and so expInfo['frameRate'] was also None.


mdcutone:
Not sure why the test are failing on this one.

aforren1:
https://github.com/psychopy/psychopy/pull/6581 only went to the dev branch?

peircej:
> https://github.com/psychopy/psychopy/pull/6581 only went to the dev branch?

No, that has also been merged back to dev branch 🤔 

Ah but @aforren1 your release branch is out of date with upstream, and your branch doesn't include that fix, so maybe do a pull from upstream and push back to your origin to update this PR

Incidentally, as we're now nearing a new major release you can send all PRs to dev branch. The next release will be built from there instead of release anyway

PR #6621, opened 2024-06-26
BF: movie component Volume inputType should be single instead of float

mh105:
`float` isn't a valid parameter `inputType`; `single` has been used for other components


TEParsons:
The actual changes look fine, I think this is just tests failing on the release branch in general currently. Could you cherry pick over and resubmit to the dev branch?

PR #6628, opened 2024-06-26
BF: Fixed microphone parameters string composition in boilerplate

mdcutone:
Fixes a missing comma in generated boilerplate for microphone settings


peircej:
I think there was also a comma needed, right? See https://github.com/psychopy/psychopy/pull/6598
As this is handled in dev branch, and the next release is from dev, we can probably close this one?

PR #6629, opened 2024-06-27
DOCS: update Japanese translation

hsogo:
Translation for upcoming 2024.2.0 (based on release branch, is that right?)


peircej:
@hsogo actually, the new release will be built off `dev` branch not `release` because of this being a new major release - we use `release` branch just for the minor updates

hsogo:
I see. So it would be better to have two branches, one is based on the dev branch to translate the upcoming version, and the other is based on the release branch to fix issues on the current version.

hsogo:
@peircej I created a new branch named "translation-japanese-dev" from the dev branch and started the translation.  I intended to use this branch only to update app/locale/ja_JP/messages.po, but I found some issues where I need to edit .py files to fix.  Should I commit fixes on .py files separately to the dev branch? Or, can I commit these fixes to "translation-japanese-dev" together with messages.po?
I think committing to separate branches makes it easier for translators of other languages to notice the fix, but It may be troublesome for you to merge (and for me, too).

peircej:
> I see. So it would be better to have two branches, one is based on the dev branch to translate the upcoming version, and the other is based on the release branch to fix issues on the current version.

Yes, in general that's the right strategy. Use dev branch for upcoming features and release branch for just bug-fix releases. 

Right now, because we're so close to a major release and there won't be another minor release occurring, you could just submit all changes to dev branch. That would save me having to merge your changes back to dev branch from release but no big deal

peircej:
Sorry, I now understand your comment about separate branches for the fixes versus the translation. It's fine for them to be on one branch assuming they are all needed. Do keep them in separate commits though of course :-)

hsogo:
I've committed fixes and updated translation to the dev branch.  Please see #6654

PR #6631, opened 2024-06-27
BF: Allow Sound to stop on a condition rather than just time

TEParsons:
This was originally limited because Sound had to have a fixed duration to load, but this hasn't been the case for a while now. Setting stop val as a condition opens up the possibility for users to have a sound which stops when `soundsOwnName.isFinished` - simulating the behaviour which recent changes by myself and @mh105 have removed (i.e. setting `.status = FINISHED` if `.isFinished`, which had a number of problems).

### DIFF
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -52,9 +52,6 @@ class SoundComponent(BaseDeviceComponent):
             "volume", "hammingWindow",  # Playback tab
         ]
         # params
-        self.params['stopType'].allowedVals = ['duration (s)']
-        self.params['stopType'].hint = _translate('The maximum duration of a'
-                                                  ' sound in seconds')
         hnt = _translate("When does the Component end? (blank to use the "
                          "duration of the media)")
         self.params['stopVal'].hint = hnt


mh105:
@TEParsons I just want to clarify on one thing to make sure I'm using the sound components correctly. After the change in 77027c5, a sound component will ALWAYS stop when `sound_1.isFinished is True` regardless of what users put into the stop parameter field, be it for `durations (s)` or `condition`, as long as its `stopVal` is not `None` (otherwise the `self.writeStopTestCode()` will not write the stopping section all together). Is my understanding correct?

So by
> Setting stop val as a condition opens up the possibility for users to have a sound which stops when `soundsOwnName.isFinished`,

if users select `condition`, they should just put `False` into the value field, and then the builder generated code will look like
```
if bool(False) or soundsOwnName.isFinished:
```
Is this what you are thinking? This may not be clear to users though.

Or should users explicitly provide a `condition` themselves? It seems to me that if you have now opened up the option to specify `condition` by the user, always adding `extra=" or %(name)s.isFinished"` under the hood may be surprising. Would it be better to ignore the `extra` parameter when users have provided a condition? This way someone can simply use `soundsOwnName.isFinished` in `condition` when they want to, and the generated builder code won't look like
```
if soundsOwnName.isFinished or soundsOwnName.isFinished:
```
which is technically still functional but not optimal. Let me know what you think!

TEParsons:
> Or should users explicitly provide a condition themselves?

This is what I'm thinking, yeah. As you spotted the boilerplate does look a bit silly (`soundsOwnName.isFinished or soundsOwnName.isFinished`) but this seemed like the least of several evils.

I wondered whether to remove the `extra` for conditional stops, as you say, but the problem then is if they're using a conditional stop that's something else (e.g. when a button is pressed or another Component finished) it would have different behaviour on playback completing than it would if it had a duration that's longer than the duration of the sound, which seems inconsistent. 

Maybe it's best to simply remove `extra=" or %(name)s.isFinished"` altogether? They can use a conditional stop if they need that behaviour, and if not then `status` always means whether the *Component* is finished.

mh105:
> Maybe it's best to simply remove `extra=" or %(name)s.isFinished"` altogether? They can use a conditional stop if they need that behaviour, and if not then `status` always means whether the _Component_ is finished.

I'm not sure about this... I personally love the idea, but my students had trouble appreciating the distinction between a `Component` versus the underlying `sound` or `movie` playback. So perhaps it's best to leave things as they are currently. This way users who think (albeit incorrectly) that `soundsOwnName.status` reflects the status of the sound file playback can still use it to achieve accurate timing behavior. I would hate to see more experiments done with subtle timing issues because users didn't read the source code to understand how PsychoPy behaves, which is probably the majority of users even in technically capable labs.

PR #6646, opened 2024-07-01
ENH: Make install behaviour consistent between plugins and packages

TEParsons:
Some of the recent improvements to `pkgtools` weren't accessible from Builder as Builder uses a different function to install plugins when the install button is clicked. This PR adds the necessary parameters to `pkgtools.installPackage` so that Builder can use that function instead (without losing the ability to do an asynchronous install)


peircej:
Are these actually bug fixes or ENHancements? It's a lot of code. Is it necessary?

TEParsons:
I guess we could figure out exactly what the difference is and modify `EnvironmentManagerDlg.installPackage` to match `pkgtools.installPackage` instead, which would be fewer lines changed, but the latter is definitely the correct behaviour so it made more sense to just use it rather than trying to recreate it.

There's also some smaller UI improvement stuff here which I'll cherry pick to a different PR so we can focus on the main change here.

PR #6648, opened 2024-07-01
FF: entrypoint may not be a file

peircej:
...in which case it won't have a `__file__` attribute

error added in fd5f50750ce1515da99b4721b4e7a30902684f98

### DIFF
--- a/psychopy/plugins/__init__.py
+++ b/psychopy/plugins/__init__.py
@@ -830,7 +830,7 @@ def loadPlugin(plugin):
 
                 return False
 
-            if '.zip' in ep.__file__:
+            if hasattr(ep, 'file') and '.zip' in ep.__file__:
                 logging.warning(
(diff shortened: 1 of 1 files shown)


psychopy/plugins/__init__.py
```diff
@@ -830,7 +830,7 @@ def loadPlugin(plugin):
 
                 return False
 
-            if '.zip' in ep.__file__:
+            if hasattr(ep, 'file') and '.zip' in ep.__file__:
```
TEParsons, line 833: Shouldn't it be `hasattr(ep, '__file__')`?

peircej: yes, it should 🤦

peircej: but at least in my case it meant the crash didn't occur 😉

PR #6657, opened 2024-07-02
BF: Allow polygon outline opacity to be persistent across frames

mdcutone:
The PR allow the user to set `opacity` once for the border colour of a shape, instead of each frame.


peircej:
@TEParsons pulled this by mistake. Matthew and I believe it's correct, but the test images need updating and I meant to hold off until those were done

PR #6663, opened 2024-07-03
ENH: Skip routine setup when 'Skip if' condition is set

rhancockn:
This changes the behavior of the Builder 'Skip if' option to also skip component initialization. This is useful for skipping routines under conditions where the content of the component might be undefined, and avoids setup work for stimuli that will never be shown.

### DIFF
--- a/psychopy/app/locale/ar_001/LC_MESSAGE/messages.po
+++ b/psychopy/app/locale/ar_001/LC_MESSAGE/messages.po
@@ -6552,7 +6552,7 @@ msgstr ""
 
 #: psychopy/experiment/components/routineSettings/__init__.py:76
 msgid ""
-"Skip this Routine if the value in this contorl evaluates to True. Leave blank to "
+"Skip this Routine if the value in this control evaluates to True. Leave blank to "
 "not skip."
 msgstr ""
 
--- a/psychopy/app/locale/de_DE/LC_MESSAGE/messages.po
+++ b/psychopy/app/locale/de_DE/LC_MESSAGE/messages.po
@@ -6694,7 +6694,7 @@ msgstr ""
 
 #: psychopy/experiment/components/routineSettings/__init__.py:76
 msgid ""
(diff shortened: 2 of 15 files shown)


TEParsons:
Given this a pretty thorough try and it all seems good! Was worried about skipped trials not showing up in the datafile as skipped but because you use the scheduler in JS and dedent just before `nextEntry` in Py the data looks perfectly fine. Nice work!

The question is whether to pull this in now or wait until after the upcoming release, as it is an enhancement rather than a fix. @rhancockn is there anything which is currently broken by running the initialisation steps of skipped routines?

rhancockn:
Thanks @TEParsons! In my use case, the current skip implementation is broken ;) One case where running the initialization is a problem is if I loop through a series of heterogenous trials. For example, a set of instructions where some instructions are text based and others show an image. Rather than defining a routine for each instruction step, I can use a list of trials like

| routine_name | stim_file                   | stim_text                       |
| ------------ | --------------------------- | ------------------------------- |
| instrText    |                             | Add your task instructions here |
| instrImage   | examples/hand_placement.png |

with skip logic `routine_name != 'instrImage'`. Since the current skip implementation always initializes the image stim, this approach fails (or a least requires a dummy image to be specified and loaded).

peircej:
Hold on this one just for now. I think there's some thinking to be done. 

The idea of skip-if was more about handling run-time dynamic decisions to skip a Routine, so I'm wondering about the flip side, where this change could break that dynamic version

For instance: 
- skip if the participant got the trial correct. Here we can't evaluate the skip-if condition during init
- skip if the experiment has moved into a particular phase or mode within the study. Like skipping feedback after first 10 trials. We might be able to evaluate it during init but we need to run the init whatever it says

So I think this is an interesting proposal/use-case that I hadn't been thinking of but I think it needs more discussion before we pull in and should wait until the 2025.1.0 release (given how close we are to 2024.2.0)

PR #6665, opened 2024-07-04
DOCS: update portuguese po file

joana-quarenta:


joana-quarenta:
DOCS: update portuguese po file2. This version is more complete than the previous one.

PR #6668, opened 2024-07-04
BF: enable mouse clickable from variable online

RebeccaHirst:
needed to add eval() so that a value from a spreadsheet (which is read in as a string) is instead converted to an object/variable name that can be clicked.

### DIFF
--- a/psychopy/experiment/components/mouse/__init__.py
+++ b/psychopy/experiment/components/mouse/__init__.py
@@ -220,7 +220,7 @@ class MouseComponent(BaseComponent):
         code = (
             "// check if the mouse was inside our 'clickable' objects\n"
             "gotValidClick = false;\n"
-            "%(name)s.clickableObjects = %(clickable)s\n;"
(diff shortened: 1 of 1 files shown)


peircej:
@RebeccaHirst might this ever have the reverse problem? i.e. are there times when we _don't_ want to evaluate (e.g. if it's _not_ a variable?)

RebeccaHirst:
@peircej the clickable field of a mouse is always an object (e.g. polygon or image) I can't think of a situation where it makes sense for a string to be clickable, it would always throw an error.

TEParsons:
It would fail an eval if it were already an object handle though, so essentially whenever it *isn't* set via a spreadsheet. This could work though if you put it inside an `if` which checks whether it's a string before evaluating.

TEParsons:
Okay so, something I didn't know about JS, you can `eval` a list, object handle, etc. with no problem! It just returns as is. So yes, this is fine actually :)

PR #6675, opened 2024-07-08
FF: App starting view wasn't being respected

TEParsons:
Previously, the app figured out which windows to open based on command line arguments (`-c` for coder, `-b` for builder, `-r` for runner), but now that we're doing a two-step process (to facilitate in-app restarting) these are obscured by the interpreter option (which is always `-c`) so Builder was never opening.

Solution is to parse command line arguments during the first step, and pass the parsed options as init arguments to the PsychoPyApp class.

### DIFF
--- a/psychopy/app/__init__.py
+++ b/psychopy/app/__init__.py
@@ -95,7 +95,7 @@ def startApp(showSplash=True, testMode=False, safeMode=False, startView=None):
     # console.
     from psychopy.app._psychopyApp import PsychoPyApp
     _psychopyAppInstance = PsychoPyApp(
-        0, testMode=testMode, showSplash=showSplash, safeMode=safeMode)
+        0, testMode=testMode, showSplash=showSplash, safeMode=safeMode, startView=startView)
 
     # After the app is loaded, we hand off logging to the stream dispatcher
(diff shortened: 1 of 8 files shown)


peircej:
Hopefully this can preserve the original default behaviour from prefs of "reopen whatever I had before"?

TEParsons:
I don't think we ever had behaviour to open the same frames as before, it was just opening the same files as before (which coincidentally meant opening the same frames, most of the time). From trying it out though I think this was lost in the current dev, so I'll try and figure out what stopped it from working.

TEParsons:
Ah okay I see what's happening - the previous files are added to the file list still, but because of Builder not opening it was never used. Next commit will fix.

peircej:
Looks like you removed the "last" option and replaced it with "all" 4 years ago
https://github.com/psychopy/psychopy/commit/e504d486300fc32fd4471c81b31cc6ec25bcb441

I'd say that "last" is what most users would expect/like if it's technically still possible

TEParsons:
Ahhh, classic 4-years-ago Todd over-refactoring :') I'm pretty sure we can add "last" back in

PR #6684, opened 2024-07-08
BF: fixes undefined variable

EyeLogicSolutions:
fixes undefined variable error when running from a github repo


peircej:
Thanks - it was actually one we already handled. Apologies ;-)

PR #6686, opened 2024-07-08
BF: Re-add distro as a runtime dependency for linux & remove deprecated function.

aforren1:
I noticed a "ModuleNotFoundError: No module named 'distro'" in the app log, and it looks like distro was not specified as a runtime dependency when switching to pdm?

It looks like distro.linux_distribution() has been deprecated since ~2021, so replaced those calls with something functionally similar. The main difference is that the original function returned a prettier codename (e.g. "Focal Fossa" instead of "focal"), but as far as I can tell that shouldn't have practical impacts?


peircej:
Thanks. Sounds good. Tests are failing for a different reason that I'm working on 😁

PR #6702, opened 2024-07-11
FF: Sound with no stop time wasn't updating its status as stop code was skipped altogether

TEParsons:
This looks like a lot of changes, but the bulk are just replacing `self.params` with `params` so that changing values during writeStart/Stop/ActiveTestCode methods doesn't change the actual param. 

The important changes are these bits:
```
if params['stopVal'].val in ('', None, -1, 'None'):
            if extra:
                # if we have extra and no stop condition, extra is the only stop condition
                params['stopType'].val = "condition"
                params['stopVal'].val = "False"
            else:
                # if we just have no stop time, don't write stop code
                return buff.indentLevel - startIndent
```

Essentially, if we have no stop time but we do have an extra condition (e.g. `or %(name)s.isFinished` for Sound) then we should still write the start/stop code, it should just be a conditional `False` (so that e.g. Sound stops if `False or %(name)s.isFinished`)

### DIFF
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -398,7 +398,6 @@ class BaseComponent:
         buff.writeIndentedLines(code % params)
         buff.setIndentLevel(+1, relative=True)
 
-        params = params
         code = (f"# keep track of start time/frame for later\n"
                 f"{params['name']}.frameNStart = frameN  # exact frame index\n"
                 f"{params['name']}.tStart = t  # local t and not account for scr refresh\n"
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -339,7 +339,7 @@ class BaseComponent:
             string.
         """
         # create copy of params dict so we can change stuff without harm
-        params = self.params.copy()
+        params = {key: param.copy() for key, param in self.params.items()}
 
         # Get starting indent level
         startIndent = buff.indentLevel
@@ -451,7 +451,7 @@ class BaseComponent:
             string.
         """
         # create copy of params dict so we can change stuff without harm
-        params = self.params.copy()
+        params = {key: param.copy() for key, param in self.params.items()}
 
         # Get starting indent level
         startIndent = buff.indentLevel
@@ -527,7 +527,7 @@ class BaseComponent:
             string.
         """
         # create copy of params dict so we can change stuff without harm
-        params = self.params.copy()
+        params = {key: param.copy() for key, param in self.params.items()}
 
         # Get starting indent level
         startIndent = buff.indentLevel
@@ -642,7 +642,7 @@ class BaseComponent:
             string.
         """
         # create copy of params dict so we can change stuff without harm
-        params = self.params.copy()
+        params = {key: param.copy() for key, param in self.params.items()}
 
         # Get starting indent level
         startIndent = buff.indentLevel
@@ -734,7 +734,7 @@ class BaseComponent:
             string.
         """
         # create copy of params dict so we can change stuff without harm
-        params = self.params.copy()
+        params = {key: param.copy() for key, param in self.params.items()}
 
         # Newline
         buff.writeIndentedLines("\n")
--- a/psychopy/experiment/components/_base.py
+++ b/psychopy/experiment/components/_base.py
@@ -338,22 +338,28 @@ class BaseComponent:
             `%(key)s` syntax to insert the values of any necessary params. Default is an empty 
             string.
         """
+        # create copy of params dict so we can change stuff without harm
+        params = self.params.copy()
 
         # Get starting indent level
         startIndent = buff.indentLevel
 
-        if self.params['startVal'].val in ('', None, -1, 'None'):
-            # If we have no start time, don't write start code
-            return buff.indentLevel - startIndent
+        if params['startVal'].val in ('', None, -1, 'None'):
+            if extra:
+                # if we have extra and no stop condition, extra is the only stop condition
+                params['startType'].val = "condition"
+                params['startVal'].val = "False"
+            else:
+                # if we just have no stop time, don't write stop code
+                return buff.indentLevel - startIndent
 
         # Newline
         buff.writeIndentedLines("\n")
 
-        if self.params['syncScreenRefresh']:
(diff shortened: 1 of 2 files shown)


TEParsons:
A note on speed: Deep copying all params in each method led to a (admittedly small) increase in time taken to write scripts, so I've set it up to just shallow copy the params dict and then deep copy the two params needed at the point they're changed.

TEParsons:
I can't add you as reviewer but @mh105 could you check this over and make sure it fits with what we've discussed with Sound Component? This means that a Sound with infinite duration will get `.status = FINISHED` once the media finishes playing, and it'll get a stop time and etc. (as the code to do that is no longer skipped when the Component has blank duration)

psychopy/experiment/components/_base.py
```diff
@@ -338,50 +338,58 @@ def writeStartTestCode(self, buff, extra=""):
...
         # Get starting indent level
         startIndent = buff.indentLevel
 
-        if self.params['startVal'].val in ('', None, -1, 'None'):
-            # If we have no start time, don't write start code
-            return buff.indentLevel - startIndent
+        if params['startVal'].val in ('', None, -1, 'None'):
+            if extra:
+                # if we have extra and no stop condition, extra is the only stop condition
+                params['startType'] = params['startType'].copy()
```
mh105, line 350: Is this extra `copy()` necessary? Isn't `params` already a copy that we can change without harm, so should we simply do:
```
params['startType'] = "condition"
params['startVal'] = "False"
```
here?

psychopy/experiment/components/_base.py
```diff
@@ -338,50 +338,58 @@ def writeStartTestCode(self, buff, extra=""):
...
+        if params['startType'].val == 'time (s)':
             # if startVal is an empty string then set to be 0.0
             if (
-                isinstance(self.params['startVal'].val, str) 
-                and not self.params['startVal'].val.strip()
+                isinstance(params['startVal'].val, str) 
+                and not params['startVal'].val.strip()
             ):
-                self.params['startVal'].val = '0.0'
+                params['startVal'].val = '0.0'
```
mh105, line 378: I'm a little concerned about this line. Originally this is updating the `self.params['startVal']` value, so if anything downstream expects a non-empty `self.params['startVal']`, it would be fine.

Now that we are not modifying in place, could this cause something to fail if something like `if self.params['startVal'] == '0.0'` gets used? I don't know all the places where component objects are used in PsychoPy, so I can't tell for sure. Maybe it's worth generating a few builder scripts after this change to ensure no unexpected bug gets caused by the modification to this line.

psychopy/experiment/components/_base.py
```diff
@@ -390,7 +398,7 @@ def writeStartTestCode(self, buff, extra=""):
         buff.writeIndentedLines(code % params)
         buff.setIndentLevel(+1, relative=True)
 
-        params = self.params
+        params = params
```
mh105: This line can now be deleted.

psychopy/experiment/components/_base.py
```diff
@@ -444,29 +452,38 @@ def writeStartTestCodeJS(self, buff, extra=""):
...
         # Get starting indent level
         startIndent = buff.indentLevel
 
-        if self.params['startVal'].val in ('', None, -1, 'None'):
-            # If we have no start time, don't write start code
-            return buff.indentLevel - startIndent
+        if params['startVal'].val in ('', None, -1, 'None'):
+            if extra:
+                # if we have extra and no stop condition, extra is the only stop condition
+                params['startType'] = params['startType'].copy()
```
mh105, line 464: Same as before

psychopy/experiment/components/_base.py
```diff
@@ -513,18 +530,26 @@ def writeStopTestCode(self, buff, extra=""):
...
         # Get starting indent level
         startIndent = buff.indentLevel
 
-        if self.params['stopVal'].val in ('', None, -1, 'None'):
-            # If we have no stop time, don't write stop code
-            return buff.indentLevel - startIndent
+        if params['stopVal'].val in ('', None, -1, 'None'):
+            if extra:
+                # if we have extra and no stop condition, extra is the only stop condition
+                params['stopType'] = params['stopType'].copy()
```
mh105, line 542: Same as before.

psychopy/experiment/components/_base.py
```diff
@@ -622,15 +647,24 @@ def writeStopTestCodeJS(self, buff, extra=""):
...
         # Get starting indent level
         startIndent = buff.indentLevel
 
-        if self.params['stopVal'].val in ('', None, -1, 'None'):
-            # If we have no stop time, don't write stop code
-            return buff.indentLevel - startIndent
+        if params['stopVal'].val in ('', None, -1, 'None'):
+            if extra:
+                # if we have extra and no stop time, extra is only stop condition
+                params['stopType'] = params['stopType'].copy()
```
mh105, line 659: Same as before

mh105 (changes requested):

psychopy/experiment/components/_base.py
TEParsons: `params` is a shallow copy of `self.params`, so the `Param` objects contained within `params` are the same objects in `self.params`. What you suggest would work in terms of not affecting `self.params`, but there's other points in here where `params['startVal'].val` are referenced, so if we did `params['startVal'] = "False"` then these would fail as the Param object is replaced by a string.

psychopy/experiment/components/_base.py
TEParsons: Good catch, I see what you mean... I think the original behaviour is actually wrong here - otherwise we'd have methods which only execute successfully after the code has been compiled once, which would definitely not be desirable. For that reason I think we've got checks downstream to handle empty startVal. I'll add some tests to the test suite which create Components with blank start/stop fields to make sure this is the case.

psychopy/experiment/components/_base.py
mh105: Ah you are right! I forgot PsychoPy uses a dedicated `Params` class object to store parameters rather than simple key-value pairs in a dictionary. Got it - this is what you meant by deep copying. Thanks a lot for explaining this! Looks good to me!

PR #6713, opened 2024-07-15
FF: Use updated port to connect to iohub

mdcutone:
Specify the correct iohub UDP port which was changed recently


mdcutone:
Confirmed that this was the cause of iohub hanging. The subprocess was not closed between successive runs of the application causing the lockup we are experiencing

PR #6718, opened 2024-07-15
BF: fix trial data writing when there are multiple key presses including None

mh105:
When there are multiple key presses in a trial, data from all trials are stored as a list of lists. However, for numerical data that underwent replacing `'None'` with `''`, this creates invalid syntaxes when converting `strVersion` back to a tuple using `eval()`.

This is because `eval("[1, , 2]")`, etc. is not a valid syntax. Current PR fixes the issue to allow an empty cell to be written into the output file for trials containing data in the form of lists of lists.

### DIFF
--- a/psychopy/data/trial.py
+++ b/psychopy/data/trial.py
@@ -423,7 +423,9 @@ class TrialHandler(_BaseTrialHandler):
                 if hasattr(tmpData, 'tolist'):  # is a numpy array
                     strVersion = str(tmpData.tolist())
                     # for numeric data replace None with a blank cell
+                    replaceNone = False
                     if tmpData.dtype.kind not in 'SaUV':
+                        replaceNone = True
                         strVersion = strVersion.replace('None', '')
                 elif tmpData in [None, 'None']:
                     strVersion = ''
@@ -442,11 +444,18 @@ class TrialHandler(_BaseTrialHandler):
                 if (len(strVersion) and
                             strVersion[0] in '[(' and
                             strVersion[-1] in '])'):
+                    if replaceNone:
+                        # Add None back so that the str is valid for eval
+                        strVersion = strVersion.replace('[,', '[None,')
+                        strVersion = strVersion.replace(', ,', ', None,')
                     tup = eval(strVersion)  # convert back to a tuple
                     for entry in tup:
                         # contents of each entry is a list or tuple so keep in
                         # quotes to avoid probs with delim
-                        thisLine.append(str(entry))
+                        currentEntry = str(entry)
+                        if replaceNone:
+                            currentEntry = currentEntry.replace('None', '')
+                        thisLine.append(currentEntry)
                 else:
                     thisLine.extend(strVersion.split(','))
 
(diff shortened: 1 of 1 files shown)


psychopy/data/trial.py
```diff
@@ -420,10 +420,12 @@ def _createOutputArray(self, stimOut, dataOut, delim=None,
             for thisDataOut in dataOut:
                 # make a string version of the data and then format it
                 tmpData = dataAnal[thisDataOut][stimN]
+                replaceNone = False
                 if hasattr(tmpData, 'tolist'):  # is a numpy array
                     strVersion = str(tmpData.tolist())
                     # for numeric data replace None with a blank cell
-                    if tmpData.dtype.kind not in ['SaUV']:
+                    if tmpData.dtype.kind not in 'SaUV':
```
TEParsons, line 427: What's the motivation for removing the []? Is it possible for `.dtype.kind` to be just part of the string `"SaUV"`, and if so do we want to catch it here?

TEParsons (approved):
I think this is fine - if I follow: You're keeping None present up to the point of the `eval` call, then removing it after so that behaviour after is unchanged from current? 

I'd also advise adding a test to `psychopy.tests.test_data.test_TrialHandler:TestTrialHandler`  to check this - we actually have one called `test_multiKeyResponses` that's been disabled for about 10 years so it would be good to have something more up to date :')

psychopy/data/trial.py
mh105: Yeah I don't think `'SaUV'` itself is a valid `dtype.kind`. I think it has always meant to check whether `.dtype.kind` is part of the string, and it was a typo to include the `[]`. Right now even for string data, `'None'` gets replaced by `''` because `tmpData.dtype.kind` is never the same as `'SaUV'`.

mh105:
> I think this is fine - if I follow: You're keeping None present up to the point of the `eval` call, then removing it after so that behaviour after is unchanged from current?

Yeah that's right. I'm adding the `'None'` back into `strVersion` so that `eval()` does not crash the script due to invalid syntax. And then after `eval()` I remove them again if they were removed previously. This way the output file behavior is the same as how it is right now. So this BF really just fixes a bug without changing any behavior.

> I'd also advise adding a test to `psychopy.tests.test_data.test_TrialHandler:TestTrialHandler` to check this - we actually have one called `test_multiKeyResponses` that's been disabled for about 10 years so it would be good to have something more up to date :')

I agree, although I still don't know well enough why this issue came up. The strange thing is that I ran into this problem by spamming a key when a response_key component is active (and it's set to store all keys). Sometimes, but not always, when I press too quickly, one of the durations becomes `None`. I still couldn't quite figure out why it would be `None`...

psychopy/data/trial.py
TEParsons: Ah okay, I don't know much about dtypes tbh so I'll take your word on it. To be clearest we should probably do `("S", "a", "U", "V")`, but this is a minor point, the string is fine :P

mh105: Might be worth checking with Jon or whoever wrote it in the first place? I don't think `"a"` is one of the valid character codes anymore: https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html

Issue #6353, opened 2024-04-07
[Bug]: Extremely slow rendering of colors


fredcallaway:
### PsychoPy Version

2024.1.0

### What OS are your PsychoPy running on?

macOS Intel

### Bug Description

My task is an interactive game with about 100 elements on the screen (see image below). I'm using a gaze contingent display, so I call flip on every frame. I'm getting about 12 frames per second on average, which makes animations look terrible, produces noticeable lag between input and response, and leads to frequent missed click events.

I did some profiling and almost all the run time is spent in `Color.__deepcopy__`. Maybe I'm naive, but it seems like rendering a red circle shouldn't take so much CPU time.

Is there anything to be done here? Or should I assume that psychopy isn't meant to handle complex graphics and switch to pure pygame?

I appreciate any help anyone can provide here!

<img width="779" alt="image" src="https://github.com/psychopy/psychopy/assets/7883156/365ad9d5-1c62-4128-9a6d-ee9381c0d076">

### Expected Behaviour

Frame rates of at least 30Hz, ideally 60.

### Steps to Reproduce

I can try to produce a minimal example if there's actually interest in working on this.

fredcallaway:
Here's a quick fix that's working for me. In `Color.render` I check if `self.contrast` is 1 and bypass the copy step if so.

```python
# Transform contrast to match rgb
contrast = self.contrast
if contrast == 1:
    return getattr(self, space)
    
contrast = np.reshape(contrast, (-1, 1))
contrast = np.hstack((contrast, contrast, contrast))
...
```

TEParsons:
@cmacmillan noticed the same thing and I've spoken to them previously about solutions, there's some discussion in this PR: https://github.com/psychopy/psychopy/pull/5790

TEParsons:
Fixed by #6720 (I think)

aforren1:
I don't think it's fixed for all cases, e.g. profiling

```
from psychopy import visual
from timeit import default_timer
from statistics import mean, median

w = visual.Window([400, 400], checkTiming=False)
txt = visual.TextBox2(w, text='Text', units='norm', size=(2,2), alignment='center', letterHeight=0.1)

times = []
for i in range(100):
    t0 = default_timer()
    txt.draw()
    times.append(default_timer() - t0)
    w.flip()

w.close()

print(f'median: {median(times)}, mean: {mean(times)}') # ~3ms with profiling turned off
```
[...]

PR #6720, opened 2024-07-16
BF: Save to the color cache when rendering.

aforren1:
The _renderCache wasn't being used, and so the color was recomputed every call regardless of previous state. This improves some cases related to https://github.com/psychopy/psychopy/issues/6353, but not a total panacea. E.g. 

```python
from psychopy import visual
from timeit import default_timer
from statistics import mean, median

w = visual.Window([400, 400], checkTiming=False)
rct = []
for i in range(100):
    rct.append(visual.Rect(w, width=0.1, height=0.1, fillColor='red'))

times = []
for i in range(100):
    t0 = default_timer()
    for r in rct:
        r.draw()
    times.append(default_timer() - t0)
    w.flip()

w.close()

print(f'median: {median(times)}, mean: {mean(times)}')
```

Improves overall CPU time from ~17ms to ~4ms.

### DIFF
--- a/psychopy/colors.py
+++ b/psychopy/colors.py
@@ -389,7 +389,8 @@ class Color:
         adj = np.clip(self.rgb * contrast, -1, 1)
         buffer = self.copy()
         buffer.rgb = adj
-        return getattr(buffer, space)
+        self._renderCache[space] = getattr(buffer, space)
+        return self._renderCache[space]
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


TEParsons (approved):
Good catch! Render cache is cleared when a new value is set so this seems all good. There's also plenty of tests in the test suite which check that changing a color value updates visually so I'm also fairly comfortable that any knock on effects from this would be caught.

PR #6724, opened 2024-07-17
BF: maxDurationReached in Routine Settings False should be false online

wakecarter:


TEParsons:
Good catch, thanks Wake. I might look into doing some basic syntax checking on compiled JS - I'm sure AST and similar can do stuff like that.

PR #6726, opened 2024-07-18
FF: Revert launcher back to original code

mdcutone:
This PR reverts the launcher code back to a state we know things worked fine. I feel there are too many issues cropping up after moving to running the app in a subprocess, causing undesirable effects which have no certain cause. Eventually we'll attempt this again. We don't need this change to make it into the next release since the application auto-restart feature is unused, which was the original motivation of this change.


peircej:
OK, will give it a go

PR #6737, opened 2024-07-22
SPANISH TRANSLATION FIXING REGIONAL IDIOMATIC VARIATIONS

carosamu:
Added new terms to resolve idiomatic regional variations . Corrections on previous commits.


carosamu:
PLEASE DISCARD LAST ATTEMPTS AS MISTAKENLY CHANGED en_US instead of es_US. This commit FIXES files

PR #6743, opened 2024-07-24
DOC: Document Linux better and tee output

larsoner:
I just tried installing PsychoPy on Linux using `pip install psychopy` on Ubuntu 24.04 and it didn't work, so I figured I'd update the install docs a bit to mention:

1. How to get an older, PsychoPy-compatible Python (3.10) on your system
2. Provide some minimal breadcrumbs about how to get wxPython installed on newer Ubuntu (the wheels they provide only go up to 22.04)
3. Only suggest `pip install psychopy` *after* the wxPython step. Locally for me wxPython would not build properly as part of `pip install wxpython` (separately or as a PsychoPy dependency), which is probably a bug at their end. But the suggestion I have in there now to `pip download` then `python setup.py bdist_wheel` *did* work. But in any case, regardless of whether wheels are available for your platform or you have to build them manually, you probably want to do it before installing psychopy.
4. Improve the command-line output of running `psychopy`.

For (4 this involved creating a little `Tee` class that both writes to a file and to `sys.__stdout__`. This was inspired by the fact that during debugging when I was missing some stuff, I would get:
```
$ psychopy  # immediate exit with no output, but:
$ echo $?
1
```
With the `_Tee` in place, now I get something more informative because the output is printed in addition to going to the `last_app_load.log`:
```
$ psychopy
  File "/home/larsoner/python/virtualenvs/psychopy/bin/psychopy", line 8, in <module>
    sys.exit(main())
  File "/home/larsoner/python/psychopy/psychopy/app/psychopyApp.py", line 105, in main
    start_app()
  File "/home/larsoner/python/psychopy/psychopy/app/psychopyApp.py", line 27, in start_app
    _ = startApp(showSplash=showSplash)  # main loop
  File "/home/larsoner/python/psychopy/psychopy/app/__init__.py", line 117, in startApp
    from psychopy.app._psychopyApp import PsychoPyApp
  File "/home/larsoner/python/psychopy/psychopy/app/_psychopyApp.py", line 10, in <module>
    from psychopy.app.colorpicker import PsychoColorPicker
  File "/home/larsoner/python/psychopy/psychopy/app/colorpicker/__init__.py", line 9, in <module>
    import wx
ModuleNotFoundError: No module named 'wx'
```
When I check, this output does indeed mirror `$ cat ~/.psychopy3/last_app_load.log`.

### DIFF
--- a/docs/source/download.rst
+++ b/docs/source/download.rst
@@ -104,20 +104,26 @@ On a MacOS machine, `brew` can be used to install |PsychoPy|::
 Linux
 ~~~~~~~~~~~~~~~~~
 
-We are aware that the procedure for installing on Linux is often rather painful. 
+We are aware that the procedure for installing on Linux is often rather painful.
 This is not the platform that the core PsychoPy developers currently use so support
 is less good than on some platforms. Feel free to jump in and help improve it as a
-contributor! :-) 
+contributor! :-)
 
 There used to be neurodebian and Gentoo packages for |PsychoPy| but these are both
-badly outdated. We'd recommend you do:
+badly outdated. We'd recommend you first make sure you have a compatible Python
+version installed (currently ``>=3.8, <3.11``). If you need an older version, you
+can on Ubuntu for example do:
 
 .. code-block:: bash
 
-    # with --no-deps flag if you want to install dependencies manually
-    pip install psychopy
+    sudo add-apt-repository ppa:deadsnakes/ppa
+    sudo apt update
+    sudo apt install python3.10-venv python3.10-dev
+    python3.10 -m venv path/to/new/psychopyenv  # choose a path of interest!
+    source path/to/new/psychopyenv/bin/activate
 
-**Then fetch a wxPython wheel** for your platform from:
+Once you have a compatible Python activated, **copy the link to a wxPython wheel** for
+your platform from:
 
 https://extras.wxpython.org/wxPython4/extras/linux/gtk3/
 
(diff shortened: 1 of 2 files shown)


peircej:
Thanks @larsoner 
Tee certainly looks useful (I think @mdcutone was actually adding something similar but maybe that got lost in a branch 😬 )

Regarding the install on linux, handling wxpython install, we've actually been developing an install script:
  https://github.com/psychopy/psychopy/blob/dev/installPsychoPy.py
and the idea then would be to provide 
- tab-based install options so that different OSs can see the other options
- use text for the linux install that will basically look like below

image

peircej:
The idea would be that, to install on linux you would run something like this:

``` bash
python3.10 -m venv ~/pp_python310  # create a virtual environment
source ~/pp_python310/bin/activate  # activate that virtual environment
python -c "$(curl -fsSL https://raw.githubusercontent.com/psychopy/psychopy/dev/installPsychoPy.py)"
```

peircej:
BUT, that install script is currently set to look for a wx wheel and then tells you to select a different version if it can't find one. It needs updating to handle building if a wheel doesn't exist

peircej:
I'm going to go ahead and merge this but keep an eye to adapting it with the new install script

larsoner:
> I'm going to go ahead and merge this but keep an eye to adapting it with the new install script

Nice, thanks!

I did see there was a script and thought about using it, would be nice to update it at some point :+1:

PR #6744, opened 2024-07-24
ENH: Python 3.12 and NumPy 2.0 compat

larsoner:
I looked briefly for open issues or PRs about NumPy 2.0 and Python 3.12 compat and didn't see any. So I started digging and opened https://github.com/aforren1/ptb-wheels/pull/10, https://github.com/Psychtoolbox-3/Psychtoolbox-3/pull/843, https://github.com/aforren1/Psychtoolbox-3/pull/1 which should fix Psychtoolbox and make 3.12 + NumPy 2.0 compatible psychtoolbox wheels available :crossed_fingers: . Locally I can at least get `psychtoolbox` to import on my 3.12 + NumPy 2.0 system with those changes.

This PR contains some additional stuff I needed to get `$ psychopy` to open when using the wheel above.

- I wasn't sure why `setuptools` was needed (and an upper bound pin) in the runtime dependencies so I removed it, but happy to put it back.
- I also bumped the Pyglet version but happy to revert that, too (I assume it was there for a reason but wanted to bring up the idea in case it wasn't -- 1.5.29 came out pretty recently so maybe just 1.5.28 was problematic or something?).

So some opinionated choices in here, happy to modify as needed or close if someone is already working on this and I missed it!


psychopy/plugins/__init__.py
```diff
@@ -60,40 +60,6 @@
...
-        for thisGroup, eps in importlib.metadata.entry_points().items():
-            # get entry points within matching group
-            if thisGroup.startswith(group):
-                # add to list of all entry points
-                entryPoints += eps
-    else:
-        # otherwise, just get the requested group
-        entryPoints += importlib.metadata.entry_points().get(group, [])
-
-    return entryPoints
```
larsoner, line 94: There was a lot of code dup here with the `psychopy.plugins.util.getEntryPoints` so I got rid of this version. I replaced it above in the one place it was used hoping it's equivalent.

mh105:
@larsoner Does this wheel of PTB work on osx-arm architect? That will be fantastic!

larsoner:
No that will be a separate paid feature sounds like, see https://github.com/Psychtoolbox-3/Psychtoolbox-3/pull/813#issuecomment-2249010600

mh105:
Hmm... that's annoying but very understandable.

peircej:
I reached out to Mario to discuss funding the work back in May and haven't heard back but maybe the email got lost in the cracks. I'll try again

peircej:
Much later(!) update on this draft/work:
- I believe the new psychtoolbox wheels fine on py3.12 (and don't appear to apply a charge and are ARM-compatible). That said, we also now have sub-ms audio, including the preemptive scheduling and zero-lag, working through `sounddevice` so we're hoping psychtoolbox won't even be needed in the future 
- PsychoPy's test suite passes on py3.12
- @HugoFara has made a number of further checks and tweaks for 3.12 and numpy 2 support

@mdcutone @TEParsons there were also some changes in this PR about the entrypoint code and redundancy reduction. Could you check those have been ported over where useful and I think we can close this long-standing PR

PR #6747, opened 2024-07-25
BF: TrialHandler2 was starting `thisRepsN` at 1 rather than 0

TEParsons:
Also added a test to make sure the indexing style remains the same for TrialHandler2 as it was for TrialHandler (at least for as long as there are two classes)


mh105:
> FAILED psychopy/tests/test_data/test_TrialHandler2.py::TestTrialHandler2::test_skipTrials_rewindTrials - AssertionError: Was expecting current trial to match all fields {'thisN': 4, 'thisRepN': 2, 'thisTrialN': 1, 'thisIndex': 1}, instead was {'type': 'trial_data', 'thisN': 4, 'thisRepN': 1, 'thisTrialN': 1, 'thisIndex': 1, 'data': {'foo': 2, 'bar': 3, 'thisN': 4, 'thisTrialN': 1, 'thisRepN': 1}} (different thisRepN)

Seems to be failing tests on `thisRepN` at the moment. Is it because of this line?https://github.com/psychopy/psychopy/blob/cecbd49236553a8e0dab197e81ba9bdbf5636780/psychopy/data/trial.py#L1686

TEParsons:
I think this is just that the tests were configured to expect `nReps` to start at 1, which makes sense as when we were adding TrialHandler2 to Builder we were checking the values using the tests... So just needs a few values changed in the exemplars

mh105:
Ah I see. You meant the answer keys need to get updated. Thanks so much @TEParsons for the prompt BFs on this!

PR #6748, opened 2024-07-29
BF: allow app always to run on M1/M2 mac (but without hdf5 data output)

peircej:
On arm64 we currently need to run the Intel python binaries (so that we can use psychtoolbox audio)

BUT If we use the Intel binaries on an M1/M2 mac it seems to find local arm64 binaries for hdf5/pytables. 
It doesn't even safely fail when trying to load these - it crashes violently swallowing all stderr outputs - so we can't even use try...import tables...except ImportError.

That prevents the app loading on most arm64 macs (most macs nowadays!) for the sake of hdf5 file outputs. 

We need to work out how to do one of the following (in order of difficulty):
- detect, in advance, that the dylib about to load is wrong arch
- work out a way to point to our packaged dylibs, which are correctly packaged as intel binaries
- get psychopy packaged for arm64 architecture with arm64 tables so that it (hopefully) won't crash

FOR NOW, I'm simply disabling the loading of tables on all arm64 macs


mh105:
@peircej I just tried on three arm64 Macs (M1, M2, M3) in the lab and couldn't reproduce this issue. Arm64 mac natively does not have binaries for hdf5 - one would have to install brew and then brew install hdf5 in order to have it appear in the sys path (the absolute path is `/opt/homebrew/opt/hdf5`) for the issue you described to show up.

In the 2024.2.0 standalone PsychoPy, the `/Applications/PsychoPy.app/Contents/Resources/lib/python3.8/tables` package has a hidden folder `.dylibs`, which copied all the needed c libraries when `tables` was pip installed. Unless there's some environment variable overriding things, the PsychoPy app python instance shouldn't be loading hdf5 binaries elsewhere other than what the `tables` holds. One potential issue is `h5py`, see below.

**I wonder instead of disabling ioHub hdf5 functionality all together for arm64 macs, could you try clearing any related environment variable (`HDF5_DIR` for example) when loading the app, and see if the app opens up correctly on your end?** 

One other question is how was the HDF5 library installed on the machine building the standalone PsychoPy across OS platforms? If it's installed through `h5py`, it might not resolve correctly by `setuptools` when building `tables`. But then again, I'm able to open up the app with no issue. If `h5py` is not used by anything else in PsychoPy (from a project-wide search it only appears in ioHub demo .rst files), I would recommend removing it from dependencies and the standalone package. I don't like the way the PyPI `h5py` copies the HDF5 .dylib, which may be messing up other packages using HDF5 (such as `tables`).

Lastly, another potential culprit is that the hdf5 binary shipped with standalone PsychoPy is version `1.12.2` for `tables` version `3.8.0`. This coincides with the info of `h5py.version.info` as built against `hdf5` version `1.12.2`. You might want to upgrade the HDF5 library from 1.12.2 to 1.14.2 or higher and see if that works more reliably across platforms.

peircej:
For the record, I think the reason it worked fine for you was that you were on the py3.8 standalone. The fail only occurs on the 3.10 binaries. That turned out to be a useful clue - the failure to load was caused by tables==3.9.2 whereas the py3.8 standalone uses tables==3.8.0. I'm now pinning tables==3.8.0 everywhere

PR #6749, opened 2024-07-29
Docs:Arabic ranslation

Ahmad-Abudoush:


peircej:
@Ahmad-Abudoush there's a syntax error so I'm having to revert

PR #6752, opened 2024-07-29
Revert "Docs:Arabic ranslation"

peircej:
Reverts psychopy/psychopy#6749

The po file doesn't compile due to syntax error


peircej:
OK, I've found the issue that meant the previous version was broken. There had been a merge conflict and that got committed with the conflict in place. We'll need to go through and see which is the correct one for each conflict :-(

PR #6754, opened 2024-07-30
BF: Fixed `--target` and `--user` argument conflict when installing plugins

mdcutone:
Fixes an issue where the argument `--target` and `--user` were both being specified in when calling `pip`, causing an error.

### DIFF
--- a/psychopy/tools/pkgtools.py
+++ b/psychopy/tools/pkgtools.py
@@ -240,27 +240,24 @@ def installPackage(
             and error from the subprocess.
         If `awaited=False`:
             Returns the job (thread) which is running the install.
-    """
-    if target is None:
-        target = prefs.paths['userPackages']
-    
+    """   
     # convert extra to dict
     if extra is None:
         extra = {}
-    # check the directory exists before installing
-    if not os.path.exists(target):
-        raise NotADirectoryError(
-            'Cannot install package "{}" to "{}", directory does not '
-            'exist.'.format(package, target))
+
 
     # construct the pip command and execute as a subprocess
     cmd = [sys.executable, "-m", "pip", "install", package]
 
     # optional args
     if target is None:  # default to user packages dir
-        cmd.append('--prefix')
-        cmd.append(prefs.paths['packages'])
+        cmd.append('--user')
     else:
+        # check the directory exists before installing
+        if not os.path.exists(target):
+            raise NotADirectoryError(
+                'Cannot install package "{}" to "{}", directory does not '
+                'exist.'.format(package, target))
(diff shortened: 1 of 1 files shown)


psychopy/tools/pkgtools.py
```diff
@@ -240,27 +240,24 @@ def installPackage(
...
+
 
     # construct the pip command and execute as a subprocess
     cmd = [sys.executable, "-m", "pip", "install", package]
 
     # optional args
     if target is None:  # default to user packages dir
-        cmd.append('--prefix')
-        cmd.append(prefs.paths['packages'])
+        cmd.append('--user')
```
mh105, line 254: I'm not sure this is correct. Doesn't this bypass the whole if statement checking if in virtual environment? This will also result in duplicated `--user` flags if not within virtual env.

psychopy/tools/pkgtools.py
```diff
@@ -240,27 +240,24 @@ def installPackage(
             and error from the subprocess.
         If `awaited=False`:
             Returns the job (thread) which is running the install.
-    """
-    if target is None:
-        target = prefs.paths['userPackages']
```
mh105, line 245: I think it makes sense to first check `prefs.paths['userPackages']` trying to set `target`. Should this actually get removed?

psychopy/tools/pkgtools.py
```diff
@@ -240,27 +240,24 @@ def installPackage(
...
-            'Cannot install package "{}" to "{}", directory does not '
-            'exist.'.format(package, target))
+
 
     # construct the pip command and execute as a subprocess
     cmd = [sys.executable, "-m", "pip", "install", package]
 
     # optional args
     if target is None:  # default to user packages dir
-        cmd.append('--prefix')
```
mh105, line 261: This then tries to use default user packages dir as `--prefix`, which is a different behavior from using `--user`. 
Should this actually get removed?

I think the correct order of checking should use another if. It's easier for me to make another PR to show what I am thinking.

mdcutone: We configure the package directory as the user site since we can just pass that down as an environment variable to the subprocess.

mh105: Are you referring to the [`[PYTHONUSERBASE]`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUSERBASE) environment variable? It looks like the pip documentation recommends using `--user` over `--prefix`. However, the `--user` scheme has an annoying directory structure on macOS framework python that does not distinguish among Python versions. I have opened #6769 for this.

PR #6755, opened 2024-07-30
FF: fix the order of install destination args for pip install

mh105:
As pointed out by #6754, the use of optional flags for `pip install` isn't cleanly organized at the moment after [#6742](https://github.com/psychopy/psychopy/pull/6742). This PR can replace #6754, as it includes what the other PR fixed and more. Current FF commit also adds more in-line comments to document the optional arg usage.

### DIFF
--- a/psychopy/tools/pkgtools.py
+++ b/psychopy/tools/pkgtools.py
@@ -249,31 +249,27 @@ def installPackage(
     cmd = [sys.executable, "-m", "pip", "install", package]
 
     # optional args
-    # priorty of install args: --target > --prefix > --user
+    # priorty of install args: --target > --user
     if target is None:  # default to user site-packages dir
         target = prefs.paths['userPackages']
-    if target is None:  # if still None, use user packages dir
-        if prefs.paths['packages']:
-            cmd.append('--prefix')
-            cmd.append(prefs.paths['packages'])
-        else:  # if still None, use OS platform user install directory
-            # check if we are in a virtual environment, if so, dont use --user
-            if hasattr(sys, 'real_prefix') or (
-                    hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix):
-                # we are in a venv
-                logging.warning(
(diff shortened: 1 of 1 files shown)


peircej:
I'm not sure about this versus #6754 
This PR preserves the use of `--prefix` but I'm not sure if that's needed ever? What does `--prefix` achieve that `--user` doesn't?

My impression was that we only needed to handle 2 options: 
- progr files install (use `--user` flag with that set to be `~/.psychopy3/packages`
- venv install ( no user/prefix flags at all)

@mdcutone has been through most iterations of implementing different versions of these flags and knows more by now about which ones worked/failed for different reasons

mh105:
Ok! I might have gotten confused when reading #6742 because it looked like it's trying to use `--target` and `--prefix` in favor of `--user`, which I thought is being viewed as the last resort. I must have misunderstood the commits; sorry about that!

I don't think `--prefix` does anything more than `--user`, except the latter requires the environment variable [`PYTHONUSERBASE`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUSERBASE) to work correctly. Looks like the standalone app is setting this variable correctly already, so it doesn't really offer any new feature over `--user` other than being more explicit (and therefore pythonic).

peircej:
I've chatted with @mdcutone and he agrees. We like your reorginisation of the code and, critically, bringing back the check for venv that was accidentally lost in #6742 but I think we just need to switch so that `--prefix` is always `--user`

mh105:
Ok let me know if the last commit achieves what you and Matt wanted. I personally prefer `--prefix` slightly more over `--user` for being more explicit and less reliant on environment variables. But since `--user` seems to work fine I have no objection at all.

mdcutone:
@mh105 Closing this because these changes are superseded by recent work on `pkgtools`

mh105:
> @mh105 Closing this because these changes are superseded by recent work on `pkgtools` 

Thanks @mdcutone ! Would you mind dropping a link to the recent changes for reference here?

PR #6766, opened 2024-08-02
ENH: Added "force end Routine" option for Sound Component

TEParsons:
Required a rewrite of the "stop" logic on the JS end to imitate isPlaying/isFinished until they're implemented in actual PsychoJS, as we need to know when the sound itself has finished if we're to end the Routine on the basis of it, but I've tested a few use cases out and it seems consistent with before.

### DIFF
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -304,7 +304,7 @@ class SoundComponent(BaseDeviceComponent):
                 )
             else:
                 code += (
-                    "%(name)s.play();\n"
+                    "%(name)s.stop();\n"
                 )
             buff.writeIndentedLines(code % self.params)
             # end Routine if asked
--- a/psychopy/experiment/components/sound/__init__.py
+++ b/psychopy/experiment/components/sound/__init__.py
@@ -290,15 +290,8 @@ class SoundComponent(BaseDeviceComponent):
             # stop playback
             code = (
                 "// stop playback\n"
+                "%(name)s.stop();\n"
             )
-            if self.params['syncScreenRefresh'].val:
-                code += (
-                    "psychoJS.window.callOnFlip(function(){ %(name)s.stop(); });\n"
-                )
-            else:
-                code += (
-                    "%(name)s.stop();\n"
-                )
             buff.writeIndentedLines(code % self.params)
             # end Routine if asked
             if self.params['forceEndRoutine']:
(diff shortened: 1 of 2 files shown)


TEParsons:
@mh105 would appreciate if you could give this a go and sanity check, having recently worked on the Sound Component :)

psychopy/experiment/components/sound/__init__.py
```diff
@@ -77,6 +78,12 @@ def __init__(self,
             updates='constant',
             hint=msg,
             label=_translate("Sync start with screen"))
+        # --- Playback params ---
+        self.order += [
```
mh105, line 80: `self.order` isn't created concisely with all the correct parameters. We should fix it. See my PR to your branch.

psychopy/experiment/components/sound/__init__.py
```diff
@@ -250,38 +291,34 @@ def writeFrameCodeJS(self, buff):
...
+        # are we stopping this frame?
+        indented = self.writeStopTestCodeJS(buff, extra=" || %(name)s.isFinished")
+        if indented:
+            # stop playback
+            code = (
+                "// stop playback\n"
+            )
+            if self.params['syncScreenRefresh'].val:
+                code += (
+                    "psychoJS.window.callOnFlip(function(){ %(name)s.stop(); });\n"
```
mh105: I don't think the end of the sound should be controlled by `syncScreenRefresh`, which is specifically for whether the start of the sound should sync to frame. When a sound is ending, it should end when it's supposed to end. See my PR to your branch.

mh105 (changes requested):
Looks great! I'm glad this `Force End Routine` option is added for Sound Component.

I have two minor suggested changes. To make it easier to see what I meant and to accept the changes, I've made a [PR](https://github.com/TEParsons/psychopy/pull/8) to your [`TEParsons:dev-enh-sound-force-end`](https://github.com/TEParsons/psychopy/tree/dev-enh-sound-force-end) branch. If you think they are acceptable, you can just merge and then merge this current PR.

psychopy/experiment/components/sound/__init__.py
TEParsons: Generally I prefer the approach of doing `+=` for each section, as it just makes it less of a monolith to maintain. e.g. if I add one param to Playback, I only have to think of it in relation to the params on the same page as it, rather than a block of all params at the top. This is just stylistic preference though, works the same either way :)

PR #6769, opened 2024-08-04
ENH: set user path for different python versions on OSX framework

mh105:
macOS framework Python uses a scheme for organizing `--user` path directories different from `posix` and `nt` schemes. We need to create a python-specific user packages directory to avoid conflicts when both py3.8 and py3.10 are installed on the same machine.

### DIFF
--- a/psychopy/preferences/preferences.py
+++ b/psychopy/preferences/preferences.py
@@ -175,7 +175,7 @@ class Preferences:
                 if err.errno != errno.EEXIST:
                     raise
 
-        # root site-packages directory for user-installed packages and add it
+        # site-packages root directory for user-installed packages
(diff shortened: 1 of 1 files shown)


mdcutone:
It seems like the problem caused by multiple Python versions on MacOS is hitting us already since we're now releasing 3.8 and 3.10 versions. We'll likely need to make this change at some point soon.

mh105:
@mdcutone Yeah agreed. I started to encounter this problem when we release 2024.2.0 in both 3.8 and 3.10 on macOS. This PR provides the change, so we can just merge it. I've been using it on my private branch, and it has been working fine on macOS for the past two months.

PR #6770, opened 2024-08-04
BF: append user packages correctly to PYTHONPATH

mh105:
When adding paths to the `PYTHONPATH` environment variable, the order of if statements matters. This is because `prefs.paths['userPackages']` is a subdirectory of `prefs.paths['packages']`, so adding the former will result in an incorrect check of using the `not in` operator.
```
_userPackages = str(prefs.paths['packages'])
if _userPackages not in env['PYTHONPATH']:
```
We need to add the parent path first to avoid this issue.


mdcutone:
I'm going to cherry pick this into release.

PR #6776, opened 2024-08-06
DOC: fill docstrings for visual.slider.Slider

skjerns:
I figured that lots of parameters of `Slider` were undocumented, so here's my fix for that

Edit: WIP, I figured there is even more outdated docstring in there


TEParsons:
The changes thus far look good, thanks! I'll pull in - if you make more changes just open another PR for them. Better to have lots of small PRs than one big one, it's easier to navigate later :)

PR #6800, opened 2024-08-19
ENH: Log all messages sent/received by Liaison

TEParsons:
Adds a subclass of _Logger which specifically handles messages sent/received by Liaison, indicating the direction of travel, and links it to the Session log file.

### DIFF
--- a/psychopy/liaison.py
+++ b/psychopy/liaison.py
@@ -13,13 +13,13 @@ zeroRPC
 
 
 import inspect
-import logging
 import asyncio
 import signal
 import json
 import sys
 import traceback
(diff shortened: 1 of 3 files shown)


apitiot:
That looks good to me.
Instead of `levelName`, I would possibly use `source` and `target`, which would be more explicit and more generic, but it's certainly not needed just yet, strictly speaking.

TEParsons:
I agree in general, I just went with labelname because it's the name of the attribute already in `LogEntry` - as that's what's ultimately being set I thought I'd keep the same name all the way up

Issue #6790, opened 2024-08-13
Documentation for psychopy.gui.Dlg outdated as of 2024.1.0, unannounced breaking change


jfkominsky:
See https://discourse.psychopy.org/t/in-new-versions-dialog-box-functionality-has-been-broken/41036/2

In short, gui.Dlg has completely changed in a way that breaks most code that used it prior to 2024.1.0 by operating in dictionaries instead of lists. For all intents and purposes this deprecates and replaces gui.Dlg for coder experiments.

The ongoing issue is that the example in documentation is now just plain wrong and will not work, and needs to be updated: https://psychopy.org/api/gui.html#psychopy.gui.Dlg

I see the logic behind the change and I see why it's better in the long run, but it's a big breaking change for any existing code, in that any pre-2024 code that uses gui.Dlg is now inoperable and any post-2024 code will probably not be able to run in any pre-2024 version of PsychoPy. It's already done and it's straightforward (though a little bothersome) to update all my code, but it would have been nice to have some warning and maybe a little more consideration of the potential consequences of this change.

peircej:
Hi there!

Sorry that the docs weren't updated. 

We do try to warn people of major compatibility changes but I'm afraid this one was overlooked, mostly because not many people are hand-writing code and even fewer make use of the basic gui.Dlg (most would use DlgFromDict if hand-writing code). We do also, for sure, try to avoid changes to the API but there are times when these changes are necessary for new features. In this case we needed more flexible inputs to the dialog fields so that we could separate the displayed text from the input key (we have a new situation where there are tags on the field names that need tracking but shouldn't be displayed to the participant) and that necessitated this change.

peircej:
To be fair, these twice-yearly major updates include a very large number of changes.[ 891 files were updated, with contributions from 28 authors](https://github.com/psychopy/psychopy/compare/2023.2.3...2024.1.0) in the 2024.1.0 release. So some things are likely to introduce breaking changes beyond our control

jfkominsky:
I totally get it! The main reason not to use DlgFromDict is that it doesn't support static text fields (i.e., addText) and is more awkward for conditional field creation in a dialog box. I just happen to be in a weird position where I'm building on PsychoPy as a library for my own tool (PyHab) and happen to need to use a lot of these darn things. I think in the long run the change to a fully dict-based structure is definitely better, the only problem is what it does to existing experiment code.

[...]

Issue #6781, opened 2024-08-09
[Bug]: gui.Dlg no longer returns same type of object, and no longer shows question text


alexholcombe:
### PsychoPy Version

2024.2.1

### What OS are your PsychoPy running on?

macOS Silicon

### Bug Description

See https://discourse.psychopy.org/t/in-new-versions-dialog-box-functionality-has-been-broken/41036

### Expected Behaviour

See screenshots at forum message https://discourse.psychopy.org/t/in-new-versions-dialog-box-functionality-has-been-broken/41036

### Steps to Reproduce

See code at forum message https://discourse.psychopy.org/t/in-new-versions-dialog-box-functionality-has-been-broken/41036

PR #6806, opened 2024-08-23
Backward compatibility for Dlg

jfkominsky:
If old gui.Dlg syntax is used for addField (i.e., no label provided), then the label is created from the key argument, and the dictionary key is just the list index of the field name.

This means that it supports the old behavior for both creating fields and accessing data from the Dlg (i.e., you can refer to it by list index in the order the fields were added, since the index is now just a numeric key for the relevant dict entry).

Addresses #6790 and #6781

### DIFF
--- a/psychopy/gui/qtgui.py
+++ b/psychopy/gui/qtgui.py
@@ -240,9 +240,11 @@ class Dlg(QtWidgets.QDialog):
 
         Returns a handle to the field (but not to the label).
         """
-        # if not given a label, use key (sans-pipe syntax)
+        # if not given a label, mimics old Dlg behavior by creating a numeric key corresponding to order added, and
+        # uses whatever string was entered as key (sans-pipe syntax)
         if label is None:
             label, _ = util.parsePipeSyntax(key)
+            key = len(self.inputFieldNames) # current length before new entry added
 
         self.inputFieldNames.append(label)
         if choices:


TEParsons (changes requested):
I see what you're going for, and it's a neat solution, but I worry that this would create unintuitive behaviour. If you did:
```
dlg = Dlg()
dlg.addField(
    key="someKey",
    initial="abcdefg"
)
data = dlg.show()
```
you'd expect to be able to access the value of `someKey` by `data['someKey']`, but with this change you'd instead have to do `data[0]`. While you could get around it by adding a value for `label`, it then means that setting the label affects the outputted data, which goes against the purpose of having a label in the first place.

We could handle it by adding the numeric index as well as the string as a key, but that has its own issues too (what happens when they do `data['someKey'] = ...`, would they then expect to see the change in `data[0]`? What if they're checking the length of `data` and find it's now longer?)

I think your initial suggestion to update the docs is the best solution; people manually initialising `Dlg` are likely to lean more towards the Coder end of things, so if there's an easily google-able solution (with a reason for the change) I think they'll be okay with updating their code. We can even link to docs for pipe syntax, to emphasize that this change comes hand in hand with useful new functionality.

jfkominsky:
My main goal here is actually not going forward (where I assume people will switch to dicts quite readily, as you say) but backwards compatibility with existing experiments for the purposes of reproducibility. For example, the current Dlg change will mean that any PyHab experiment made before this year will never work w/PsychoPy 2024 or later without some code modification. I guess the question is whether we should prioritize intuitive behavior for the future or backwards compatibility. Either way it involves a documentation update - the alternative is to change the docs to explain clearly when to use numeric indexes and when to use strings to retrieve data, and as you say, people manually programming Dlgs can probably google it.

TEParsons:
Hmm, maybe there's a way to have our cake and eat it. Two solutions that come to mind:
- We could make a subclass of dict (`IndexDict` maybe?) which can accept either a numeric index or a key and still set/get the correct value by overloading `__getitem__` and `__setitem__`. 
- We could add a subclass of Dlg in the `psychopy-legacy` plugin which returns a list like it used to, and have a method which assigns this subclass to the name `psychopy.gui.Dlg` (e.g. `psychopy_legacy.activateLegacyDlg()`). That way things like PyHab can be updated to work in either version like so:
```python
if Version(psychopy.__version__) >= Version("2024.2.0"):
    try:
        from psychopy_legacy import activateLegacyDlg
    except ModuleNotFoundError:
        raise ModuleNotFoundError("PyHab requires the plugin psychopy-legacy to work in PsychoPy version 2024.1.0 onwards. You can install this by either `pip install psychopy-legacy` or using the plugin manager in PsychoPy Builder.")
    else:
        activateLegacyDlg()
```

jfkominsky:
I was contemplating something like an IndexDict as well (probably a subclass of OrderedDict?), but frankly wasn't confident I could figure out how to incorporate it into the larger structure of psychopy's code in order to make it accessible in all the right ways.

I think that would be the ideal solution. There's one weird edge case that I don't think anyone would actually do in creating Dlgs, which is the fact that normal dicts can use integers as keys and this IndexDict won't be able to do that and will probably need some kind of safeguard against it (possibly as simple as converting int to str if set as key in `__setitem__`). Other than that wrinkle, I think it's the best solution because it supports both old code and new code seamlessly.

Do you have a thought about how/where to implement it?

TEParsons:
I think handling integers as keys should be okay, I would expect that numeric indices explicitly supplied as keys would take precedence over positional indices implied by the order of keys. I'll have a go at implementing it in a fresh branch, and will add a test to make sure it behaves as expected. 

For implementing it into the rest of PsychoPy, that's what I'm cautious about too. So long as it's a subclass of dict (or OrderedDict, which is itself a subclass of dict I think?) it should largely behave the same as normal - it'll still return True for `isinstance(data, dict)`, will still report the same `len()` and  will still respond the same to `"key" in data`, so everything onwards should treat it exactly the same as a dict.

jfkominsky:
Made obsolete by #6809

PR #6809, opened 2024-08-27
BF: Make output of Dlg.show backwards compatible with older versions

### LINKED ISSUE #6781: [Bug]: gui.Dlg no longer returns same type of object, and no longer shows question text

TEParsons:
Implementing the class discussed in #6806, allowing data returned from Dlg.show to be indexed by either keys (as in a dict) or indices (as in a list). In situations where there's conflict, i.e. where one of the keys in the dict is explicitly states as an integer, the explicit key takes priority over the positional index. I've added some tests to be extra sure it's behaving as expected.

Flagging @jfkominsky to give this a look over, if you could pull down the branch and give it a try with your scripts that would be great

Fixes #6781 and #6790

### DIFF
--- a/psychopy/gui/qtgui.py
+++ b/psychopy/gui/qtgui.py
@@ -8,6 +8,7 @@
 # Distributed under the terms of the GNU General Public License (GPL).
 import importlib
 from psychopy import logging, data
+from psychopy.tools.arraytools import IndexDict
 from . import util
 
 haveQt = False  # until we confirm otherwise
@@ -167,7 +168,7 @@ class Dlg(QtWidgets.QDialog):
         self.inputFields = []
         self.inputFieldTypes = {}
         self.inputFieldNames = []
-        self.data = {}
+        self.data = IndexDict()
         self.irow = 0
         self.pos = pos
         # QtWidgets.QToolTip.setFont(QtGui.QFont('SansSerif', 10))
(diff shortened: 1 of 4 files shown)


jfkominsky:
Can confirm this works. I threw it a couple of challenging use cases with dynamic list indexes and it handled them perfectly. Thank you so much!

jfkominsky:
I decided to also check with my updated code that uses the new dict keys as well. Interestingly it did generate an error here, but it still worked and did what it was supposed to do. Here's the error message:
29.0356     ERROR     Error in handleLineEditChange: inputFieldName=Height of movie stimuli in pixels, type=<class 'int'>, value=, error=invalid literal for int() with base 10: ''

The field in question is one that has an "initial" value of 600 (an integer) and I changed it to 720. The inputFieldName is the label, the key is 'movieHeight'. It still recorded the change and saved the output correctly, but I wanted to note this in case it indicates a potential issue.

PR #6830, opened 2024-09-04
ENH: added the option for an expVersion into the SettingsComponent

peircej:
In general this is useful to keep track of the experiment version but this is a first draft and we should check if it works (e.g. does it show up in the csv file). Very much untested as yet


peircej:
This seems to be working fine. By including in expInfo it makes it into the data file and it seems pretty straightforward and useful

PR #6843, opened 2024-09-11
NF: Automatic microphone polling

mdcutone:
This PR allows the microphone to poll asynchronously by getting samples using a background thread. When `autoPolling=True`, the background thread will get samples at a rate specified by `pollInterval`. This value just needs to be less than the stream buffer size to prevent samples from being missed.  Calls to `poll()` in existing scripts will still work as they always have. However, the user never actually needs to call `poll()` explicitly when `autoPolling` is enabled, as the stream buffer is effectively infinite and all data is contains will be automatically written to the recording buffer when the stream is stopped.

This also has the benefit of keeping the audio stream alive if the device drivers tend to timeout and go into an inactive state.


mdcutone:
Updated to allow polling continuously between `open()/close()` calls. Also various other improvements and documentation changes

PR #6858, opened 2024-09-16
BF: Linux iohub keyboard relative import error

zeyus:
When running psychopy on linux, using the iohub mouse device results in an importerror:

```text
ImportError: cannot import name 'Keyboard' from 'psychopy.iohub.devices'
```

This fixes the import error by aligning the linux mouse device variant to the other OS variants.

### DIFF
--- a/psychopy/iohub/devices/mouse/linux2.py
+++ b/psychopy/iohub/devices/mouse/linux2.py
@@ -6,7 +6,8 @@
 from ctypes import cdll
 
 from . import MouseDevice
-from .. import Keyboard, Computer, Device, xlib
+from .. import Computer, Device, xlib
+from ..keyboard import Keyboard
 from ...constants import MouseConstants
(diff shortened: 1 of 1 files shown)


TEParsons:
Thanks for catching this! The fix looks good to me, @mdcutone can you confirm?

peircej:
Is this not a problem in release branch too (or is it actually a FF not a BF)?

zeyus:
> Is this not a problem in release branch too (or is it actually a FF not a BF)?

Yeah I noticed it first in release, should I have made the PR against main?

peircej:
Yes, for a bug fix, should really check out release branch and PR back to release branch but no worries. We make bug-fix releases from that branch and dev branch is kept for major versions (ie. a bug fix to dev will only get released in Feb or June)

 I'll cherry-pick to release

PR #6869, opened 2024-09-23
FF: fix startView across OS platforms that opens psychopyApp.py

mh105:
Same fix as #6789. I found out today that this issue of automatically opening `psychopyApp.py` is happening on Windows standalone app as well. I'm not sure what is trying to open `psychopyApp.py` when launching the app, but this FF commit can at least make the `startView` setting in preference work correctly and not always open Coder view due to the `psychopyApp.py`.


TEParsons:
I've just realised this is the same error I was fixing in #6875... The two solutions don't clash though, so I'll pull this in too as a "belt and braces" approach :)

PR #6873, opened 2024-09-24
BF: if PsychoPy version includes "postN" then strip from PsychoJS version number

peircej:
Originally we only expected to use `rcN`, `aN`, `bN`, `devN` tags but recently we used `postN` which broke the PsychoJS experiments


JartanFTW:
:+1: PR looks good to me. Very little code to check :stuck_out_tongue:

PR #6877, opened 2024-09-26
ENH: Better error catching when loading plugins

TEParsons:
Currently, because we use `return False` rather than `continue` when an entry point fails to load, if one entry point fails then none of the rest of the entry points for that plugin are registered. There also isn't any detail given, so the logs just look like:
```
ERROR Failed to load <entry point> from <plugin-name> for unknown reason. Skipping
```

This PR means that when an entry point fails, only that specific entry point is skipped, and we'll log the reason.

### DIFF
--- a/psychopy/experiment/plugins.py
+++ b/psychopy/experiment/plugins.py
@@ -1,3 +1,6 @@
+from psychopy import logging
+
+
 class PluginDevicesMixin:
     """
     Mixin for Components and Routines which adds behaviour to get parameters and values from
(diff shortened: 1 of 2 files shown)


mdcutone:
Wouldn't this result in partially loaded plugins? Should we want to stop loading entry points for a specific plugin if there is a fail and warn the user? Partial loading of plugins could lead to undefined behaviour. I agree the logging needs to be improved around this.

TEParsons:
It would, but I think a partially loaded plugin is better than an aborted plugin. For example, the thing that motivated this was the Cedrus plugin failing to load due to an error in Riponda VoiceKey - meaning the old Cedrus Button Box Component wasn't loading, despite it not being at all dependent on the Riponda backend. If an entry point can still load without error I think we should allow it to, even if there's a problem with another entry point in the same plugin.

PR #6885, opened 2024-09-30
BF: Handle when record of last open app windows is an invalid value (e.g. from old version)

### LINKED ISSUE #6812: [Bug]: Builder fails to open after splash screen
### PsychoPy Version

2024.2.1

Linux-based systems

### Bug Description

After installing PsychoPy in a fresh environment on a fresh system and launching from the command line via e.g. `psychopy`, the splash screen & other first-time messages show, but no builder/coder/runner windows appear.

### Expected Behaviour

Builder/coder/runner should open.

### Steps to Reproduce

Essentially https://github.com/psychopy/psychopy/issues/6768 for setup. Can work around by explicitly running `app.startApp()` with a specific view (which puts it in a good state for subsequent launches), and deleting the .psychopy3 folder causes the behavior to occur again.

### Additional context

I don't _think_ this is https://github.com/psychopy/psychopy/issues/6768, since it's just hanging instead of segfaulting? I think the issue for this one is that the `lastFrame` value in the config here https://github.com/psychopy/psychopy/blob/bed821597c7ff73b9dd0413dd258cb5923a6fde1/psychopy/app/appData.spec#L2 
hasn't been valid for awhile, and when https://github.com/psychopy/psychopy/pull/6675 changed the defaultView from 'all' to 'last' it became apparent?

This is in an Ubuntu 24.04 VM, Python 3.10

TEParsons:
Should fix #6812 (though I was unable to replicate the issue on Windows so please give this a try @aforren1 and let me know if it fixes the problem)

### DIFF
--- a/psychopy/app/_psychopyApp.py
+++ b/psychopy/app/_psychopyApp.py
@@ -532,6 +532,8 @@ class PsychoPyApp(wx.App, handlers.ThemeMixin):
             if self.prefs.app['defaultView'] == 'all':
                 startView = ["builder", "coder", "runner"]
             elif self.prefs.app['defaultView'] == "last":
+                if self.prefs.appData['lastFrame'] == "both":
+                    self.prefs.appData['lastFrame'] = "builder-coder-runner"
                 startView = self.prefs.appData['lastFrame'].split("-")
             elif self.prefs.app['defaultView'] in ["builder", "coder", "runner"]:
                 startView = self.prefs.app['defaultView']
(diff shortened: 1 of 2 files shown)


aforren1:
Thanks @TEParsons, I'll check today or tomorrow.

aforren1:
Seems to work for me-- I could reproduce on Windows by deleting my existing `...\AppData\Roaming\psychopy3` folder, and trying to launch the psychopy app from a new conda env (e.g. `conda create -n test python=3.10` + `pip install psychopy` + `psychopy`). The current release briefly shows the splash screen & exits, and this branch lets the app fully launch.

The standalone 2024.2.1 didn't seem affected by this issue (??)

PR #6895, opened 2024-10-04
ENH: Added error type for devices not found to make it easier to catch

TEParsons:
Makes it easier to distinguish between the requested device not being connected and initialisation failing for other reasons


TEParsons:
Closing and cherry picking in #6954 to deal with merge conflicts

PR #6912, opened 2024-10-11
RF: Various Numpy 2 and Python 3.10+ compatibility fixes

mdcutone:
This PR makes the appropriate changes for compatibility with Numpy 2 based on the depreciation warnings that come up in the build logs.


mdcutone:
@TEParsons Added fixes for escape characters

PR #6918, opened 2024-10-16
FF: Updated `componsTemplate.txt`

mdcutone:
Updates the `componsTemplate.txt` file to clear most warnings in the build logs.


mdcutone:
Needed an update to ignore random unicode errors in the output file

PR #6931, opened 2024-10-23
RF: Interpret --builder, --coder, etc. arguments *before* restoring previous files / windows from prefs

### LINKED ISSUE #6914: [Bug]: v2024.2.* not respecting arguments like --coder or --builder
### PsychoPy Version

`>=2024.2.0`

Linux-based systems

### Bug Description

I tested on PopOs.  2024.2.* versions just open all 3 windows. 

### Expected Behaviour

Open only the specified window.

### Steps to Reproduce

1. Install version >=2024.2.0
2. use with --coder
3. Opens all 3 windows
...

### Additional context

_No response_

TEParsons:
Fixes #6914, as now we only reopen files/frames from the last time the app was opened if not actively told to open a specific frame.

This slipped through because we have arg parsing happening manually in multiple locations, so I moved it all to `psychopy.app.psychopyApp:main` and used Python's `argparse` handler as it's more robust than us manually iterating through `sys.argv`.

### DIFF
--- a/psychopy/app/__init__.py
+++ b/psychopy/app/__init__.py
@@ -61,7 +61,8 @@ def startApp(
         safeMode=False, 
         startView=None,
         startFiles=None,
-        firstRun=False
+        firstRun=False,
+        profiling=False,
     ):
     """Start the PsychoPy GUI.
 
@@ -130,7 +131,8 @@ def startApp(
         safeMode=safeMode, 
         startView=startView,
         startFiles=startFiles,
-        firstRun=firstRun
+        firstRun=firstRun,
+        profiling=profiling,
     )
 
     # After the app is loaded, we hand off logging to the stream dispatcher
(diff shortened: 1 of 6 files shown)


mdcutone:
We'll need to double-check if there are any environment variables that need to be set prior to running using the `-x` param. Stuff like pointing to user packages for plugins

TEParsons:
I guess if they're running a .py file it'll either:
- have been generated from Builder, so will have `plugins.activatePlugins` at the beginning and have reference to the relevant folders via `prefs`
- have been written by hand, in which case `plugins.activatePlugins` is probably something best left to their discretion (as they may actively want to run it without plugins if they're not needed and don't want any delay while they load)

I was thinking actually, if given a `.psyexp` file with `-x`, we could compile and run then-and-there, which could be useful

PR #6937, opened 2024-10-23
NF: Set caret position in `Textbox2` with cursor

mdcutone:
This PR adds support for mouse interaction with `Textbox2`. If the `Textbox2` stimulus is `editable==True` and `clickable==True` (new param), the user will be able to set the focus and caret position of the textbox with the mouse.

### DIFF
--- a/psychopy/visual/basevisual.py
+++ b/psychopy/visual/basevisual.py
@@ -1661,6 +1661,69 @@ class WindowMixin:
         self._updateListShaders()
 
 
+class PointerMixin:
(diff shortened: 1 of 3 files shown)


peircej:
Very cool @mdcutone - this will be very useful
I feel like we can probably vectorise that for...loop into a numpy array operation. That isn't easy for non-square objects but, for a square, because the `contains()` decision is a simple min<x<max decision, I think we can find the row where the verts fit the equation in a find. For a large textbox (e.g. 1000) it will surely be a lot faster

PR #6938, opened 2024-10-23
BF: Fixed ioHub process staying alive after crash

mdcutone:
This PR catches errors that come from the iohub server process and prints them instead of simply crashing. This is a WIP so the dev team can discuss this change before merging.


mdcutone:
Added an `atexit` handler which shuts down any active connection when the program exits. This should also handle cases where the app exits abnormally

TEParsons (approved):

PR #6944, opened 2024-10-25
PKG: Automatically label PR's according to their title

TEParsons:
Currently `release.yml` categorises everything as "Other changes" unless we've manually gone and labelled it. This action checks for the tags from [our commit spec](https://psychopy.org/developers/repository.html#psychopy-commit-messages) and assigns the appropriate label when the PR is opened.

### DIFF
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -22,3 +22,6 @@ changelog:
       labels:
         - "📓 docs & demos" 
         - "🔨 tests"
+    - title: "Other Changes"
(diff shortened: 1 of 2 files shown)


TEParsons:
I can't work out why it's still complaining about permissions - from what I'm finding online it sounds like `pull-requests: write` should be the only permission it needs. @peircej is there a setting on the PsychoPy repo about actions accessing/editing pull requests?

PR #6945, opened 2024-10-28
BF: Sending experiment to Runner was sometimes (wrongly) toggling pilot/run mode

TEParsons:
Solution is to split up the code to update the buttons from the code to trigger the event, set modified, etc. so that selecting a new experiment on add can *just* update the buttons, not alter the experiment file or raise an event.

@mh105 could you pull down and try this out to confirm it fixes the bug you described?


mh105:
Works beautifully. Thanks @TEParsons !

PR #6947, opened 2024-10-28
ENH: Added case-sensitive search and search-as-you-type in Builder search

jderrfuss:
- Search as you type.

- Case-sensitive is no longer the default, but is still possible by ticking a checkbox.

- The name of the routine where a component is located is now also displayed in the search results.


TEParsons:
All looks good! Only worry I had was that on-typing search would feel laggy, particularly for big experiments, but actually testing this out on my mid-range laptop it's absolutely fine. I guess it's ultimately just doing basic string matching in an array so seems to return in a matter of milliseconds. So yep happy to pull in once tests pass, thanks for contributing!

PR #6948, opened 2024-10-29
FF: skip ioHub event callback without PsychoPy window

mh105:
There is a very niche bug within ioHub at the end of an experiment on Windows. When the PsychoPy window has closed but the rest of close-down steps are still going, if we move the mouse at this moment, it triggers a Traceback error due to trying to access `.key()` on the closed psychopy window, which is now `None`.

p.s. It's still a bug on Linux but the Linux implementation is wrapped by a `try` statement so does not lead to a crush of the ioHub server subprocess.

This FF allows the `_nativeEventCallback` method to `return True` and ignore the event callback if no PsychoPy window is currently open. This fix allows both the ioHub server subprocess and PsychoPy main process to close out gracefully.

### DIFF
--- a/psychopy/iohub/devices/mouse/linux2.py
+++ b/psychopy/iohub/devices/mouse/linux2.py
@@ -90,6 +90,10 @@ class Mouse(MouseDevice):
                 logged_time = currentSec()
                 event_array = event[0]
 
+                if self._iohub_server._psychopy_windows is None:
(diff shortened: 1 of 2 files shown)


mh105:
Actually I realized that the problem is not that the window got closed but the fact that the ioHub server itself got closed.

PR #6954, opened 2024-10-31
ENH: Added error type for devices not found to make it easier to catch

TEParsons:


TEParsons:
Closed in favour of #7019

PR #6956, opened 2024-11-01
ENH: Add "Find" buttons to Builder and Coder ribbons

TEParsons:


TEParsons:
@jderrfuss @peircej following our conversation by email, here's the necessary changes to add Find to the ribbon (also added for Coder)

PR #6958, opened 2024-11-01
RF: Use threshold values from 0 to 1 rather than 255

TEParsons:


TEParsons:
Closing in favour of #7122 (resolving merge conflicts)

PR #6961, opened 2024-11-04
BF: Fixed `cielch2rgb` handling of multidim arrays

### LINKED ISSUE #6925: [Bug]: multidimensional arrays break `cielch2rgb`
### PsychoPy Version

2024.2.1

macOS Silicon

### Bug Description

When passing in a multidimensional numpy array to `cielch2rgb`, this produces the following traceback error:

```python
  File "/Users/x/opt/anaconda3/envs/psychopy/lib/python3.10/site-packages/psychopy/tools/colorspacetools.py", line 376, in cielch2rgb
    lab[:, 1] = lch[:, 1] * numpy.math.cos(numpy.math.radians(lch[:, 2]))
TypeError: only length-1 arrays can be converted to Python scalars
```

I believe the fix is pretty simple: replace all calls to functions in `numpy.math` (e.g., from `numpy.math.cos` to `numpy.cos`). Calling any function in `numpy.math` also results in a deprecation warning (even when running numpy `1.26.4`, so not an issue due to any breaking changes introduced in `2.x`), so it may be worthwhile to update these function calls regardless of this bug.

### Expected Behaviour

The expected behavior is for the function to return a multidimensional numpy array specifying RGB coordinates.

### Steps to Reproduce

Minimal example below:

```python
from psychopy.tools.colorspacetools import cielch2rgb
import numpy as np

# For displaying a colorwheel
texture_res = 256
colorwheel_cielch = np.ones([texture_res, texture_res, 3], dtype=float)
colorwheel_cielch[:, :, 0] = 73  # Lightness
colorwheel_cielch[:, :, 1] = 39  # Chroma
colorwheel_cielch[:, :, 2] = np.linspace(0, 360, texture_res, endpoint=False)

colorwheel_rgb = cielch2rgb(colorwheel_cielch)
```

### Additional context

_No response_

mdcutone:
Fixes #6925 by using vectorized functions

### DIFF
--- a/psychopy/tests/test_tools/test_colorspacetools.py
+++ b/psychopy/tests/test_tools/test_colorspacetools.py
@@ -114,34 +114,35 @@ def test_srgbTF():
 
 
 @pytest.mark.colorspacetools
-def test_cielab2rgb():
-    """Test the CIE-Lab to RGB function using the sRGB transfer function.
-
+def test_cielch2rgb():
+    """Test the CIE-LCh to linear RGB function.
     """
-    # preset CIE-Lab colors using a D65 white-point and expected RGB values
-    cielabD65 = np.array([
-        [0.0, 0.0, 0.0],   # black
-        [100.0,  0.0,  0.0],    # white
-        [53.24, 80.09, 67.20],  # red
-        [87.73, -86.18, 83.18],   # green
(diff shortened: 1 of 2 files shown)


mdcutone:
Added a test and reduced the number of operations in the conversion function

PR #6962, opened 2024-11-04
FF: Rework logic for speaker stream initialization

mdcutone:
Reworks the logic when creating a new audio playback stream. This provides better handling of default devices, makes parameter handling more robust, and bypasses unused calls in the `sound` package


peircej:
@mdcutone regarding `useWASAPI=True` maybe we should be honoring the recent addition to prefs that allows that to be optional? https://github.com/psychopy/psychopy/pull/6922

psychopy/hardware/speaker.py
```diff
@@ -136,10 +185,10 @@ def createStream(self):
...
             self.profile['DefaultSampleRate'], 
             # if that fails, try some common sample rates
-            48000,
-            44100, 
-            22050, 
-            16000
+            48000.0,
+            44100.0, 
+            22050.0, 
+            16000.0
```
TEParsons: Is this going to lead to the opposite problem - it'll fail when ptb is expecting an int? Should we just add both (seeing as we're silencing fails anyway)? Like:
```
48000,
48000.0,
44100,
44100.0,
etc.
```

psychopy/hardware/speaker.py
```diff
@@ -181,6 +230,11 @@ def createStream(self):
                 "Failed to setup a PsychToolBox audio stream for device %(DeviceName)s "
                 "(%(DeviceIndex)s)." % self.profile
             )
+
+        logging.warning(
```
TEParsons: Shouldn't this be `info` or `debug`?

psychopy/sound/__init__.py
```diff
@@ -198,8 +198,18 @@
 if Sound is not None:
     audioLib = thisLibName
     init = backend.init
-    if hasattr(backend, 'getDevices'):
+    getDevices = None
+    if audioLib != 'ptb' and hasattr(backend, 'getDevices'):
```
TEParsons, line 202: Isn't this already handled by `SpeakerDevice.getAvailableDevices`? As that imports `sound.getDevices` which then handles the different backends

TEParsons (changes requested):

psychopy/hardware/speaker.py
peircej: Or an Alert?

TEParsons: The same applies to alert - if the message is just "Created stream for speaker device" I don't think we should be alerting them to it as if something's gone wrong

psychopy/hardware/speaker.py
mdcutone: PTB returns floats for these values so I assumed it wants them back as that. I've been bitten by passing ints to extensions since they typically don't handle implicit conversion well.

mdcutone: Anyways I restored the ints

psychopy/hardware/speaker.py
mdcutone: Yes, that should be a lower log level. Changing it.

psychopy/sound/__init__.py
mdcutone: I don't know what your referring to, there is no reference to `getDevices()` in `Sound.__init__.py` or the file/module for the PTB backend. I tried to match the original functionality by referencing the newer stuff in `hardware.speaker`

TEParsons (approved):

PR #6966, opened 2024-11-11
BF Rewind to first trial

wakecarter:
Previous code was rewinding to trial 1 not trial 0 when the first trial was requested


TEParsons:
Closing in favour of #6971

PR #6967, opened 2024-11-11
RF: Refactor `Joystick` class

mdcutone:
Made some changes the the Joystick class:

* Module doesn't import libraries other than pyglet unless needed
* Added methods for scaling axes values
* New attributes query the device value on access
* Added docstrings
* Added static method `getAllJoysticks` to get a human-readable list of devices
* Changed `id` -> `deviceIndex` to avoid overriding built-in, you can still pass `id` but we shouldn't allow that in a future version
* Joystick plugins are possible
* Joystick mappings can be set


mdcutone:
Added attributes for joysticks/gamepads that support spatial and motion tracking (e.g. Oculus Touch controllers).

PR #6974, opened 2024-11-20
RF: Move `RigidBodyPose` and `BoundingBox` to `mathtools`

mdcutone:
Moves the `RigidBodyPose` and `BoundingBox` classes to `mathtools`. This is helpful since the RBP class will be needed for future VR integration, motion trackers (e.g. OptiTrack), and joysticks (Oculus Touch). Making them usable without needing to import `visual.stim3d` (and by extension `visual`).

References have been added to `stim3d` for compatibility.


mdcutone:
Moved `lookAt` over to `mathtools` to avoid circular import. It also doesn't always apply to view transformations. There is a reference in `viewtools` still

PR #6980, opened 2024-11-22
NF: Add option for custom calibration points in Eyetracker Calibration

TEParsons:
- The EyetrackerCalibration object can now take any visual stimulus as its `target` parameter, not just a TargetStim (thanks to the `serialize` and `actualize` methods from #6801)
- The Eyetracker Calibration Routine in Builder now has a checkbox "use custom" in its Target tab, if ticked it gives a textbox in which you just put the name of the visual Component you want to use as a calibration point

Here's it in action, using an ImageComponent with `default.png` in its `image` param (using MouseGaze backend):

https://github.com/user-attachments/assets/6e034f8a-59e9-4830-b95a-d67a240f1ece


TEParsons:
May be of interest @mh105 :)

mh105:
Nice!! Thanks for flagging. Definitely very relevant, especially in infant studies.

PR #6981, opened 2024-11-22
BF: restrict glob matching to rename pattern in handleFileCollision()

mh105:
Otherwise trial summary files also match into the pattern of `rootName*extension` and creates confusing counter in the renamed files. This additional use of regular expression can avoid that behavior.

### DIFF
--- a/psychopy/tools/fileerrortools.py
+++ b/psychopy/tools/fileerrortools.py
@@ -33,7 +33,7 @@ def handleFileCollision(fileName, fileCollisionMethod):
                "fileCollisionMethod to overwrite.")
         raise IOError(msg % fileName)
     elif fileCollisionMethod == 'rename':
-        rootName, extension = os.path.splitext(fileName)
+        rootName, extension = os.path.splitext(os.path.normpath(fileName))
         matchingFiles = glob.glob("%s*%s" % (rootName, extension))
 
         # Keep only files matching the specific pattern used in renaming
--- a/psychopy/tools/fileerrortools.py
+++ b/psychopy/tools/fileerrortools.py
@@ -9,6 +9,7 @@
 """
 import os
 import glob
+import re
 
 from psychopy import logging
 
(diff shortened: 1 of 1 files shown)


TEParsons:
Funnily enough I actually fixed the same bug in dev recently (#6917), pretty much the same way only using `pathlib.Path` to do the regex. If you try your test cases out in the current dev branch, do you still get the issue?

Also did you get chance to play with the test suite? We have a tests for file collision behaviour (`psychopy.tests.test_tools.test_fileerrortools`) but they're fairly minimal, if you're in the neighbourhood then this is an area we could certainly be testing more thoroughly :)

mh105:
@TEParsons Amazing. I just pulled the commits you made to the `dev` branch, and the `handleFileCollision()` works correctly in my use cases.

I like your solution a lot more, since counting the list length of the `glob` result is not bullet-proof. For example, if I manually ran the same subject ID first with session ID `001` and then with session ID `001_2`. If I run it again with session ID `001`, the old (as well as mine here) `handleFileCollision()` will fail to rename the files correctly, crush, and lose the data.

The iterative checking you implemented does not have this issue, and will correctly take the `001_1` spot in this case for saving data files.

Re: test suites. Unfortunately I haven't. I promise once I'm progressed beyond postdoc I'll look at the PsychoPy test suites - probably also when I will be adding test suites to help folks in my lab. For now, I'm swamped by too many other things, which make me really appreciate the awesome work that you and Jon are putting to improve PsychoPy. I hope I can contribute more to the testing of PsychoPy in the near future (which is one of my favorite parts of coding, but unfortunately takes too much time to test things exhaustively. Honestly it is a bit addictive for me so it sucks up my time).

TEParsons:
Don't worry about it, any contributions you make are appreciated! There's no pressure to do more if you're swamped elsewhere, you're a volunteer after all.

mh105:
Oh well, it is a crucial coding practice to include test cases. If I'm telling that to my students, I'd better adhere to that myself, or come up with some sort of excuse like above ;)

PR #6982, opened 2024-11-22
RF: Summary csv files from loops now have a _ between the filename and the loop name

mh105:
This is super minor, but given the way duplicated files are renamed and how the session ID is appended to the rest of file name using underscores, I feel that the summary csv files should also use `_[loopName]` for naming. Totally fine and harmless if not worth changing.

### DIFF
--- a/docs/source/builder/settings.rst
+++ b/docs/source/builder/settings.rst
@@ -108,6 +108,9 @@ Save Excel file
 Save csv file (summaries)
     If this box is checked a summary file will be created with one row corresponding to the entire loop. If a keyboard response is used the mean and dtandard deviations of responses across trials will also be stored.
 
+.. note::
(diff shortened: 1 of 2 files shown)


TEParsons:
The only worry I'd have is people's analysis pipelines - if someone had it set up to read in all the summaries and found them by looking for e.g. `rf"{filename}_{time_regex_pattern}{loop_name}\.csv` then they'd need to add an underscore in newer versions. I think if we communicate that clearly in the release notes / documentation that's fine though, the important thing is that when they go looking for a reason why their analysis stopped working they find one (and an easy solution).

I think `docs/source/builder/settings.rst` would be a good place, under the param for saving summaries? Just a note about how the filename is put together and that it's changed in 2025.1.0

mh105:
@TEParsons Sounds good. Added a note to the Sphinx doc page - let me know if it looks good.

Issue #6369, opened 2024-04-12
[Bug]: Two Counterbalance routines with the same name can be added / Conditions file is not updated in Counterbalance routine


hsogo:
### PsychoPy Version

2024.1.1

### What OS are your PsychoPy running on?

Windows 11

### Bug Description

Sorry for reporting two bugs in a single issue, but both concerns with the Counterbalance routine.

1. Adding a new Counterbalance routine to an experiment which already has Counterbalance routine, two tabs with the same name appear in the Routine pane.
2. Updating the name of Conditions file using "File" button is not effective until further changes are made to the Counterbalance routine.

### Expected Behaviour

1. Two property tabs with the same name should not be created.
2. Changes in the name of Conditions file using "File" button should be effective immediately.

### Steps to Reproduce

Bug 1
1. Create a new experiment, put a counterbalance routine, insert the routine to flow and save the experiment.
2. **Close and reopen** the experiment, and try to add a counterbalance routine.

https://github.com/psychopy/psychopy/assets/2094930/9a5e1d94-a65e-490d-b412-f984ccab552d

Bug 2
1. Create a new experiment, put a counterbalance routine, insert the routine to flow and save the experiment.  Prepare a condition file for the counterbalance component.
2. Open the property of the counterbalance routine, set "Groups from..." to "Conditions file" and press "Specify file..." button on "Condition" property to select the condition file.
4. Click "Write python" button on the Builder ribbon to generate Python script. Then check the parameter of `data.utils.importConditions()`.  The conditions file name is empty.
5. Return to the Builder and edit property of the counterbalance routine (editing textboxes, toggling checkboxes, and so on).  Then generate Python script and check the parameter of `data.utils.importConditions()` again.  Now the conditions file name is filled.

https://github.com/psychopy/psychopy/assets/2094930/15b0628f-41d1-4327-85d9-86470bed2e76

PR #6999, opened 2024-12-04
BF: Standalone Routine names weren't handling name conflicts properly

Lance-Buescher1:
Loaded routine names with tags not equal to "Routine" weren't added to the user namespace, so adding a duplicate name wouldn't be recognized.

fixes GH-6369 -- BUG 1 (there are 2 bugs listed in this issue)

### DIFF
--- a/psychopy/experiment/_experiment.py
+++ b/psychopy/experiment/_experiment.py
@@ -770,6 +770,34 @@ class Experiment:
 
         return exp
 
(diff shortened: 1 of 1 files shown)


TEParsons:
Good catch, I think this is a good solution :) I've just renamed the PR for the sake of the changelog, to make it more layperson friendly

PR #7002, opened 2024-12-05
RF/EN: Pyglet2 and OpenGL 3 support

mdcutone:
PR which updates PsychoPy's rendering pipeline to use modern functionality in OpenGL needed to support Pyglet 2 and platforms which don't support the OpenGL 2.1 API


mdcutone:
Due to some bugs upstream, we cannot ship with Pyglet 2 this release, however most of the stimuli classes will work if installed. Full support should be available in the next major release.

iPsych:
@mdcutone Does Pyglet 2 is supposed to resolve Metal related issue (broken Vsync, etc...) in recent MacOS after Monteray?
 The frame dropping and huge variance in frame timing is continously observable since Ventura. 
 I tested MacOS versions from Ventura, and Asahi Ubuntu in the same M1 machine, and pretty sure that it's an OS related issue, more exactly, discontinued openGL support, and Metal compatibility of backend python libraries (like PyGlet).

mdcutone:
> @mdcutone Does Pyglet 2 is supposed to resolve Metal related issue (broken Vsync, etc...) in recent MacOS after Monteray? The frame dropping and huge variance in frame timing is continously observable since Ventura. I tested MacOS versions from Ventura, and Asahi Ubuntu in the same M1 machine, and pretty sure that it's an OS related issue, more exactly, discontinued openGL support, and Metal compatibility of backend python libraries (like PyGlet).

Unlikely, this just updates the OpenGL calls and rendering techniques to remove deprecated/unsupported features where possible

PR #7003, opened 2024-12-06
FF: Allow scripts to fetch the current Liaison server rather than creating a new one

TEParsons:
This means external applications can call a new script and have it map its classes to Liaison without having to create a new Liaison server just for it, like so:
```
from psychopy import liaison
server = liaison.WebSocketServer.servers[host]
```


TEParsons:
Closed in favour of 7004

PR #7007, opened 2024-12-11
DOC: More search-engine-optimized message from PluginStub

TEParsons:
Also moved PluginStub to `psychopy.plugins` to make it easier to find

### DIFF
--- a/psychopy/hardware/cedrus.py
+++ b/psychopy/hardware/cedrus.py
@@ -19,13 +19,14 @@ instead (bundled with Standalone PsychoPy)::
 """
 
 import psychopy.logging as logging
-from psychopy.tools.pkgtools import PluginStub
+from psychopy.plugins import PluginStub
 
 
 class RB730(
     PluginStub, 
     plugin="psychopy-cedrus", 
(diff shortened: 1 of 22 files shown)


peircej:
I think this will be a big improvement to searching @TEParsons but the tests are currently failing with
```
________ ERROR collecting psychopy/tests/test_hardware/test_emulator.py ________
psychopy/tests/test_hardware/test_emulator.py:11: in <module>
    from psychopy.hardware import emulator
psychopy/hardware/emulator.py:25: in <module>
    class SyncGenerator(
E   TypeError: __init_subclass__() missing 2 required positional arguments: 'docsHome' and 'docsRef'
```

TEParsons:
Ah yeah looks like I missed a few - I thought they all had `docsLink` so I went and split it into `docsHome` and `docsRef` (allowing the constructed string to have both a link to the plugin's homepage and to the specific class docs), but looks like the ones which don't have docs yet I'd just left `docsLink` blank so I didn't catch them when I CTRL+SHIFT+F'd for `docsLink=`. Fixing now :)

Also, good that the test suite catches this sort of thing!

PR #7039, opened 2025-01-06
FF: packages symlink only if the folder doesn't exist

peircej:
We were testing if the folder was already a symlink but maybe it's already a folder

Not sure why but that occurred in our devops build


peircej (commented):
@mdcutone I'm not sure why this happened but on our build VM this failed because the folder already existed as a regular folder. It seems in that case we should not create the symlink and just carry on as before

mdcutone (approved):

PR #7049, opened 2025-01-13
RF: Use more layperson friendly terms for photodiode ("light sensor") and voicekey ("sound sensor")

TEParsons:
As I'm just updating documentation & names, this should mostly not affect how things work. However, plugins will need to be updated to target `VisualValidatorRoutine` rather than `PhotodiodeValidatorRoutine` when adding a backend.


TEParsons:
Matching PR's opened in BBTK and Cedrus plugins:
- https://github.com/psychopy/psychopy-bbtk/pull/15
- https://github.com/psychopy/psychopy-cedrus/pull/23

PR #7058, opened 2025-01-20
RF: Remove max size param from Microphone

TEParsons:
Param is rarely used and was causing problems online - it's also somewhat irrelevant now that we have control over "policy when full"


mdcutone:
Making changes to the microphone class for indefinite recording duration for audio. Users will not be required to specify the audio sample storage buffer size in advance. They can still specify limits but it's no longer a requirement

PR #7065, opened 2025-01-24
BF: Avoid error when providing intensity to QuestPlusHandler.addResponse() on first trial

laurabpaulsen:
If a custom intensity is provided on the first trial previously and IndexError was raised as it tries to self.intensities.pop() on an empty list. Fixed by checking the length of self.intensities and only popping if length is not 0.

Same solution was already used in QuestHandler!


peircej:
Good point @laurabpaulsen 
Thanks

PR #7077, opened 2025-01-30
NF: Asynchronous audio file writer

mdcutone:
Adds the `psychopy.tools.audiotools.AudioFileWriter` class which allows for asynchronous audio file writing to disk. Passing audio samples to the class will write data to a file in a separate thread to disk. If the application quits unexpectedly, partial data will be saved


BXYMartin:
I am new to this repo but I am thinking, since the current solution is using threading and when the experiment ends, the current implementation will check if the thread implements running and stop attributes and will call stop() if it is still running

https://github.com/psychopy/psychopy/blob/e27e2389c3f8385dd578de61ab4ace09cc5a95f7/psychopy/core.py#L83

Would it make more sense to also implement these attributes here so that the audio file will be saved properly whenever the experiment ends? If so, we can also enable the daemon=True parameter for this thread which makes it safer.

PR #7078, opened 2025-01-31
Update installPsychoPy.py

sjentsch:
l. 17-21: removed import os, sys (imported individually some lines below) and import requests (not used; needs otherwise to be installed first) and sorted the modules alphabetically (to make it easier to keep track)
l. 82: added minor version number and correct to python3.xx-dev (you need to update yum which is incorrect too); adding the minor version ensures that there is no mismatch between the default Python version on the system and the Python version used for psychopy (e.g., when running in a venv)
l. 84: changing the minor version for libwebkit2gtk-4.x-dev into a placeholder (e.g., Ubuntu 24.04 comes with version libwebkit2gtk-4.1-dev, and the placeholder ensures that libwebkit2gtk-4.0-dev used in earlier versions is matched too, wxPython seems to install properly with either version).

### DIFF
--- a/installPsychoPy.py
+++ b/installPsychoPy.py
@@ -14,12 +14,11 @@ shortcuts, checking/recommending virtual envs etc.
 
 # Author: Jonathan Peirce, based on work of Flavio Bastos and Florian Osmani
 
-import os, sys
-import pathlib
-import subprocess
 import os
+import pathlib
+import platform
+import subprocess
 import sys
-import requests
 
 _linux_installer = None  # will be apt-get or yum depending on system
 
@@ -80,9 +79,9 @@ def apt_install(*packages):
     def find_package(package):
         # check pacakage name according to apt/yum
         packages_lookup = {
-            'python3-dev': {'apt':'libgtk-3-dev', 'yum':'gtk3-devel'},
+            'python3-dev': {'apt':'python3.' + str(sys.version_info[1]) + '-dev', 'yum':'gtk3-devel'},
             'libgtk-3-dev': {'apt':'libgtk-3-dev', 'yum':'gtk3-devel'},
-            'libwebkit2gtk-4.0-dev': {'apt':'libwebkit2gtk-4.0-dev', 'yum':'webkit2gtk3-devel'},
+            'libwebkit2gtk-4.0-dev': {'apt':'libwebkit2gtk-4.*-dev', 'yum':'webkit2gtk3-devel'},
             'libxcb-xinerama0': {'apt':'libxcb-xinerama0', 'yum':'libxcb-xinerama'},
             'libegl1-mesa-dev': {'apt':'libegl1-mesa-dev', 'yum':'mesa-libEGL-devel'},
         }


sjentsch:
In addition, you may want to add the following lines before `# create a virtual environment and activate it` (on https://www.psychopy.org/download.html), perhaps with a comment that those apply to newer Ubuntu versions, e.g., 24.04:
```
sudo add-apt-repository ppa:deadsnakes/ppa
sudo apt update
sudo apt install python3.10 python3.10-venv
```
Thanks and cheers, Sebastian

peircej:
Thanks. I should say that there's also this by @wieluk :
https://github.com/wieluk/psychopy_linux_installer
which is rather more advanced, e.g. in the range of linux distros it can handle. I'm in favour of making that the recommended way to install to linux, although regarding the idea to encourage use of venv I think this install also doesn't do that

sjentsch:
Thanks Jon!
Just as a brief comment. I tried the script (the modified version of it, i.e. this PR) on Ubuntu 22.04 and it works there as well.
I may try https://github.com/wieluk/psychopy_linux_installer the next time I install it. Thanks for the suggestion.

wieluk:
Hi Jon (@peircej), the installer is designed to always run within a venv, with no option to install outside of one. The venv folder is automatically named based on the selected PsychoPy and Python versions: `psychopy_${PSYCHOPY_VERSION}_py_${PYTHON_VERSION}`.

peircej:
Thanks @wieluk 
I confess I haven't tried it in earnest yet, so wasn't clear on that part. I do think [the install docs](https://github.com/psychopy/psychopy/blob/dev/docs/source/download.rst) should be updated on psychopy.org to indicaet your installer

wieluk:
@peircej  I believe it's ready now. I just released 1.5.0, and I'll be pushing 1.5.1 soon. At this point, I'm confident that it works across a wide range of distributions—both those explicitly listed in the README.md and likely others (e.g., Arch).

I'm unsure about the best way to integrate the code into the PsychoPy repository. I really appreciate having full control over the runners and GitHub runner environment variables for testing. Additionally, having its own release is beneficial for providing prebuilt Python and wxPython binaries, making installations much faster.

peircej:
Sounds great!

Regarding integration that's fine. The key thing is about giving a sense of credibility, but if the PsychoPy official install docs point to wieluk installer then hopefully that's enough credibility on its own, at least for starters. We can see whether we get worried people asking if it's legitimate

So let's update PsychoPy docs. The other question is whether the install can be a single command like this without the chmod step:
``` bash
bash (curl -LOs https://github.com/wieluk/psychopy_linux_installer/releases/latest/download/psychopy_linux_installer) --gui
```
or this without the gui flag
``` bash
bash (curl -LOs https://github.com/wieluk/psychopy_linux_installer/releases/latest/download/psychopy_linux_installer)
```

(I'm thinking that this is what's done when [installing homebrew on macos](https://brew.sh/) for instance)

wieluk:
I conducted some testing and found that the one-line installation method does not consistently handle arguments. Instead, I recommend downloading and running the script separately. The script has a built-in automatic update feature and will prompt the user to update if a new version is available, ensuring that outdated script versions are not an issue.

PR #7079, opened 2025-01-31
ENH: compile *.po files when building the package

hsogo:
Because recent Standalone Installers for Windows didn't include compiled translation files (messages.mo), label strings in the applications are not translated.  I guess that `buildWinDistributions.ps1` is intended to compile *.po files when building the Standalone Installer, but  it doesn't seem to work.

In this pull request, I tried to compile *.po files when building the package.  This is achieved by putting `pdm_build.py` in the root directory of the project.  The content of this script is almost the same as `building/compile_po.py`.  The pdm backend automaticall run this script when building the package (i.e. running `python -m build .`).

I also tried to add `custom-hook = "building/compile_po.py"` to `[build-system]` of `pyproject.toml` to run `building/compile_po.py` directly, but in this case `building/compile_po.py` is included in the built package even though `excludes = ["building/"]` is in `[build-system]` of `pyproject.toml`.

### DIFF
--- a/pdm_build.py
+++ b/pdm_build.py
@@ -1,0 +1,26 @@
+import polib
+import pathlib
+
+root = pathlib.Path(__file__).absolute().parent / 'psychopy/app/locale'
+
+def compilePoFiles(root=root, errIfEmpty=True):
+    """Looks for all paths matching **/*.po and compiles to a .mo file using
+    python polib
+
+    :param: root
+    """
+    po_files = list(pathlib.Path(root).glob('**/*.po'))
+
+    for popath in po_files:
+        mopath = popath.with_suffix(".mo")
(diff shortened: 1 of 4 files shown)


peircej:
Apologies for the delayed response to the issue previously @hsogo 
Actually, the build process is driven, these days, by a separate private devops repository so that we can connect it with privately hosted Github Runners and domain secrets. The setupApp script(s) need cleaning up which I'll try and do this week.

You were right that I had missed the step of rebuilding the mo files in the devops script though and that has now been added in

For the commit you add here, I don't actually use `pdm build` to make the standalone apps, but I'll look into whether that will also help the system. Happy to include the file here though on the basis that some 3rd-party developers may use that in their pipeline and will therefore benefit

PR #7123, opened 2025-02-21
DOC: Correct zh_CN translation for user feedback message

myd7349:
This patch corrects translation from '练习' to '联系' for user feedback message.

- 练习 (liànxí): This word generally refers to "practice" or "exercise."
- 联系 (liánxì): This word means "contact" or "communication."

They share the same pinyin, so it's possible that this error was introduced when using a pinyin input method.


TEParsons:
Awesome, thanks for catching this :)

PR #7153, opened 2025-03-07
BF: translation of the Preference Dialog was broken

hsogo:
Translation of the Preference dialog was broken probably since `_localized` was removed from preferencesDlg.py.
I restored `_localized` from an old version of preferencesDlg.py, added new strings to it, and moved it to localizedStrings.py.

While doing this, I noticed that localizedStrings.py is jast a collection of strings now and the code importing `_localized` from `psychopy.localization` and updating it doesn't work.  In fact, `psychopy.localization._localized` no longer exists and the Builder components are not where the code expects them to be.  So I removed these unnecessary codes from localizedStrings.py.

Probably for historical reasons, localizedStrings.py is in psychopy/app/builder/.  However, now this file includes not only Builder strings but also Coder and Runner strings.  Because this file is not imported by other modules in PsychoPy, we can safely move this file to other location.  I think that psychopy/app/ is a better location (I haven't moved it yet).  What do you think?


peircej:
Hi @hsogo yes, I agree. I think part of this comes from refactoring to get localization out of the core psychopy into the app, but I think it hasn't gone far enough. We're trying to reduce inter-dependency of the app and the lib because a) some people only want the lib and write their own scripts form vscode and b) the app will load faster if it doesn't rely so heavily on the lib.

hsogo:
I've moved localizedStrings.py to psychopy/app/.  It's nice to hear you are trying to speed up loading the app because it takes a long time on low-spec machines that some of my students have.

psychopy/app/preferencesDlg.py
```diff
@@ -121,7 +121,7 @@ def addPage(self, label, name, sections=(), bitmap=None):
                                     list(sections))
         self.pageNames[name] = self.pageIdx
         self.lstPrefPages.InsertItem(
-            self.lstPrefPages.GetItemCount(), label, self.pageIdx)
+            self.lstPrefPages.GetItemCount(), _translate(label), self.pageIdx)
```
TEParsons: Wouldn't this essentially be translating twice? As the label is drawn from `_localized` so is already translated

TEParsons (changes requested):
Not so much changes as just clarification - see my comment in this review. Overall looks good so happy to pull in either way, just want to minimise duplication if possible

psychopy/app/preferencesDlg.py
hsogo: Hi @TEParsons ,
Thanks to you I realized that `addPage()` receives trasnlated labels on call in the current version.  For this fix, I referred to the version 2023.2.3, where `addPage()` received "raw" labels and had to translate them.  Either of translating before or after calling `addPage()` would be fine.  Which do you think is better?

hsogo: I found some label changes and fix them.

TEParsons: I think before makes the most sense :)

TEParsons (approved):

psychopy/app/preferencesDlg.py
hsogo: I see.😄 I've removed the `_translate()` in line 124.

Issue #7173, opened 2025-03-23
[Bug]: Only one dependent Param of type "choice" can be dynamically populated


rlaboiss:
### PsychoPy Version

2024.2.5

### What OS are your PsychoPy running on?

Linux-based systems

### Bug Description

I am trying to create a new routine for PsychoPy with three parameters, say `Foo`, `Bar`, and `Baz`. The three parameters have `inputType ` equal to `choice`. In my design, the lists of choices of parameters  `Bar` and `Baz` should be dynamically repopulated when parameter `Foo` is changed. However, it is not possible to implement this behavior.

### Expected Behaviour

The expected behavior is described in the example below. Note that both parameters `Bar` and `Baz` have a `dependsOn` field set to `Foo`. So, when the value of `Foo` changes, then the lists of choices for **both** `Bar` and `Baz` must be repopulated.

### Steps to Reproduce

1. Put the following in the `__init__` method of a new routine class (derived from `BaseStandaloneRoutine`):

``` python
        self.params["foo"] = Param(
            "",
            valType="str",
            inputType="choice",
            categ="Basic",
            allowedVals=["1", "2"],
            label="Foo"
        )

        def get_values():
            return [f"{self.params['foo'].val}-{x}" for x in range(2)]

        self.params["bar"] = Param(
            "",
            valType="str",
            inputType="choice",
            categ="Basic",
            allowedVals=get_values,
            label="Bar"
        )

        self.depends.append({
            "dependsOn": "foo",
            "condition": "",
            "param": "bar",
            "true": "populate",
            "false": "populate",
        })

        self.params["baz"] = Param(
            "",
            valType="str",
            inputType="choice",
            categ="Basic",
            allowedVals=get_values,
            label="Baz"
        )

        self.depends.append({
            "dependsOn": "foo",
            "condition": "",
            "param": "baz",
            "true": "populate",
            "false": "populate",
        })
```

2. Add that routine to an experiment.
3. Open the experiment in Builder.
4. Navigate to the `Basic` panel.
5. Change the value of the `Foo` parameter.
6. Only the list of choices for the `Bar` parameter will be repopulated. The list of choices for the `Baz` parameter will remain unchanged.

### Additional context

[...]

PR #7174, opened 2025-03-23
BF: Fix conflict when a Component has multiple parameters with live-populated dropdowns

rlaboiss:
This commit is intended to address issue #7173. In sum, the bug happens when there are two parameters that have inputType = "choice" and have also declared dependsOn relationships with a third parameter.

In the current state of the code, only one of the two dependent parameters has its list repopulated. This happens because, in the loop over the declared dependencies (self.parent.element.depends) for a given element (in method CategoryPage.checkDepends), the value of dependencyParam.val is prematurely set.

With this commit, the repopulation, as well as the setting of dependencyParam.val, are delayed after the for loop completes.

I hope that this solutions will be acceptable to the upstream authors.

### DIFF
--- a/psychopy/app/builder/dialogs/__init__.py
+++ b/psychopy/app/builder/dialogs/__init__.py
@@ -657,6 +657,9 @@ class ParamNotebook(wx.Notebook, handlers.ThemeMixin):
              "false": "Disable",  # permitted: hide, show, enable, disable
             }"""
             isChanged = False
+            # This list will be used to delay the populate actions on the
+            # dependent params
+            populateCtrls = []
             for thisDep in self.parent.element.depends:
                 if not (
                         thisDep['param'] in list(self.ctrls) + ['start', 'stop']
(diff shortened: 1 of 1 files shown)


peircej:
@TEParsons from a quick look, this seems good to me, but it's your code to know the history 😃

TEParsons:
Well caught, yep this fix all looks good - I've just pulled down to test and it seems to work great. Thanks @rlaboiss!

rlaboiss:
Thanks for merging, @TEParsons !

I see at least one piece of code in PsychoPy that will immediately benefit from the fix, It is in file `psychopy/experiment/components/camera/__init__.py`, class `CameraComponent`, where the parameters `resolution` and `frameRate` both depend on the parameter `device`:
 
https://github.com/psychopy/psychopy/blob/68c2850703458ada61b934baa08e3a6fee1678ae/psychopy/experiment/components/camera/__init__.py#L240-L287

PR #7207, opened 2025-04-22
ENH: Allow overriding user preferences directory with PSYCHOPY_CONFIG_FOLDER environment variable

wieluk:
This pull request changes `psychopy/preferences/preferences.py` to let users override PsychoPy’s config folder by setting `PSYCHOPY_CONFIG_FOLDER`. The new logic checks that the override exists and is readable/writable. It currently applies only to non‑Windows systems (it would be easy to enable on Windows, I’m just not sure if we want to). It works with:

```bash
export PSYCHOPY_CONFIG_FOLDER=/path/to/config
psychopy
```

but the `monitors` subfolder still defaults to `~/.psychopy3/monitors`. The code doesn’t issue a message on invalid/succesfull overrides. Adding one could be a useful follow‑up.

### DIFF
--- a/psychopy/app/psychopyApp.py
+++ b/psychopy/app/psychopyApp.py
@@ -95,9 +95,11 @@ depends on the type of the [files]:
     )
     # parse args
     args, startFilesRaw = argParser.parse_known_args(sys.argv)
-    
-    # setup prefs
+
+    # import app and prefs now that args have been parsed
     from psychopy.preferences import prefs
+    from psychopy.app import startApp, quitApp
+    # setup prefs
     prefs.getPaths(userDir=args.userDir)
     # insert fallbacks from prefs for unsupplied call args
     if args.showSplash is None:
--- a/psychopy/app/psychopyApp.py
+++ b/psychopy/app/psychopyApp.py
@@ -19,9 +19,6 @@ import psychopy.locale_setup  # noqa
 
 
 def main():
-    from psychopy.app import startApp, quitApp
-    from psychopy.preferences import prefs
-
     # parser to process input arguments
     argParser = argparse.ArgumentParser(
         prog="PsychoPyApp", description=(
@@ -80,7 +77,7 @@ depends on the type of the [files]:
     )
     # add option to hide splash
     argParser.add_argument(
-        "--no-splash", dest="showSplash", action="store_false", default=prefs.app['showSplash'], help=(
+        "--no-splash", dest="showSplash", action="store_false", default=None, help=(
             "Suppresses splash screen"
         )
     )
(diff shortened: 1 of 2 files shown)


wieluk:
I’ve already created a corresponding branch in psychopy_linux_installer that would work with this implementation of the config folder, if this PR is accepted:
[wieluk/psychopy_linux_installer@start-wrapper](https://github.com/wieluk/psychopy_linux_installer/tree/start-wrapper)

wieluk:
I would have preferred to add a `--config-folder=...` flag instead of relying on an environment variable. However, this approach seems more complicated, since the config folder is needed before command-line arguments are parsed.

peircej:
@TEParsons @mdcutone this seems sensible to me (for other OSs as well as linux, actually)

It would mean, for instance, that we can launch a psychopy process with a deliberately different set of preferences by setting that as an env var while launching the process

TEParsons:
> I would have preferred to add a `--config-folder=...` flag instead of relying on an environment variable. However, this approach seems more complicated, since the config folder is needed before command-line arguments are parsed.

This was my first thought too - I think it's workable still as the only thing `prefs` is needed for in the arg parsing stage is a default for `showSplash`. We could have that default to None, then import and setup `prefs` after parsing args, then say `if args.showSplash is None: args.showSplash = prefs.app['showSplash']`.

I have an idea of how this could look so I'll submit a PR to your branch :)

TEParsons:
https://github.com/wieluk/psychopy/pull/1

wieluk:
Thanks for your help @TEParsons!

However, with your version, the .psychopy3 folder is still being created in the home directory when PsychoPy starts, even when I specify a different directory using `--user-dir=/any/other/folder`.

As it stands, the folder ends up being created in both locations.
This also results in a strangely split configuration: for example, if I change the theme in Preferences and then delete the .psychopy3 folder in HOME, the theme resets to default on the next launch — even though that preference should have been saved in `/any/other/folder`.

It seems like PsychoPy is still relying on the home directory for certain settings despite the `--user-dir` override.

TEParsons:
Ah, I see what you mean... I was testing it by checking that `self.paths['userPrefsDir']` was populated correctly, so I missed that the actual prefs weren't changed. I think it's because `loadAll` is called when prefs is initialised, but isn't called again when we do `getPaths` with the new directory. 

Maybe rather than calling the two separately in `__init__`, we could have `getPaths` be called inside `loadAll` & call that when setting the custom path? I'll update the PR (and test it properly this time!)

wieluk:
I've updated the PR. It appears to be functioning correctly now. However, it still creates a `.psychopy3` directory in both locations. Under `/home/.psychopy3`, everything except the `monitors` and `themes` folders is now empty in `/home`. Preventing this behavior would likely require a significant refactor of the preferences.

TEParsons:
I think that's okay :) It'll be creating the folder when initially importing the class, but so long as it doesn't use that folder when told not to it's fine for it to exist.

Issue #7206, opened 2025-04-17
[Feature Request]: Automatically update psychopy/app/locale/**/*.po files and ping translators


peircej:
### Description

We aren't currently auto-updating the locale/**/messages.po files so translations get out of date (see example below). 

![Image](https://github.com/user-attachments/assets/9584bff4-2786-4952-8729-155e7dc4686f)

### Purpose

Keep translations up to date by notifying translators of updated `_translate()` strings

### Describe the solution you would like

Ideally we would
1. run a script to update *.po files automatically after a successful merge to dev
2. if the *.po files have changed then create a new PR with the updated files
3. notify known translators when the entries have changed so they know to update the translations

### Describe alternatives that you have considered

As a first step we could simply have a script that updates all the **/messages.po files even if this doesn't operate from a GH action

peircej:
I **think** all the above steps are doable:

1. @hsogo I have a memory that you once wrote a script using python gettext to update the po files, so you solved this already, but I've lost track of where that went
2. a new GH action that follows push on dev or release should be easy enough
3. we could git add all and then detect whether anything got added? Create the PR using [create-pull-request](https://github.com/marketplace/actions/create-pull-request) action
4. If the commit message says `"@psychopy/translators this commit may have introduced new/changed phrases"` then we should be able to create a "team" called translators that contains the user id's of relevant known contributors

peircej:
The @psychopy/translators team already exists, and I've just added some people that we know have contributed in the past) so this part is begun

peircej:
Ah, I found the script
https://github.com/psychopy/psychopy/blob/dev/psychopy/localization/generateTranslationTemplate.py

peircej:
OK, so the script [generateTranslationTemplate.py](https://github.com/psychopy/psychopy/blob/dev/psychopy/localization/generateTranslationTemplate.py) written by @hsogo does generate the .pot file but then we would need a step to update the various `.po` files from the `.pot` which I think could be done with

```
pybabel update -i messages.pot -o ../app/locale/<LOCALE HERE **>/LC_MESSAGE/messages.po --width 79 -l <LOCALE HERE **> --ignore-obsolete --previous
```
I'm using the tags 
- `--ignore-obsolete` so that it won't delete unused translations (I don't see the need to purge but willing to be pursuaded)
- `--previous` so that previous entries (e.g. comments in the po file) are not changed
[...]

PR #7230, opened 2025-05-07
ENH: update PO and POT files by a script

hsogo:
generateTranslationTemplates.py was modified to updates PO files in addition to POT file.
With -c option, the script makes a commit if any files are updated.
Please see also #7206.

Quesition:
- The script commits the changes with a title of "ENH: Translation template is updated" (see afc0f4d42a0193bbe38881096f0ff385b0abb78d for example: this commit was made by the script).  "ENH" category is appropriate?
- The script updates `Project-Id-Version` (PsychoPy version is stored) and `POT-Creation-Date` of all PO files.  Is it Ok?  Are there other metadata that should be updated?


peircej:
oh @hsogo can your script change the wrap width to 79? It used the babel default of 76, by the looks, but because poedit defaults to 79 it means a lot of the apparent changes in the git file are actually just re-wraps.

hsogo:
@peircej 
I modifed the scripts to set wrap width 79.  Since this pull request has been closed, do I need to make a new pull request?

By the way, in my environment (Windows11, Poedit free version 3.6.1, "wrap at:"=79), text is not wrapped even if "wrap at:" option is set in Poedit.  For example, a long sentence in messages.po generated by my script is wrapped as following:

```
#: ../app/connections/updates.py:344
#, python-format
msgid ""
"You are running PsychoPy (%(running)s), which is ahead of the latest official"
" version (%(latest)s)"
msgstr "PsychoPy %(running)s を実行中です 公式の最新バージョン(%(latest)s)より新しいバージョンです"
```
After editing this messages.po with Poedit, this entry is not wrapped as following:

```
#: ../app/connections/updates.py:344
#, python-format
msgid "You are running PsychoPy (%(running)s), which is ahead of the latest official version (%(latest)s)"
msgstr "PsychoPy %(running)s を実行中です 公式の最新バージョン(%(latest)s)より新しいバージョンです"
```

**EDITED at 2025/07/09**
I wrote that the "Preserve formatting of existing files" option of the Poedit might be involved, but I'm not sure the exact solution at this time.

PR #7333, opened 2025-07-24
FF: Don't poll camera mic if it's closed

TEParsons:


peircej:
@mdcutone @TEParsons I've pulled this in (tests have passed) before review so it can be tested in the build (meaning it does still need Matthew's eyeball but I want it building)

PR #7336, opened 2025-07-24
FF: Fixes for MovieStim in builder experiments

mdcutone:
Fixes the issue where `MovieStim` does not correctly reset its internal state when movies are played in sequence from the same instance


psychopy/experiment/components/movie/__init__.py
```diff
@@ -281,15 +281,16 @@ def writeFrameCode(self, buff):
...
+            # if self.params['backend'].val not in ('moviepy', 'avbin', 'vlc'):
+            #     code += "%(name)s.stop()\n"
+            # buff.writeIndentedLines(code % self.params)
         # to get out of the if statement
         buff.setIndentLevel(-indented, relative=True)
 
         # do force end of trial code
         if self.params['forceEndRoutine'].val is True:
-            code = ("if %s.isFinished:  # force-end the Routine\n"
+            code = ("if %s.status == FINISHED:  # force-end the Routine\n"
```
peircej, line 293: My understanding of this was that `isFinished` is the status set by the movie itself (e.g. at EOF) whereas `status==FINISHED` refers to stuff set by the Builder script. Is that not the case anymore? Does the movie itself change its status field?

peircej (approved):

psychopy/experiment/components/movie/__init__.py
TEParsons: That's my understanding too, so this would mean the routine only ends when the component reaches its end condition and not when the file is finished?

TEParsons: Ah actually, Builder sets `movie.status = FINISHED` if `movie.isFinished` from its each frame code, so this is fine :)

PR #7344, opened 2025-07-31
BF: MacOS related fixes for Camera recordings

mdcutone:
Fixes issues around camera enumeration/selection via device manager and incorrect timestamps being generated by the codec. Still an issue around the experiment stalling when shutting down with the camera remaining open, however recordings are being saved with the correct duration

I also improved how textures are processed by `ImageStim` allowing for objects exposing the `colorTexture` attribute to pass data over without creating a texture within the ImageStim class

This PR needs to be tested on Windows machines to see if the changes are compatible

### DIFF
--- a/psychopy/hardware/camera/__init__.py
+++ b/psychopy/hardware/camera/__init__.py
@@ -474,7 +474,7 @@ class CameraDevice(BaseDevice):
             for profile in self.getAvailableDevices(False):
                 if profile['device'] == device:
                     foundProfile = profile
-                    device = profile['device']
+                    device = profile['deviceName']
                     break
         elif isinstance(device, str):
             # if device is a string, use it as the device name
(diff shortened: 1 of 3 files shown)


psychopy/hardware/camera/__init__.py
```diff
@@ -579,7 +579,7 @@ def isSameDevice(self, other):
         elif isinstance(other, Camera):
             return getattr(other, "_capture", None) == self
         elif isinstance(other, dict) and "device" in other:
-            return other['device'] == self._device
+            return other['deviceName'] == self._device
```
TEParsons, line 582: Wouldn't this mean that `getProfile` returns the first profile matching the device name rather than the specific profile (resolution and all) which we want?

TEParsons (changes requested):

TEParsons (approved):

PR #7348, opened 2025-08-01
FF: More fixes for camera recordings on MacOS

mdcutone:
Some additional fixes for camera recordings on MacOS. Uses FFMPEG directly for saving output video files. Still have an issue where the microphone fails to close when `Camera.close()` is called, that's under investigation


TEParsons (approved):

TEParsons:
I'm gonna pull this in and start a build so we can get it tested ASAP, any bugs we can attend to in a separate PR

PR #7352, opened 2025-08-06
DOC: Update stable/beta release, and deprecate Compatability+

TEParsons:


peircej (commented):
Certainly we can remove the Compatibility+ option for 2025.2.0 (because we haven't built one) but for 2025.1.1 it is still available so we may as well leave that option available as a button? Or does it mess up the layout for one to have it and the other not?

peircej:
I'm just thinking that in future we can then point to this release and say 

"The `useVersion` feature in this release support PsychoPy versions back to 2023.1.x. If you need `useVersion` support for studies in older releases than that you should fetch the `Stable 2025.1.1 Compatibility+` installer"

TEParsons:
The layout can handle it, so I'll add it back in for 2025.1.1 & will also put a link in the section below about compatibility installers, which can stay there to point people who need py3.8 to the last compatibility installer after 2025.2.n becomes the stable release.

PR #7374, opened 2025-08-21
BF: ExperimentHandler not saving extraInfo to wide text

davidmwatson:
I recently updated my PsychoPy, and noticed the `ExperimentHandler` was no longer saving the fields in the `extraInfo` dictionary to the wide text/csv file. I think this is due to a recent change where the `extraInfo` gets stored as PsychoPy's new `IndexDict` object. This was causing checks for the type of the `extraInfo` object to fail, as it isn't a normal Python dictionary any more. Seeing as `IndexDict` inherits from `dict`, just changing the checks to use `isinstance` rather than `type == dict` seems to fix the problem.


TEParsons:
Seems a good solution, thanks for catching this!

PR #7381, opened 2025-09-02
BF: Fix view and projection matrices not being applied before first flip

mdcutone:
Fixes the issue where visual stimuli appear incorrectly scaled prior to the first flip call. This was caused by the projection matrix for 2D rendering being incorrect until cleared and reset by a flip call. How the default ortho projection matrix is computed has been changed, no longer accounting for window dimensions since that's dealt with by the stimuli's vertex data container

I added method `setDefaultView()` to put matrices in the correct state for PsychoPy's 2D stim classes and now have `setOrthographicView` separate for those who wish to scale coordinates that way

### DIFF
--- a/psychopy/visual/window.py
+++ b/psychopy/visual/window.py
@@ -625,7 +625,7 @@ class Window():
         self._projectionMatrixNeedsUpdate = True
         self._viewMatrixNeedsUpdate = True
 
-        self.setOrthographicView()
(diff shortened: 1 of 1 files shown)


TEParsons (approved):
I'm not sure how to test the bug this fixes but the changes seem sound and don't appear to affect running normally so I'm happy to pull in :)

PR #7409, opened 2025-09-26
BF: Importing `psychopy.app` when writing scripts was slowing things down and breaking Use Version

### LINKED ISSUE #7406: UseVersion shouldn't import the app - now breaks due to dependency on pypi-search
The `experiment` lib used to live in the app but it shouldn't. It's needed to load/write experiments but we don't need all the app machinery for that! That means:
- we were loading way more files than were needed (the app machinery like wxpython shouldn't be needed to write a script!)
- it's now broken due to the missing pypi-search which the app needed in 2024 but isn't needed to write an experiment file

This should be fixed by the below line from:
`from psychopy.app.builder import experiment`
to be
`from psychopy import experiment`

https://github.com/psychopy/psychopy/blob/1bce77fd63c23ffaa7a6293217054c9e8f12623d/psychopy/scripts/psyexpCompile.py#L146

TEParsons:
Fixes #7406

### DIFF
--- a/psychopy/scripts/psyexpCompile.py
+++ b/psychopy/scripts/psyexpCompile.py
@@ -143,7 +143,7 @@ def compileScript(infile=None, version=None, outfile=None):
             The experiment object used for generating the experiment script
         """
         # import PsychoPy experiment and write script with useVersion active
-        from psychopy.app.builder import experiment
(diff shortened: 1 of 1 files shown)


peircej (approved):
Yes, all good

PR #7414, opened 2025-09-29
BF: Allow multiple joysticks when using the Pyglet backend

### LINKED ISSUE #7351: [Bug]: Two Joysticks, getting no data from frist one
### PsychoPy Version

2025.1.1

Linux-based systems

### Bug Description

Unfortunately using two joysticks there are no data  for the first joystick (device=0) e.g.  (x,y axes),
second one works fine. Both devices are opened correctly with different device ids, initialization is OK
Looks like a mismatch with the  inputMappings 
Testing from the command-line using  jstest simultaneously  both joysticks are working fine

OS: Linux Mint  
lsusb:
Bus 001 Device 026: ID 044f:0406 ThrustMaster, Inc. T.A320 Copilot
Bus 001 Device 027: ID 044f:0405 ThrustMaster, Inc. T.A320 Pilot

### Expected Behaviour

I would expect that psychopy can handle more than one joystick device.
A work around might be to poll the devices outside the pyglet event loop, 
closer to the evdev implementation.

### Steps to Reproduce

1 apply jstest for both joys
2. switching the order of joysticks and reconnect
3. example code: https:// jugit.fz-juelich.de/f.boers/psychopy-test
...

mdcutone:
Changes allow multiple joysticks to be polled in parallel when using the pyglet backend. Tested using a Thrustmaster HotasX and Xbox gamepad

Fixes #7351


psychopy/hardware/joystick/backend_pyglet.py
```diff
@@ -125,6 +118,17 @@ def open(self):
         self._device.open()
         self._isOpen = True
 
+        # register with pyglet event loop
```
TEParsons, line 121: How come this is in `getAvailableDevices`? This function is called when getting a list of devices in Device Manager from Builder so there often won't be a window open

TEParsons (changes requested):

psychopy/hardware/joystick/backend_pyglet.py
mdcutone: Looked over this today and I'm not sure what you mean here, the code you're referring to is not within the `getAvailableDevices()` routine, but rather the `open()` method which it is required

TEParsons (approved):

PR #7425, opened 2025-10-09
BF: Don't import audio backend until needed

### LINKED ISSUE #7397: [Bug]: device manager crashes psychopy on mac 15.7
### PsychoPy Version

2025.2.0

macOS Silicon

### Bug Description

When I open a new experiment in 2025.2.0 and try to add a device in Device Manager, psychopy immediately crashes. I get the attached termination message if that helps.

[psychopy_device_erorr.txt](https://github.com/user-attachments/files/22453904/psychopy_device_erorr.txt)

### Expected Behaviour

Open device manager, click add device, select device

### Steps to Reproduce

1. Open psychopy
2. Select Device Manager
3. Select 'Add Device'
4. Psychopy closes
...

### Additional context

_No response_

TEParsons:
Fixes #7397

Rather than importing sound backends to check whether they're available, use plugin entry points to check available & import from those entry points when a Sound object is initialised.


psychopy/sound/sound.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+        Get all available Sound backends (by name)
+
+        Returns
+        -------
+        dict[str:importlib.metadata.EntryPoint]
+            Dict mapping backend names to backend entry points - call `.load` on an entry point to 
+            import the relevant module.
+        """
+        # start off with builtin backends
```
mdcutone, line 34: I would import and catch errors here for each backend since they might be removed in the future to plugins.

mdcutone:
@TEParsons We should also consider cherry-picking this into the dev branch

PR #7434, opened 2025-10-16
Correctly get the number of joystick devices

rlaboiss:
The method `getNumJoysticks()` of the class `hardware.joystick` has been removed from file `psychopy/hardware/joystick/__init__.py` in commit edca46d4b15a91091339296734950f416e7e0124. However, the code in the file `experiment/components/joystick/__init__.py` (injection in the `_lastrun.py` file) has not been updated accordingly. This causes any experiment that includes the joystick component to fail, displaying the error message `No joystick/gamepad device found`. The commit in this PR seems to fix the problem.

It also resolves a similar issue in the `experiment.components.joyButtons` class.

### DIFF
--- a/psychopy/experiment/components/joyButtons/__init__.py
+++ b/psychopy/experiment/components/joyButtons/__init__.py
@@ -132,7 +132,7 @@ class JoyButtonsComponent(BaseComponent):
         buff.writeIndentedLines(code % self.params)
 
         buff.setIndentLevel(+1, relative=True)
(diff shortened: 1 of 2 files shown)


TEParsons:
This seems good to me, we're using `getAvailableDevices` across various other devices so this seems consistent. Adding @mdcutone as a reviewer as he knows most about the Joystick class.

mdcutone (approved):

PR #7457, opened 2025-10-31
BF: Fixed param control text color selection when using system dark theme

mdcutone:
Adds system wide UI theming detection (whether we're using light/dark mode) to ensure parameter text remains legible in either setting.


psychopy/app/builder/dialogs/paramCtrls.py
```diff
@@ -39,6 +39,16 @@
 EVT_PARAM_CHANGED = wx.PyEventBinder(wx.IdManager.ReserveId())
 emptyNamespace = NameSpace(experiment.Experiment())
 
+# get if dark theme
+_appearanceSettings = wx.SystemSettings.GetAppearance()
```
TEParsons: This seems to be breaking the test suite as it's querying wx before there's an app. I think it would make sense to move this code to a function (`getDarkMode`?) somewhere in utils, then import and call it when needed (e.g. in `styleValid` below)

TEParsons (changes requested):

psychopy/app/builder/dialogs/paramCtrls.py
mdcutone: Addressed this by using the dark mode flag in the app's main instance

TEParsons (approved):

PR #7460, opened 2025-11-05
BF: If Pavlovia API can't find a user, use pygitlab to get their user ID

TEParsons:


peircej (changes requested):
That's good but if we *still* don't get a gitlabId then we still want to raise an error here rather than being puzzled later. So still do the `assert` but do it after the gitlab users search

peircej (approved):

PR #7462, opened 2025-11-05
BF: Fixed `Mouse.getPos()` returning `float64` typed objects instead of `float`

mdcutone:
Mouse coordinates are now once again returned as builtin Python `float` types instead of NumPy `float64` objects. This mainly affected how values were written to the data file, appearing as `np.float64(<value>)` instead of just the value.


TEParsons (changes requested):
From the error messages in the test suite, I think you need to catch when `self.lastPos` is `None`

mdcutone:
Seems like the tests assume that they're getting numpy objects by performing vectorized math on them, I'll need to update all those now

PR #7469, opened 2025-11-13
DOC: Add demo of dual window design to Builder demos

kimDundas:
Added to Design Templates Builder demos folder


TEParsons:
@kimDundas tests are failing on this with the following message:
```
AssertionError: Builder demos should be saved in Pilot mode, so that they start off piloting when users unpack them. Please change run mode on dual_window.psyexp
```
If you switch to pilot mode and save, then commit and push again, tests should pass :)

PR #7475, opened 2025-11-19
BF: clamp StairHandler intensity when step size is negative

### LINKED ISSUE #7188: [Bug]: StairHandler minVal and maxVal will be ineffective when set stepSizes to negative
### PsychoPy Version

2024.2.5

Linux-based systems

### Bug Description

see following

### Expected Behaviour

 see following

### Steps to Reproduce

```python
  from psychopy import data

  handler = data.StairHandler(
      startVal=0,
      minVal=0,
      stepType="lin",
      stepSizes=-1,
  )
  i = 20
  for e in handler:
      handler.addResponse(0)
      print(e)
      i = i - 1
      if i <= 0:
          break
```

It should not print negative numbers, but 0.

### Additional context

_No response_

stephenmjerge:
Fixes #7188

## Summary
- enforce `StairHandler` min/max clamping regardless of step direction, covering negative step sizes
- add regression test to ensure negative `stepSizes` with `minVal` do not drop below bounds

## Testing
- python -m pytest psychopy/tests/test_data/test_StairHandlers.py::TestStairHandler::test_negative_step_clamped_to_min

### DIFF
--- a/psychopy/data/staircase.py
+++ b/psychopy/data/staircase.py
@@ -413,9 +413,7 @@ class StairHandler(_BaseTrialHandler):
             self._nextIntensity *= 10.0**self.stepSizeCurrent
         elif self.stepType == 'lin':
             self._nextIntensity += self.stepSizeCurrent
-        # check we haven't gone out of the legal range
-        if (self.maxVal is not None) and (self._nextIntensity > self.maxVal):
-            self._nextIntensity = self.maxVal
+        self._clampIntensity()
         self.correctCounter = 0
 
     def _intensityDec(self):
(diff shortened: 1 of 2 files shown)


stephenmjerge:
Thanks for reviewing! I checked the failing Ubuntu CI job and it appears
unrelated to this PR.

The failure occurs during test collection in:

    psychopy/tests/test_app/conftest.py

With the following error:

    pytest.PytestRemovedIn9Warning: Marks applied to fixtures have no effect

It looks related to pytest 9’s change around fixture marking rather than the
StairHandler code. My patch only touches StairHandler and its tests.

Happy to adjust anything or help with the test config if that would be useful—
just wanted to flag that the CI failure seems to be from existing test marks on
Linux rather than this bugfix.

peircej:
Yes, we're seeing this across the board right now and working on the pytest update to fix it.
Thanks, we'll pull in once we've fixed the test suite 👍

TEParsons:
The test suite is fixed now (as of #7481) so if you pull down dev, merge in and push again tests should pass for you :)

stephenmjerge:
Rebased onto latest dev (pytest fix) and re-ran targeted StairHandler test locally, pushing to rerun CI.

PR #7490, opened 2025-12-03
EN: Object detection in camera streams

mdcutone:
This PR adds object detection capabilities to the camera interface. The user can define classifiers and pass them to an object detection method which returns data about any detected objects (position, size, etc.)

This PR also adds a new `CameraFrame` object which encapsulates frame data and contains the new `detectObjects()` method. To perform object detection, first create a new classifier:

```
faceDetector = imagetools.HaarCascadeObjectRecognizer('haarcascade_frontalface_default.xml', name='face')
```

Then pass the classifier to the `detectObjects()` method of desired frame to detect objects:

```
# using the most recent frame (i.e. lastFrame)
if camera.lastFrame is not None:
    facePos = camera.lastFrame.detectObjects(faceDetector)
````

Returns the results as such:

```
{'face': {'pts': 16.064490300006582, 'count': 1, 'objects': [{'rect': (31, -167, 151, 151), 'center': (31, 16), 'bbox': ((271, 153), (422, 153), (422, 304), (271, 304))}]}}
```

This also supports batch object detection which allows the user to specify a set of classifiers and pass them all at once to `detectObjects` as a list. This is more efficient than calling the method multiple times since it avoids multiple color conversions.

### CODE REVIEW

**Review by TEParsons — APPROVED** [review]
All looks good and seems to be working great! For the record this is the code I used to test it:
```
from psychopy import visual
from psychopy.tools import imagetools
from psychopy.hardware import camera
import time

# setup camera
for profile in camera.CameraDevice.getAvailableDevices():
    cam = camera.Camera(profile['device'])
    break
# setup detector
faces = imagetools.HaarCascadeObjectRecognizer('haarcascade_frontalface_default.xml', name='face')
# setup window
win = visual.Window(checkTiming=False)
# setup tracking dot
dot = visual.Rect(win, fillColor="red", units="norm")

# start camera
cam.start()
# start frame loop
start = time.time()
while time.time() - start < 10:
    # get camera frames
    cam.poll()
    # do detection
    if cam.lastFrame:
        found = cam.lastFrame.detectObjects(faces)
        # if we found a face...
        if found:
            if len(found['face']['objects']):
                # get size and position
                pos = found['face']['objects'][0]['center']
                size = found['face']['objects'][0]['size']
                # set dot pos (adjusting for unit space)
                dot.pos = [
                    (i * 2) / total
                    for i, total in zip(pos, cam.lastFrame.frameSize)
                ]
                # set dot size (adjusting for unit space)
                dot.size = [
                    (i * 2) / total
                    for i, total in zip(size, cam.lastFrame.frameSize)
                ]
                # draw dot
                dot.draw()
    # draw to win
    win.flip()
    # sleep so other stuff can happen
    time.sleep(0.01)
# stop camera
cam.stop()
```

and the dot on screen seemed to move (albeit flipped horizontally) in the same way I moved my head around in front of the camera.

PR #7510, opened 2025-12-27
DOCS: add validator notes to timing page and details on light sensor

RebeccaHirst:
Added details about validator components to the general > timing page and also added details on using light sensors.

### DIFF
--- a/docs/source/builder/components/VisualValidatorRoutine.rst
+++ b/docs/source/builder/components/VisualValidatorRoutine.rst
@@ -7,7 +7,21 @@
 Visual Validator Routine
 -------------------------------
 
-Use a light sensor to confirm that visual stimuli are presented when they should be.
+Use a light sensor to confirm that visual stimuli are presented when they should be. Before using a validator component in your experiment it is important to ensure that you have a photodiode device available for the routine to use. If you do not have a photodiode to hand you can use an "emulator" temporarily until you have access to a real device.
+
+To configure your photodiode with an emulator go to Device Manager > Add device > Screen Buffer Sampler (Debug) > Light Sensor Emulator. Alternatively if you have a photodiode connected you can search for the name of your actual device in the list of available devices.
+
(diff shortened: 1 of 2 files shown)


docs/source/builder/components/VisualValidatorRoutine.rst
```diff
@@ -7,7 +7,21 @@
...
+
+In the "Device" section of the component you can select the name of your photodiode device from the dropdown list. If you have multiple photodiodes connected to your system you may also need to specify a channel number to distinguish between them.
+
+By default "Find best threshold?" and "Find sensor?" " are both set to True. This means that when you run your experiment the visual validator routine will first run a brief routine to find the best threshold for the light sensor, followed by another brief routine to find the size and position of the light sensor on the screen. Once these two brief routines have been run the main visual validator routine will run for the remainder of your experiment. 
+
+Recommended debugging steps for using the visual validator routine in your experiment:
+
+* Use an external piece of software to check that your computer can detect the external hardware. For example, if using Cedrus you can install `Xidon <https://cedrus.com/support/xid/xidon.htm>`_ to check that your computer can detect the device.
```
TEParsons, line 22: Might be worth mentioning here something like "External software can sometimes block PsychoPy from accessing the device while open, so make sure to close it before setting up the device in PsychoPy."

(otherwise I'm worried about getting loads of forum posts like "PsychoPy can't detect my Riponda" because they've checked it with Xidon and then left Xidon open)

TEParsons (changes requested):
Just one change, see comment below, otherwise looks good :)

docs/source/builder/components/VisualValidatorRoutine.rst
peircej: I've added this comment in the [psychopy-docs](https://github.com/psychopy/psychopy-docs) repo in this commit:
https://github.com/psychopy/psychopy-docs/commit/89a6b99c583eab709ad8e82569228d62a295c57b

PR #7513, opened 2026-01-05
BF: Fix key error & missing refresh token when logging in from saved details

TEParsons:


peircej:
Will users know what "logging in again from scratch" means? Is it something we can do/trigger on their behalf rather than needing them to go and find dialogs after this message?

PR #7518, opened 2026-01-14
DOC: fix hyperlink to match rst format

Emil-Ant:
### What 
- Use the the correct formatting for `.rst` hyperlinks for text on the `/developers/testsuitecontributing.html` 

### Why
- It seems like `.md` hyperlink format is being used and the "here" is not actually a hyperlink. 

This is actually currently visible on the page: 

image


Emil-Ant:
closing as I used the wrong base as mentioned per the psychopy documentation!

PR #7519, opened 2026-01-14
DOCS: fix codecov hyperlink

Emil-Ant:
### What 
- Use the the correct formatting for `.rst` hyperlinks for text on the `/developers/testsuitecontributing.html` 

### Why
- It seems like `.md` hyperlink format is being used and the "here" is not actually a hyperlink.

This is actually currently visible on the page: 

image


TEParsons:
Thanks for catching this! I think it was me who made that mistake in the first place, my markdown muscle memory is too strong :')

PR #7529, opened 2026-01-22
Adding scores to trial data

alexOpenScience:
Adding score trials to data so frontend only needs to make one request. 
Uses UUID's to match to the trial


alexOpenScience:
@TEParsons I've just removed the setTrialScore and now call setTrialData 3 times

PR #7531, opened 2026-01-23
Add missing setTrialData to session

alexOpenScience:


alexOpenScience:
@TEParsons I missed adding this method to the session

PR #7532, opened 2026-01-23
BF: Allow iohub errors in clients to be logged

mdcutone:
Fixes issue where tracebacks arising from unhanded exceptions in ioHub subprocesses were not being properly logged


peircej:
When I do this from a builder-generated experiment I don't seem to see anything happen (not even the expected creation of iohub_last_session)

PR #7555, opened 2026-02-13
TEST: Fix linux test runners

TEParsons:


peircej:
@TEParsons this conflicts but I think is also superseded by https://github.com/psychopy/psychopy/pull/7563
I'm closing this for now

PR #7563, opened 2026-02-23
Removing app code for its own repo

peircej:
App code moved to https://github.com/psychopy/psychopy-app
so removing the code and dependencies from here


peircej:
@TEParsons @mdcutone 
There's currently a test fail on ubuntu with core dump (so no useful information) whereas the test does pass on macos runner. We need to work out what's going on here but I want to get the refactor into the repo before we start getting merge conflicts

PR #7568, opened 2026-03-02
BF: Skip camera devices with no available formats.

aforren1:
I have some virtual cameras (e.g. OBS) on my system that have no available formats, and this leads to e.g.:

```python
>>> x = CameraDevice.getAvailableDevices()
File "...\psychopy\hardware\camera\__init__.py", line 614, in getAvailableDevices
    minFrameRate = max(28, min([cam.frameRate for cam in allCams]))
ValueError: min() arg is an empty sequence
```

As far as I can tell, this is the only place where non-emptiness is assumed?

Tested vs. 2026.1.1 on Windows 11. This lets the app device manager successfully list the built-in webcam.

### DIFF
--- a/psychopy/hardware/camera/__init__.py
+++ b/psychopy/hardware/camera/__init__.py
@@ -600,6 +600,9 @@ class CameraDevice(BaseDevice):
         profiles = []
         # iterate through cameras
         for cams in CameraDevice.getCameras().values():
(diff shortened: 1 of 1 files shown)


mdcutone:
This indirectly fixed another bug that seemed unrelated that I couldn't track down. Thanks for your help @aforren1

PR #7590, opened 2026-03-24
FF: use string concatenation for default JS (#7588)

### LINKED ISSUE #7588: [Bug]: Builder exports invalid Python-style string syntax in JS code for dataFileName, causing SyntaxError
### PsychoPy Version

2026.1.2

Windows 11

### Bug Description

When creating an experiment with PsychoPy Builder (using default settings for the participant component without any modifications), exporting the experiment to JavaScript and running it in a local browser results in the following syntax error: Uncaught SyntaxError: Unexpected string

The error occurs on this auto-generated line of JS code:
psychoJS.experiment.dataFileName = (("." + "/") + (u'data/%s_%s_%s' % [expInfo['participant'], expName, expInfo['date']]));

Image

### Expected Behaviour

When exporting to JS, PsychoPy Builder should automatically convert Python-style string formatting to valid JavaScript syntax. 

### Steps to Reproduce

1. White any experiment in Builder and export HTML
2. Send to Runner and pilot in browser
3. Get Uncaught SyntaxError: Unexpected string
...

### Additional context

_No response_

billybillymc:
Fixes #7588.

## Tested on
- PsychoPy 2026.1.2
- Windows 11
- Python 3.11, pip install (no metapensiero, see note on dependency below)

## Problem
The default Data filename uses Python `%` stringformatting. The transpiler relies on `metapensiero`, an undeclared dependency not installed via `pip` and stored only as a repository on GitLab. Without it, `expression2js` silently emits raw Python as JavaScript, causing `SyntaxError` in the browser. Because metapensiero comes packaged with the standard installation of PsychoPy available on PsychoPy.org, the bug does not come up on that executable, but does come up when packaging the files from the GitHub repository.

## Fix
Change the default from `%` formatting to `+` concatenation (valid in both Python and JS, needs no transpilation).

## Scope
This fixes new experiments only. Existing `.psyexp` files store the old format in XML and would need a migration path or a fix to the silent `except: pass` in `py2js.py:89`.

### DIFF
--- a/psychopy/experiment/components/settings/__init__.py
+++ b/psychopy/experiment/components/settings/__init__.py
@@ -143,9 +143,9 @@ class SettingsComponent:
         # if filename is the default value fetch the builder pref for the
         # folder instead
         if filename is None:
-            filename = ("u'xxxx/%s_%s_%s' % (expInfo['participant'], expName,"
-                        " expInfo['date'])")
-        if filename.startswith("u'xxxx"):
+            filename = ("'xxxx/' + expInfo['participant'] + '_' + expName + '_'"
+                        " + expInfo['date']")
+        if filename.startswith("'xxxx"):
             folder = self.exp.prefsBuilder['savedDataFolder'].strip()
             filename = filename.replace("xxxx", folder)
 


peircej:
Thanks. Sounds good.

For info regarding metapensiero not being specified (but highly recommended for anyone using online studies):
- latest official releases of metapensiero require an old version of dukpy isn't compatible with recent Python versions
- [my PR for an updated version of dukpy](https://github.com/metapensiero/metapensiero.pj/pull/85) works for us (in py3.10) but results in a test fail on py3.11
- so we can install a working metapensiero from that gitlab fork but not the official release
- BUT if we include that as a dependency in the wheel we upload to pypi it gets blocked because only pypi dependencies are allowed in pypi wheels

Your solution is a fair way to handle the immediate issue. Other options: 
- provide a clear Alert to users that don't have metapensiero installed because if they are pushing studies online then they probably want this
- get metapensiero fixed for py3.11 so that the PR can be merged and we can start using official releases again. This will actually become more important this year because we also need PsychoPy to move to 3.11 (3.10 is end of life in Oct). The metapensiero team themselves don't have time to work on it - they don't use it anymore - so it's up to us

@TEParsons any further thoughts?

billybillymc:
> * get metapensiero fixed for py3.11 so that the PR can be merged and we can start using official releases again. This will actually become more important this year because we also need PsychoPy to move to 3.11 (3.10 is end of life in Oct). The metapensiero team themselves don't have time to work on it - they don't use it anymore - so it's up to us

I will take a look at what needs fixing on metapensiero as well. I'll check out the GitLab repo and see how it looks with regard to its usage inside psychopy.

peircej:
> I will take a look at what needs fixing on metapensiero as well. I'll check out the GitLab repo and see how it looks with regard to its usage inside psychopy.

Sounds great! Thanks. I _suspect_ the immediate fix so that metapensiero supports py3.11/3.12 will be a relatively small issue

If metapensiero isn't maintained in future (which seems the case) we do need a plan for how to translate py->js for users
- we could take over metapensiero (or at least the parts the we need)
- we could switch to something else like dukpy (which metapensiero uses) or something new like [Groovy](https://github.com/gradio-app/groovy) (but this isn't nearly as complete as metapensiero, like it can't transpile "print")
- we could switch to doing it with AI but the lack of control/reproducibility scares me (at least for now). The deterministic nature of our current solution, even though it doesn't generate the neatest code ever, is something I find reassuring!

apitiot:
Note that I have also implemented a restricted version of f-strings (it's in experiment/py2js_transpiler.py). It's not complete but it offers another way to deal with % formatting, i.e. to convert the expression to an f-string.

PR #7593, opened 2026-03-26
TESTS: updated runners and fixed fails

peircej:
In particular
- libwebkit2gtk needed updating for ubuntu 24.04
- ubuntu setcap needed update
- no need for pyWinhook
- better handling of using uv for selecting venv (set once and use)


HugoFara:
Hi @peircej . I think there is an issue in with `test_ori` (RMS=31.2, threshold=20) caused by stale exemplar images, not by the runner changes in this PR.

PR #7592 changed the textbox rotation math in `_layout()` from `self.size` to `self.contentBox._size.pix`, which changes the rendered output, but the 12 reference images in `tests/data/Test_textbox/test_ori_*.png` were not regenerated. They still reflect the old rotation behavior from June 2024.

This is why `dev` CI passes (it doesn't have #7592's fix yet) while `release` CI fails.

I could fix by regenerating the exemplar images on `release` after #7592. Happy to open a PR for that if convenient.

peircej:
Yes, exactly what I guessed too. I was going to get @TEParsons to do update the test artefacts (he updated the orientation behaviour) but if you're willing to jump on it that's great. Thanks for your help!

PR #7606, opened 2026-04-22
ENH: Automatically refresh expired Pavlovia authentication when needed

TEParsons:


psychopy/projects/pavlovia.py
```diff
@@ -472,11 +487,58 @@ def getNamespace(self, namespace):
...
+            params={
+                'client_id': client_id,
+                'refresh_token': refreshToken or self.__dict__['refreshToken'],
+                'grant_type': "refresh_token",
+                'redirect_uri': redirect_url,
+                'code_verifier': code_verifier
+            }
+        ).json()
+        # start again with new token
+        print(resp)
```
peircej: `print(resp)` looks like a debugging statement. Maybe remove (or convert to be `logging.debug(resp)`)

TEParsons: Oops, yes that's leftover from when I was debugging. Removed now.

PR #7616, opened 2026-04-30
BF: Find user-level installed fonts on Linux

TEParsons:
`/usr/share/fonts` is the standard install location for fonts installed system-wide, but there's a different standard location for user-level fonts, and user-level install is the default (on GNOME at least) if you install by just double clicking a .ttf file and pressing "Install".

This path is common to GNOME and KDE from what I can tell from the docs, so should cover the majority of distros (at least as much so as the existing search paths).

### DIFF
--- a/psychopy/tools/fontmanager.py
+++ b/psychopy/tools/fontmanager.py
@@ -35,6 +35,7 @@ _X11FontDirectories = [
     "/usr/X11/lib/X11/fonts",
     # here is the new standard location for fonts
     "/usr/share/fonts",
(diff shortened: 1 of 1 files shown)


peircej:
@TEParsons there was an issue in the past with enumerating fonts on linux crashing the app but I think that was in a different point. Do you remember? Could you just confirm?

TEParsons:
I think I remember that - wasn't it `matplotlib.font_manager.findSystemFonts` which failed? I think that's why we now only do that if `sys,platform == "win32"`

PR #7617, opened 2026-05-08
fix: replace 35 bare except clauses with except Exception

KeloYuan:
## Summary

Replaces **35** bare `except:` with `except Exception:` across **26 files**.

Key areas: preferences, tools, experiment components, iohub client, visual backends, demos.

## Why

Bare `except:` catches `SystemExit` and `KeyboardInterrupt`.

### DIFF
--- a/building/searchCopyrightYear.py
+++ b/building/searchCopyrightYear.py
@@ -60,7 +60,7 @@ targetFiles = 0 # count of files to be updated
 tmpFile = './replaceCopyright'+oldYear+'_'+newYear+'.sh'
 try:
     del files[files.index(tmpFile)]
-except:
+except Exception:
     pass
 tmp = open(tmpFile, 'w')
 tmp.write('#!/bin/sh \necho Updating...\n')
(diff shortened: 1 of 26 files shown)


HugoFara:
Hi @KeloYuan, thanks for this! Just a heads up: since this was opened, #7604 and #7605 both merged and cover most of the same ground, which is why this is now showing as conflicting.

I went through and diffed against current `dev`: of the bare `except:` sites this PR touches, 4 are still genuinely unfixed (`building/searchCopyrightYear.py`, one in `preferences.py`, one in `tools/fontmanager.py`), and the rest are now redundant with what #7604/#7605 already merged.

I also found 3 more that neither this PR nor the prior sweeps caught -- an identical `resolveBackend()` classmethod duplicated across `psychopy/sound/sound.py`, `psychopy/hardware/microphone/__init__.py` and `psychopy/hardware/speaker/__init__.py` (the latter two look like they were reintroduced when those modules became packages, after the original sweep had already fixed the old single-file versions).

Opened #7707 covering all 7 remaining sites, so feel free to close this one -- sorry for the overlap, this is just a case of two people hitting the same cleanup independently while the earlier PRs were still in flight.

PR #7676, opened 2026-06-17
PKG: don't import (unused) psychopy.projects.sshkeys to reduce dependencies

peircej:
sshkeys depends on the cryptography package which otherwise isn't needed


peircej:
@TEParsons we haven't used the sshkeys for years (we sync git projects using https which doesn't need this) and this is the only place we use the cryptography package to my knowledge

PR #7684, opened 2026-06-19
FF: Fixes to audio backend

TEParsons:


peircej:
Sorry @TEParsons I think this should be `except ImportError` which is what's currently occuring. That includes "didn't exist" and "exists but fails to load"

peircej:
ach, I see you've merged it. OK, I guess in the interests of getting this out it's fine, but we should change that really not to have bare `except`

TEParsons:
I'll put in a separate PR to change that for the first bug fix release

PR #7688, opened 2026-06-27
BF: remove undefined _getSvnVersion/_getHgVersion references in info.py

HugoFara:
## Summary

`_getSvnVersion` and `_getHgVersion` were removed in 895f4cbf0 ("PKG: getHgVersion and getSvnVersion no longer useful"), but two loose ends were left behind in `psychopy/info.py`:

- their entries in `__all__`, and
- their call sites in `RunTimeInfo._setExperimentInfo`.

Because the call sites are wrapped in `try/except Exception`, the `NameError` raised by the missing functions has been **silently swallowed** ever since — so the SVN/Mercurial revision capture (`experimentScript.svnRevision`, `experimentScript.hgChangeSet`, …) has been dead code with no observable effect.

`pyflakes psychopy/info.py` reports these as undefined names:

```
psychopy/info.py:13:1: undefined name '_getHgVersion' in __all__
psychopy/info.py:13:1: undefined name '_getSvnVersion' in __all__
psychopy/info.py:276:33: undefined name '_getSvnVersion'
psychopy/info.py:285:27: undefined name '_getHgVersion'
```

## Changes

- Remove the dangling `__all__` entries and the dead SVN/Mercurial blocks, completing the removal started in 895f4cbf0. The other private names in `__all__` (`_getUserNameUID`, `_getHashGitHead`, `_getSha1hexDigest`, `_thisProcess`) remain defined and are kept.
- Add a regression test (`test_all_names_defined`) asserting every name in `info.__all__` resolves, so a future dangling export is caught.

No behaviour change: the removed code never executed successfully.

## Testing

```
python -m pytest psychopy/tests/test_misc/test_info.py -q   # 2 passed
python -m pyflakes psychopy/info.py                         # no undefined-name warnings
```

Found via static analysis; not linked to an existing issue.

### DIFF
--- a/psychopy/info.py
+++ b/psychopy/info.py
@@ -21,8 +21,6 @@ __all__ = [
     '_getUserNameUID',
     '_getHashGitHead',
     '_getSha1hexDigest',
-    '_getHgVersion',
-    '_getSvnVersion',
     '_thisProcess'
 ]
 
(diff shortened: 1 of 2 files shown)


HugoFara:
It seems one pipeline is not passing. As far as I can tell, it is a flaky test on a pixel-precise location. Rerunning the pipeline should fix it, but I don't think it's worth the effort.

peircej:
> It seems one pipeline is not passing. As far as I can tell, it is a flaky test on a pixel-precise location. Rerunning the pipeline should fix it, but I don't think it's worth the effort.

I think it was actually an issue with google fonts server, which is also likely to be a brief glitch. Easy to rerun though

HugoFara:
Hi, pipelines now pass, any news on this PR?

PR #7691, opened 2026-06-29
FF: sample-accurate (sub-block) onset scheduling in the SoundDevice backend

HugoFara:
# FF: sample-accurate (sub-block) onset scheduling in the SoundDevice backend

> **Draft / RFC — opening this for coordination, not asking for a merge yet.**

### Context

I'm building an auditory frequency-tagging EEG task (periodic stimulus onsets at a fixed rate, where per-onset jitter directly smears the tagged neural response) so I needed precise audio onset timing. Digging into the SoundDevice backend I saw you're actively working on exactly this (the new backend in #7584 and the "Initial work of precision audio onset timing" commit). I didn't want to step on in-progress work, so I'm opening this as a **draft** to share what I found plus a tested patch, and to ask how you'd like to coordinate. If sub-block placement is already in a local branch of yours, I'm happy to drop this or rebase onto whatever you have.

### What I found

`_SoundStream._callback` starts a scheduled sound at the beginning of the first output block whose `outputBufferDacTime` has reached `_tSoundRequestPlay`:

```python
if thisSound._tSoundRequestPlay > outputBufferDacTime:
    continue                      # not time yet
dat = thisSound._nextBlock(blockSize)
toSpk[:datSize] += dat            # ...then always written at sample 0 of the block
```

So the onset is quantised to the audio block boundary. There's no intra-block sample offset. Onset error is uniform in `[0, blockSize/sampleRate)`:

| blockSize | mean / max onset jitter @ 44.1 kHz |
|-----------|------------------------------------|
| 128       | 1.44 / 2.76 ms                     |
| 256       | 2.89 / 5.51 ms                     |
| 512       | 5.78 / 11.0 ms                     |

It varies per onset (whenever the requested time isn't block-aligned), which is the worst case for frequency-tagging and any onset-locked measure.

### The change

Place the onset on the exact requested sample:

- `_callback` computes the sub-block sample offset of the onset within the current block and writes the (partial) first block at that offset → residual < 1 sample (~0.02 ms).
- `_nextBlock` advances `self.t` by the samples actually produced (not the configured `blockSize`), so the partial first block stays gapless.
- the end-of-stream check compares to the samples requested this block (`nReq`), not the full buffer, so a partial first block isn't mistaken for the end.
- `samplesLeft` is rounded, not truncated, so an off-grid time cursor (a consequence of sub-block offsets) doesn't drop the final sample.

### Verification

- Adds `psychopy/tests/test_sound/test_sd_onset_offset.py`: headless (no audio device) tests that drive the **real** `_callback` / `_nextBlock` with synthetic PortAudio timing — asserting sample-accurate onset across block phases, gapless continuity, and full reconstruction through EOS. They **fail on `dev`** (onset at the block boundary, `assert 64 == 20`) and **pass with this change**.
- Existing `test_sound/test_sound.py` still collects cleanly; I haven't run the playback tests (no device on my side). A CI run would be appreciated.

### Still open (why this is a draft)

There is a few tests I would have wanted to do, but I'm blocked on the hardware side. I'm putting them here as the part I estimate "genuinely missing validation".

- [x] Full `test_sound/` playback suite on hardware/CI — these changes touch shared paths (generated tones, file streaming, the Hamming window, looping / multi-sound), which the new test doesn't cover.
- [x] Hardware loopback measurement of the *acoustic* onset (the unit tests only prove buffer placement; I have a measurement script ready).
- [x] `_tSoundRequestPlay` is set from `time.monotonic()` while `_callback` compares against PortAudio `outputBufferDacTime`. These appear to share an epoch on Linux (`CLOCK_MONOTONIC`) but I'm not certain that holds on macOS/Windows — worth a check, and if it's shaky maybe convert via the stream clock. Your call on whether that belongs in this PR.

### Coordination

This is your backend and clearly active work, so just let me know how you'd like to proceed — take it forward as-is, fold it into your branch, or close it if it overlaps something you already have. cc @TEParsons (audio timing).

### DIFF
--- a/psychopy/sound/backend_sounddevice.py
+++ b/psychopy/sound/backend_sounddevice.py
@@ -276,11 +276,22 @@ class _SoundStream:
         outputBufferDacTime = timepoint.outputBufferDacTime
 
         self.frameN += 1
+        blockDur = blockSize / float(self.sampleRate)
         for thisSound in self.sounds.copy():
-            if thisSound._tSoundRequestPlay > outputBufferDacTime:
-                continue  # not time to play this sound yet
+            # Where does this sound's requested onset fall relative to the start
+            # of this output block? Place the onset with sub-block (sample)
+            # precision instead of quantising it to the block boundary, which
+            # otherwise adds up to one blockSize of onset jitter.
+            tToStart = thisSound._tSoundRequestPlay - outputBufferDacTime
+            if tToStart >= blockDur:
+                continue  # onset falls in a later block - not time yet
+            # sample offset of the onset within this block (0 if already due)
+            offset = int(round(tToStart * self.sampleRate)) if tToStart > 0 else 0
+            if offset >= blockSize:
+                continue
 
-            dat = thisSound._nextBlock(blockSize)
+            nReq = blockSize - offset  # samples remaining in this block from offset
+            dat = thisSound._nextBlock(nReq)
             if dat is None:  # no data for some reason (e.g., sound finished)
                 continue
 
@@ -290,20 +301,22 @@ class _SoundStream:
 
             datSize = len(dat)
             datDims = len(dat.shape)
+            end = offset + datSize  # write the block starting at the sample offset
 
-            # old method
             if self.channels == 2 and datDims == 2:
-                toSpk[:datSize, :] += dat  # add to out stream
+                toSpk[offset:end, :] += dat  # add to out stream
             elif self.channels == 2 and datDims == 1:
-                toSpk[:datSize, 0] += dat 
-                toSpk[:datSize, 1] += dat  
+                toSpk[offset:end, 0] += dat
+                toSpk[offset:end, 1] += dat
             elif self.channels == 1 and datDims == 2:
-                toSpk[:datSize, :] += dat
+                toSpk[offset:end, :] += dat
             else:
-                toSpk[:datSize, 0:self.channels] += dat 
+                toSpk[offset:end, 0:self.channels] += dat
 
-            # check if that was a short block (sound is finished)
-            if datSize < len(toSpk[:, :]):
+            # check if that was a short block (sound is finished). Compare to the
+            # number of samples requested this block (nReq), not the full buffer,
+            # so a partial first block (offset > 0) is not mistaken for the end.
+            if datSize < nReq:
                 self.remove(thisSound)
                 thisSound._EOS()
                 # check if that took a long time
@@ -762,7 +775,10 @@ class SoundDeviceSound(_SoundBase):
         if not self.isPlaying:
             return
         
-        samplesLeft = int((self.duration - self.t) * self.sampleRate)
+        # round (not truncate): with a sub-block start offset self.t no longer
+        # lands on the integer-sample grid, so a bare int() could drop the final
+        # sample (e.g. 19.9999 -> 19).
+        samplesLeft = int(round((self.duration - self.t) * self.sampleRate))
         blockSize = blockSize or self.blockSize
         nSamples = min(blockSize, samplesLeft)
         
@@ -817,7 +833,10 @@ class SoundDeviceSound(_SoundBase):
                 else:
                     block *= thisWin[0:len(block)]
 
-        self.t += self.blockSize / float(self.sampleRate)
+        # advance by the number of samples actually produced (not the configured
+        # blockSize) so a partial block - e.g. the first block of a sound started
+        # at a sub-block sample offset - keeps the time cursor continuous.
+        self.t += len(block) / float(self.sampleRate)
 
         return block
 
--- a/psychopy/tests/test_sound/test_sd_onset_offset.py
+++ b/psychopy/tests/test_sound/test_sd_onset_offset.py
@@ -1,0 +1,146 @@
+"""Sub-block (sample-accurate) onset scheduling for the SoundDevice backend.
+
+Regression test for the fix that places a sound's onset at the exact sample
+within an output block, rather than quantising it to the block boundary
+(which added up to one blockSize of onset jitter).
+
+The test drives the real ``_SoundStream._callback`` and ``SoundDeviceSound.
+_nextBlock`` with synthetic PortAudio timing info, so it needs no audio device
+and is safe in headless CI. We build the objects with ``object.__new__`` to
+avoid opening a hardware stream.
+"""
+import types
+import numpy as np
+import pytest
+
+sd = pytest.importorskip("sounddevice")
+pytest.importorskip("soundfile")
+
+import psychopy.sound.backend_sounddevice as bsd
+from psychopy.sound.backend_sounddevice import _SoundStream, SoundDeviceSound
+
+
+SR = 44100
+BLOCK = 64
+OFFSET = 20          # requested onset, in samples, inside the first block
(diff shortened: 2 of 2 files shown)


mdcutone:
@HugoFara Thanks for working on this feature, looks good so far. You can take up working on this and I can review it when done

About the stream clock, I've done timing tests using photodiodes and at least on MacOS and Linux the DAC output times are close enough to `time.monotonic()` to schedule onset times precisely

HugoFara:
Ran this on real macOS hardware (MacBook Air, arm64, macOS 26.1, Python 3.11, sounddevice 0.5.5) to close out some of the "still open" items above:

**Unit tests** — `test_sd_onset_offset.py`: all 7 pass unchanged (`7 passed in 26.57s`).

**Clock epoch question, resolved** — measured `time.monotonic()` against PortAudio's `currentTime`/`outputBufferDacTime` from inside a live callback on a real output stream:

| | |
|---|---|
| `time.monotonic()` vs `currentTime` | track together, offset ~0.0001–0.0002s (just measurement jitter) |
| `outputBufferDacTime` vs `time.monotonic()` | stable **+14.7ms**, converging tightly across samples |

The +14.7ms is consistent with normal DAC/output buffering latency for a 512-sample block @ 44.1kHz, not epoch drift. **`time.monotonic()` and PortAudio's stream clock share the same epoch on macOS**, same as on Linux — comparing `_tSoundRequestPlay` against `outputBufferDacTime` is valid here. I think that item can come off the list.

**Real hardware playback** — built a `SoundDeviceSound` directly (forced `audioLib=['sounddevice']` to bypass the unrelated PTB path), scheduled a 440Hz tone 150ms out via `play(when=0.15)`. It played through the MacBook Air speakers over a real PortAudio stream (`44100_2_128`) with no exceptions — the sub-block onset path runs cleanly on real macOS Core Audio.

**Existing `test_sound.py` suite** — still errors here, but for an unrelated reason: it hardcodes `contextName='ptb'`, and the `psychtoolbox` backend fails to import (`PsychHID.so` needs `libusb`, not installed on this machine, no admin rights to brew-install it). This doesn't touch the `sounddevice` backend code this PR changes — pre-existing environment gap on the PTB path, not a regression from this change.

Only remaining item from the checklist is the acoustic loopback measurement.

HugoFara:
Follow-up: added a second, independent acoustic loopback measurement on Linux (ALSA/PipeWire), same method as the macOS run above (built-in speaker -> built-in mic, single process, sharp `hamming=False` click, sample-level threshold detection in the continuously-recorded mic stream).

| | macOS (MacBook Air, CoreAudio) | Linux (ALSA/PipeWire) |
|---|---|---|
| n detected | 12/12 | 11/12 (trial 0 missed - first-call stream warm-up, benign) |
| mean latency | +50.8 ms | **-6.25 ms** |
| std | 0.601 ms | **0.165 ms** |
| jitter (max-min) | 1.87 ms | **0.45 ms** |

Two independent OSes, audio backends (CoreAudio vs ALSA/PipeWire), and machines, same qualitative result: **tight, phase-independent onset latency** across trials landing at effectively random block phases (natural scheduling jitter between trials). The Linux run is even tighter (sub-millisecond jitter).

The mean latency differs in sign and magnitude between the two (+50.8ms vs -6.25ms) - expected, not a concern: that's just each platform's own DAC+ADC+processing latency as reported by `outputBufferDacTime`/measured round-trip, which naturally differs by backend. What validates sub-block placement is the *spread* across trials, not the absolute offset, and both are sub-2ms.

One unrelated usability note from setting this up on Linux: `SoundDeviceSpeakerDevice`'s "no preference -> first device found" fallback (`speaker_sounddevice.py::getAvailableDevices()[0]`) doesn't filter out output-only devices with no real audio path (e.g. unconnected HDMI outputs), so on a desktop with multiple audio devices it can silently pick a device nothing is plugged into. I had to pass `speaker='default'` explicitly to get real hardware. Not a blocker for this PR, but worth a separate look at some point.

This closes out the last item on my "still open" checklist (acoustic loopback measurement). Converting this out of draft.

peircej:
Hmmm, I am a bit surprised there's a lag of 50ms on macos. Psychtoolbox backend has been able to hit a start time with a sub-ms lag (See the Audio Lag column in https://psychopy.org/timing/2020/table2.html). There's a known lag for the screen refresh on MacOS but the audio processing has been good so far

peircej:
> I had to pass speaker='default' explicitly to get real hardware. Not a blocker for this PR, but worth a separate look at some point.

Good to know that worked, at least.

And good work on the test configuration!

HugoFara:
> Hmmm, I am a bit surprised there's a lag of 50ms on macos. Psychtoolbox backend has been able to hit a start time with a sub-ms lag (See the Audio Lag column in https://psychopy.org/timing/2020/table2.html). There's a known lag for the screen refresh on MacOS but the audio processing has been good so far

Thanks a lot for the table. Better to hold off this PR then, I will investigate further, and let you know if I can find a good explanation before calling it "ready to merge".

mdcutone:
@HugoFara The screen refresh stuff is being dealt with (see #7794) so `getFutureFlipTime()` returns the point when the buffer actually shows up on the display. So that should allow for scheduling the sound to coincide with the stimulus appearing on-screen.

Please continue working on this PR if you wish. I need this feature in soon since it was planned for the next release and we need to test it internally. If it's not complete by the 30th, I'll be taking over development of this feature and closing this PR.

HugoFara:
@mdcutone Thanks for the heads-up on #7794. I've rebased onto `dev` (now including #7794), and this is ready for review. I also added a test that `play(when=win)` starts the sound on the exact sample of the flip time from `getFutureFlipTime()`. It fails on `dev` whenever the flip falls mid-block.

I redid the acoustic loopback on Linux with a more careful method: a 20 ms chirp located by cross-correlation instead of a threshold, 30 onsets per condition, and `dev` measured alongside this PR. Onset spread:

| block | this PR: std / range | `dev`: std / range |
|---|---|---|
| 128 | 0.26 / 0.89 ms | 1.61 / 6.28 ms |
| 256 | 0.60 / 1.92 ms | 1.37 / 4.77 ms |
| 512 | 0.13 / 0.45 ms | 2.96 / 9.79 ms |

**Revisiting my earlier comment.** I described the +50.8 ms (macOS) and −6.25 ms (Linux) means as each platform's own latency, but it turns out to be less simple than that (the negative Linux value was a hint). When the recording is timestamped with the input stream's `inputBufferAdcTime`, the mean offset equals the error in PortAudio's reported output latency *plus* the error in its reported input latency. This kind of test can't separate the two.

To check, I ran a single input+output stream, where input and output share one sample clock and the true round trip can be measured exactly. On this laptop, PortAudio reports a round trip 2.0–2.3 ms longer than the true one at blocks 128/256, and 8.0 ms longer at 512. That matches this run's mean offsets (−2.2 / −2.3 / −7.9 ms). So the sound may come out a few ms early relative to the requested time, or the error may be on the input side. I can't tell which.

@peircej For the same reason, the macOS +50.8 ms may be output latency, input latency, or both. I no longer have access to that Mac, so I can't narrow it down. A photodiode + mic measurement against a flip would answer it directly (@mdcutone, you mentioned having that setup). Whatever the answer, it's about how accurate `outputBufferDacTime` is, which this PR doesn't change: the fix only decides where the onset falls inside a block.

PR #7727, opened 2026-07-29
ENH: Speed up getting Component profiles

TEParsons:
Mostly by removing unecessary calls to other parts of the library within psychopy.experiment, instead moving them inside functions which import them *as needed* rather than at the very beginning.

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -22,20 +22,38 @@ jobs:
           contents: write 
         steps:
 
-        - uses: actions/checkout@v6
+        - name: Checkout PsychoPy (release)
+          uses: actions/checkout@v6
           if: ${{ startsWith(github.ref, 'refs/tags/') }}
           with:
             ref: ${{ github.ref_name }}
+            path: psychopy
 
-        - uses: actions/checkout@v6
+        - name: Checkout PsychoPy (tag)
+          uses: actions/checkout@v6
           if: ${{ github.event.inputs.tag }}
           with:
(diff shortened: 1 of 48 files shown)


peircej:
I certainly approve of the approach, once tests pass

If we're concerned, by the way, about later sluggishness (i.e. when the function is first called and does its import), we could create an additional function like `utils.loadAncilliaryLibs()` for importing things that aren't urgent enough to be needed immediately, but slow enough that we want to load them in an idle period once the UI is up and running

peircej:
Could set a timer to `utils.loadAncilliaryLibs()` after say a minute of the app being open (or maybe there's a flag we can set when the app gets to a state of being idle)

TEParsons:
> Could set a timer to `utils.loadAncilliaryLibs()` after say a minute of the app being open (or maybe there's a flag we can set when the app gets to a state of being idle)

Polling / flagging idle is tricky, I can't think of an easy way to do it, but we could go after a set time as you say. What I'd be worried about with that though is that slowdown would become less predictable - if the app says "Loading Cedrus devices..." for a while because we're importing `pyxid2` that's not great, but if the app randomly takes forever on something which should be basically instant because it's decided now is a good time to import `pyxid2`, that feels worse?

We *could* make Liaison execute commands asynchronously (currently it listens asynchronously but executes them one at a time in the main thread), but I expect there'll be new issues arising from that if we do. Maybe it's worth it though?

TEParsons:
For now I'll pull this in, but will keep mulling over `loadAncillaryLibs`

PR #7730, opened 2026-08-06
BF: Add expInfo code to survey routine

wakecarter:
I've added "%(name)s.setVariables(expInfo);\n" to the survey routine __initi__.py
Manually editing the code works in 2026.2.2 (different code is needed in 2026.2.1)

### DIFF
--- a/psychopy/experiment/routines/pavlovia_survey/__init__.py
+++ b/psychopy/experiment/routines/pavlovia_survey/__init__.py
@@ -129,7 +129,7 @@ class PavloviaSurveyRoutine(BaseStandaloneRoutine):
         code = (
             "});\n"
             "%(name)sClock = new util.Clock();\n"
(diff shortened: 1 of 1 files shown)


HugoFara:
Thanks @wakecarter I gave it a check.

`excludedNames` only got a default (`= []`) in psychojs 2026.2.2. On 2026.2.1 and earlier (checked back to 2023.2.3) the one-arg call hits `undefined.indexOf(...)` and throws. Since *Use PsychoJS version* lets people pin an older lib, passing the list explicitly is safer and identical on 2026.2.2:

```suggestion
     "%(name)s.setVariables(expInfo, []);\n"
```

wakecarter:
Thanks @HugoFara . I had the extra argument in my original test of 2026.2.1 but I hadn't checked it would work in 2026.2.2 without giving an error. I've now checked that and added , [] into the merge request.

PR #7744, opened 2026-08-24
feat: Make it easier to use CountdownTimer in for/while loops

TEParsons:
This comes out of me writing fairly repetitive code for quick tests of visual stimuli and realising there's a way to slightly streamline it... Extremely low priority, I just thought this would be nice & basically harmless.

### Example
Displaying a basic Polygon for 5s

Before:
```
from psychopy import visual
import time

win = visual.Window()
obj = visual.Polygon(win)

started = time.time()
while time.time() - started < 5:
    obj.draw()
    win.flip()
```

After:
```
from psychopy import clock, visual

win = visual.Window()
obj = visual.Polygon(win)

for t in clock.CountdownTimer(5):
    obj.draw()
    win.flip()
```

### DIFF
--- a/psychopy/clock.py
+++ b/psychopy/clock.py
@@ -372,6 +372,62 @@ class CountdownTimer(Clock):
         if start:
             self.reset()
 
+    def __iter__(self):
+        return self
+    
+    def __next__(self):
+        """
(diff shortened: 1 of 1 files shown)


TEParsons:
@claude review

PR #7751, opened 2026-08-26
BF: Fixed 'add' blend mode for text rendering

mdcutone:
Changes the the shader `TextBox2` uses to allow for correct opacity and color blending when the window is in 'add' mode


peircej:
@claude review

Issue #7755, opened 2026-08-31
[Bug]: Unsaved data recording if dataFileName contains folder ending with space under windows


suxpert:
### PsychoPy Version

All version up to 2026.2.3

### What OS are your PsychoPy running on?

Windows 11

### Bug Description

In our experiment workflow (following the BIDS specification), we typically store each participant’s data in a separate folder, with the participant ID entered via a dialog box.

On one occasion, the experimenter accidentally included a trailing space in the participant ID. This ID was used by `ExperimentHandler` to create the data folder. At first, the experiment proceeded without any visible error. However, when the experiment ended and `ExperimentHandler` attempted to write the data, a `FileNotFoundError` was raised. As a result, all behavioral data collected during those runs were lost, with no opportunity for recovery.

### Expected Behaviour

Although it is not common practice to end folder names with a space, such names are technically valid in many file systems. The core issue appears to be platform‑specific: on Windows, when creating a folder with a trailing space, Python (via the underlying OS API) silently strips that space, producing a folder name without the trailing space. Yet the file path used later still contains the original folder name with the space, causing the file to be inaccessible.

### Steps to Reproduce

Here is a minimal reproducible example with psychopy:
```python
from psychopy import data

def main():
    participant_id = '001 '
    filepath = f'{participant_id}/{participant_id}-event.tsv'
    record = data.ExperimentHandler(dataFileName=filepath, extraInfo={'participant': participant_id})

if __name__ == '__main__':
    main()
```

MRE with python-minimal:
```python
import os

path = '001 '
file = f'{path}/somefile.txt'
os.makedirs(path, exist_ok=True)

with open(file, 'w') as f:
    f.write('Hello, world!')

```

### Additional context

While this behavior seems to stem from Windows itself, we believe there is a strong case for addressing it at the application level. At the very least, an exception should be raised **before** the experiment begins---rather than allowing a full run to complete and then discarding all recorded data and raised an error.

Or always follow the users' request, to support such folder, as might be in Linux & macOS.

Yi-111-a:
I looked into this and traced it to `checkValidFilePath()` in `psychopy/data/utils.py`; I've opened PR #7773 against the `release` branch for it.

What actually happens on Windows (measured on Windows 11):

- `os.path.abspath('001 /x.tsv')` keeps the trailing space
- `os.path.isdir('<...>/001 ')` is `False`, so `checkValidFilePath()` calls `os.makedirs()`
[...]

PR #7773, opened 2026-09-11
BF: Strip trailing spaces from data file paths in handleFileCollision

Yi-111-a:
## What this fixes

Issue #7755: on Windows, if the participant ID (used to build the data folder) contains a
trailing space, the whole experiment runs without complaint and only fails when the data is
saved — at which point **all collected data is lost**.

## Root cause

Windows silently strips trailing spaces from folder names when they are created, so a
requested path like `001 /x.tsv` ends up in a folder called `001` on disk while the path
held by `ExperimentHandler` still contains the space — and the final `open()` raises
`FileNotFoundError` after all data has been collected.

## The fix

Per maintainer feedback, instead of rejecting such paths we now strip the spaces ourselves:

- `handleFileCollision()` (in `psychopy/tools/fileerrortools.py`) removes trailing spaces
  from each path component before the overwrite/rename/fail logic runs, and logs a warning
  when the path is changed. `dataFileName` always passes through this function before being
  stored on `ExperimentHandler`, so the stored name matches what is actually created on
  disk.
- `checkValidFilePath()` is back to its original behaviour — it no longer rejects
  trailing-space paths and the `platform` test hook is gone.
- `ExperimentHandler` now validates `self.dataFileName` (the stripped name) rather than the
  raw argument, so `makedirs` runs on the corrected path.

## Tests

- `psychopy/tests/test_tools/test_fileerrortools.py`: new test covering trailing-space
  stripping of folder and file-name components for both `overwrite` and `rename`, plus a
  check that internal spaces are left untouched. Existing collision tests still pass.
- `psychopy/tests/test_data/test_utils.py`: the previous rejection test is replaced by a
  simple `checkValidFilePath` sanity check.

```
$ pytest psychopy/tests/test_tools/test_fileerrortools.py  -> 7 passed
$ pytest psychopy/tests/test_data/test_utils.py            -> 13 passed
```

**Not run**: the repository's full test suite (needs wxPython/pygame and a display).

### DIFF
--- a/psychopy/data/utils.py
+++ b/psychopy/data/utils.py
@@ -37,15 +37,42 @@ haveXlrd = False
 _nonalphanumeric_re = re.compile(r'\W')  # will match all bad var name chars
 
 
-def checkValidFilePath(filepath, makeValid=True):
+def checkValidFilePath(filepath, makeValid=True, platform=None):
     """Checks whether file path location (e.g. is a valid folder)
 
     This should also check whether we have write-permissions to the folder
     but doesn't currently do that!
 
+    On Windows a path component ending in a space is stripped when the
+    folder is created, so the folder on disk no longer matches the requested
+    path and saving data fails at the end of the run, after all data has
+    been collected (see issue #7755). Such paths are rejected here so that
+    the experiment fails straight away instead. A trailing period is also
+    stripped by Windows, but it resolves those consistently, so it is not
+    rejected.
+
+    `platform` is exposed so that the Windows behaviour can be exercised
+    from any OS (the CI matrix has no Windows runner).
+
     added in: 1.90.00
     """
+    if platform is None:
+        platform = os.name
     folder = os.path.split(os.path.abspath(filepath))[0]
+    if platform == 'nt':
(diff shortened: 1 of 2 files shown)


TEParsons:
I think we can go a step further; rather than rejecting filepaths with a space in, why not strip them ourselves? Before `dataFileName` is stored in the experiment handler, we pass it through `handleFileCollision`, so we could remove any trailing spaces inside this function.

Yi-111-a:
Thanks @TEParsons — good suggestion, that's cleaner than rejecting. Done in a570997:

- `handleFileCollision()` now strips trailing spaces from each path component before the collision logic runs, and logs a warning when it changes the path.
- `checkValidFilePath()` is reverted to its original behaviour — no more rejection, and the `platform` test hook is removed.
- `ExperimentHandler` now validates the stripped `self.dataFileName`, so the folder is created under the corrected name.

Tests added for stripping folder/file-name components under both `overwrite` and `rename` (`test_fileerrortools.py`: 7 passed, `test_utils.py`: 13 passed).

Issue #7754, opened 2026-08-31
[Bug]: Builtin demo scripts incompetible with psychopy, can not run correctly.


suxpert:
### PsychoPy Version

v2025.2.1, v2026.2.3

### What OS are your PsychoPy running on?

Windows 11

### Bug Description

Builtin demos do not always work, some of them may depends on deprecated features.

### Expected Behaviour

We would expect the builtin demos are starting point of learning how to use PsychoPy and use it efficiently, rather than meeting different kinds of errors.

### Steps to Reproduce

Here are some examples that do not run:
1. `testSoundLatency.py`:  `ModuleNotFoundError: No module named 'labjack'`
2. `soundStimuli.py`: `AttributeError: module 'psychopy.sound' has no attribute 'audioLib'. Did you mean: 'audioclip'?`
3. `variousVisualStims.py`: `AttributeError: module 'psychopy.visual' has no attribute 'MovieStim3'. Did you mean: 'MovieStim'?`

### Additional context

If it depends on additional packages/plugins, perhaps the `secondOrderGratings.py` is a good example:
> psychopy.plugins.util.PluginRequiredError: Support for `psychopy.visual.noise` is not available this session. Please install `psychopy-visionscience` and restart the session to enable support.

TEParsons:
Good catch, thanks! 

The reason we can handle `psychopy.visual.noise` that way is that it's a submodule of `psychopy`, so we can add a PluginStub class which raises that error when initialised. Because `labjack` is an entire package, we can't do the same. I think the solution in this case is to move that demo out to the `psychopy-labjack` plugin and add it to `psychopy.demos.coder` by an entry point; so if the plugin isn't installed then that demo simply won't be available.

For the others I think they're just outdated, updating them would be a nice way to get into contributing so I'll mark this as Good First Issue

Yi-111-a:
Thanks for marking this as a good first issue — I've taken the first one: PR #7774 fixes
`variousVisualStims.py`, where `visual.MovieStim3` no longer exists (only `MovieStim` is
exported now). `MovieStim` takes the same `win`/`filename`/`units`/`pos`/`size` arguments and
inherits `ori` from `BaseVisualStim`, so the rest of that demo works unchanged.

I left the other two alone for now:

- `soundStimuli.py` uses `psychopy.sound.audioLib`, which is gone. I wasn't sure what the
  intended replacement is — is it enough to just report the backend name in some other way, or
[...]

PR #7774, opened 2026-09-11
BF: Update variousVisualStims demo to use MovieStim

Yi-111-a:
Fixes one of the three demos listed in #7754 (`variousVisualStims.py`). You marked that issue
as a good first issue, so I hope this is the kind of thing you had in mind — happy to take the
other two as well if you'd like them in one go (notes below).

## The problem

```python
movie = visual.MovieStim3(win, 'jwpIntro.mp4', units='pix', pos=[100, 100], size=[160, 120])
```

`psychopy.visual` no longer has `MovieStim3` — `psychopy/visual/__init__.py` only exports
`MovieStim` — so the demo dies with
`AttributeError: module 'psychopy.visual' has no attribute 'MovieStim3'` on the first frame.

## The change

`MovieStim3` → `MovieStim`, nothing else. I checked that `MovieStim` actually covers what this
demo does:

- its `__init__` takes `win, filename, units, pos, size`, which is exactly what the demo passes
  (same argument order and names);
- the demo animates `movie.ori` later on, and `MovieStim` sets `self.ori` via `BaseVisualStim`,
  so that still works.

`jwpIntro.mp4` is still present in the demo folder, so the stimulus file is not missing.

I used the existing `MovieStim.py` demo in the same folder as the reference for current usage
rather than inventing anything.

## Not touched here (the other two from #7754)

- `soundStimuli.py` — uses `psychopy.sound.audioLib`, which no longer exists. I'd need to know
  what the intended replacement is (just reporting the backend name, or something else) before
  changing it.
- `testSoundLatency.py` — imports `labjack`. Per your comment on the issue this one needs to
  move out to the `psychopy-labjack` plugin and be added via an entry point, which is a
  different kind of change from this one.

Say the word if you'd rather I did either of those, or folded them into this PR.

## Validation

Source-level only, I'm afraid: I could not actually run the demo. There is no display available
here, and PsychoPy would not install in this environment (tried the normal install,
`--prefer-binary`, and `--no-build-isolation`; the last one fails building numpy from source
with no compiler available). So this rests on:

- `psychopy/visual/__init__.py` exports `MovieStim` and nothing called `MovieStim3`
- `MovieStim.__init__` signature accepts `win`, `filename`, `units`, `size`, `pos`
- `MovieStim` sets `self.ori`, so the demo's rotation still applies

If you can run it, the demo should now show the gabor, movie, text and face stimuli rotating
with mouse movement.

### DIFF
--- a/psychopy/demos/coder/stimuli/variousVisualStims.py
+++ b/psychopy/demos/coder/stimuli/variousVisualStims.py
@@ -11,7 +11,7 @@ import numpy
 win = visual.Window([600, 600], color='black')
 
 gabor = visual.GratingStim(win, mask='gauss', pos=[-0.5, -0.5], color=[0, 0, 1], sf=5, ori=30)
-movie = visual.MovieStim3(win, 'jwpIntro.mp4', units='pix', pos=[100, 100], size=[160, 120])
+movie = visual.MovieStim(win, 'jwpIntro.mp4', units='pix', pos=[100, 100], size=[160, 120])
 txt = u"unicode (eg \u03A8 \u040A \u03A3)"
 text = visual.TextStim(win, pos=[0.5, -0.5], text=txt, font=['Times New Roman'])
 faceRGB = visual.ImageStim(win, image='face.jpg', pos=[-0.5, 0.5])


TEParsons:
The change looks good, `MovieStim3` -> `MovieStim` is certainly correct. I'll look into the other errors preventing you from running it separately.

> I'd need to know what the intended replacement is (just reporting the backend name, or something else) before
changing it.

I think `Sound.backend` is what you want here

> Say the word if you'd rather I did either of those, or folded them into this PR.

Go for it :)

PR #7782, opened 2026-09-19
RF: Move movie writing classes to `movietools` from `camera`

mdcutone:
PR moves the movie writer classes from the camera module to movietools. They're general purpose for writing image frames to file asynchronously and can be used in other parts of PsychoPy without needing to import the camera module to access them.


mdcutone:
Needed to catch the OpenCV version on MacOS since there is a bug in the current package which makes it not ship with the necessary extensions built-in. See: https://github.com/opencv/opencv-python/issues/1156

PR #7790, opened 2026-09-21
PKG: Set `requires-python` to include Python 3.14.X

mdcutone:
Adds Python 3.14+ as a supported Python version. Allows for UV installs to work on that platform


peircej:
Could you also set 
```
ffpyplayer; python_version < "3.14",
```
since they haven't added support for 3.14

and remove moviepy entirely, since they're now requiring a version of pillow that raises security flags

PR #7794, opened 2026-09-21
RF: Rework of display sync on MacOS

mdcutone:
This PR reworks the DisplayLink logic for synchronizing buffer flips with actual screen refreshes. It is not possible to eliminate the frame lag entirely, however these changes improve stability. I also added a change that has `getFutureFlipTime` report the authoritative timestamp from the system when a frame will appear next on screen. So you can now synchronize audio, cameras, etc. with that timestamp to get precise multi-modal synchronization


mdcutone:
@peircej I'll let you test this one out at the Notts office with a photodiode before pulling it in.

peircej:
@claude review

peircej:
testing out Claude's reviews on this one

mdcutone:
Applied the suggestion made by the agent. However, I needed something more robust since framerate is unbound after timeouts. This would affect experiments that rely on frame counting for timing. Added software pacing to keep framerates consistent during those periods until the DisplayLink is reacquired

